Issue #167, opened 2011-02-12
Add merge option to phpunit.xml configuration


havvg:
# Behavior

**Given** there is a phpunit.xml.dist<br />
**And** there is a phpunit.xml<br />
**When** the phpunit.xml is set to merge changes<br />
**Then** all changes between phpunit.xml and phpunit.xml.dist should be merged into result configuration

# Goal

The goal to achieve with this enhancement: remove the need to update the phpunit.xml every time the phpunit.xml.dist changes e.g. by adding new testsuites.

# Example

## phpunit.xml.dist

    <?xml version="1.0" encoding="UTF-8"?>
    <phpunit
      colors="false"
      convertErrorsToExceptions="true"
      convertNoticesToExceptions="true"
      convertWarningsToExceptions="true"
      stopOnFailure="true">
    
      <filter>
        <blacklist>
          <directory>config</directory>
          <file>*.yml</file>
        </blacklist>
      </filter>
    
      <testsuites>
        <testsuite name="Unit Tests">
          <directory>test/unit/</directory>
        </testsuite>
      </testsuites>
    </phpunit>

## phpunit.xml

    <?xml version="1.0" encoding="UTF-8"?>
    <phpunit merge="true" colors="true">
      <logging>
        <log type="coverage-html" target="./web/coverage" charset="UTF-8" yui="false" highlight="true" lowUpperBound="65" highLowerBound="85"/>
      </logging>
    </phpunit>

## result configuration to use

    <?xml version="1.0" encoding="UTF-8"?>
    <phpunit
      colors="true"
      convertErrorsToExceptions="true"
      convertNoticesToExceptions="true"
      convertWarningsToExceptions="true"
      stopOnFailure="true">
        
      <filter>
        <blacklist>
          <directory>config</directory>
          <file>*.yml</file>
        </blacklist>
      </filter>
    
      <testsuites>
        <testsuite name="Unit Tests">
          <directory>test/unit/</directory>
        </testsuite>
      </testsuites>
    
      <logging>
        <log type="coverage-html" target="./web/coverage" charset="UTF-8" yui="false" highlight="true" lowUpperBound="65" highLowerBound="85"/>
      </logging>
    </phpunit>

ryanneufeld:
Can you provide an example, or link to the docs for this? I've been trying to get it working, but it doesn't seem to be working for me.

sebastianbergmann:
Dear contributor,

[...]

PR #190, opened 2011-03-10
Configuration inheritance

whatthejeff:
Added basic configuration inheritance as described in #167.

### DIFF
--- a/PHPUnit/Util/Configuration.php
+++ b/PHPUnit/Util/Configuration.php
@@ -177,6 +177,9 @@ class PHPUnit_Util_Configuration
     protected $xpath;
     protected $filename;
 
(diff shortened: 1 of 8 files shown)


beporter:
Did anything like this ever get implemented? It sure would be convenient to not have to specify all of the same configs from `phpunit.xml.dist` in my local `phpunit.xml` file. If I could instruct phpunit to load the dist file, then apply my overrides on top of it, it'd keep things a lot more DRY.

sebastianbergmann:
Have not found the time yet to implement #913. More important things got/get in the way.

beporter:
Wow, looking over that thread, I don't blame you-- it's a whopper of a project. Thanks for the response and the pointer.

PR #240, opened 2011-05-04
Fixes XML string escaping by removing invalid characters

lastzero:
Details are described in my commit note. I also added a test that checks all chars from 0 to 255 using DOMDocument::loadXML() - not sure how "standard" this implementation is, but for sure it is common in a PHP environment. Sorry for the delay.


lastzero:
Sorry, I didn't notice your reply. Is it still needed?

PR #754, opened 2012-12-17
Not automatically using the autoloader in GlobalState.php during class_exists

brobie:
We found this need because:
1. We do not have PHP_Invoker installed.
2. We have multiple autoloaders registered
3. When the class_exists was checking for "PHP_Invoker" it was
automatically trying to autoload it and failing in a way that fatally
ended our phpunit tests.

### DIFF
--- a/PHPUnit/Util/GlobalState.php
+++ b/PHPUnit/Util/GlobalState.php
@@ -406,7 +406,7 @@ class PHPUnit_Util_GlobalState
      */
     protected static function addDirectoryContainingClassToPHPUnitFilesList($className, $parent = 1)
     {
-        if (!class_exists($className)) {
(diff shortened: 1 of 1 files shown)


spencer-aerian:
Come  on Sebastian. Invoker is an optional package (according to the documentation) but you are making it a prerequisite. This is breaking lots of developers if its not loaded (and if you do load it you then have to muck about installing an optional php extension locally and on integration servers.)
I think the patch 'should be applied'.

sebastianbergmann:
I agree that the current situation is not ideal and am open to suggestion for how to improve it. Messing with the autoloader, however, because of broken autoloaders (see my comment in https://github.com/sebastianbergmann/phpunit/pull/754#issuecomment-13894665) is not an option.

PR #1060, opened 2013-11-22
Clone test arguments

### LINKED ISSUE #1057: Dependency hierarchy branches affect eachother
I may be doing something wrong or maybe I'm missing something. This is a very simplified version of the type of class I'm trying to test:

``` php
class SomeStack
{
    private $items = array();

    public function addItem($item)
    {
        $this->items[] = $item;
    }

    public function getItems()
    {
        return $this->items;
    }
}
```

And the test:

``` php
class SomeStackTest extends \PHPUnit_Framework_TestCase
{
    public function testStackInitiallyEmpty()
    {
        $stack = new SomeStack();

        $this->assertEmpty($stack->getItems());

        return $stack;
    }

    /**
     * @depends testStackInitiallyEmpty
     */
    public function testAddItem($stack)
    {
        $stack->addItem('SomeItem');

        $this->assertNotEmpty($stack->getItems());
    }

    /**
     * @depends testStackInitiallyEmpty
     */
    public function testSomethingElseThatWantsAnEmptyStack($stack)
    {
        $this->assertEmpty($stack->getItems()); // Fails
    }
}
```

In the example there's two different 'dependency hierarchy branches'. I was expecting to receive an empty stack in `testSomethingElseThatWantsAnEmptyStack`, but instead it seems that the earlier branch affected the state of the stack. Is this intended behavior? Am i doing something wrong? Is this a known limitation that is unfortunately a necessary evil for performance reasons? (I think the dependency would need to be rerun for every branch?)

There are workarounds of course, but they don't seem as elegant as having multiple tests that depend on the same 'parent' test.

baileylo:
If a test uses `@depenends`, clone the arguments that are objects to prevent unintended modification of the objects  by other functions.

My biggest fear here is an bump in memory usage by phpunit, as we now duplicate multiple objects.

Fixes #1057

### DIFF
--- a/PHPUnit/Framework/TestCase.php
+++ b/PHPUnit/Framework/TestCase.php
@@ -1837,7 +1837,11 @@ abstract class PHPUnit_Framework_TestCase extends PHPUnit_Framework_Assert imple
                         return FALSE;
                     }
 
-                    $this->dependencyInput[] = $passed[$dependency]['result'];
+                    if (is_object($passed[$dependency]['result'])) {
+                        $this->dependencyInput[] = clone $passed[$dependency]['result'];
+                    } else {
+                        $this->dependencyInput[] = $passed[$dependency]['result'];
+                    }
                 } else {
                     $this->dependencyInput[] = NULL;
                 }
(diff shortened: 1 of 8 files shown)


sebastianbergmann:
Dear contributor,

Sebastian

PR #1066, opened 2013-12-04
Implemented --version-check

frqnck:
This adds a `--version-check` option to the console -- convenient to all package distributions (inc. Phar).

It uses tags from the main PHPUnit repo on Github (sebastianbergmann/phpunit) to extract the latest (and stable) version numbers to compare against.

I could move this code to `PHPUnit_Runner_Version` if it is preferable?

For the story, I started to implement `PHPUnit_Util_SelfUpdate` in **3.7** and  I was about to PR -- just to find out that **master** had a an implementation too...


sebastianbergmann:
Implemented in f796a70200be4db1cec9fddf8129a971345c0e46

PR #1404, opened 2014-08-21
Do not silently discard exceptions from after methods

eelf:
In case of error in after test method (i.e. tearDown) add it to result


eelf:
I've noted that it is a bad idea to call testResultObject->addError somewhere in test (tearDown is a part of test too) before output buffering is closed.
I want phpunit to be able to trigger more than one failure/error in one test. It seems that it should be done using status and statusMessage changed to statusesList and statusMessagesList.

PR #1458, opened 2014-10-06
only enable colors when output to a terminal (not to a file)

remicollet:


Fludimir:
Looks like this commit breaks colors display when running on windows in cygwin terminal

GrahamCampbell:
@Fludimir Lots of things have changed since this.

Fludimir:
May be, but after update from phpunit 4.3.5 to 4.4.1 colors aren't displayed in console and probably no way to display it.

GrahamCampbell:
Yeh, try out phpunit alpha, 4.6.x. It's got the new behaviour in.

Fludimir:
Tried, works only with --colors=always in console. So the only way to display colors in cygwin console will be to manually add this option each run?

PR #1506, opened 2014-11-17
Adding support to testdox for showing incomplete tests in text output

jforzano:
I've been using phpunit's testdox to output our test results, but not being able to differentiate between failed and incomplete tests in the output has made it difficult to understand our results. This updates allows you to pass an  option `--testdox-text-show-incomplete` which makes the differentiation and marks incomplete tests with `[i]`.

Would love to get some feedback on this solution!

### DIFF
--- a/src/TextUI/Command.php
+++ b/src/TextUI/Command.php
@@ -115,6 +115,7 @@ class PHPUnit_TextUI_Command
       'testdox' => null,
       'testdox-html=' => null,
       'testdox-text=' => null,
+      'testdox-text-show-incomplete' => NULL,
       'test-suffix=' => null,
       'no-configuration' => null,
       'no-globals-backup' => null,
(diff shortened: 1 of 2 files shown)


src/Util/TestDox/ResultPrinter/TextShowIncomplete.php
```diff
@@ -0,0 +1,77 @@
+<?php
+/**
+ * PHPUnit
```
GrahamCampbell: incorrect doc

src/TextUI/Command.php
```diff
@@ -115,6 +115,7 @@ class PHPUnit_TextUI_Command
       'testdox' => null,
       'testdox-html=' => null,
       'testdox-text=' => null,
+      'testdox-text-show-incomplete' => NULL,
```
GrahamCampbell: powercse require

PR #1507, opened 2014-11-20
Intercept SIGINT signal on tests suites

henriquemoody:
If the `pcntl_signal()` function exists, creates a signal handler for `SIGINT` which prints the summary of test suite and generates code coverage reports.
Move all code which prints results and generates coverage from `doRun()` method to `doReports()` method on `PHPUnit_TextUI_TestRunner` class.

Here some benchmark of PHPUnit's test suite, without [tests/TextUI/sigint.phpt](https://github.com/henriquemoody/phpunit/blob/sigint-intercept/tests/TextUI/sigint.phpt) file:

**Without SIGINT interception**

| PHP Version | Time | Memory |
| --- | --- | --- |
| 5.3.3 | 28.6 seconds | 31.50Mb |
| 5.3 | 17.85 seconds | 31.50Mb |
| 5.4 | 18.63 seconds | 23.00Mb |
| 5.5 | 19.14 seconds | 23.25Mb |
| 5.6 | 19.21 seconds | 23.50Mb |

**With SIGINT interception**

| PHP Version | Time | Memory |
| --- | --- | --- |
| 5.3.3 | 20.04 seconds | 31.75Mb |
| 5.3 | 21.18 seconds | 31.75Mb |
| 5.4 | 21.76 seconds | 23.25Mb |
| 5.5 | 18.62 seconds | 23.50Mb |
| 5.6 | 20.23 seconds | 23.50Mb |

**See it working**: https://asciinema.org/a/14111

### DIFF
--- a/src/TextUI/TestRunner.php
+++ b/src/TextUI/TestRunner.php
@@ -27,6 +27,7 @@ class PHPUnit_TextUI_TestRunner extends PHPUnit_Runner_BaseTestRunner
     const SUCCESS_EXIT   = 0;
     const FAILURE_EXIT   = 1;
     const EXCEPTION_EXIT = 2;
+    const INTERRUPT_EXIT = 3;
 
     /**
      * @var PHP_CodeCoverage_Filter
@@ -325,6 +326,7 @@ class PHPUnit_TextUI_TestRunner extends PHPUnit_Runner_BaseTestRunner
             $codeCoverageReports = 0;
         }
 
+        $codeCoverage = null;
         if ($codeCoverageReports > 0) {
             $codeCoverage = new PHP_CodeCoverage(
                 null, $this->codeCoverageFilter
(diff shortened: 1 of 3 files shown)


henriquemoody:
@sebastianbergmann, it has been a while since the last comment on this pull request.

Looks like 4.6 has been released so this change will only be available (if accepted) on 6.7 now...

@derickr said there is no problem with ticks declaration and I have already explained about `pcntl_signal_dispatch()`, also I have made benchmarks (they are in the description now) often I rebase it to be up to merge so what is missing here?

sebastianbergmann:
PHPUnit 4.6 was not released but you are right that if this were merged now then it would go into PHPUnit 4.7. However, I do not really want to merge this. The negative implication of activating ticks on performance outweighs the benefit IMHO.

henriquemoody:
Got it.

PR #1516, opened 2014-11-27
add support for cygwin

BreiteSeite:
When running in cygwin, `__DIR__` resolves to vendor/bin/ and therefore phpunit can't find the autoload.php from composer.


sebastianbergmann:
Duplicate of #1613.

Issue #719, opened 2012-11-15
--debug flag has no effect when using --stderr


wenslayer:
In trying to work around the "headers already sent" problem, I'm using --stderr to get my tests to pass. Unfortunately, using --stderr seems to disable the --debug flag I'm also passing to see the details of which tests are running at a given point.

I couldn't find any mention that these flags were incompatible -- are they? If so, should the help be updated? If they should be compatible, perhaps this is a bug then?

Thanks for your help,
-Marc

euoia:
@edorian perhaps we should consider reopening this?

henriquemoody:
Which version are you using @euoia?  A lot of things changed after 302b465. Maybe this bug got back before ea1fea3.

Ps.: When #1550 get merged we can create PHPT tests to avoid this behaviour come back.

euoia:
@henriquemoody 
```
phpunit --version
PHPUnit 4.4.1 by Sebastian Bergmann.
```

whatthejeff:
@henriquemoody @euoia I think the original issue only concerned the CLI options. Feel free to open a new issue, though.

PR #1537, opened 2014-12-19
Fix problem of `--stderr` with `--tap` or `--testdox` options

### LINKED ISSUE #1455: --testdox does not work with stderr="true" but with --stderr
```
$ cat phpunit.xml
<phpunit stderr="true"/>


$ phpunit --testdox Test         
PHPUnit 4.2.6 by Sebastian Bergmann.

Configuration read from /home/sb/phpunit.xml

.

Time: 77 ms, Memory: 4.25Mb

OK (1 test, 0 assertions)


$ cat phpunit.xml       
<phpunit stderr="false"/>


$ phpunit --testdox --stderr Test
PHPUnit 4.2.6 by Sebastian Bergmann.

Configuration read from /home/sb/phpunit.xml

 [x] One
```

henriquemoody:
I don't think this pull request is the best solution to fix #1455, but I'm sure it solves the problem.

Actually, the main problem, IMO, is because there is no centralized point to create printers instances. `PHPUnit_TextUI_Command` creates and configures printers objects but `PHPUnit_TextUI_TestRunner` too. 

I will be glad to create a Factory for these printers but it requires a quite of change.

PS.: I can not test this, because seems that PHPT does not recognize STDERR.

### DIFF
--- a/src/TextUI/Command.php
+++ b/src/TextUI/Command.php
@@ -470,12 +470,12 @@ class PHPUnit_TextUI_Command
                 break;
 
                 case '--tap': {
-                    $this->arguments['printer'] = new PHPUnit_Util_Log_TAP;
+                    $this->arguments['printer'] = 'PHPUnit_Util_Log_TAP';
                     }
                 break;
 
                 case '--testdox': {
-                    $this->arguments['printer'] = new PHPUnit_Util_TestDox_ResultPrinter_Text;
+                    $this->arguments['printer'] = 'PHPUnit_Util_TestDox_ResultPrinter_Text';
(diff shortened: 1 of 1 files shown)


henriquemoody:
I think I made a mistake here. Since it was a bug fix I had to sent it against `4.4` instead of `master`.
Can someone cherry-pick ea1fea3 to `4.4` and `4.5`?

BTW, this also fix #719 which should be reopened.

whatthejeff:
Alright, cherry-picked into `4.4` and `4.5`. Thanks for the heads up!

PR #1548, opened 2014-12-23
Create factory implementation for printers

henriquemoody:
This pull request centralizes creation of printers instances into a single point removing duplication from `PHPUnit_TextUI_Command` and `PHPUnit_TextUI_TestRunner`.
Now `PHPUnit_TextUI_Command` has nothing to do with printers, it only populates the arguments used by `PHPUnit_TextUI_TestRunner`.

This pays the technical debts of #1537.

### DIFF
--- a/src/TextUI/Command.php
+++ b/src/TextUI/Command.php
@@ -622,25 +622,6 @@ class PHPUnit_TextUI_Command
                 $this->handleBootstrap($phpunit['bootstrap']);
             }
 
-            /**
-             * Issue #657
-             */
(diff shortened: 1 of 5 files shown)


henriquemoody:
Is anything missing?

whatthejeff:
I haven't had a chance to review it in detail yet. Looks pretty good to me at a glance.Hopefully @sebastianbergmann or I will have time soon.

Thanks for the effort and patience!

henriquemoody:
After the plans of dropping support for older PHP versions and probably change a lot of things on PHPUnit, I think this is not necessary (actually it never really _necessary_). 

Should I close this?

PR #1550, opened 2014-12-24
Redirect STDERR to STDOUT when running PHPT tests

### LINKED ISSUE #1169: phpt runner doesn't look at STDERR
On 4.0.3 + HHVM master, this causes tests/TextUI/fatal.phpt and tests/Regression/GitHub/873.phpt to fatal. - from facebook/hhvm#1904

henriquemoody:
This is the current behaviour of the [run-tests.php](https://github.com/php/php-src/blob/master/run-tests.php) script which runs all entire PHP source test suite.

Fix #1169.

### DIFF
--- a/src/Extensions/PhptTestCase.php
+++ b/src/Extensions/PhptTestCase.php
@@ -108,6 +108,9 @@ class PHPUnit_Extensions_PhptTestCase implements PHPUnit_Framework_Test, PHPUnit
 
         $result->startTest($this);
 
+        // Redirects STDERR to STDOUT
+        $php->setUseStderrRedirection(true);
+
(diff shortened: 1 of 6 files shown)


henriquemoody:
Any thoughts about this?

henriquemoody:
@sebastianbergmann, it will be awesome if you give some review and perhaps merge it and #1552.

I'm using PHPT indeed, and those features are basic to use this kind of testing.

I know you're a busy person and I will understand if you don't want this on PHPUnit's core I just want some feedback about this (and my others pull requests too, if possible).

GrahamCampbell:
:(

PR #1558, opened 2014-12-31
Update license and copyright in all files

### LINKED ISSUE #1557: Files with licence headers
A new year is coming and I guess all PHPUnit's files are going to be updated because the copyright year.
Symfony has some different approach:

```php
/*
 * This file is part of the Symfony package.
 *
 * (c) Fabien Potencier <fabien@symfony.com>
 *
 */
```
Every file, in Symfony project, contains this in the header of the file. How about do the same in PHPunit?
According to my research there is no such restriction in [BSD-3-Clause](http://opensource.org/licenses/BSD-3-Clause) license, am I wrong?

henriquemoody:
Fix #1557

### DIFF
--- a/phpunit
+++ b/phpunit
@@ -1,38 +1,12 @@
 #!/usr/bin/env php
 <?php
-/* PHPUnit
+/*
+ * This file is part of PHPUnit.
  *
- * Copyright (c) 2001-2014, Sebastian Bergmann <sebastian@phpunit.de>.
- * All rights reserved.
+ * (c) Sebastian Bergmann <sebastian@phpunit.de>
  *
- * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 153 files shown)


phpunit
```diff
@@ -1,38 +1,12 @@
 #!/usr/bin/env php
 <?php
-/* PHPUnit
+/*
+ * This file is part of the PHPUnit.
```
GrahamCampbell: Surely `This file is part of PHPUnit.` rather than `This file is part of the PHPUnit.`?

henriquemoody: Agreed, I will change as soon as I can

henriquemoody:
I'm updating right now.

henriquemoody:
@sebastianbergmann was faster than me :stuck_out_tongue_closed_eyes:

PR #1572, opened 2015-01-09
British English: colour

franzholz:


GrahamCampbell:
This is confusing now having the flag in American english, but the description in British.

franzholz:
Should the flag option also be changed from 'color' into 'colour'?

GrahamCampbell:
We can't do that in 4.x.

whatthejeff:
I really don't see the value in changing a command line option that's been around since 2008. Seems a bit excessive considering most people likely understand the meaning regardless of the regional spelling.

franzholz:
Maybe it is a solution to silently (undocumented) keep the American writing as well.

whatthejeff:
I discussed this with @sebastianbergmann on gitter. We decided to stick with the current spelling since American English is more ubiquitous in the PHP world.

PR #1579, opened 2015-01-15
redefine instructions to checkout from source

pscheit:
add php-token-stream to checkout
add double quotes to windows bat file


sebastianbergmann:
PHPUnit 3.6 has not been maintained in years, won't merge.

pscheit:
okay, just for the reference if someone needs to install this from source and finds this info useful.

best regards
philipp

Issue #1582, opened 2015-01-20
Use class autoloading in assertClassHasAttribute


bigfoot90:
I'm having error ```PHPUnit_Framework_Exception: Argument #2 (No Value) of PHPUnit_Framework_Assert::assertClassHasAttribute() must be a class name```

```php
class ManyToOneTest extends \PHPUnit_Framework_TestCase
{
    public function testProperties()
    {
        $this->assertClassHasAttribute('nullable', 'Mapping\Annotations\ManyToOne');
    }
}
```

```php
namespace Mapping\Annotations;

final class ManyToOne extends Reference
{
    /**
     * @var boolean
     */
    public $nullable = false;
}
```

whatthejeff:
Closed via b925393.

PR #1583, opened 2015-01-20
Use class autoloading in assertClassHasAttribute

bigfoot90:
Solve https://github.com/sebastianbergmann/phpunit/issues/1582


whatthejeff:
:+1: Thanks!

PR #1606, opened 2015-02-06
Cache vendor directory on Travis

henriquemoody:
A similar change was made in #1556, but it was reverted because HHVM was using sudo.

Now, after 5985512, HHVM build is no longer using sudo, but the caching made in #1556 was not reapplied.

This pull request is sent against `4.5` branch in order to improve builds from `4.5`, `4.6` and `4.7` versions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,9 @@
 language: php
 
+cache:
+  directories:
+    - vendor
+
 php:
   - 5.3.3
   - 5.3


whatthejeff:
From what I remember, caching the `vendor` directory can create problems with dependency resolution. I think we can safely cache the archives in the home directory though.

/cc @stof @GrahamCampbell

GrahamCampbell:
Yeh, we shouldn't be caching the vendor folder.

henriquemoody:
Good to know. Can you elaborate, @GrahamCampbell?

whatthejeff:
@henriquemoody There's more detailed information here: phpspec/phpspec#595

henriquemoody:
Thanks @whatthejeff!

whatthejeff:
No problem :). I think we can safely cache the archives in `$HOME/.composer/cache`.

henriquemoody:
I will do that.

henriquemoody:
But then we have to remove `--prefer-source` will that be a problem?

GrahamCampbell:
> But then we have to remove --prefer-source will that be a problem?

No. We're only doing this to prevent downloading loads of zips. If we're caching them, we're not doing this. :)

PR #1607, opened 2015-02-06
Store cache of Composer's cache directory

henriquemoody:
Related to #1606.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,9 @@
 language: php
 
+cache:
+  directories:
+    - $HOME/.composer/cache
+
 php:
   - 5.3.3
   - 5.3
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -20,7 +24,7 @@ before_install:
   - composer self-update
```
GrahamCampbell: We don't need this self-update either. Can it be removed @whatthejeff?

henriquemoody: Sorry, I've removed my fork because I thought it doesn't need changes, if we have to remove this I can send another pull request.

whatthejeff:
Looks good to me. Thanks!

.travis.yml
whatthejeff: Yeah, let's just make a new PR and remove the self-update :)

whatthejeff: Sorry for the inconvenience, @henriquemoody.

henriquemoody: No problem.

henriquemoody: #1609 is ready.

Issue #1613, opened 2015-02-11
Composer PHPUnit can't find the autoload.php


rainphp:
I've just installed Laravel 5 and run composer install (composer 1.0.0-dev), which installed phpunit in vendor/bin/phpunit.

When I run vendor/bin/phpunit I get this message:
``` You need to set up the project dependencies using the following commands:```

The autoload.php file is in vendor/autoload.php.

PHPunit is searching for autoload.php in the wrong folder:
https://github.com/sebastianbergmann/phpunit/blob/4a9fb9c9d7f980b00da53eb809d82f9f1a9fe56b/phpunit#L16

This can be fixed by simply as adding ``` __DIR__ . '/../autoload.php' ``` into the array in the foreach.

**Comment by whatthejeff** [comment]
Which platform are you using? Sounds like this is possibly the same as #1516.

**Comment by rainphp** [comment]
I'm on OSX Yosemite. __DIR__ resolve as ``` vendor/bin ``` but that's correct because the files are in:
``` vendor/bin/phpunit ```
``` vendor/autoload.php ```

To resolve the issue in my project I added ``` __DIR__ . '/../autoload.php' ``` to the array.

**Comment by whatthejeff** [comment]
Is `vendor/bin/phpunit` not a symlink on your system? For me it looks like this:

~~~
$ ls -lo vendor/bin/phpunit 
lrwxr-xr-x  1 jeff  26 Feb 11 18:11 vendor/bin/phpunit -> ../phpunit/phpunit/phpunit
~~~

**Comment by rainphp** [comment]
Not a symlink.

```
$ ls -lo vendor/bin/phpunit 
-rwxr-xr-x  1 federico  960 Feb 11 17:45 vendor/bin/phpunit
```

**Comment by rainphp** [comment]
FYI this is the composer.json file:
```
{
	"name": "laravel/laravel",
	"description": "The Laravel Framework.",
	"keywords": ["framework", "laravel"],
	"license": "MIT",
	"type": "project",
	"require": {
		"laravel/framework": "5.0.*"
	},
	"require-dev": {
		"phpunit/phpunit": "~4.0",
		"phpspec/phpspec": "~2.1"
	},
	"autoload": {
		"classmap": [
			"database"
		],
		"psr-4": {
			"App\\": "app/"
		}
	},
	"autoload-dev": {
		"classmap": [
			"tests/TestCase.php"
		]
	},
	"scripts": {
		"post-install-cmd": [
[...]

Issue #1433, opened 2014-09-13
phpunit/phpunit sets global $file


cmb69:
The [Fix](https://github.com/sebastianbergmann/phpunit/commit/dbc451e7381e6532f2ce904b500735d2593f7ba4) introduced a regression bug, because it sets the global variable $file, what will cause unit tests relying on this global to be falsy to fail.

I suggest to unset $file in line 44 in the current stable branch.

Sorry for not making a pull request, but I have no experience with Git (still stuck with SVN).

PR #1615, opened 2015-02-14
Fix phpunit not searching all places for composer's autoload

hojgr:
In my installation of latest Laravel (5) the phpunit bin script is located in vendor/bin/phpunit. When it looks for composer's autoload.php it fails because it doesn't look for it in ../

phpunit bash script location - PROJECT_ROOT/vendor/bin/phpunit
autoload.php location - PROJECT_ROOT/vendor/autoload.php

After this change it searches 1 level above too solving my issue

### DIFF
--- a/phpunit
+++ b/phpunit
@@ -13,7 +13,14 @@ if (!ini_get('date.timezone')) {
     ini_set('date.timezone', 'UTC');
 }
 
-foreach (array(__DIR__ . '/../../autoload.php', __DIR__ . '/../vendor/autoload.php', __DIR__ . '/vendor/autoload.php') as $file) {
+$directories = array(
+    __DIR__ . '/../../autoload.php',
+    __DIR__ . '/../vendor/autoload.php',
+    __DIR__ . '/vendor/autoload.php',
+    __DIR__ . '/../autoload.php'
+);
+
+foreach ($directories as $file) {
     if (file_exists($file)) {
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
This is targeted at the wrong branch.

> In my installation of latest Laravel (5) the phpunit bin script is located in vendor/bin/phpunit. When it looks for composer's autoload.php it fails because it doesn't look for it in ../

I can't replicate this problem.

whatthejeff:
@GrahamCampbell I haven't had the chance to look into it yet, but I think this happens when the binary isn't symlinked.

phpunit
```diff
@@ -13,7 +13,14 @@ if (!ini_get('date.timezone')) {
     ini_set('date.timezone', 'UTC');
 }
 
-foreach (array(__DIR__ . '/../../autoload.php', __DIR__ . '/../vendor/autoload.php', __DIR__ . '/vendor/autoload.php') as $file) {
+$directories = array(
```
whatthejeff: Global variables need to be unset or they can break existing tests (see #1433).

hojgr:
@whatthejeff fixed
@GrahamCampbell as @whatthejeff said the binary is not symlinked. Not sure how to replicate it

sebastianbergmann:
Duplicate of #1613.

PR #1625, opened 2015-02-23
Autoloading of Functions into `composer.json`

Ocramius:


Ocramius:
Note that I'm not sure if this was intended for user-facing API, but I found myself in a project using these functions after manually requiring them, which is really odd.

PR #1626, opened 2015-02-23
Feature percentageskip

blockjon:
Sometimes PHPUnit fatals before the end of a long test run. By using "--percentage-skip 50", you can skip the first 50% of the tests for example allowing the developer to continue the test run from before a problem area.

Example:
http://screencast.com/t/hG8goERc

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -547,7 +547,7 @@ abstract class PHPUnit_Framework_TestCase extends PHPUnit_Framework_Assert imple
      * Skips the current test if --percentage-skip was used and the current PHPUnit test run is not the required
      * percentage complete.
      */
-    protected function checkPercentageSkip()
+    private function checkPercentageSkip()
     {
         /**
          * @var $result \PHPUnit_Framework_TestCase
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -553,7 +553,13 @@ abstract class PHPUnit_Framework_TestCase extends PHPUnit_Framework_Assert imple
          * @var $result \PHPUnit_Framework_TestCase
          */
         $result = $this->getTestResultObject();
+        if (!is_object($result)) {
+            return;
+        }
         $topTestSuite = $result->topTestSuite();
+        if (!is_object($topTestSuite)) {
+            return;
+        }
         $numTestsInTotal = $topTestSuite->count(true);
         $numTestsRun = $topTestSuite->getNumTestsRun();
         $percentageSkip = $result->getPercentageSkip();
@@ -843,7 +849,10 @@ abstract class PHPUnit_Framework_TestCase extends PHPUnit_Framework_Assert imple
             }
         }
 
-        $this->getTestResultObject()->topTestSuite()->incNumTestsRun();
+        if (is_object($this->getTestResultObject()) && is_object($this->getTestResultObject()->topTestSuite())) {
+            $topTestSuite = $this->getTestResultObject()->topTestSuite();
+            $topTestSuite->incNumTestsRun();
+        }
 
         // Workaround for missing "finally".
         if (isset($e)) {
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -544,7 +544,8 @@ abstract class PHPUnit_Framework_TestCase extends PHPUnit_Framework_Assert imple
     }
 
     /**
-     * @since Method available since Release 3.6.0
+     * Skips the current test if --percentage-skip was used and the current PHPUnit test run is not the required
+     * percentage complete.
      */
     protected function checkPercentageSkip()
     {
@@ -555,11 +556,11 @@ abstract class PHPUnit_Framework_TestCase extends PHPUnit_Framework_Assert imple
         $topTestSuite = $result->topTestSuite();
         $numTestsInTotal = $topTestSuite->count(true);
         $numTestsRun = $topTestSuite->getNumTestsRun();
-        $skipPercentage = $result->getPercentageSkip();
+        $percentageSkip = $result->getPercentageSkip();
         $percentageOfTestsExecuted = floor(($numTestsRun / $numTestsInTotal) * 100);
 
-        if ($skipPercentage > $percentageOfTestsExecuted) {
-            $this->markTestSkipped("Skipped by jon");
+        if ($percentageSkip > $percentageOfTestsExecuted) {
+            $this->markTestSkipped("Skipped because --percentage-skip is set to $percentageSkip but the current PHPUnit run is only $percentageOfTestsExecuted complete.");
         }
     }
 
@@ -737,7 +738,6 @@ abstract class PHPUnit_Framework_TestCase extends PHPUnit_Framework_Assert imple
         $hookMethods = PHPUnit_Util_Test::getHookMethods(get_class($this));
 
         try {
-
             $hasMetRequirements = false;
             $this->checkRequirements();
             $this->checkPercentageSkip();
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -544,6 +544,26 @@ abstract class PHPUnit_Framework_TestCase extends PHPUnit_Framework_Assert imple
     }
 
     /**
+     * @since Method available since Release 3.6.0
+     */
+    protected function checkPercentageSkip()
+    {
+        /**
+         * @var $result \PHPUnit_Framework_TestCase
+         */
+        $result = $this->getTestResultObject();
+        $topTestSuite = $result->topTestSuite();
+        $numTestsInTotal = $topTestSuite->count(true);
+        $numTestsRun = $topTestSuite->getNumTestsRun();
+        $skipPercentage = $result->getPercentageSkip();
+        $percentageOfTestsExecuted = floor(($numTestsRun / $numTestsInTotal) * 100);
+
+        if ($skipPercentage > $percentageOfTestsExecuted) {
+            $this->markTestSkipped("Skipped by jon");
+        }
+    }
+
+    /**
      * Returns the status of this test.
      *
      * @return integer
@@ -717,8 +737,10 @@ abstract class PHPUnit_Framework_TestCase extends PHPUnit_Framework_Assert imple
         $hookMethods = PHPUnit_Util_Test::getHookMethods(get_class($this));
 
         try {
+
             $hasMetRequirements = false;
             $this->checkRequirements();
+            $this->checkPercentageSkip();
             $hasMetRequirements = true;
 
             if ($this->inIsolation) {
@@ -820,6 +842,8 @@ abstract class PHPUnit_Framework_TestCase extends PHPUnit_Framework_Assert imple
                 $e = $_e;
             }
         }
+
+        $this->getTestResultObject()->topTestSuite()->incNumTestsRun();
 
         // Workaround for missing "finally".
         if (isset($e)) {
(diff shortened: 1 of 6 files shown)


sebastianbergmann:
* Skipping tests based on a percentage feels weird to me
* Skipping the first N tests could be useful
* Either way, though, the implementation proposed does not take test dependencies into account (question is: should it?)
* Commit f43ea95 (changing the Composer package name) is, of course, unacceptable and must be removed from the pull request
* The test suite does not pass at the moment because the tests for `phpunit --help` were not updated

src/Framework/TestSuite.php
```diff
@@ -404,14 +420,19 @@ public function addTestFiles($filenames)
     /**
      * Counts the number of test cases that will be run by this test.
      *
+     * @Param boolean $preferCache Indicates if cache is preferred.
      * @return integer
      */
-    public function count()
+    public function count($preferCache = false)
```
sebastianbergmann: Is this introduction of a cache for the return value of `count()` related to the proposed feature? If not, this should be a separate pull request.

src/Framework/TestCase.php
```diff
@@ -544,6 +544,33 @@ protected function checkRequirements()
     }
 
     /**
+     * Skips the current test if --percentage-skip was used and the current PHPUnit test run is not the required
+     * percentage complete.
+     */
+    protected function checkPercentageSkip()
```
sebastianbergmann: Please make this `private`. By default, new non-public methods should be private. We can always open it up later to `protected` if need be but we cannot close it down from `protected` to `private`.

src/Framework/TestResult.php
```diff
@@ -22,6 +22,11 @@
 class PHPUnit_Framework_TestResult implements Countable
 {
     /**
+     * @var integer
+     */
+    protected $percentageSkip = 0;
```
sebastianbergmann: Please make this `private`. Potential child classes can use the `public` accessor and mutator methods the patch also introduces.

src/Framework/TestSuite.php
```diff
@@ -47,6 +47,20 @@
 class PHPUnit_Framework_TestSuite implements PHPUnit_Framework_Test, PHPUnit_Framework_SelfDescribing, IteratorAggregate
 {
     /**
+     * Tracks number of tests run.
+     *
+     * @var integer
+     */
+    protected $numTestsRun = 0;
```
sebastianbergmann: Please make this `private`.

src/Framework/TestSuite.php
```diff
@@ -47,6 +47,20 @@
...
+
+    /**
+     * Last count of tests in this suite.
+     *
+     * @var integer|null
+     */
+    protected $cachedNumTests;
```
sebastianbergmann: Please make this `private`.

src/Framework/TestSuite.php
```diff
@@ -131,6 +145,8 @@ class PHPUnit_Framework_TestSuite implements PHPUnit_Framework_Test, PHPUnit_Fra
      */
     public function __construct($theClass = '', $name = '')
     {
+        $this->numTestsRun = 0;
```
sebastianbergmann: This assignment is superfluous as the attribute is initialized to this value upon declaration.

spriebsch:
Personally, I think you might be throwing too much solution at the problem. How about just skipping the test(s) that make(s) PHPUnit crash?

spriebsch:
If skipping is not viable, how about adding a feature to PHPUnit that tracks which test(s) crash, and have an option to automagically skip those? I am thinking about an optional feature where PHPUnit writes a log about executed tests (similar to a DB transaction log). If it logs each test that is started and ended, then it if it crashes, since it cannot log that that test has ended, on a subsequent run it could figure out which tests has caused the crash and just skip it.

sebastianbergmann:
* I think that @spriebsch's suggestion for (optionally) keeping track of which test is currently being executed in a temporary file for debugging (finding out quickly which test triggers a fatal error) makes sense
* The more I think about it the more I lean towards not wanting to support skipping based on a percentage or number of tests
* It may be useful to add support for starting the execution of a test suite based on the name of the first test is to be run (ie. skip tests until a test of a given name is to be executed)

src/Framework/TestSuite.php
blockjon: Ok. I have moved this to https://github.com/sebastianbergmann/phpunit/pull/1627.

src/Framework/TestCase.php
blockjon: Ok done.

src/Framework/TestResult.php
blockjon: Ok done.

src/Framework/TestSuite.php
blockjon: Ok done.

src/Framework/TestSuite.php
blockjon: Ok. I have removed this.

sebastianbergmann:
Thank you, @blockjon, for incorporating the feedback I provided. Can you comment on @spriebsch's feedback and my response to that? Thanks!

blockjon:
* Usually when a test fails from a thrown exception, PHPUnit provides the developer with useful information about the failed test and it proceeds by continuing with the test run. That's good.
* It is unfortunate that when PHP encounters a fatal error that PHPUnit tends to just immediately stop.
* By default, when PHPUnit encounters a fatal error, the whole test run just stops dead in its tracks without any useful information emitted into the console. This is probably because the PHP language dumps out to the registered shutdown handler and logs the fatal to the error log and then stops processing further PHP instructions.
* I feel as though this default behavior is very confusing to new PHPUnit users. Reasonable engineers easing into PHP & PHPUnit will have an expectation to have PHPUnit elegantly handle this scenario and continue if possible just like how it handles a regular thrown exception. 
* The following code encountered during a PHPUnit run causes PHPUnit to crash with a fatal error:
```
$x = null;
$x->foo();
```
* My project has thousands of tests. When I encounter a fatal exception after having waited 15 minutes, the last thing I usually see on screen looks like this:
```
......................... (97%)
......................... (98%)
......................... (98%)
........  <- Fatal exception happens here but the user has no idea what caused it
>
```
In this scenario, I really don't want to run thousands of tests again. If all I know it that something stopped somewhere after 98%, I want some way to quickly identify the PHPUnit test file and test function name which died so I can triage the problem. Once fixed, I would like to continue the test run from around 98% to ensure the remainder of the test run is smooth.
* In my main codebase, we added a call to this subroutine during our setUp() function:
```
 protected function commonSetupActivities()
 {
  $reflector = new ReflectionClass($this);
  $GLOBALS['phpunit_last_test_name'] = $this->getName();
  $GLOBALS['phpunit_last_test_file'] = $reflector->getFileName();
  $GLOBALS['phpunit_last_test_class'] = get_class($this);
 }
```
In our registered shutdown function, we have code that looks like this https://gist.github.com/blockjon/fb3682acfc9db76ae8f7

blockjon:
@spriebsch If there were some way to keep track of all of the tests which generated a fatal crash, that would be great. I think the problem though is that a fatal causes the PHPUnit run to stop. That means that at most, we can only capture one fatal. It is as if an option called --stop-on-fatal is always enabled which of course nobody would want. Do you know of some way to keep the test run going even after a fatal is encountered in order to capture *all* of the crashes in a crash report?

sebastianbergmann:
@blockjon If you run your test suite in process isolation then each test will be executed in a separate PHP process. If that child process exits with a fatal error this will be recorded and the execution of the test suite continues.

That being said, using process isolation is prone to creating more problems than solving them.

spriebsch:
@blockjon I know this is gonna *sound* not helpful, but in fact I'm trying to be. So please bear with me as I'm asking a few questions:

What is causing those crashes in the first place?

If syntax errors are the problem, did you consider running a lint check on the codebase before executing tests?

You also might want to consider using the @group annotation to ease selectively executing only parts of the testsuite.

Generally, are you using a feature like --filter? I'm assuming that tests break where you have worked on the code, so you could selectively run only the relevant tests first.

My apologies if you feel that I am taking this thread off topic. I'm trying to help you, and my gut feeling tells me that you are trying to solve a problem with this patch that should be solved somewhere else. Feel free to email me at stefan@thephp.cc if you want to continue this discussion (cc sebastian@thephp.cc to keep him in the loop).

blockjon:
@spriebsch At our shop, we have thousands of tests, hundreds of test files, and dozens of developers of writing tests. We use --filter often and we use @group occasionally in order to target specific tests while developing a feature.

The problem we are trying to solve for is not related to syntax.

Some of what the devs do is more than a traditional unit test... we have many complex functional tests which exercise the interaction between classes. Here's what the scenario looks like that I'm trying to improve:

1) Dev creates a feature branch
2) Dev develops feature and a set of PHPUnit tests to support the behaviors behind the feature
3) Finally, in order to verify all previous app features still work, the dev runs all of the PHPUnit tests with something like "phpunit ."

Naturally, a dev will occasionally accidentally break something else in the codebase. Sometimes, the faulty code leads to a function call on a non-object value. I don't think a linter can reliably find this. In this contrived example, PHP will crash with a fatal... sometimes after 15 minutes of waiting for the tests to pass. I'm trying to save a developer's time so they are not forced to rerun 15 minutes of tests in order to triage the problem area of a huge test run.

blockjon:
Assume you see a fatal after 99%, take a look at this screencast and maybe you can get a sense of how this is useful in order to continue from where you left off:

http://screencast.com/t/hG8goERc

spriebsch:
Thanks for creating that screencast. I have to admit that I may (still) not fully understand your use case: the failing test is the one you would want to fix first, right? But after that, given the fact that "a dev will occasionally accidentally break something else" you will have to rerun *all* tests. First up, my impression was that you wanted to use the "skip percentage" feature to focus on certain tests, rather than executing all of them again. Now I'm not so sure about that any more. Maybe you can clarify, thanks!

blockjon:
1) A dev waits 20 minutes to run PHPUnit. The screen fatals with something like that looks like this just as it's almost finished:
```
......................... (97%)
......................... (98%)
......................... (98%)
....
```
2) Now the dev has no idea what happened. The PHP error log only shows you which source code file and line number the fatal exception was detected in. *The dev has no idea which test is broken*.
3) The dev is forced to do a full run of PHPUnit again with something like "--debug" and is now *forced to watch the screen for 20 more minutes* re-doing everything we already know is healthy in order to get to the problem area.

* Since the only useful piece of information is "98%", I would much rather run phpunit with something like "phpunit --debug --percentage-skip 98" in order to get it to start running from the last known healthy area so I don't have to wait 20 minutes running tests which I already know work.

sebastianbergmann:
Just for completeness' sake: it's not just `(98%)`, though, but rather `XXXXX / YYYYY ( 98%)`. And `XXXXX` plus (manually counted dots on the next line) will give you the number of the test at which you want to start.

I will think about this some more. I am sure we can find a good solution for the problem.

sebastianbergmann:
Quick idea: if the TAP logger flushes after each test then you could see which test has the failure. This should save you the 20 minutes you're interested in.

blockjon:
Shouldn't PHPUnit register a shutdown function which observes if there is a fatal exception that has been thrown and then emit the unit test file and test name which has failed to the console? We wrote that by hand in our child classes but I'm pretty sure that's what everybody would want who uses PHPUnit.

sebastianbergmann:
Shutdown function will not help when there is a segfault.

blockjon:
Correct ... shutdown function will not help when there is a segfault but I believe it would be helpful in almost every other fatal scenario I have come across.

PR #1627, opened 2015-02-24
Add option to cache the count of tests since the strategy for the count() function is not efficient.

blockjon:
Created this PR in response to https://github.com/sebastianbergmann/phpunit/pull/1626/files#r25231977

I have a new PHPUnit feature branch which uses this function often during the PHPUnit run. Without this technique, my test run goes significantly slower.


sebastianbergmann:
Thanks!

Issue #1407, opened 2014-08-22
test fails with hhvm and @runInSeparateProcess


NeoBlack:
Hi,
I has a problem with phpunit on travis in hhvm env.
I think before copy all output here, take a look at the output on travis:

https://travis-ci.org/PhileCMS/Phile/jobs/33320542

I am not sure if this is a problem with phpunit or hhvm, with 5.4 and 5.5 the tests running fine.

Issue #1214, opened 2014-04-07
hhvm - tests marked @runInSeparateProcess fail


mugwhump:
Hello,

I know @runInSeparateProcess caused tests to fail in phpunit 3.7 because hhvm needs to be called with certain options to run code from stdin (see #1090). Since 4.0 officially supports hhvm I thought I'd try again. I ran the following test:

``` php
<?php

class CSeparateProcessTest extends PHPUnit_Framework_TestCase 
{
     /**
      * @runInSeparateProcess
      */
    public function testSeparateProcess() {
        $this->assertTrue(true);
    }
}
```

It failed with this:

```
$ hhvm phpunit-4.0.13/phpunit.phar CSeparateProcessTest.php 
PHPUnit 4.0.13 by Sebastian Bergmann.

E

Time: 716 ms, Memory: 9.10Mb

There was 1 error:

1) CSeparateProcessTest::testSeparateProcess
PHPUnit_Framework_Exception: PHPUnit 4.0.13 by Sebastian Bergmann.

Usage: phpunit [options] UnitTest [UnitTest.php]
       phpunit [options] <directory>

Code Coverage Options:
**followed by the rest of phpunit's help text**

Caused by
ErrorException: Unable to unserialize: [PHPUnit 4.0.13 by Sebastian Bergmann.

**help text again**

]. Expected ':' but got 'H'.


FAILURES!
Tests: 1, Assertions: 0, Errors: 1.
```

It's still failing, but now it's spitting out phpunit's no-argument helptext instead of hhvm's no-argument text (Notice: Nothing to do etc).

EDIT: upon further investigation this seems to be an issue with my installation. Calling phpunit from the phar doesn't work, but it worked after installing phpunit with composer.

DOUBLE-EDIT: actually maybe this is a bug. I would assume it's not supposed to fail when running the phar directly. I'll let you decide whether to close this.

Issue #1090, opened 2014-01-04
[hhvm] Process isolation does not work


Maks3w:
I've open this issue just for track the progress in the adoption of HHVM

When use process isolation or the annotation `@runInSeparateProcess` HHVM fails with the following message

> PHPUnit_Framework_Exception: HipHop Notice: Nothing to do. Either pass a .php file to run, or use -m server

PR #1631, opened 2015-03-01
Bump the php-code-coverage requirement

duncan3dc:
php-code-coverage:2.0.11 was the first version to require php-token-stream:~1.3 which had a fix for the `@runInSeparateProcess` on hhvm (#1090, #1214, #1407)

This was discovered by running tests using the composer option of `--prefer-lowest`.
So at the moment projects that rely on `@runInSeparateProcess` and test on hhvm with `--prefer-lowest` have to add another dependency of `php-code-coverage:2.0.11` or `php-token-stream:~1.3` to the project.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
         "php": ">=5.3.3",
         "phpunit/php-file-iterator": "~1.3",
         "phpunit/php-text-template": "~1.2",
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
Cherry-picked into `4.5`.

duncan3dc:
Thanks @sebastianbergmann

Issue #1637, opened 2015-03-11
Feature: Assertion of Array Items by Inexact Match


sshymko:
## The Problem

PHPUnit does offer assertions for almost every occasion. However, this is not the case for arrays (or traversables to be more generic). Only few assertions work with array contents: `assertArrayHasKey` `assertContains`, `assertContainsOnly`, `assertContainsOnlyInstancesOf`. There are also their negative counterparts. But that's pretty much it, array item/key support ends here.

Existing assertions that target arrays allow only for *exact match*. However, there are cases when exact match would overly complicate a test or make it too fragile. Other cases are impossible to cover by exact matching.

Actually, the problem is even bigger - not only inexact match is unavailable for array items, but *arbitrary constraints* cannot be used to assert array items.

### Known Use Cases

Elaborate assertions of array items are fairly common. Some generic use cases are listed below.

#### Asserting Headers
Asserting that certain header is present among all headers. For example, a test may be interested in verifying the fact that redirection has been set, but not interested in matching exact target URL (to avoid test fragility). Unless matching exact header (case sensitive) is desirable, `assertContains` cannot be used here.

#### Asserting Filenames
Imagine a code that traverses file system and collects files/dirs. A test for such a code would contain fixture files/dirs. Asserting exact match would complicate such a test, because it would have to include a full fixture directory path. It also would encourage to dynamically build expected values, again, because of necessity to include fixture directory path.

Actually, asserting that files in array exist is currently impossible. It's not an example of inexact match, but rather unavailability of existing constraints (`PHPUnit_Framework_Constraint_FileExists` in this case) for array items.

#### Asserting Messages
It is typical for an application to collect error/warning/notification messages and display them to an end user. Of course, a test would be interested in verifying presence of a specific message. But also absence of error messages of any kind may need to be verified. If plain text messages are used, assertion would come down to verifying absence of such keywords as "error", "fail", etc.

The same is applicable to asserting log entries as well.

## Proposed Solution

The idea is to to allow matching array items/keys against any constraint, i.e. an instance of `PHPUnit_Framework_Constraint`. In such a way the whole variety of existing constrains becomes available for asserting array items.

### Proposed Syntax

At first, it appeared to make sense to name assertions around `array` keyword. More detailed analysis suggests more generalization to comply with traversable instances. Existing assertion `assertContains` seems to be a good prototype to borrow the naming from.

Usage example of `assertContainsMatching`:
```php
$actualHeaders = array(
[...]

Issue #1292, opened 2014-06-16
Decide what to do with assertTag() and assertSelect*().


whatthejeff:
These assertions are simply broken. For reference, see:

 * #1278
 * #1158
 * #995
 * #878
 
The implementation is also quite poor. Most notably:

 * The CSS selector parser is not proper.
   * It doesn't comply to any known CSS specification.
   * It doesn't specify which features it intends to support and which it doesn't.
   * It doesn't validate syntax, fail or report errors.
 * The node selecting is done by recursively walking through the DOM in userland which is inefficient when compared to something more conventional like XPath.
 * The CSS node selecting functionality is done in reverse which makes it effectively broken for combinators (see: #878).
 
My opinion is that the implementations for these features should be replaced by something like [symfony/DomCrawler](https://github.com/symfony/DomCrawler). Personally, I would like to see these features deprecated in core and ported to a 3rd-party plugin as the current APIs are cumbersome and unintuitive–but I understand that might not be practical.

whatda:
There haven't been any updates for months now... assertTag of whatthejeff's package is still not implemented (see https://github.com/phpunit/phpunit-dom-assertions/issues/3).

I think it was a bad idea to remove the assertions. They should rather have seen bugfixes or at least been ported to a separate package.

Our problem now is that we need to use an old version of PHPUnit, since we depend on assertTag and IMO there aren't any good alternatives.

Nevertheless, thanks for your efforts and have a nice day.

PR #1648, opened 2015-03-21
Assertion of array items/keys against arbitrary constraint

sshymko:
## Introduction

This is an improvement to the support of traversables by PHPUnit. New assertions allow testing of array items against an arbitrary constraint, i.e. an instance of `PHPUnit_Framework_Constraint`. Any existing constraint can be used to assert not only array items, but optionally also array keys.

The problem and the solution have been originally described in the feature request #1637.
### Assertions

Assert that a haystack contains a needle matching a constraint:

``` php
assertContainsMatching (
    PHPUnit_Framework_Constraint $needleConstraint , string $haystack [, string $message ]
)
```

Assert that a haystack contains only needles matching a constraint:

``` php
assertContainsOnlyMatching (
    PHPUnit_Framework_Constraint $needleConstraint , string $haystack [, string $message ]
)
```

Negative assertions `assertNotContainsMatching` and `assertNotContainsOnlyMatching` are provided as well.
### Factory Methods

The following constraint factory methods/functions are available:

``` php
PHPUnit_Framework_Constraint_TraversableContainsMatching containsMatching (
    PHPUnit_Framework_Constraint $itemConstraint [, PHPUnit_Framework_Constraint $keyConstraint ]
)
```

``` php
PHPUnit_Framework_Constraint_TraversableContainsMatchingOnly containsOnlyMatching (
    PHPUnit_Framework_Constraint $itemConstraint [, PHPUnit_Framework_Constraint $keyConstraint ]
)
```

The factory methods support optional constraint for array keys in addition to array items. It is intentionally omitted from the assertion interface to keep the syntax concise for the primary use case, that is asserting array items. Please refer to [Asserting Array Keys](#assertingArrayKeys) for the usage guidelines.
## Usage

The examples below demonstrate usage of the array assertions for scenarios that are otherwise impossible to implement with PHPUnit.
### Asserting Inexact Match

Asserting array items by inexact match can be handy for testing HTTP headers, for example:

``` php
$actualHeaders = array(
    'Date: Sat, 21 Mar 2015 07:13:31 GMT',
    'Content-Type: text/html; charset=utf-8',
    'Content-Length: 42137',
    // ...
);
$this->assertContainsMatching(
    $this->stringStartsWith('Content-Type: text/html'),
    $actualHeaders,
    'HTML page is expected.'
);
```
### Asserting Files

Existing constraint `PHPUnit_Framework_Constraint_FileExists` can be particularly useful to test an array of filenames, for instance:

``` php
$actualFilenames = array(
    '/project/tests/_files/fixture/path/1.txt',
    '/project/tests/_files/fixture/path/2.txt',
    '/project/tests/_files/fixture/path/sub/dir/3.txt',
    // ...
);
$this->assertContainsOnlyMatching(
    $this->fileExists(),
    $actualFilenames,
    'Only existing files are expected.'
);
```
### Asserting Multi-Dimensional Arrays

The syntax comes with the flexibility of constraining array items on any nesting level, for example:

``` php
$actual = array(
    array('a11', 'a12', 'a13'),
    array('a21', 'a22', 'a23'),
    // ...
);
$this->assertContainsOnlyMatching(
    $this->containsOnlyMatching(
        $this->matchesRegularExpression('/^a\d+$/')
    ),
    $actual
);
```

Note that `containsMatching/containsOnlyMatching` can be recursively nested as much as needed to cover array of any depth.
### <a name="assertingArrayKeys"></a>Asserting Array Keys

The constraints implementing the array assertions go beyond validating array items. In particular, optional asserting of array keys is available. It can be used in the following way:

``` php
$actual = array(
    '/project/tests/_files/fixture/path/1.txt' => 'File contents 1',
    '/project/tests/_files/fixture/path/2.txt' => 'File contents 2',
    '/project/tests/_files/fixture/path/sub/dir/3.txt' => 'File contents 3',
    // ...
);
$this->assertThat(
    $actual,
    $this->containsOnlyMatching(
        $this->logicalNot($this->isEmpty()), // item constraint
        $this->stringEndsWith('.txt') // key constraint
    ),
    'Only non-empty text files are expected.'
);
```

Moreover, constraint of array items can be completely dismissed by using `$this->anything()` in place of the required argument. In such a way it is possible to assert solely array keys.

### DIFF
--- a/src/Framework/Assert.php
+++ b/src/Framework/Assert.php
@@ -398,6 +398,25 @@ abstract class PHPUnit_Framework_Assert
     }
 
     /**
+     * Asserts that a haystack does not contain a needle matching a constraint.
+     *
+     * @param PHPUnit_Framework_Constraint $needleConstraint
+     * @param string                       $haystack
+     * @param string                       $message
+     */
+    public static function assertNotContainsMatching(PHPUnit_Framework_Constraint $needleConstraint, $haystack, $message = '')
+    {
+        if (!(is_array($haystack) || is_object($haystack) && $haystack instanceof Traversable)) {
+            throw PHPUnit_Util_InvalidArgumentHelper::factory(
+                2,
+                'array or traversable'
+            );
+        }
+
+        self::assertThat($haystack, self::logicalNot(self::containsMatching($needleConstraint)), $message);
+    }
+
+    /**
      * Asserts that a haystack contains only needles matching a constraint.
      *
      * @param PHPUnit_Framework_Constraint $needleConstraint
@@ -414,6 +433,25 @@ abstract class PHPUnit_Framework_Assert
         }
 
         self::assertThat($haystack, self::containsOnlyMatching($needleConstraint), $message);
+    }
+
+    /**
+     * Asserts that a haystack does not contain only needles matching a constraint.
+     *
+     * @param PHPUnit_Framework_Constraint $needleConstraint
+     * @param string                       $haystack
+     * @param string                       $message
+     */
+    public static function assertNotContainsOnlyMatching(PHPUnit_Framework_Constraint $needleConstraint, $haystack, $message = '')
+    {
+        if (!(is_array($haystack) || is_object($haystack) && $haystack instanceof Traversable)) {
+            throw PHPUnit_Util_InvalidArgumentHelper::factory(
+                2,
+                'array or traversable'
+            );
+        }
+
+        self::assertThat($haystack, self::logicalNot(self::containsOnlyMatching($needleConstraint)), $message);
     }
 
     /**
--- a/src/Framework/Assert.php
+++ b/src/Framework/Assert.php
@@ -379,6 +379,44 @@ abstract class PHPUnit_Framework_Assert
     }
 
     /**
+     * Asserts that a haystack contains a needle matching a constraint.
+     *
+     * @param PHPUnit_Framework_Constraint $needleConstraint
+     * @param string                       $haystack
+     * @param string                       $message
+     */
+    public static function assertContainsMatching(PHPUnit_Framework_Constraint $needleConstraint, $haystack, $message = '')
+    {
+        if (!(is_array($haystack) || is_object($haystack) && $haystack instanceof Traversable)) {
+            throw PHPUnit_Util_InvalidArgumentHelper::factory(
+                2,
+                'array or traversable'
+            );
+        }
+
+        self::assertThat($haystack, self::containsMatching($needleConstraint), $message);
+    }
+
+    /**
+     * Asserts that a haystack contains only needles matching a constraint.
+     *
+     * @param PHPUnit_Framework_Constraint $needleConstraint
+     * @param string                       $haystack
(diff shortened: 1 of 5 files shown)


sshymko:
@sebastianbergmann 
This PR has been completely ready with tests and everything for a month already.
IMO, it contributes the only significant feature that is missing in PHPUnit.
Could you please take a look at it.
Thanks.

sshymko:
It's been 3 months and no updates whatsoever...
@sebastianbergmann, could you please accept or dismiss the PR.
Thank you.

orlangur:
How is that useful in core? From my experience even existing array assertions are not so useful compared to exact matching with `assertEquals` and such. 

In your examples tests are simply not reproducible due to improper fixtures. Proper test isolation would allow to perform exact assertions.

To avoid situation like https://github.com/sebastianbergmann/phpunit/issues/1292#issuecomment-46163048 I believe the right place for such doubtful PHPUnit API extension would be separate repository similar to https://github.com/phpunit/phpunit-dom-assertions

sebastianbergmann:
I did not have time to look at / think about the proposed assertions when the ticket was opened. It appeared on my radar again only after the comment in April. And to be honest, I stopped reading after "it contributes the only significant feature that is missing in PHPUnit". You are, of course, entitled to this opinion and I understand that these assertions are important to you.

I am not convinced that these assertions add enough value to be added to PHPUnit. Why don't you create a separate project that offers them in the form of a trait, for instance? Not everything needs to be in PHPUnit's "core".

sshymko:
@sebastianbergmann,
Thank you for getting back to me.
The statement "only significant feature that is missing in PHPUnit" referred to completeness of assertions. It comes from the assumption that any existing constraint may potentially be useful to assert array items/keys. That's what this contribution was about.
I will consider re-implementing it as a small separate project, although the chances of me working on it are way lower than 3 months ago when the PR was submitted.
Thank you.

sshymko:
@sebastionbergmann,
One last thing before abandoning this thread.
I completely agree with minimalism of assertions along with the fact that PHPUnit is already somewhat overloaded with rarely used assertions.
However, IMO what differentiates an excessive assertion from the true "core" one is whether it is generic enough to simplify the framework. The proposed assertions replace the following assertions:
assertContains
assertContainsInstanceOf
assertArrayHasKey

sshymko:
@orlangur,
Thank you for the link to an example of the project adding assertions to PHPUnit.
Yes, it can be done as a separate small project.
However, the decision from the PHPUnit team was necessary to take this route.
The only significant downside of such a side project is inability to embed new assertions into `PHPUnit_Framework_TestCase`.
New test case class' descendant needs to be introduced and it greatly limits flexibility of the test cases inheritance.
It will be impossible to use new assertions in tests that inherit the test case class `PHPUnit_Framework_TestCase` indirectly.
For example, the tests derived from `PHPUnit_Extensions_Selenium2TestCase` and the like.

bishopb:
+1 for the good work, definitely helpful in the use cases you've mentioned...
and +1 for keeping these in separate repo.

PR #1649, opened 2015-03-21
add refuteX syntax to alias assertNotX

jaredh159:
@sebastianbergmann this is per our brief twitter exchange.

This is just code to add an alias for one method: assertNotSame(). Before I went through and did all of the inverse assertions, I wanted to make sure you were OK with this approach or if you wanted me to modify how I'm doing it.

If this looks good, I'll go through and add the refuteX methods for all of the inversions.

### DIFF
--- a/src/Framework/Assert.php
+++ b/src/Framework/Assert.php
@@ -117,6 +117,19 @@ abstract class PHPUnit_Framework_Assert
     }
 
     /**
+     * Refutes that an array has a specified key. Alias of self::assertArrayNotHasKey()
+     *
+     * @param mixed             $key
+     * @param array|ArrayAccess $array
+     * @param string            $message
+     * @since  Method available since Release 4.7.0
+     */
+    public static function refuteArrayHasKey($key, $array, $message = '')
+    {
+        return self::assertArrayNotHasKey($key, $array, $message);
+    }
+
+    /**
      * Asserts that a haystack contains a needle.
      *
      * @param mixed   $needle
@@ -213,6 +226,22 @@ abstract class PHPUnit_Framework_Assert
         }
 
         self::assertThat($haystack, $constraint, $message);
+    }
+
+    /**
+     * Refutes that a haystack contains a needle.
+     *
+     * @param mixed   $needle
+     * @param mixed   $haystack
+     * @param string  $message
+     * @param boolean $ignoreCase
+     * @param boolean $checkForObjectIdentity
+     * @param boolean $checkForNonObjectIdentity
+     * @since  Method available since Release 4.7.0
+     */
+    public static function refuteContains($needle, $haystack, $message = '', $ignoreCase = false, $checkForObjectIdentity = true, $checkForNonObjectIdentity = false)
+    {
+        return self::assertNotContains($needle, $haystack, $message, $ignoreCase, $checkForObjectIdentity, $checkForNonObjectIdentity);
     }
 
     /**
--- a/src/Framework/Assert.php
+++ b/src/Framework/Assert.php
@@ -1208,6 +1208,20 @@ abstract class PHPUnit_Framework_Assert
     }
 
     /**
+     * Refutes that two variables have the same type and value.
+     * Used on objects, it refutes that two variables reference
+     * the same object. Alias for self::assertNotSame().
+     *
+     * @param mixed  $expected
+     * @param mixed  $actual
+     * @param string $message
+     */
+    public static function refuteSame($expected, $actual, $message = '')
+    {
+        return call_user_func_array('self::assertNotSame', func_get_args());
+    }
+
+    /**
      * Asserts that a variable and an attribute of an object do not have the
      * same type and value.
      *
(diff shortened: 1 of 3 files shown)


src/Framework/Assert.php
```diff
@@ -1208,6 +1208,20 @@ public static function assertNotSame($expected, $actual, $message = '')
...
+     * Used on objects, it refutes that two variables reference
+     * the same object. Alias for self::assertNotSame().
+     *
+     * @param mixed  $expected
+     * @param mixed  $actual
+     * @param string $message
+     */
+    public static function refuteSame($expected, $actual, $message = '')
+    {
+        return call_user_func_array('self::assertNotSame', func_get_args());
```
sebastianbergmann: Why not simply use `self::assertNotSame($expected, $actual, $message)` instead of using `call_user_func_array()`?

jaredh159: When I'm aliasing a function/method, I usually do it this way because it is future-compatible with any argument signature changes.  That said, I don't see these arguments changing much in the future, so if you prefer I call it directly, I'm totally fine with that.  Just let me know.

jaredh159: The other thing I considered was replicating the method body directly to prevent another line in any stack trace, but that didn't seem very DRY.

sebastianbergmann: Yes, please. Thanks!

jaredh159:
@sebastianbergmann removed call_user_func_array, as requested.  That's cleaner, I agree. Let me know if I'm OK to rip through the rest of the assertions, or if you have other thoughts, I'm all ears.

By the way, we really love PHPUnit.  I can't thank you enough for all the hard work you've put in over the years. It is truly appreciated. It's a great tool.

sebastianbergmann:
Go ahead and prepare a complete pull request. Please also update the `ChangeLog-4.7.md`.

jaredh159:
@sebastianbergmann sorry, another question.  I started working through these and realized I was basically copy-pasting tons of lines of codes to duplicate all of the test cases for the new refuteX- style assertions. See the above commit for an example. Are you OK with this type of non-DRY duplication in the AssertTest.php test suite? Or does that seem like a code smell? I'm not sure how I feel and wanted your direction before I continue.

I figure I could a) just barrel on and replicate all of the tests, or, b) I could not write new tests for the aliases and just very carefully check that I passed through all of the arguments correctly, or c) somehow re-write all of the tests that use assertNotX() in a DRY way to test both refuteX and assertNotX().  

Thoughts?

jaredh159:
@sebastianbergmann did you have a chance to think about the code duplication issue I mentioned above?

sshymko:
How about eliminating code duplication like the following:
```php
    /**
     * @dataProvider refuteArrayHasStringKeyProvider
     */
    public function testRefuteArrayHasStringKey($assertionMethod)
    {
        $this->$assertionMethod('bar', array('foo' => 'bar'));
        try {
            $this->$assertionMethod('foo', array('foo' => 'bar'));
        } catch (PHPUnit_Framework_AssertionFailedError $e) {
            return;
        }
        $this->fail();
    }

    public function refuteArrayHasStringKeyProvider()
    {
        return array(
            'refute' => array('refuteArrayHasKey'),
            'assert' => array('assertArrayNotHasKey'),
        );
    }
```

At first glance the approach above does not seem to be very clean. However, reuse through delegation to a shared protected method does help with these kinds of tests that differ in a name of tested method only.

jaredh159:
@sshymko Thanks, yeah, I was thinking along the same lines.  I think that's probably the best approach.  The only downside I see is that it means I will have to create a bunch of dataProvider methods that just return the two method names, but that seems better than duplicating the tests.

I messed around for a few minutes to see if I could auto-generate those providers by setting the dataProvider to a non-existent method name and using __call() to hand back the assert/refute pairs based on the missing method name, but the dataProvider reflection system would need to be modified for that to work, and I'm guessing @sebastianbergmann would not want that.

sshymko:
@jaredh159 BTW, the idea to replace awkwardness of `assertNotX` with `refureX` is amazing! Even statements `assert` and `refute` are of the same length! Great job!

jaredh159:
@sebastianbergmann did you have a chance to look at the discussion above?

sebastianbergmann:
Not yet, sorry.

sebastianbergmann:
Looks good so far.

jaredh159:
@sebastianbergmann so you're OK with the approach outlined above by sshymko where I would use data providers to provide the two synonymous method names?

sebastianbergmann:
For testing the synonymous methods? Yes.

keradus:
Creating so many methods that are only aliases - same interface, same job, no new functionality.
The only thing will have will be that one will use `assertNotSame`, while other `refuseSame`, and finally, in your big PHP project, you will have both... no consistency at all... :(

IMO `refuseSame` would be great if `assertNotSame` will be deprecated.

sebastianbergmann:
Dear contributor,

Sebastian

PR #1661, opened 2015-03-28
Update Prophecy dependency

ciaranmcnulty:
Prophecy has tagged `1.4.0` which is currently not installable.

As Prophecy is a SemVer project there is no reason to constrain the dependency to `1.3.x`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,7 @@
         "phpunit/php-code-coverage": "~2.0,>=2.0.11",
         "phpunit/php-timer": "~1.0",
         "phpunit/phpunit-mock-objects": "~2.3",
-        "phpspec/prophecy": "~1.3.1",
+        "phpspec/prophecy": "~1.3,>=1.3.1",
         "symfony/yaml": "~2.1|~3.0",
         "sebastian/comparator": "~1.1",
         "sebastian/diff": "~1.2",


GrahamCampbell:
Could we have this going into an earlier version than phpunit 4.7 please? How about 4.6?

ciaranmcnulty:
Ah, probably opening this against `master` is a mistake?

GrahamCampbell:
@sebastianbergmann Can merge this into what ever branch he chooses I guess, just not though the github gui.

sebastianbergmann:
Cherry-picked into `4.5`.

GrahamCampbell:
@sebastianbergmann Is a 4.5.1 release planned before 4.6.0?

ciaranmcnulty:
Thanks @sebastianbergmann!

ciaranmcnulty:
@sebastianbergmann I have to decide whether the upcoming PhpSpec release can depend on Prophecy 1.4.0 or whether it has to have a bunch of conditionals in it. I want both tools to be installable together.

Do you have any idea when a PHPUnit release will be made that contains this change?

sebastianbergmann:
10 seconds ago :)

ciaranmcnulty:
/me removes lots of conditionals from phpspec ;)

GrahamCampbell:
:)

PR #1682, opened 2015-04-17
report PHP version used for test

remicollet:
I think this can be useful
- For user of multiple PHP versions on the same computer (especially when environment module are used to switch)
- For bug report, when the reporter copy/paste the output of phpunit command.


remicollet:
Thanks

PR #1700, opened 2015-04-29
KISS principle

ftwbzhao:


sebastianbergmann:
I agree that this part of the code could be nicer. But this patch does not really improve it.

PR #1702, opened 2015-04-29
Extra check for composer installer

jildertmiedema:
If the phpunit file is in the vendor/bin/phpunit. It will now find the autoload.php

### DIFF
--- a/phpunit
+++ b/phpunit
@@ -13,7 +13,7 @@ if (!ini_get('date.timezone')) {
     ini_set('date.timezone', 'UTC');
 }
 
-foreach (array(__DIR__ . '/../../autoload.php', __DIR__ . '/../vendor/autoload.php', __DIR__ . '/vendor/autoload.php') as $file) {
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
Works for me from `vendor/bin/phpunit` without this patch. Why do you think this is needed?

jildertmiedema:
if __DIR__ equals app/vendor/bin. none of the current patterns match.

GrahamCampbell:
That's a very stange setup though?

sebastianbergmann:
Indeed.

jildertmiedema:
```__DIR__``` points to the directory of the current script.

In ```myapp/vendor/bin/phpunit``` :  ```__DIR__ = "myapp/vendor/bin/"``` so the ```autoload.php``` will be in ```__DIR__ . '/../autoload.php'```

jildertmiedema:
The default ```bin-dir``` for composer is ```vendor/bin```

jildertmiedema:
Sorry its working now

jildertmiedema:
Extra info: By using the laravel installer the symblic link to the phpunit file in vendor/bin gets lost. The file is copied instead of linked. There for the ```__DIR__``` reference was incorrect.

PR #1706, opened 2015-05-03
Avoid hardcode URI to phpunit.xsd

Maks3w:
Using URI reference to local phpunit.xsd avoid the need of update the XSD for each release.


sebastianbergmann:
Cherry-picked into `4.6`.

PR #1712, opened 2015-05-12
Create PHPUnit bash-completion script

henriquemoody:
I've created this script [long time ago](https://gist.github.com/henriquemoody/5014805). 
I'm not sure if it should or shouldn't be in the PHPUnit repository, but here it is.

---

I've tested it on Bash 4.3.11 (Ubuntu) and 4.3.33 (OSX).

### DIFF
--- a/phpunit.bash-completion
+++ b/phpunit.bash-completion
@@ -1,0 +1,46 @@
+# This file is part of PHPUnit.
+#
+# (c) Sebastian Bergmann <sebastian@phpunit.de>
+#
+# For the full copyright and license information, please view the LICENSE
+# file that was distributed with this source code.
+#
+# Author: Henrique Moody <henriquemoody@gmail.com>.
+#
+
+_phpunit()
+{
+    COMPREPLY=()
+
+    local cur="${COMP_WORDS[COMP_CWORD]}"
+    local prev="${COMP_WORDS[COMP_CWORD-1]}"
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
I don't think this should be part of the PHPUnit project.

marcioAlmada:
> I don't think this should be part of the PHPUnit project.

@sebastianbergmann why?

BTW, in case the issue is maintenance, is there a centralized place where we can gather information about commands, sub commands and flags so the completion script generation could be automated?

@henriquemoody  would be really nice to have this in another repository :)

sebastianbergmann:
Separation of concerns; it has nothing to do with the main project, IMHO.

henriquemoody:
@marcioAlmada, I will probably send a pull request to [homebrew-php](https://github.com/Homebrew/homebrew-php) if you have more ideas I will be glad.

@sebastianbergmann, despite I've seem some repositories ([drush](https://github.com/drush-ops/drush/blob/master/drush.complete.sh), [wp-cli](https://github.com/wp-cli/wp-cli/blob/master/utils/wp-completion.bash), [git](https://github.com/git/git/blob/master/contrib/completion/git-completion.bash), [juju](https://github.com/juju/juju/blob/master/etc/bash_completion.d/juju-core), [npm](https://github.com/creationix/nvm/blob/master/bash_completion), etc) with the bash-completion script into the same repository, I can understand your point.

Issue #920, opened 2013-05-21
--no-coverage option to ovewrite config file


drgomesp:
The default configuration for all my projects is set to generate html code coverage reports.

Is there anyway I can run the phpunit command at a given time and not generate the code coverage reports without having to change the config file?

This would be something like a `--no-coverage option`.

PR #1727, opened 2015-05-26
Disable code coverage from the command line

ozh:
This PR allows to disable code coverage from the command line, to override a setting defined in the XML config file.
Usage: `phpunit --disable-coverage` or `phpunit --disable-coverage -- tests/some/tests`

I know this was [not on the agenda more than a year ago](https://github.com/sebastianbergmann/phpunit/issues/920) but I feel this would be a valuable addition when you want to quickly run one particular test or test suite and don't want to wait for the coverage generation

Thanks for your consideration (and thanks in any case for this most excellent tool)

### DIFF
--- a/src/TextUI/Command.php
+++ b/src/TextUI/Command.php
@@ -845,7 +845,8 @@ Code Coverage Options:
   --coverage-text=<file>    Generate code coverage report in text format.
                             Default: Standard output.
   --coverage-xml <dir>      Generate code coverage report in PHPUnit XML format.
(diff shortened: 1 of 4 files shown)


ozh:
Hu, just realized there were also tests to update -- will update my PR asap

sebastianbergmann:
Implemented in f625f6dc58a0cf31ef7d4ec53e40af5e5b09ac2c

ozh:
Thanks!

Issue #1151, opened 2014-02-23
@testWith feature


farafiri:
It would be great to have @testWith (like JUnit) feature:

```php
/**
 * @testWith (1, 2, 3)
 *           (10, 11, 21)
 */
function testAdd($a, $b, $result)
{
  $this->assertEquals($result, $sut->add($a, $b));
}
```

I know that I can use @dataProvider but it overcomplicated and require to much code for most tasks.

giorgiosironi:
@farafiri do you mind if I build on bf1c416 and open a PR with a complete implementation?

farafiri:
It would be great. Thank you.

marcinrogacki:
Hello,

I came here to propose such feature. That nice it is already done. 

A few thoughts:

1) Why do not use old good `@dataProvider` annotation instead of adding new one `@testWith`? In my opinion distinguish between function and data set shouldn't be complicated.

2) Simplify syntax (more space for large data list, faster to write, easier to read) 

```php
@dataProvider
0, 1, 1
1, 2, 3
20, 22, 42
``` 

3) Support complex types (inspared by https://code.google.com/p/zohhak/). Requirements: parse test method signature to find complex types, parse class file to find desired function which will create complex type. Search could support `@return` phpdoc.

Example 1
```php
    /**
     * @dataProvider
     *  "EUR", "FR", 
     *  "GBP", "UK"
     *  "USD", "US"
     */
    public function testCountryUsesCurrency(Currency $currency, Country $country)
    {
        $this->assertTrue($country->isDefaultCurrency($currency));
    }

   /**
    * @return Currency
    */
   public function createCurrency($currency)
   {
       return new Currency($currency);
   }

   /**
    * @return Country
    */
   public function createCountry($country)
   {
       return new Country($country);
   }
```

Example 2

```php
    /**
     * @dataProvider
     *  [49.95, "EUR"], "ACCEPTED",  49.95, "EUR"
     *  [39.95, "GBP"], "REFUSED", 0, "GBP"
     */
    public function testCapturePayment(
        Namespace_Order $order, Zend_Soap_Client $gateway,
       $expectedAmount, $expectedCurrency
    ) {
        $payment = new Payment($gateway, $order);
        $transaction = $payment->capture();
        $this->assertEquals($expectedAmount, $transaction->getAmount());
        $this->assertEquals($expectedCurrency, $transaction->getCurrency());
    }

   /**
    * @return Zend_Soap_Client
    */
[...]

PR #1728, opened 2015-05-26
@testWith annotation

giorgiosironi:
Implementing #1151.

The scope is reduced to a sequence of lines following the annotation, each containing a JSON array.

In case of annotation parsing error this is signaled through a standard PHPUnit exception. The test for this case has to use `setExceptedExceptionRegExp()` because `@testWith` becomes a reserved keyword inside docblocks.

### DIFF
--- a/src/Util/Test.php
+++ b/src/Util/Test.php
@@ -29,6 +29,7 @@ if (!function_exists('trait_exists')) {
 class PHPUnit_Util_Test
 {
     const REGEX_DATA_PROVIDER      = '/@dataProvider\s+([a-zA-Z0-9._:-\\\\x7f-\xff]+)/';
+    const REGEX_TEST_WITH          = '/@testWith\s+/';
     const REGEX_EXPECTED_EXCEPTION = '(@expectedException\s+([:.\w\\\\x7f-\xff]+)(?:[\t ]+(\S*))?(?:[\t ]+(\S*))?\s*$)m';
     const REGEX_REQUIRES_VERSION   = '/@requires\s+(?P<name>PHP(?:Unit)?)\s+(?P<value>[\d\.-]+(dev|(RC|alpha|beta)[\d\.])?)[ \t]*\r?$/m';
     const REGEX_REQUIRES_OS        = '/@requires\s+OS\s+(?P<value>.+?)[ \t]*\r?$/m';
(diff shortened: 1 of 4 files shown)


sebastianbergmann:
The tests fail with PHP 5.3. If you want this in PHPUnit 4.8 then it needs to be compatible with PHP 5.3. Otherwise I'll merge it to `master` once `4.8` has been branched off.

giorgiosironi:
Ah, was just muscular memory, will update it to work with 5.3

sebastianbergmann:
Thanks!

marcinrogacki:
Hello,

I came here to propose such feature. That nice it is already done. 

Few thoughts:

1) Why do not use old good `@dataProvider` annotation instead of adding new one `@testWith`? In my opinion distinguish between function and data set shouldn't be complicated.

2) Simplify syntax (more space for large data list, faster to write, easier to read) 

```php
@dataProvider
0, 1, 1
1, 2, 3
20, 22, 42
``` 

3) Support complex types (can be added later - complicated). Requirements: parse test method signature to find complex types, parse class file to find desired function which will create complex type. Search could support `@return` phpdoc.

Example 1
```php
    /**
     * @dataProvider
     *  "EUR", "FR", 
     *  "GBP", "UK"
     *  "USD", "US"
     */
    public function testCountryUsesCurrency(Currency $currency, Country $country)
    {
        $this->assertTrue($country->isDefaultCurrency($currency));
    }

   /**
    * @return Currency
    */
   public function createCurrency($currency)
   {
       return new Currency($currency);
   }

   /**
    * @return Country
    */
   public function createCountry($country)
   {
       return new Country($country);
   }
```

Example 2

```php
    /**
     * @dataProvider
     *  [49.95, "EUR"], "ACCEPTED",  49.95, "EUR"
     *  [39.95, "GBP"], "REFUSED", 0, "GBP"
     */
    public function testCapturePayment(
        Namespace_Order $order, Zend_Soap_Client $gateway, $expectedAmount, $expectedCurrency
    ) {
        $payment = new Payment($gateway, $order);
        $transaction = $payment->capture();
        $this->assertEquals($expectedAmount, $transaction->getAmount());
        $this->assertEquals($expectedCurrency, $transaction->getCurrency());
    }

   /**
    * @return Zend_Soap_Client
    */
   public function createStubPaymentGateway($status)
   {
        $stub = $this->getMockBuilder('Zend_Soap_Client')->getMock();
        $stub->method('getResult')->willReturn($status);
        return $stub;
   }

   /**
    * @return Namespace_Order
    */
   public function createOrder($amount, $currency)
   {
       return new Namespace_Order($amount, $country);
   }
```

PR #1733, opened 2015-05-28
Allow custom CLI options

### LINKED ISSUE #634: Allow custom CLI options
I want to run some Selenium tests with PHPUnit. For this, I need to know what browser and what base-URL to use inside my tests. Since I boot-up phpunit with my own CLI runner wrapper script, I wanted to do something like this:

```php
<?php

define('PHPUnit_MAIN_METHOD', 'PHPUnit_TextUI_Command::main');

// gotta love Composer...
$loader = require __DIR__.'/../src/vendor/autoload.php';
$loader->add('', __DIR__.'/../tests/unit');

// get custom CLI params
$opts = getopt('', array(...));

// define constants based on the params
define('MYPROJECT_TEST_BASE_URL', ...);

// hide options
unset($_SERVER['argv'][5]);

// run PHPUNit
PHPUnit_TextUI_Command::main();
```

Unfortunately, PHPUnit does not allow any other paramters that its own. The "PHPUnit-way" would be to use ```<php>``` sections inside the XML config files (I use XML config files, by the way) to define env variables. But this would make me have a ton of config files (Firefox + http://domainA/, Firefox + Domain B, Chrome + Domain A, Chrome + Domain B, etc.)

I understand the reasoning behind not allowing unknown options. But it should be possible to allow all options starting with ```--custom-``` (or something like that). This would make the CLI runner much more useful and still be relatively safe against typos and deprecated/removed options.

Currently, as a extremely ugly hack, I remove my elements from ```$_SERVER['argv']``` to make PHPUnit work, but I would like to remove this hack some time in the future.

Maybe I'm missing something from the docs about how to run PHPUnit with custom options?

clphillips:
Fixes #634

3 years is long enough.

### DIFF
--- a/src/Util/Getopt.php
+++ b/src/Util/Getopt.php
@@ -164,7 +164,5 @@ class PHPUnit_Util_Getopt
 
             return;
         }
-
-        throw new PHPUnit_Framework_Exception("unrecognized option --$opt");
     }
 }


sebastianbergmann:
This causes more problems than it solves.

clphillips:
Such as? Let's name them so we can resolve them.

sebastianbergmann:
Removing the check in `PHPUnit_Util_Getopt` means that a user will no longer get an error message when he or she mistyped a commandline switch. It should be possible to implement what you want without losing this feature.

clphillips:
What if we continue to detect unrecognized switches, and acknowledge that by printing the "unrecognized option" message, but don't halt execution (i.e. continue with the unrecognized switches pushed through)?

In that case:

1. Users that don't RTFM are notified that what they entered may not do what they intend, and
2. Custom switches no longer halt execution

sebastianbergmann:
It should have been clear from my previous comment that that would be fine by me. Sorry if I wasn't clear enough.

clphillips:
Excellent. Thank you.

sebastianbergmann:
Dear contributor,

Sebastian

PR #1739, opened 2015-06-04
Add support the parsing of options after the arguments

yvoyer:
Allow usage of options after the arguments (if any).

```
$phpunit --debug path/to/test.php --colors
```

Before: would ignore the `--colors` option
After: Parse all the options after the path

### DIFF
--- a/src/Util/Getopt.php
+++ b/src/Util/Getopt.php
@@ -54,8 +54,8 @@ class PHPUnit_Util_Getopt
 
             if ($arg[0] != '-' ||
                 (strlen($arg) > 1 && $arg[1] == '-' && !$long_options)) {
-                $non_opts = array_merge($non_opts, array_slice($args, $i));
-                break;
+                $non_opts[] = $args[$i];
+                continue;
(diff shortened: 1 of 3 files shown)


sebastianbergmann:
Merged manually.

yvoyer:
@sebastianbergmann I just noticed a [note](https://phpunit.de/manual/current/en/textui.html) (at the bottom) that no longer apply with this PR.

I made a [PR](https://github.com/sebastianbergmann/phpunit-documentation/pull/306) to keep it consistent with this PR.

ivicaned:
Before version 4.8.0 I was able to add custom parameters that was used by my own test suite, for example:
phpunit TestSuite --exclude-benchmarks
and it would exclude benchmark tests(based on same internal directory structure and code in TestSuite.php).
Now after this pull request is accepted, when I try to add additional parameters for our internal use, we get error:
unrecognized option --exclude-benchmarks

Is there a way to allow additional parameters to phpunit, that might be used only internally(so in case if phpunit do not recognize, just skip them and provide warning message to users once tests are completed)?

PR #1753, opened 2015-06-18
added assertFinite, assertInfinite and assertNan

tronsha:


sebastianbergmann:
Merged manually, thanks.

sebastianbergmann:
Would be great if you could document (https://github.com/sebastianbergmann/phpunit-documentation/issues/308) these new assertions, @tronsha.

src/Framework/Assert.php
```diff
@@ -968,6 +968,42 @@ public static function assertNull($actual, $message = '')
...
+    /**
+     * Asserts that a variable is nan.
+     *
+     * @param mixed  $actual
+     * @param string $message
+     */
+    public static function assertNan($actual, $message = '')
+    {
+        self::assertThat($actual, self::isNan(), $message);
+
```
GrahamCampbell: cs

sebastianbergmann: Was already fixed in 182be7c2e78b7f05f2416e833e9a0b2815fbacc3

PR #1754, opened 2015-06-18
Don't declare missing native constant JSON_PRETTY_PRINT

GromNaN:
Declaring system constants in PHPUnit can result is side effects in projects checking constant definition.

Example in LemonRestBundle, the code assumes that if `JSON_PRETTY_PRINT` is defined, the other constant `JSON_UNESCAPED_SLASHES` is also defined.
https://github.com/stanlemon/rest-bundle/blob/0.8.0/DependencyInjection/Extension.php#L29

### DIFF
--- a/src/Util/Log/JSON.php
+++ b/src/Util/Log/JSON.php
@@ -8,10 +8,6 @@
  * file that was distributed with this source code.
  */
 
-if (!defined('JSON_PRETTY_PRINT')) {
-    define('JSON_PRETTY_PRINT', 128);
-}
-
 /**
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
We don't even need this at all on the master since it's php 5.6+.

GromNaN:
Ok, I didn't know that.
So can I submit this PR on branch 2.7 or 2.8 ?

sebastianbergmann:
It has been taken care of.

GromNaN:
Thank you. It happened so fast. :dash:

PR #1764, opened 2015-06-23
Fix issue with annotations in parent classes

### LINKED ISSUE #1763: before and after annotations don't work in some traits
If you use a trait in a class then extend that class then phpunit doesn't process the before and after annotations for those methods in that trait.

NB, experiencing this with phpunit 4.7.5.

GrahamCampbell:
This is a major issue and affects all laravel users who choose to have a base testcase class.

----
Closes #1763.

### DIFF
--- a/src/Util/Test.php
+++ b/src/Util/Test.php
@@ -701,10 +701,6 @@ class PHPUnit_Util_Test
                 $class = new ReflectionClass($className);
 
                 foreach ($class->getMethods() as $method) {
-                    if ($method->getDeclaringClass()->getName() != $className) {
-                        continue;
-                    }
-
                     if (self::isBeforeClassMethod($method)) {
                         self::$hookMethods[$className]['beforeClass'][] = $method->getName();
                     }


jbrooksuk:
:+1: this issue affected us when writing tests for Cachet.

sebastianbergmann:
This basically reverts f99f4763faae01b6ab0d0a2ec6fbe810ec4d945a. Unfortunately I do not remember why I made that change but I am sure that there was a good reason for it.

GrahamCampbell:
As far as I can see, that change introduces way more issues than it surely can be fixing?

ciaranmcnulty:
What's the use case you have in your base class?

GrahamCampbell:
Well, if I can have traits in php 5.3, I have to use a base class to achieve what I want.

GrahamCampbell:
See my example in https://github.com/sebastianbergmann/phpunit/issues/1763#issuecomment-114516437. :)

ciaranmcnulty:
PHP 5.3 is EOL a while ago...

GrahamCampbell:
Yeh, I know, but it doesn't mean people aren't still using it. Besides that, people may still want base classes in php 5.4. Out of the box laravel has a base class for unit tests and it requires php 5.5.9+.

GrahamCampbell:
It's common for people to load their traits in here: https://github.com/laravel/laravel/blob/master/tests/TestCase.php, only if they want annotations, phpunit's broken.

GrahamCampbell:
Something like https://github.com/laravel/framework/blob/5.1/src/Illuminate/Foundation/Testing/DatabaseMigrations.php if loaded in that testcase class is currently broken.

GrahamCampbell:
Here's an example of how people are affected using lumen: https://github.com/laravel/lumen-framework/issues/108.

Issue #1748, opened 2015-06-14
EngineExceptions (eg TypeException) can not be detected using @expectedException (PHP 7)


throup:
PHP 7 introduces `EngineException`s in place of fatal errors.  These exceptions are not detected when annotated with `@expectedException`. In fact, the test runner will fail unless these exceptions are manually caught.

As a simple example, consider the class:
``` php
<?php
declare(strict_types = 1);

class ScalarType {
    public function expectsInteger(int $integer) {
        // Do nothing
    }
}
```

and the testcase:
``` php
<?php
declare(strict_types = 1);

class ScalarTypeTest extends PHPUnit_Framework_TestCase {
    /**
     * @test
     * @expectedException TypeException
     */
    public function passStringWhenIntegerExpected() {
        $sut = new ScalarType();
        $sut->expectsInteger('Not an integer');
    }
}
```

The test runner output:
``` bash
$ phpunit tests/ScalarTypeTest.php 
PHPUnit 4.7.3 by Sebastian Bergmann and contributors.

PHP Fatal error:  Uncaught TypeException: Argument 1 passed to ScalarType::expectsInteger() must be of the type integer, string given, called in .../PHPUnitTests/tests/ScalarTypeTest.php on line 11 and defined in .../PHPUnitTests/src/ScalarType.php:5
Stack trace:
#0 .../PHPUnitTests/tests/ScalarTypeTest.php(11): ScalarType->expectsInteger('Not an integer')
#1 [internal function]: ScalarTypeTest->passStringWhenIntegerExpected()
#2 .../PHPUnitTests/vendor/phpunit/phpunit/src/Framework/TestCase.php(863): ReflectionMethod->invokeArgs(Object(ScalarTypeTest), Array)
#3 .../PHPUnitTests/vendor/phpunit/phpunit/src/Framework/TestCase.php(741): PHPUnit_Framework_TestCase->runTest()
#4 .../PHPUnitTests/vendor/phpunit/phpunit/src/Framework/TestResult.php(605): PHPUnit_Framework_TestCase->runBare()
#5 .../PHPUnitTests/vendor/phpunit/phpunit/src/Framework/TestCa in .../PHPUnitTests/src/ScalarType.php on line 5

[...]

PR #1766, opened 2015-06-24
Allow onNotSuccessfulTest() to accept Throwable

brownoxford:
Remove "Exception" type-hint for onNotSuccessfulTest() and replace with inline checking for proper type based on PHP major version. The calling context for this method was changed in #1748 such that both objects of type Exception and interface Throwable can be passed.

Throwable was introduced in 7.0.0-alpha2, so changing the type-hint to "Throwable" is a BC break; this fix attempts to do the right thing while maintaining BC.

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -1852,13 +1852,19 @@ abstract class PHPUnit_Framework_TestCase extends PHPUnit_Framework_Assert imple
     /**
      * This method is called when a test method did not execute successfully.
      *
-     * @param Exception $e
+     * @param Exception|Throwable $e
      * @since Method available since Release 3.4.0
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
Merged manually.

mnapoli:
Note for people stumbling on this like me: this was merged in the 5.0 branch, so you need to install PHPUnit 5.0.

sebastianbergmann:
But keep in mind that PHPUnit 5.0 is not stable yet.

bweston92:
@sebastianbergmann so if it's not stable what is the work around for PHPUnit 4?

GrahamCampbell:
There isn't one. You have to use 5.x to get this stuff.

PR #1772, opened 2015-06-29
Added possibility to use a own config loader

attrib:
We needed the possibility to change the test suite configuration, when starting a test. We didn't wanted to use the xml file, as we also needed to add more data for each test in a test suite.

Therefore we create the possibility to have a separate config loader. In our current environment we just extend the default loader and only overwrite getTestSuiteConfiguration(). But for others maybe it becomes handy that you also can overwrite the rest.

To define a separate loader a new argument was added (loaderConfig), where you define the name. To always use it, we copied the phpunit startup script in our own directory and added the argument there. As this is about the config, I didn't had a better idea how you maybe could define it somewhere else. The name is not ideal, but nothing beginning with config can be used (otherwise some regression tests fail).

Looking forward for tips and change requests, so this gets maybe accepted at one time. :)

### DIFF
--- a/src/TextUI/Command.php
+++ b/src/TextUI/Command.php
@@ -38,6 +38,7 @@ class PHPUnit_TextUI_Command
         'colors=='             => null,
         'columns='             => null,
         'configuration='       => null,
+        'loaderConfigs='       => null,
         'coverage-clover='     => null,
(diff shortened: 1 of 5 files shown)


sebastianbergmann:
Dear contributor,

Sebastian

PR #1775, opened 2015-06-30
fixes #1771 - don't print colors to stdout in text coverage summary when 'never' is used

### LINKED ISSUE #1771: coverage text do not honor --colors setting
According to doc

https://phpunit.de/manual/current/en/logging.html
```
When used from the command-line by writing to php://stdout this will honor the --colors setting
```

But it doesn't: version 4.7.5

```xml
<phpunit
        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
        xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/4.5/phpunit.xsd"
        colors="false"
        stopOnError="true"
        stopOnFailure="true"
        bootstrap="./tests/bootstrap.php">

    <testsuite name="Unit test suite">
        <directory>./tests</directory>
    </testsuite>

    <filter>
        <whitelist>
            <directory suffix=".php">./lib</directory>
        </whitelist>
    </filter>

    <groups>
        <exclude>
            <group>ignore</group>
            <group>tests</group>
        </exclude>
    </groups>

    <logging>
        <log type="coverage-text" target="php://stdout" showOnlySummary="true" showUncoveredFiles="false" />
    </logging>

</phpunit>
```

zrzut ekranu 2015-06-29 11 10 25

abacaphiliac:
coverage-text-summary-only-without-colors - do not print colors to stdout in text coverage summary when \PHPUnit_TextUI_ResultPrinter::COLOR_NEVER is used.

@sebastianbergmann i'm new to phpt large tests. is this the way you would cover that change? i'd like to get this merged so that i can write a similar test to make sure that codeCoverageIgnore is disabled in a new pull request that makes use of that feature (https://github.com/sebastianbergmann/php-code-coverage/pull/338).

### DIFF
--- a/src/TextUI/TestRunner.php
+++ b/src/TextUI/TestRunner.php
@@ -496,6 +496,9 @@ class PHPUnit_TextUI_TestRunner extends PHPUnit_Runner_BaseTestRunner
                 if ($arguments['coverageText'] == 'php://stdout') {
                     $outputStream = $this->printer;
                     $colors       = $arguments['colors'];
+                    if ($colors == PHPUnit_TextUI_ResultPrinter::COLOR_NEVER) {
+                        $colors = false;
+                    }
                 } else {
                     $outputStream = new PHPUnit_Util_Printer($arguments['coverageText']);
                     $colors       = false;
--- a/tests/TextUI/full-coverage-text-summary-only-without-colors.phpt
+++ b/tests/TextUI/full-coverage-text-summary-only-without-colors.phpt
@@ -1,0 +1,26 @@
+--TEST--
+phpunit --configuration=../TextUI/full-coverage-text-summary-only-without-colors.xml FullCoverageClassTest ../_files/FullCoverageClassTest.php
+--FILE--
+<?php
+$_SERVER['argv'][1] = '--configuration=' . dirname(__FILE__) . '/../TextUI/full-coverage-text-summary-only-without-colors.xml';
+$_SERVER['argv'][2] = 'FullCoverageClassTest';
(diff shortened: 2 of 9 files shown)


henriquemoody:
`always`, `never` and `auto` are available only on command-line, see #1529.

abacaphiliac:
hello @henriquemoody , thanks for the reply! here is some context for the proposed change:

i want to use summary-only coverage-text to `php://stdout` with `colors=never` for a large test to support a new option i will add to phpunit soon. i discovered that summary-only coverage-text to `php://stdout` with `colors=never` actually does print to screen with color, which is not the desired behavior.

when `colors=never` is used, it is set to the `$colors` variable in `PHPUnit_TextUI_TestRunner ` here:
https://github.com/sebastianbergmann/phpunit/blob/4.7.6/src/TextUI/TestRunner.php#L497

... which is passed to the text-coverage processor here:
https://github.com/sebastianbergmann/phpunit/blob/4.7.6/src/TextUI/TestRunner.php#L511

... and in `PHP_CodeCoverage_Report_Text ::process()` `$showColors` is used as a boolean value:
https://github.com/sebastianbergmann/php-code-coverage/blob/2.1.7/src/CodeCoverage/Report/Text.php#L62

... so you can see that when `never` is passed it will be evaluated as truthy and will cause the incorrect behavior.

my proposed change is meant to convert `never` to a boolean value specifically for use with the text-coverage processor and nothing more.

is there some other logic i should add for `always` or `auto`? it seemed to me that these should continue to function as they already do.

abacaphiliac:
closed in favor of https://github.com/sebastianbergmann/phpunit/pull/1841

PR #1781, opened 2015-07-04
Improved JSON detection: Empty string is not treated as a valid JSON string anymore

shochdoerfer:
When passing an empty string to json_decode() null gets returned and no "error" is detected by json_last_error(). I added an explicit check to PHPUnit_Framework_Constraint_IsJson::matches() to detect an empty string and fail in this case.

### DIFF
--- a/src/Framework/Constraint/IsJson.php
+++ b/src/Framework/Constraint/IsJson.php
@@ -24,6 +24,10 @@ class PHPUnit_Framework_Constraint_IsJson extends PHPUnit_Framework_Constraint
      */
     protected function matches($other)
     {
+        if ($other === '') {
+            return false;
+        }
+
         json_decode($other);
         if (json_last_error()) {
             return false;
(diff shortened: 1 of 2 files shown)


sebastianbergmann:
Merged manually, thanks.

rflavien:
According to https://tools.ietf.org/html/rfc8259#section-1 empty strings are allowed. This change introduced a wrong assertion result.
```php
// Given
$this->assertJson("");
// I expect true, even if the PHP function json_decode handle it weirdly.
```

shochdoerfer:
@rflavien I think your example is slightly wrong, you are not asserting an empty string but an empty JSON document and that is not valid. According to [RFC 8259](https://tools.ietf.org/html/rfc8259#section-2) a JSON document needs to contain a value that maps to either being an object, array, number, or string, or a literal (false, null, true). A string is defined to start and end with a quotation mark. That means your example should look like this `$this->assertJson('""');` to be correct and the assertion evaluates as true then.

I tested this behaviour with [JSONLint](https://jsonlint.com) and [jsonlint](https://www.npmjs.com/package/jsonlint). For empty documents, I get an error in both tools. For a document containing an empty string both tools succeed.

PR #1782, opened 2015-07-06
Fix: Acess vs. Access

localheinz:
This PR
- [x] fixes a small typo where `Acess` should be `Access`


sebastianbergmann:
Cherry-picked into `4.7`, thanks.

PR #1805, opened 2015-07-25
correct autoload.php path inclusion

rockymontana:
when installing with composer this files get copied to $vendorDir/bin and the autoload.php-file is directly in $vendorDir.
So the the file_exists-statement should include that path too.


sebastianbergmann:
Likely duplicate of #1613 or similar issues.

PR #1823, opened 2015-08-13
Columns attribute in phpunit.xml is ignored

m157y:


sebastianbergmann:
Merged manually, thanks.

PR #1825, opened 2015-08-14
Redirect STDERR to STDOUT when running PHPT tests

### LINKED ISSUE #1169: phpt runner doesn't look at STDERR

henriquemoody:
This is the current behaviour of the [run-tests.php](https://github.com/php/php-src/blob/master/run-tests.php) script which runs all entire PHP source test suite.

Related to #1550

Fix #1169


sebastianbergmann:
Merged manually, thanks.

PR #1828, opened 2015-08-18
Rename a parameter of function assertJson

lancechentw:
Rename `$expectedJson` to `$actualJson` according to commit
672d9c912be0962832a91bd6fda3403bebfe38ce. Also, rename param `$filename` in
docblock, which seems to be a typo, to `$actualJson`.


sebastianbergmann:
Merged manually, thanks.

PR #1836, opened 2015-08-23
Fix test status being overwritten to error

### LINKED ISSUE #1835: Skipped test reported as errored since 4.7.4
Hi,
we use custom TestListener and log results of each test. But I started to notice skipped tests are in recent PHPUnit version reported as having status *4* (= `STATUS_ERROR`), instead of *1* (= `STATUS_SKIPPED`). I found the regression appears since PHPUnit 4.7.4:

SkippedTest.php:
```php
class SkippedTest extends \PHPUnit_Framework_TestCase
{
    public function testSkipped()
    {
        $this->markTestSkipped('Skipped test');
    }
}
```

TestStatusListener.php
```php
...
    public function endTest(PHPUnit_Framework_Test $test, $time)
    {
        printf("Test '%s' ended with status %d\n", $test->getName(), $test->getStatus());
    }
...
```

Output of up to version 4.7.3:
```
PHPUnit 4.7.3 by Sebastian Bergmann and contributors.

STest 'testSkipped' ended with status 1
```

Output since 4.7.4:
```
PHPUnit 4.7.4 by Sebastian Bergmann and contributors.

STest 'testSkipped' ended with status 4
```

Also tried with current PHPUnit master (0d7b83e), it still returns `4`.

I guess this is not intentional?

OndraM:
Fixes #1835 and also adds tests to ensure test status is being properly set for all the cases.

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -766,9 +766,9 @@ abstract class PHPUnit_Framework_TestCase extends PHPUnit_Framework_Assert imple
             $e = $_e;
         }
 
-        if (isset($e)) {
+        if (isset($_e)) {
             $this->status        = PHPUnit_Runner_BaseTestRunner::STATUS_ERROR;
-            $this->statusMessage = $e->getMessage();
+            $this->statusMessage = $_e->getMessage();
         }
 
         // Clean up the mock objects.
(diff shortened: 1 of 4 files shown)


sebastianbergmann:
Cherry-picked into `4.8`, thanks.

PR #1838, opened 2015-08-27
Closes #1802

Shaked:
Few comments: 
- I have added the `warning` as a separated thing than a `failure`. 
- The example in #1802 shows a very specific scenario for a warning, I have taken this a step farther as I saw that a warning can be shown from a different reasons. 
- As well, I kept the `1) Warning` title as I assumed that it was dropped by mistake. However, the fix for that is simple in case its required though I believe this will create a bit of a mess if thats true only for warnings while its not true for other scenarios. 

I hope this would do the job but feel free to comment out if anything comes in mind. 
Cheers
Shaked

PR #1840, opened 2015-08-29
Closes #1802

### LINKED ISSUE #1802: Introduce separate class for warnings
## Example

```php
<?php
class Test extends PHPUnit_Framework_TestCase
{
}
```

## Actual

```
$ phpunit Test                                  
PHPUnit 5.0-g6caa139 by Sebastian Bergmann and contributors.

F

Time: 58 ms, Memory: 4.50Mb

There was 1 failure:

1) Warning
No tests found in class "Test".

FAILURES!
Tests: 1, Assertions: 0, Failures: 1.
```

## Expected

```
$ phpunit Test                                  
PHPUnit 5.0-g6caa139 by Sebastian Bergmann and contributors.

W

Time: 58 ms, Memory: 4.50Mb

There was 1 warning:

1) No tests found in class "Test".

WARNINGS!
Tests: 1, Assertions: 0, Warnings: 1.
```

Shaked:
Opened this pull request here as a request from @sebastianbergmann. For more information see https://github.com/sebastianbergmann/phpunit/pull/1838.

### DIFF
--- a/phpunit.xsd
+++ b/phpunit.xsd
@@ -202,6 +202,7 @@
     <xs:attribute name="processIsolation" type="xs:boolean" default="false"/>
     <xs:attribute name="stopOnError" type="xs:boolean" default="false"/>
     <xs:attribute name="stopOnFailure" type="xs:boolean" default="false"/>
+    <xs:attribute name="stopOnWarning" type="xs:boolean" default="false"/>
     <xs:attribute name="stopOnIncomplete" type="xs:boolean" default="false"/>
     <xs:attribute name="stopOnRisky" type="xs:boolean" default="false"/>
     <xs:attribute name="stopOnSkipped" type="xs:boolean" default="false"/>
--- a/src/Extensions/TicketListener.php
+++ b/src/Extensions/TicketListener.php
@@ -109,7 +109,7 @@ abstract class PHPUnit_Extensions_TicketListener implements PHPUnit_Framework_Te
      */
     public function startTest(PHPUnit_Framework_Test $test)
     {
-        if (!$test instanceof PHPUnit_Framework_Warning) {
+        if (!$test instanceof PHPUnit_Framework_WarningTestCase) {
             if ($this->ran) {
                 return;
             }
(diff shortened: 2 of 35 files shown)


GrahamCampbell:
Looks pretty good to me. :+1:

sebastianbergmann:
Merged manually, thanks.

keradus:
Extending the `TestListener` interface on 5.1 (MINOR) update have broke the interface :(:(:(

GrahamCampbell:
Not the only breaking change on 5.1 either. :(

sebastianbergmann:
You are right, @keradus, and that should not have happened. However, test listeners should **extend** `BaseTestListener` instead of **implementing** `TestListener` directly.

sebastianbergmann:
@GrahamCampbell Huh?

GrahamCampbell:
The changing the call order for the before annotations stuff has broken thousands of Laravel applications.

GrahamCampbell:
> Two months of beta and nobody reported any breaking change.

That;s because we're all rolling with a stable version. ;)

keradus:
@sebastianbergmann but since the interface is public, it is still OK to implement it too, isn't it? And a lot of libraries used to typehint the interface, so it is a problem of BC breaker without MAJOR update. I know how to change the code to be able to run with changed interface, but it is better to fix it in the source library than in all external usages.

Any chances to get 5.x line fixed ?

sebastianbergmann:
@keradus I am working on a fix.

keradus:
Great to hear that! Thank you!

sebastianbergmann:
I just released PHPUnit 5.1.1 which has 073c4de6d013353df368ccb25f9de37f13f61d2d. Thank you for bringing this to my attention!

beberlei:
@GrahamCampbell the execution order of decoupled event listeners is *not* something you can rely on when its undefined. If this break testing in Laravel core, then this must be because there is high temporal coupling, where there shouldn't be. This means its a bug in Laravel.

GrahamCampbell:
> @GrahamCampbell the execution order of decoupled event listeners is not something you can rely on when its undefined.

But we're NOT relying on that. Only that the setup method is run first. I don't care what order the before methods are run, so long as it's after the setup method.

beberlei:
@GrahamCampbell but the order of setup or annotations first was undefined as well, PHPUnit never guaranteed that one runs before the other. If you rely on it, then you rely on undefined behavior and that can change without being a BC break.

GrahamCampbell:
Then it shouldn't matter to you if this is reverted by the same argument...

PR #1861, opened 2015-09-18
Fixed #1860

### LINKED ISSUE #1860: Not well-formed XML strings are always asserted equal by Assert::assertXmlStringEqualsXmlString
I tried to compare two XML strings and my test always passes if both XML strings are not well-formed.

simohammedhttp:
SimpleXMLElement will trigger E_WARNING error messages if XML strings to be compared are not well-formed.


sebastianbergmann:
There must be a better way to fix #1860.

Issue #1355, opened 2014-07-25
Replace Util/GetOpt with Symfony Console


sun:
The current CLI options/arguments parsing in `PHPUnit_Util_GetOpt` offers a suboptimal UX, because options have to be specified before arguments.

Proposal: Replace `PHPUnit_Util_GetOpt` with [Symfony Console](https://github.com/symfony/console).

Console has an excellent options/argument parser, which allows you to specify options anywhere in the command line.  Also used by [Composer](https://github.com/composer/composer).

However, for this proposal, we probably need to consider two additional aspects:
1. Similar to #1342, Console uses PHPUnit for its own tests, so what happens if Console has a bug that causes commands to not be run?  How do you debug that?
2. If PHPUnit depends on Console, then projects, which depend on PHPUnit for testing, but also on Console, will have to specify the dependency in a way that is compatible with PHPUnit's.

sergeyz:
Considering we are moving to PSR2\PSR4 I suggest to replace arguments like class+file `phpunit [options] UnitTest [UnitTest.php]` with multiple files

Idrinth:
I'm wondering what the allowed range of symfony/console should be.
- `^4`
- `^3|^4`
- `^2|^3|^4`

I think versions 3 and 4 are pretty much required to make any composer based installs possible.
2 is very similar to 3, so that might only be a tiny addition in workload

sebastianbergmann:
I have decided not to do this.

Issue #1330, opened 2014-07-14
Allow non-ambiguous shortened longopts


tolry:
Hi there,

since this is a quite common behaviour of cli-tools, I was expecting phpunit to act the same. I am so used to this, that I made the following mistake already twice :frowning: 

using
`phpunit --config foo.xml`
instead of
`phpunit --configuration|-c foo.xml`

Since this didn't throw any kind of error or warning message on the cli, it took me quite some time (longer than I care to admit) to figure out the source of the problem.

Would this be considered a bug fix or a feature (since behaviour change is introduced)?

Or am I missing some obvious explanation due to why phpunit behaves like this?

If I could get hinted the right direction, I would be happy to start a PR on this.

so long,
Tobias

PR #1868, opened 2015-09-24
[RFC] Add InputDefinition to define Arguments and Options instead of using GetOpt

yvoyer:
This PR introduce a way to start using the Symfony Console component. 
I would like to have your opinions on the implementation and the suggested usage.

This relates with #1355, 
- The first step of this PR is to define all supported arguments and options
- The next step would be to transform the `PHPUnit_TextUI_Command` into a Symfony `Command` or `Application`.

**Suggested usage**: 

I propose to fetch the arguments in `PHPUnit_TextUI_Command::handleArguments()` using the `Symfony\Component\Console\Input\ArgvInput`.
Once done, we could set the inner attribute `$this->arguments[<option>] = $input->getOption(<option>);` instead of relying on the `PHPUnit_Util_Getopt` class.


sebastianbergmann:
Thank you for starting to work on this. It would be great to have a clean CLI test runner. Ideally this would be a completely new test runner developed in a separate repository as a separate component. In a year from now, for PHPUnit 6, I would like `phpunit/phpunit` to simply be a "meta package" that pulls in two new components, `phpunit/test-runner` and `phpunit/framework`.

src/TextUI/InputDefinition.php
```diff
@@ -0,0 +1,56 @@
...
+ */
+
+use Symfony\Component\Console\Input\InputDefinition;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Input\InputArgument;
+
+/**
+ * @author Yannick Voyer <star.yvoyer@gmail.com> (http://github.com/yvoyer)
+ *
+ * @since Class available since Release 5.0.0
```
sebastianbergmann: `master`, where this has to go, is PHPUnit 6.0.0

yvoyer:
@sebastianbergmann The bug #1330 would probably be broken if we were to use the Console component, since any ambigous option would trigger an exception. How would you like me to manage that?

I managed to implement (on my local machine) a way to add the console component with the version 4.8 as the head. Would you like to see it in another PR, or do you prefer to keep it for 6.0?

Currently this branch supports all the options already found in the help. I'm having some issues with the arguments `UnitTest [UnitTest.php]`, but I'll check it out soon.

sebastianbergmann:
Sorry, but this has to go into `master` as I fear that it will take longer than a release cycle to get right. We can always backport it later.

src/TextUI/Argument/Argument.php
```diff
@@ -0,0 +1,20 @@
...
+ * (c) Sebastian Bergmann <sebastian@phpunit.de>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use Symfony\Component\Console\Input\InputArgument;
+
+/**
+ * @author Yannick Voyer <star.yvoyer@gmail.com> (http://github.com/yvoyer)
```
sebastianbergmann: We do not use the `@author` annotation.

src/TextUI/Command.php
```diff
@@ -248,7 +250,7 @@ protected function handleArguments(array $argv)
         foreach ($this->options[0] as $option) {
             switch ($option[0]) {
                 case '--colors':
-                    $this->arguments['colors'] = $option[1] ?: PHPUnit_TextUI_ResultPrinter::COLOR_AUTO;
+                    $this->arguments['colors'] = $input->getOption('colors');
```
yvoyer: @sebastianbergmann I implemented the parsing using symfony console for the `colors` option, so that you could see how its done. 

I'll work later on all the other options.

src/TextUI/Option/Colors.php
```diff
@@ -0,0 +1,28 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use Symfony\Component\Console\Input\InputOption;
+
+/**
+ * @since Class available since Release 6.0.0
+ */
+final class PHPUnit_TextUI_Option_Colors extends PHPUnit_TextUI_Option_Option
```
yvoyer: @sebastianbergmann What are your point of view with the `final` keyword?

fyi, [Here](http://verraes.net/2014/05/final-classes-in-php/) is why I put all my class final in my template.

yvoyer:
I have started to migrate options from the old parsing to the new parsing, while adding tests to make sure that each option is still supported the same way it previously was.

I'm having some issues with the expectation of output with colors (Fail tests with `--colors` option, or `--coverage-text`). @sebastianbergmann Do you know how we can fix those expectations?

There is also this [issue](https://github.com/symfony/symfony/pull/12773) that blocks the resolution of the `--coverage-text` since there is no way yet to know if an option is set with a default value, or not set.

For the resolution of `--coverage-text`, either we wait for Symfony community to merge the fix, or we could define a new option switch (`--coverage-text-stdout`) that would default to `php://stdout`. The `--coverage-text` option would still accept path to a file, and nothing else. What do you think?

sebastianbergmann:
As long as we do not have a working solution (something like https://github.com/webmozart/php-scoper) to use multiple versions of the same library within a PHP process I am afraid that PHPUnit (core) must not depend on a popular component such as Symfony Console.

PR #1874, opened 2015-09-26
Avoid exiting the Travis build script in case the status code testing fails

stof:
`exit` should not be used in the .travis.yml file steps, as it would exit the whole build, leading to no feedback and missing the other steps: http://docs.travis-ci.com/user/customizing-the-build/#How-does-this-work%3F-%28Or%2C-why-you-should-not-use-exit-in-build-steps%29

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ install:
 
 script:
   - ./phpunit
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
Merged manually, thanks.

PR #1876, opened 2015-09-27
add --atleast-version command option

remicollet:
This option mimics the pkg-config one (same name used).

Probably not very useful for most user which use composer or phpunit.phar, but will be appreciated when "phpunit" is not under the user control.

```
if phpunit --atleast-version 5; then
   phpunit --verbose
else
   echo "PHPUnit is too old"
fi
```

Especially for Package in Fedora/EPEL (same spec file used in all branches) as we are going to be sticked to 4.x (in EPEL), because of PHP dependencies.

Of course, will be nice to have in 4.8...  but as "phpunit --atleast-version 5" fails (unrecognized option) this should be enough.

### DIFF
--- a/src/TextUI/Command.php
+++ b/src/TextUI/Command.php
@@ -34,6 +34,7 @@ class PHPUnit_TextUI_Command
      * @var array
      */
     protected $longOptions = [
+        'atleast-version='        => null,
         'bootstrap='              => null,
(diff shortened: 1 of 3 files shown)


sebastianbergmann:
Merged manually, thanks.

remicollet:
Thanks

PR #1878, opened 2015-09-27
Removed processUncoveredFilesFromWhitelist

MajorCaiger:
Hi @sebastianbergmann 

Here are the changes I made to phpunit to remove the processUncoveredFilesFromWhitelist flag. I needed to do this before I could properly test my changes to code coverage. Not sure exactly what branch this needs merging into.

Thanks

Rob

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -19,7 +19,7 @@
   </testsuites>
 
   <filter>
-    <whitelist processUncoveredFilesFromWhitelist="true">
+    <whitelist>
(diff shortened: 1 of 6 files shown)


sebastianbergmann:
Thanks. I will merge this into the right branch if/when the respective branch for php-code-coverage is merged.

Maks3w:
Unfortunately this can't happen until next major version of php-code-coverage and phpunit

sebastianbergmann:
I don't think that this is needed anymore, closing.

PR #1884, opened 2015-10-01
Avoid passing Error to onNotSuccesfulTest on PHP 7

### LINKED ISSUE #1882: onNotSuccessfulTest is broken on PHP 7 in PHPUnit 4.8
PHPUnit 4.8 is catching ``Throwable`` and then passing it to ``onNotSuccessfulTest``, which is typehinted as ``Exception``.
Given that removing the typehint is not possible (it would break BC), I suggest that instances of ``Error`` are rethrown directly instead of calling ``onNotSuccessfulTest``

See for such an error https://travis-ci.org/symfony/symfony/jobs/83059828#L3640. this makes the failure report much less useful.

stof:
onNotSuccessfulTest is typehinted as Exception, so only Exception throwables should be passed to it.

Closes #1882

This does not allow `onNotSuccessfulTest` to be called for `Error` on PHP 7, but this is impossible because of the typehint.
This code simply rethrow the error (which is the default method of the method), so that we get the display of the error instead of seeing the display of a TypeError triggered in PHPUnit.

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -848,6 +848,11 @@ abstract class PHPUnit_Framework_TestCase extends PHPUnit_Framework_Assert imple
                 $e = new PHPUnit_Framework_AssertionFailedError($e->getMessage());
             }
 
+            if (!$e instanceof Exception) {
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
Merged manually, thanks.

Issue #1013, opened 2013-09-30
PHP settings not recognized when using process isolation


mburtscher:
I configured some `$_SERVER` variables to be set for tests using the XML file:

``` xml
<phpunit>
  <php>
    <server name="SERVER_PORT" value="80" />
  </php>
</phpunit>
```

Works fine but not with process isolation. When marking a test with `@runInSeparateProcess` I get an "Undefined index: SERVER_PORT" error.

kyale:
The related issue #757 has nothing to to with the issue described above. I have a fix for this issue and would like to submit a pull request.

PR #1885, opened 2015-10-01
#1013 added configuration path to globals for reinitializing php values in tpl

kyale:
This is a fix for issue #1013 which causes tests that are ran in isolation. to lose any php value configured in configuration xml. This only happens when preserve globals is disabled.


sebastianbergmann:
Merged manually, thanks.

Issue #63, opened 2010-11-05
Invalid @covers annotations should produce a test error (E) instead of aborting phpunit


dharkness:
If any @covers annotation specifies an invalid class, method, function, etc. the tests stop due to an uncaught RuntimeException. While this relates to code coverage, the fix belongs in PHPUnit itself.

To turn this into a test error, replace this line in PHPUnit_Framework_TestResult::run():

```
$this->codeCoverage->append($data, $test, $filterGroups);
```

with this block:

```
try {
    $this->codeCoverage->append($data, $test, $filterGroups);
} catch (Exception $cce) {
    // @todo Create specific CC exception?
    $error = TRUE;
    if (!isset($e)) {
        $e = $cce;
    }
}
```

PR #1893, opened 2015-10-04
Remove workaround for phpab bug

theseer:
Remove workaround for theseer/Autoload bug [#63](https://github.com/theseer/Autoload/issues/63)


sebastianbergmann:
Merged manually, thanks.

PR #1895, opened 2015-10-05
Only use the first word after @use or @covers

### LINKED ISSUE #1857: 'uses' and 'covers' should only take a single word
I'd like to comment, why it's OK for a test to `@uses` a different class, or method like

```php
/**
 * @uses Foo\Bar This is a value object
 */
```
This fails with

    Trying to @cover or @use not existing class or interface "Foo\Bar This is a value object".

Because it is not possible to refer to a class, or method with a whitespace PHPUnit should only take the first word after the annotation and should leave everything else as a comment

dstockto:
Attempt to fix #1857

### DIFF
--- a/src/Util/Test.php
+++ b/src/Util/Test.php
@@ -141,6 +141,8 @@ class PHPUnit_Util_Test
 
             $element = preg_replace('/[\s()]+$/', '', $element);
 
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
Merged manually, thanks.

Issue #1892, opened 2015-10-03
coverage text do not honor --colors setting


mgansler:
Issue #1771 still exists with PHPUnit 4.8.10 as well as with 5.0.3

`--coverage-text` shows colors even when `--colors=never` is used.

sebastianbergmann:
Since I neither use `--colors` nor `--coverage-text` myself it is unlikely that I will work on a fix for thix myself. I would review and then merge a pull request that fixes this, of course.

llaville:
@sebastianbergmann I'm sorry to tell you that PHPUnit 4.8.13 is still affected by this issue (not fixed)

https://github.com/sebastianbergmann/phpunit/blob/4.8/src/TextUI/TestRunner.php#L506-L513

sebastianbergmann:
Nothing to be sorry about. The fix has not been backported / cherry-picked to `4.8` yet.

PR #1906, opened 2015-10-07
coverage-text: do not use colors if --colors=never

gkralik:
Don't use colors when using `--coverage-text` along with `--colors=never`.

See #1771 and #1892


gkralik:
Sorry, this was intended for the 5.0 branch, not the 5.1. Closing this.

PR #1907, opened 2015-10-07
coverage-text: do not use colors if --colors=never

gkralik:
Don't use colors when using `--coverage-text` along with `--colors=never`.

See #1771 and #1892 

(replaces #1906)

### DIFF
--- a/src/TextUI/TestRunner.php
+++ b/src/TextUI/TestRunner.php
@@ -489,10 +489,7 @@ class PHPUnit_TextUI_TestRunner extends PHPUnit_Runner_BaseTestRunner
             if (isset($arguments['coverageText'])) {
                 if ($arguments['coverageText'] == 'php://stdout') {
                     $outputStream = $this->printer;
-                    $colors       = $arguments['colors'];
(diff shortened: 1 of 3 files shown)


src/TextUI/TestRunner.php
```diff
@@ -490,6 +490,9 @@ class_exists($arguments['printer'], false)) {
                 if ($arguments['coverageText'] == 'php://stdout') {
                     $outputStream = $this->printer;
                     $colors       = $arguments['colors'];
+                    if($colors == PHPUnit_TextUI_ResultPrinter::COLOR_NEVER) {
```
GrahamCampbell: why not just `$colors       = $arguments['colors'] && !PHPUnit_TextUI_ResultPrinter::COLOR_NEVER`?

GrahamCampbell: just thinking about avoiding a double assignment - maybe your way is more readable though :)

gkralik: yeah, just thought about the same thing... will change it

GrahamCampbell:
:+1:

sebastianbergmann:
Merged manually, thanks.

GrahamCampbell:
:+1:

PR #1909, opened 2015-10-09
Added 'flowId' parameter to each Teamcity message for parallel tests

philipp-kolesnikov:
TeamCity can't handle correctly output from several instances of PHPUnit running in parallel. In my case some tests are reported with "Unknown test name"

According to https://confluence.jetbrains.com/display/TCD8/Build+Script+Interaction+with+TeamCity#BuildScriptInteractionwithTeamCity-MessageFlowId it's possible to use 'flowId' parameter in service message to help TeamCity to do this.


sebastianbergmann:
@SvetlanaZem Is this okay to merge?

SvetlanaZem:
@philipp-kolesnikov Are you using printer class to run tests on TeamCity or with PhpStorm?

philipp-kolesnikov:
I use this on TeamCity only.

SvetlanaZem:
@philipp-kolesnikov function getmypid() may return false in case of errors. Don't we want to process the case? Otherwise parameter flowId will be empty string.

philipp-kolesnikov:
We can use uniqid(getmypid()) like this:
```
diff --git a/src/Util/Log/TeamCity.php b/src/Util/Log/TeamCity.php
index 1fd97c7..3c3326d 100644
--- a/src/Util/Log/TeamCity.php
+++ b/src/Util/Log/TeamCity.php
@@ -29,6 +29,11 @@ class PHPUnit_Util_Log_TeamCity extends PHPUnit_TextUI_ResultPrinter
     private $startedTestName;

     /**
+     * @var string
+     */
+    private $flowId;
+
+    /**
      * @param string $progress
      */
     protected function writeProgress($progress)
@@ -166,6 +171,8 @@ public function printIgnoredTest($testName, Exception $e)
      */
     public function startTestSuite(PHPUnit_Framework_TestSuite $suite)
     {
+        $this->flowId = uniqid(getmypid());
+
         if (!$this->isSummaryTestCountPrinted) {
             $this->isSummaryTestCountPrinted = true;

@@ -272,7 +279,7 @@ private function printEvent($eventName, $params = [])
     {
         $this->write("\n##teamcity[$eventName");

-        $params['flowId'] = getmypid();
+        $params['flowId'] = $this->flowId;

         foreach ($params as $key => $value) {
             $escapedValue = self::escapeValue($value);
```
This code will generate same flowIds if both parallel processes return false from getmypid() at same microsecond. 

If this is not stable enough we can use file-lock mechanism like [here] (http://php.net/manual/en/function.getmypid.php#93753), but I'm afraid that this solution is slower.

SvetlanaZem:
The provided code will treat each test suite (e.g. each test class) as separate process, but if we have single test running in separate process it won't be handled properly.

As I understand the most common problem with "getmypid()" is that is could be disabled in php.ini, therefore we could check it inside constructor (I see decision with "ini_get('disable_functions')", but probably there's something better) and call "getmypid()" only if it's not disable.

In other cases I think it would be enough to return "undefined" if getmypid() == false. TeamCity should process empty strings correctly but it's always better to have checks on both sides.

philipp-kolesnikov:
As I understand, even if we want to run each test method in parallel, we have to have the same flowId as for test suite so TeamCity can decide to which test suite belongs test case.

BTW, can you tell me how to run test methods in parallel and in different process with test suite? I want to play with this on TeamCity and can't find a way to do this.
Thanks!

SvetlanaZem:
It's important to set flowId to unique id related to current thread, otherwise parallel tests output could be messed up. If you have two test methods with the same flowId running in parallel the output from the first test may appear in the second one.   

I'm not TeamCity developer and unfortunately I don't know the answer for your question. The best way would be ask TeamCity team (https://teamcity-support.jetbrains.com). We have really great support team, they'll help you with any of your questions and/or create feature request if the functionality is not implemented yet.

philipp-kolesnikov:
TeamCity has no dedicated PHPUnit runner, so from TeamCity point of view PHPUnit is just a command-line application. As I know, PHPUnit is unable to run two test methods from one suite in parallel in different processes, correct me if I'm wrong.

Also, according to TeamCity documentation "All the individual test messages are to appear between testSuiteStarted and testSuiteFinished (in that order) with the same name attributes.". So, even if we have two parallel test methods we have to print them like this:
 ```
##teamcity[testSuiteStarted name='suite1' flowId='1']
##teamcity[testStarted name='test1' flowId='1']
##teamcity[testFinished name='test1' duration='1' flowId='1']
##teamcity[testSuiteFinished name='suite1' flowId='1']

##teamcity[testSuiteStarted name='suite1' flowId='2']
##teamcity[testStarted name='test2' flowId='2']
##teamcity[testFinished name='test2' duration='1' flowId='2']
##teamcity[testSuiteFinished name='suite1' flowId='2']
```
All messages inside <testSuiteStarted>/<testSuiteFinished> have to have the sane flowId. Otherwise TeamCity will be unable to understand what  is test suite test case belongs to.

**Comment by SvetlanaZem** [comment]
You can use attribute "@runInSeparateProcess" for single tests too.

For the code below 'getmypid()' will be different for 'testFirst' and 'testSecond'
```php
class NewProcessTest extends PHPUnit_Framework_TestCase
{
    /**
    * @runInSeparateProcess
    * @preserveGlobalState disabled
    */
    public function testFirst()
    {
        var_dump("runInSeparateProcess: " . getmypid());
    }

    public function testSecond()
    {
        var_dump(getmypid());
    }
}
```

> "All the individual test messages are to appear between testSuiteStarted and testSuiteFinished (in that order) with the same name attributes."

From my point of view it means only messages order, not flowId, e.g. it can printed as 
```
##teamcity[testSuiteStarted name='suite1' flowId='1']
##teamcity[testStarted name='test1' flowId='1']
##teamcity[testFinished name='test1' duration='1' flowId='1']
##teamcity[testStarted name='test2' flowId='2']
##teamcity[testFinished name='test2' duration='1' flowId='2']
##teamcity[testSuiteFinished name='suite1' flowId='1']
```

I think the best way would be to check it on real TeamCity installation to be sure that the fix will solve your problem. Is it possible to do it?

SvetlanaZem:
BTW there is another approach to run PHPUnit on TeamCity (https://github.com/JetBrains/meta-runner-power-pack/tree/master/php), it's suitable for previous PHPUnit versions too.

philipp-kolesnikov:
@runInSeparateProcess annotation only means that test methods will run in separate processes and not in parallel but one by one. Therefore, it is not necessary to use different ‘flowId’ parameter for each case.

To illustrate this I wrote sample test project with three test classes that I run in parallel. You can find source code here https://github.com/philipp-kolesnikov/phpunittest

Parallel1Test.php and Parallel2Test.php contains test methods with similar names to check how TeamCity will handle parallel test methods with similar names from parallel test suites.
Parallel3Test.php contains test method that uses data provider.
In some methods I added sleeps to check if they are run in parallel or not.

I run this test project using two different approaches how to use ‘flowId’ parameter:
- Each test method uses its own ‘flowId’. Since getmypid() used in TeamCity.php returns pid of phpunit process but not pid of test method, I use suite name plus test name as ‘flowId’. I didn’t run Parallel3Test.php here, but this is not important for result.
- All test methods inside test suite use the same ‘flowId’ and this ‘flowId’ is used for test suite related service messages

I run this project on latest TeamCity.

In first case TeamCity found all test names correctly but was unable to find out which test suite method belongs to. 
first_approach
In second case TeamCity found all test names and test suites correctly.
second_approach
As you can see, it’s necessary to have the same ‘flowId’ parameter on all test flow starting from ‘testSuiteStarted’ message.
To ensure that test methods inside test class executed one by one and in separate process it’s better to run project from command line. Output will be something like this:
```
PHPUnit 5.0.3-14-gd31d78c by Sebastian Bergmann and contributors.
 ##teamcity[testCount count='3' flowId='43256283d70ba28f']
 ##teamcity[testSuiteStarted name='ParallelTest2' locationHint='php_qn://C:\Users\Philipp\work\phpunittest\tests\Parallel2Test.php::\ParallelTest2' flowId='43256283d70ba28f']
 PHPUnit 5.0.3-14-gd31d78c by Sebastian Bergmann and contributors.
 ##teamcity[testCount count='3' flowId='476056283d70aae85']
 ##teamcity[testSuiteStarted name='ParallelTest1' locationHint='php_qn://C:\Users\Philipp\work\phpunittest\tests\Parallel1Test.php::\ParallelTest1' flowId='476056283d70aae85']
 ##teamcity[testStarted name='testFirst' locationHint='php_qn://C:\Users\Philipp\work\phpunittest\tests\Parallel1Test.php::\ParallelTest1::testFirst' flowId='476056283d70aae85']
 PHPUnit 5.0.3-14-gd31d78c by Sebastian Bergmann and contributors.
 ##teamcity[testCount count='3' flowId='1077256283d70bbcb0']
 ##teamcity[testSuiteStarted name='ParallelTest3' locationHint='php_qn://C:\Users\Philipp\work\phpunittest\tests\Parallel3Test.php::\ParallelTest3' flowId='1077256283d70bbcb0']
 ##teamcity[testSuiteStarted name='testFirst' locationHint='php_qn://C:\Users\Philipp\work\phpunittest\tests\Parallel3Test.php::\ParallelTest3::testFirst' flowId='1077256283d70bbebb']
 ##teamcity[testStarted name='testFirst' locationHint='php_qn://C:\Users\Philipp\work\phpunittest\tests\Parallel2Test.php::\ParallelTest2::testFirst' flowId='43256283d70ba28f']
 ##teamcity[testStarted name='testFirst with data set #0' locationHint='php_qn://C:\Users\Philipp\work\phpunittest\tests\Parallel3Test.php::\ParallelTest3::testFirst with data set #0' flowId='1077256283d70bbebb']
 ##teamcity[testFinished name='testFirst' duration='0' flowId='43256283d70ba28f']
 ParallelTest2::testFirst pid is 7304
 ##teamcity[testStarted name='testSecond' locationHint='php_qn://C:\Users\Philipp\work\phpunittest\tests\Parallel2Test.php::\ParallelTest2::testSecond' flowId='43256283d70ba28f']
 ##teamcity[testFinished name='testFirst with data set #0' duration='0' flowId='1077256283d70bbebb']
 ParallelTest3::testFirst with data set "0" pid is 2028
 ##teamcity[testStarted name='testFirst with data set #1' locationHint='php_qn://C:\Users\Philipp\work\phpunittest\tests\Parallel3Test.php::\ParallelTest3::testFirst with data set #1' flowId='1077256283d70bbebb']
 ##teamcity[testFinished name='testFirst with data set #1' duration='0' flowId='1077256283d70bbebb']
 ParallelTest3::testFirst with data set "1" pid is 12112
 ##teamcity[testStarted name='testFirst with data set #2' locationHint='php_qn://C:\Users\Philipp\work\phpunittest\tests\Parallel3Test.php::\ParallelTest3::testFirst with data set #2' flowId='1077256283d70bbebb']
 ##teamcity[testFinished name='testFirst with data set #2' duration='0' flowId='1077256283d70bbebb']
 ParallelTest3::testFirst with data set "2" pid is 8928
 ##teamcity[testSuiteFinished name='testFirst' flowId='1077256283d70bbebb']
 ##teamcity[testSuiteFinished name='ParallelTest3' flowId='1077256283d70bbebb']
 Time: 1.44 seconds, Memory: 4.00Mb
 OK (3 tests, 0 assertions)
 ##teamcity[testFinished name='testFirst' duration='2000' flowId='476056283d70aae85']
 ParallelTest1::testFirst pid is 8768
 ##teamcity[testStarted name='testSecond' locationHint='php_qn://C:\Users\Philipp\work\phpunittest\tests\Parallel1Test.php::\ParallelTest1::testSecond' flowId='476056283d70aae85']
 ##teamcity[testFinished name='testSecond' duration='0' flowId='476056283d70aae85']
 ParallelTest1::testSecond pid is 11296
 ##teamcity[testStarted name='testLast' locationHint='php_qn://C:\Users\Philipp\work\phpunittest\tests\Parallel1Test.php::\ParallelTest1::testLast' flowId='476056283d70aae85']
 ##teamcity[testFinished name='testSecond' duration='2000' flowId='43256283d70ba28f']
 ParallelTest2::testSecond pid is 10888
 ##teamcity[testStarted name='testLast' locationHint='php_qn://C:\Users\Philipp\work\phpunittest\tests\Parallel2Test.php::\ParallelTest2::testLast' flowId='43256283d70ba28f']
 ##teamcity[testFinished name='testLast' duration='0' flowId='476056283d70aae85']
 ParallelTest1::testLast pid is 10536
 ##teamcity[testSuiteFinished name='ParallelTest1' flowId='476056283d70aae85']
 Time: 3.47 seconds, Memory: 3.75Mb
 OK (3 tests, 0 assertions)
 ##teamcity[testFinished name='testLast' duration='0' flowId='43256283d70ba28f']
 ParallelTest2::testLast pid is 1852
 ##teamcity[testSuiteFinished name='ParallelTest2' flowId='43256283d70ba28f']
 Time: 3.53 seconds, Memory: 3.75Mb
 OK (3 tests, 0 assertions)
```

philipp-kolesnikov:
Actually, we already use this wrapper https://github.com/maartenba/phpunit-runner-teamcity with flowId added to each message as in second case in comment above. This wrapper internally looks the same as TeamCity logger in PHPUnit 5.0 so I wanted to move to builtin solution.

philipp-kolesnikov:
It looks like using uniqid(getmypid()) is also bad solution. In case of nested test suites only last one will have the same flowId with test cases and will be handled correctly because uniqid() depends on time.

Examples:
Test reporting with uniqid(getmypid())
second_approach
Test reporting with getmypid() only
withoutuniqid

Second case looks much better :)

So, I want to summarize:
1. All messages produced by one running instance of phpunit should have same flowId
2. We use getmypid() as flowId in our testing and I've never seen that getmypid() returned false. Actually, I can't imagine case when running process can't get its own pid
3. But I agree that since getmypid() can return false in some case, we can try to find out another unique identifier.

As for me personally use getmypid() is stable enough solution.

SvetlanaZem:
I think we could use "getmypid()", but we should check that it's not disabled in php.ini. It looks like it's the most common case when "getmypid()" returns false. "flowId" option is not necessary for tests running in PhpStorm, therefore it would be annoying to have a lot of PHP warning (e.g. "Warning: getmypid() has been disabled for security reasons").
I think the check could be implemented in constructor with "ini_get('disable_functions')" and don't add "flowId" option at all if the function is disabled. But if you know better way please go ahead with it.

philipp-kolesnikov:
I checked getmypid() implementation in PHP source code. It never returns false on all POSIX and all supported Windows systems starting from XP. So it's safe to use this function.
I also added check that getmypid() is not disabled in php.ini

SvetlanaZem:
Great! 
The only thing, the check "!stripos(ini_get("disable_functions"), "getmypid")" should be replaced with "stripos(ini_get("disable_functions"), "getmypid") === false". Otherwise stripos() function might return 0 and it will be silently cast to false.

philipp-kolesnikov:
Done!

SvetlanaZem:
The fix is ok for me. 
@sebastianbergmann Would you please merge the change?

sebastianbergmann:
Merged manually, thanks.

PR #1910, opened 2015-10-13
Return simple object mock

chris1312:
When I want to create a simple mock which would not invoke original constructor, I need to call 

``` php
$this->getMock($class, $methods, ['someMethod'], '', false);
```

This is the most popular case (at least for me) so it really simplifies the things.

Now I can write just 

``` php
$this->getSimpleMock('MyClasToBeMocked');
```

or

``` php
$this->getSimpleMock('MyClasToBeMocked', ['someMethod');
```

without calling getMock() with a ton of parameters.


sebastianbergmann:
`getMockWithoutInvokingTheOriginalConstructor()` already exists and does what you want.

PR #1915, opened 2015-10-15
Allow @coversDefaultClass annotation to work on traits

Firehed:
Using a `@coversDefaultClass` annotation would previously fail when used with a trait with the message `Trying to @cover or @use not existing class or interface "Name\Of\Trait"`. This minor change allows traits to be used as a default coverage target.

If setting up a different annotation is preferable I can look into adding it, but traits already work as expected with the standard `@covers` annotations.


sebastianbergmann:
Please send a new pull request against the `5.2` branch.

PR #1934, opened 2015-11-04
allow unordered occurence of phpType elements

BreiteSeite:
With this change, the ordering of the elements under the `<php>` element in the xml configuration does not matter anymore.

Prior this change, the XML did not validate against the `phpunit.xsd` if - for example - a `env` tag was inserted after a `server` tag inside the `php` tag.

Thank you for considering this pull request.

### DIFF
--- a/phpunit.xsd
+++ b/phpunit.xsd
@@ -161,17 +161,19 @@
   </xs:attributeGroup>
   <xs:complexType name="phpType">
     <xs:sequence>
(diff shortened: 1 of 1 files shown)


sun:
:+1: Very good idea - thanks for contributing!

BreiteSeite:
I thought you closed it without merging it, but the commit is in the master branch with the ID 496745aeba741e63b7149da3e1f712d441751182. Unfortunately, this does not show up in this PR. Weird.

PR #1938, opened 2015-11-06
Path coverage switch

Maks3w:
Requires php-code-coverage v3.1 (https://github.com/sebastianbergmann/php-code-coverage/pull/400)

Turned off by default because is resource expensive and requires Xdebug >= 2.3.2

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -3,6 +3,7 @@
          xsi:noNamespaceSchemaLocation="phpunit.xsd"
          bootstrap="tests/bootstrap.php"
          backupGlobals="false"
+         enablePathCoverage="true"
          verbose="true">
   <testsuites>
     <testsuite name="small">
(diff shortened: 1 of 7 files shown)


phpunit.xml.dist
```diff
@@ -19,7 +19,7 @@
   </testsuites>
 
   <filter>
-    <whitelist processUncoveredFilesFromWhitelist="true">
+    <whitelist processUncoveredFilesFromWhitelist="true" enablePathCoverage="true">
```
sebastianbergmann: This does not belong on the `<whitelist>` element. We do not, however, have an element for configuring generic code coverage settings like this. So the only place to put this for now would be the root `<phpunit>` element unfortunately.

Maks3w:
@sebastianbergmann done

sebastianbergmann:
This pull request was automatically closed because I deleted the `master` branch and then renamed `5.2` to `master`. It was too confusing for many developers that new development would not go into `master` for the time being. Sorry for any inconvenience caused, please open a new pull request against `master`.

Issue #1342, opened 2014-07-23
Replace PHP process isolation code with Symfony Process component


sun:
Platform-specific process isolation bugs like #1340 are exceptionally hard to debug.

The [Symfony Process component](https://github.com/symfony/process) is a decoupled and very lightweight library.  In essence, a robust wrapper around `proc_open()`.

Proposal:  Replace the `PHPUnit_Util_PHP_*` classes with Symfony `Process`.

Doing this should be relatively straightforward.  However, I don't know whether the backwards compatibility policy also applies to rather internal code of the test runner itself?

sebastianbergmann:
I am no longer convinced that this is the right way to go.

allan-simon:
@sebastianbergmann  Hello  , is it possible to why it's not the right way to go and if you have an other idea on how to tackle the issue of `@runInSeparateProcess` being extremly hard to debug

I spend several days , without success , trying to debug a test failing only in CI environment

the test failed with this error (I put it here in case it help people googling it finding this ticket) 

```
There was 1 error:
1) App\Tests\IntegrationTest\EndUserRegisterTest::testEndUserThatRegistedAndManuallyEnterInvitationCanAccessDashboard
PHPUnit\Framework\Exception: <!-- Serialization of 'Closure' is not allowed (500 Internal Server Error) -->
<!DOCTYPE html>
[lot of Symfony generic error 500 HTML]
*]]>*/        </script>
    </body>
</html>
<!-- Serialization of 'Closure' is not allowed (500 Internal Server Error) -->
Caused by
ErrorException: unserialize(): Error at offset 0 of 53820 bytes in /builds/rosaly-fintech/backend-php/vendor/phpunit/phpunit/src/Util/PHP/AbstractPhpProcess.php:289
Stack trace:
#0 [internal function]: PHPUnit\Util\PHP\AbstractPhpProcess::PHPUnit\Util\PHP\{closure}(8, 'unserialize(): ...', '/builds/rosaly-...', 289)
#1 /builds/rosaly-fintech/backend-php/vendor/phpunit/phpunit/src/Util/PHP/AbstractPhpProcess.php(289): unserialize('<!-- Serializat...')
[...]

PR #1942, opened 2015-11-08
Always write the template in order to make it easier to debug isolation mode

dawehner:
Debugging the isolation mode can be tricky as xdebug doesn't help you
with the initial $job piped into the php process.

As solution this PR uses the same technique as the windows util,
which is dumping the template into a temporary file and include it from there.


sebastianbergmann:
I think #1342 is the better solution.

dawehner:
@sebastianbergmann 
This is an additional / orthogonal issue. #1342 won't solve the problem how we deal with the actual executed PHP code, though I'm happy to work on #1342, actually have a local branch for that already

PR #1949, opened 2015-11-10
Throw InvalidCoversTargetException for extended covers tags

oradwell:
When an invalid class is given in `@covers` tag with `<extended>` no exception is thrown.

This causes a warning to be triggered:

```
class_implements(): Class NotExistingClass does not exist and could not be loaded
```

Also see: https://github.com/ockcyp/covers-validator/issues/5
Thanks @willemstuursma for finding it out


ockcyp:
will create new one for 5.2

PR #1950, opened 2015-11-11
Rollback version of DbUnit for PHP 5.3

nishigori:
It's revert to part of 54ee96a

DbUnit ~2.0 is supporting PHP version >= 5.4.
DbUnit ~2.0 is using array-short-syntax (`[]`), case PHP 5.3, it's occurred syntax error.
PHPUnit 4.8.17 bundles DbUnit 2.0.2..

PHPUnit 4.8 supports PHP version >= 5.3 now.

Hope to continue using DbUnit 1.4 **for PHP 5.3 user**
(I think something is not one's real intention, too..)

or PHPUnit over then 4.8.16 will be not supporting PHP 5.3?

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -91,7 +91,7 @@
         <exec executable="php">
             <arg value="composer.phar"/>
             <arg value="require"/>
-            <arg value="phpunit/dbunit:~2.0.2"/>
+            <arg value="phpunit/dbunit:~1.4"/>
             <arg value="phpunit/phpunit-selenium:~1.4"/>
             <arg value="phpunit/php-invoker:~1.1"/>
         </exec>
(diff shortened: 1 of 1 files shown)


nishigori:
using PHP 5.3, output detail exam:

```sh
$ php --version
PHP 5.3.3 (cli) (built: Jul  9 2015 15:04:50)
Copyright (c) 1997-2010 The PHP Group
Zend Engine v2.3.0, Copyright (c) 1998-2010 Zend Technologies

$ php phpunit.phar --version
PHP Parse error:  syntax error, unexpected '[' in phar:///var/tmp/phpunit.phar/dbunit/Extensions/Database/DataSet/AbstractDataSet.php on line 34
```

PR #1951, opened 2015-11-12
Throw InvalidCoversTargetException for extended covers tags

oradwell:
When an invalid class is given in `@covers` tag with `<extended>` no exception is thrown.

This causes a warning to be triggered:

```
class_implements(): Class NotExistingClass does not exist and could not be loaded
```

Also see: https://github.com/ockcyp/covers-validator/issues/5
Thanks @willemstuursma for finding it out

### DIFF
--- a/src/Util/Test.php
+++ b/src/Util/Test.php
@@ -978,6 +978,18 @@ class PHPUnit_Util_Test
             $classes = [$element];
 
             if ($extended) {
+                if (!class_exists($element) &&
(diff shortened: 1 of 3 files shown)


sebastianbergmann:
This pull request was automatically closed because I deleted the `master` branch and then renamed `5.2` to `master`. It was too confusing for many developers that new development would not go into `master` for the time being. Sorry for any inconvenience caused, please open a new pull request against `master`.

PR #1956, opened 2015-11-17
Fix insulated tests with phpdbg

### LINKED ISSUE #1955: @runTestsInSeparateProcesses fails when running tests with phpdbg -qrr
Running tests in a separate process does not work when run with the `phpdbg -qrr phpunit`

nicolas-grekas:
Fixes #1955

### DIFF
--- a/src/Util/PHP/Default.php
+++ b/src/Util/PHP/Default.php
@@ -30,9 +30,14 @@ class PHPUnit_Util_PHP_Default extends PHPUnit_Util_PHP
     public function runJob($job, array $settings = array())
     {
         $runtime = new Runtime;
+        $runtime = $runtime->getBinary() . $this->settingsToParameters($settings);
+
+        if ('phpdbg' === PHP_SAPI) {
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
Nice fix. :+1:

sebastianbergmann:
Thanks. I do think, though, that this should go into the `Runtime` class. Or?

nicolas-grekas:
It could but it's not trivial: -d options needs to be before the eval-stdin.php on the command line, which means that simple concatenation doesn't work...

sebastianbergmann:
Okay. Sounds like one more reason to work on #1342 then.

bwoebi:
This, … well https://bugs.php.net/73484

This doesn't fix the bug, it just ends up not running the tests at all due to the use of php://input instead of php://stdin.

sebastianbergmann:
Pull requests welcome! :-)

Truth be told, neither process isolation nor PHPDBG compatibility are high priority topics for me.

bwoebi:
Yeah, see #2356 ;-)

PR #1963, opened 2015-11-21
Fixes --filter

marcioAlmada:
Fixes --filter option since migration to strict types. Right now this code is emitting type errors:
- `strlen` expects a string
- `debug_backtrace` expects an int not bool


sebastianbergmann:
Please do not send pull requests against `master` at this point in time.

PR #1968, opened 2015-11-24
Throw exception when @testWith annotation contains invalid datasets

acoulton:
The existing code was silently ignoring invalid JSON if it had
found at least one valid dataset. It was using this behaviour
to detect the end of the @testWith annotation (see for eg the
testTestWithSimpleTextAfter test, which fails if a simple
throw on invalid JSON is used).

This fix treats a line as being part of the @testWith if it begins
with a `[` character - since all datasets are required to be JSON
arrays this should be reliable.

It can then reliably throw if invalid JSON is found for any of the
datasets.

### DIFF
--- a/src/Util/Test.php
+++ b/src/Util/Test.php
@@ -469,9 +469,15 @@ class PHPUnit_Util_Test
             $data              = array();
             foreach (explode("\n", $annotationContent) as $candidateRow) {
                 $candidateRow = trim($candidateRow);
+                if ($candidateRow[0] !== '[') {
+                    break;
+                }
                 $dataSet      = json_decode($candidateRow, true);
                 if (json_last_error() != JSON_ERROR_NONE) {
-                    break;
+                    $error = function_exists('json_last_error_msg') ? json_last_error_msg() : json_last_error();
+                    throw new PHPUnit_Framework_Exception(
+                        'The dataset for the @testWith annotation cannot be parsed: '.$error
+                    );
                 }
                 $data[] = $dataSet;
             }
--- a/tests/Util/TestTest.php
+++ b/tests/Util/TestTest.php
@@ -346,10 +346,22 @@ class Util_TestTest extends PHPUnit_Framework_TestCase
     {
         $this->setExpectedExceptionRegExp(
             'PHPUnit_Framework_Exception',
-            '/^The dataset for the @testWith annotation cannot be parsed.$/'
+            '/^The dataset for the @testWith annotation cannot be parsed:/'
         );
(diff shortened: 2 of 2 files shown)


giorgiosironi:
What about including `json_last_error_msg()` in the exception message in the PHP versions where this function is available (checking with `function_exists`)?

acoulton:
@giorgiosironi thanks for the suggestion, that's a good idea - I've updated the PR.

@sebastianbergmann any chance of this being merged? The prospect of having test cases that are silently skipped feels like quite a major issue to me. 

I've checked and the build passes apart from on 5.3, which looks like a pre-existing composer / travis issue with the 4.8 branch.

acoulton:
@sebastianbergmann I was bitten by this (test case silently skipped because of a typo in the JSON, meaning a build that should have been red was green) again today. Any chance you could look at it?

Otherwise I may need to migrate from `@testWith` back to explicit PHP data providers as the current implementation just isn't robust enough for me.

acoulton:
@sebastianbergmann sorry to nag but 7 months later the JSON `@testWith` annotations are still highly unreliable - one or more test cases will be silently skipped if there are any parse errors after the first line.

It would be really good to see this merged (or another fix applied). Please let me know if I can do/provide anything further.

src/Util/Test.php
```diff
@@ -469,9 +469,15 @@ public static function getDataFromTestWithAnnotation($docComment)
             $data              = array();
             foreach (explode("\n", $annotationContent) as $candidateRow) {
                 $candidateRow = trim($candidateRow);
+                if ($candidateRow[0] !== '[') {
+                    break;
+                }
                 $dataSet      = json_decode($candidateRow, true);
                 if (json_last_error() != JSON_ERROR_NONE) {
-                    break;
+                    $error = function_exists('json_last_error_msg') ? json_last_error_msg() : json_last_error();
```
sebastianbergmann: Why would `json_last_error_msg()` not be available?

acoulton: It was only added in PHP 5.5.

sebastianbergmann: Right, and PHPUnit 4.8 still needs to support that. Repressed that fact, sorry.

acoulton: No worries, it's so long ago I had repressed it at first too :)

sebastianbergmann:
Merged manually, thanks.

acoulton:
That's great, thanks very much.

acoulton:
And sorry about the coding standards issues.

sebastianbergmann:
No worries.

PR #1977, opened 2015-11-30
Disable code coverage ignore phpunit5

abacaphiliac:
- add new --disable-ignore-coverage flag to disable @codeCoverageIgnore annotation in coverage reports.
- add --disable-ignore-coverage to help menu, and update help.phpt and help2.phpt tests to support the change.
- add two new phpt tests to demonstrate functionality and output when --disable-ignore-coverage is included and when it is excluded by printing the coverage summary text to stdout.


sebastianbergmann:
Merged manually, thanks.

PR #1983, opened 2015-12-05
Allow skipping of testcases that have have @expectedException annotation

dshafik:
This allows for something like:

``` php
/**
  * @expectedException \PHPUnit_Framework_Error_Warning
  */
function testSomething() {
     if (!version_compare(PHP_VERSION, '7.0.0', '<')) {
         $this->markTestSkipped("PHP < 7.0.0 is required");
     }
}
```

Currently it fails with:

`Failed asserting that exception of type "PHPUnit_Framework_SkippedTestError" matches expected exception "\PHPUnit_Framework_Error_Notice". Message was: "PHP < 7.0.0 is required"`

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -917,7 +917,7 @@ abstract class PHPUnit_Framework_TestCase extends PHPUnit_Framework_Assert imple
         if (isset($e)) {
             $checkException = false;
 
-            if (is_string($this->expectedException)) {
+            if (!($e instanceof PHPUnit_Framework_SkippedTestError) && is_string($this->expectedException)) {
                 $checkException = true;
 
                 if ($e instanceof PHPUnit_Framework_Exception) {


dshafik:
This allows for something like:

```php
/**
  * @expectedException \PHPUnit_Framework_Error_Warning
  */
function testSomething() {
     if (version_compare(PHP_VERSION, '7.0.0', '>=')) {
         $this->markTestSkipped("PHP < 7.0.0 is required");
     }
}
```

Currently it fails with:

`Failed asserting that exception of type "PHPUnit_Framework_SkippedTestError" matches expected exception "\PHPUnit_Framework_Error_Notice". Message was: "PHP < 7.0.0 is required"`

sebastianbergmann:
Would the right way not be to use `@requires PHP 7.0.0`?

dshafik:
In this case I want it to be skipped for >= 7.0.0 (e.g. only run on 5.x); but that's just one potential of an infinite number of reasons why you would want to skip a test and you can't if it has an `@expectedException` annotation.

sebastianbergmann:
That could be expressed using `@requires < PHP 7`.

dshafik:
I don't think it can... the check is a `version_compare()` with an explicit `<` check; unless I am mistaken and [this](https://github.com/sebastianbergmann/phpunit/blob/master/src/Util/Test.php#L229) is not it?

Also, what if you did:

```php
if (ini_get('safe_mode')) {
    $this->markTestSkipped('Safe mode not supported');
}
```

The reason for skipping could be completely arbitrary, so the @requires annotation is not the answer to the problem this PR solves.

dshafik:
(FWIW, I'd welcome the ability to do that, and am happy to come up with a patch; should be reasonably trivial and possible to keep BC easily)

sebastianbergmann:
Damn, you're right. `getMissingRequirements()` should be extended to support arbitrary operators for `version_compare()`. Would be great if you could implement that for PHPUnit 5.3 (`master`).

I need to think some more about this pull request.

dshafik:
@sebastianbergmann if you want to open up a ticket for that change and assign it to me I'll get it done tomorrow (Saturday) :)

sebastianbergmann:
I created https://github.com/sebastianbergmann/phpunit/issues/1984 for you, @dshafik.

rentalhost:
I think that this PR continues valid, even if `@requires` is improved. I think in cases where I can skip tests by my own system implementation. For instance:

```
/**
 * @expectedException App\SomeException
 */
public function testSomeException() {
    if (System::checkSomething() === false) {
        $this->markTestSkipped('unavaiable on system');
    }

    Does::something();
}
```

It could depends of external factors that phpunit can't handle alone. A system env variable, for instance.

rentalhost:
Maybe the unique way of it fails is if you try use `@expectedException \PHPUnit_Framework_SkippedTestError`, but it make no sense for general purposes.

dharkness:
I agree that skipping a test should *not* result in a failure. Skipping should *always* mark the test skipped.

sebastianbergmann:
Merged manually, thanks.

Issue #1984, opened 2015-12-05
Add support for comparison operators to @requires annotation


sebastianbergmann:
Currently only

```php
/**
 * @requires PHP 7.0.0
 */
```

is supported, for instance and evaluated by

```php
version_compare(PHP_VERSION, $required['PHP'], '<')
```

in https://github.com/sebastianbergmann/phpunit/blob/master/src/Util/Test.php#L229.

In addition to `@requires PHP 7.0.0` also the following should be supported

* `@requires >= PHP 7.0.0` (has to be the equivalent of `@requires PHP 7.0.0` for backwards compatibility)
* `@requires = PHP 7.0.0`
* `@requires != PHP 7.0.0`
* `@requires < PHP 7.0.0`
* `@requires <= PHP 7.0.0`
* `@requires > PHP 7.0.0`

PR #1986, opened 2015-12-05
Add support for [optional] operators in @requires and @extension annotations

dshafik:
This adds support for optional operators in both `@requires` for PHP and PHPUnit, as well as for `@extension`.

Default is `>=` and the `version_compare` has been negated to account for that (previously was `<`) — this is so that the error message
makes sense (`PHP >= 7.0.0 is required` vs `PHP < 7.0.0 is required`) without needed to juggle based on the operator.

I can add the ability to change the operators to words e.g.:
- `>` becomes `PHP 7.0.0 or older is not supported`
- `>=` becomes `PHP 7.0.0 or greater is required` 
- `<` becomes `PHP 7.0.0 or greater is not supported`
- `<=` becomes `PHP 7.0.0 or older is required`
- `=` and `==` becomes `PHP 7.0.0 is required`
- `!=` and `<>` becomes `PHP 7.0.0 is not supported`

Refs #1984


sebastianbergmann:
Merged manually, thanks!

PR #1990, opened 2015-12-07
Ensure the dependencies work on php 5.3.3

GrahamCampbell:


composer.json
```diff
@@ -61,6 +61,11 @@
             "tests/_files/CoveredFunction.php"
         ]
     },
+    "config": {
+        "platform": {
+            "php": "5.5.3"
```
stof: typo: should be 5.3.3, not 5.5.3

GrahamCampbell: OOPS!

GrahamCampbell: I'm so used to writing 5.5. ;)

PR #1991, opened 2015-12-07
Allow throwing any Throwable

trowski:
Simple update related to https://github.com/sebastianbergmann/phpunit-mock-objects/pull/282 so instances of Error can also be thrown.


sebastianbergmann:
Merged manually, thanks.

Issue #1999, opened 2015-12-10
Handler is inherited from previous custom option with handler


eelf:
With own TextUI_Command and handler, handler for one option is called for other option. See PR

sebastianbergmann:
Fixed via 39bb51bdcbd743df8e03a353870f125349394a30.

PR #2000, opened 2015-12-10
Reset previous handler before processing option

eelf:
Fixes https://github.com/sebastianbergmann/phpunit/issues/1999


sebastianbergmann:
Merged manually into `5.7`, `6.0`, and `master`.

PR #2006, opened 2015-12-11
Fixed php version constraint

GrahamCampbell:
`~5.4` already includes all 5.x versions greater or equal to 5.4.0, so `~5.5` means nothing since it's a subset of `~5.4`.

`^5.3.3` is probably the shortest way to express what you want here.


vinkla:
:+1:

Issue #2013, opened 2015-12-21
Testsuite with name="directory" causing fatal error


redaxmedia:
Hello Sebastian,

please rename PHPUnit's internal **small** testsuite to **directory** to reproduce following error:

<pre>
henry:~/Projekte/phpunit$ php phpunit.phar --configuration phpunit.xml --testsuite directory
PHP Fatal error:  Uncaught PHPUnit_Framework_Exception: Class "Directory" does not extend PHPUnit_Framework_TestCase.

phar:///home/henry/Projekte/phpunit/phpunit.phar/phpunit/Framework/TestSuite.php:167
phar:///home/henry/Projekte/phpunit/phpunit.phar/phpunit/Util/Configuration.php:823
phar:///home/henry/Projekte/phpunit/phpunit.phar/phpunit/Util/Configuration.php:804
phar:///home/henry/Projekte/phpunit/phpunit.phar/phpunit/TextUI/Command.php:663
phar:///home/henry/Projekte/phpunit/phpunit.phar/phpunit/TextUI/Command.php:117
phar:///home/henry/Projekte/phpunit/phpunit.phar/phpunit/TextUI/Command.php:106

  thrown in phar:///home/henry/Projekte/phpunit/phpunit.phar/phpunit/Framework/TestSuite.php on line 167</pre>

Can you confirm this is a valid error? Tested with latest version of PHPUnit.

sebastianbergmann:
Can you provide a minimal, reproducing example?

redaxmedia:
Just rename PHPUnit's internal testsuite **small** testsuite to **directory**:

<pre>
&lt;testsuite name=&quot;directory&quot;&gt;
  &lt;directory suffix=&quot;Test.php&quot;&gt;tests/Framework&lt;/directory&gt;
  &lt;directory suffix=&quot;Test.php&quot;&gt;tests/Extensions&lt;/directory&gt;
  &lt;directory suffix=&quot;Test.php&quot;&gt;tests/Runner&lt;/directory&gt;
   &lt;directory suffix=&quot;Test.php&quot;&gt;tests/Util&lt;/directory&gt;
&lt;/testsuite&gt;
</pre>

original-brownbear:
Added PR #2022 

I don't think this should be fixed, only documented. If I understand the existing test cases for this constructor correctly it is expected behavior to throw an exception in case the test suit name collides with an existing class that is not a subclass of PHPUnit_Framework_TestCase.
[...]

PR #2022, opened 2016-01-04
Adding test defining issue #2013

original-brownbear:
_Added test for test suit name "Directory".
*_Indicated the use of PHP internal classes being treated the same way
as using a test suit by the name of an existing class that does not
subclass PHPUnit_Framework_TestCase.


sebastianbergmann:
`Directory` is a built-in class. What is described in #2013 is expected behavior in case a string containing a class name is passed to the constructor of `PHPUnit_Framework_TestSuite` (https://github.com/sebastianbergmann/phpunit/blob/master/src/Framework/TestSuite.php#L129).

PR #2032, opened 2016-01-08
Reverted broken change

GrahamCampbell:


GrahamCampbell:
1.5.0 doesn't exist. It's breaking all installs that require stable dependencies.

PR #2036, opened 2016-01-13
Add test failure data to JSON output

matijn-frank:
When using the --log-json, it does not have the actual and expected data when an assertEquals test fails. With this fix it does, which can be very helpful. (We use the json output in automated tests)


sebastianbergmann:
Is this related to #2037?

matijn-experty:
Yes

sebastianbergmann:
Closing this in favor of #2037.

PR #2037, opened 2016-01-13
Log more information about failures in JSON output

inwaar:
Current JSON logger doesn't log comparison information when there is a failure on assertEquals. This will add logging of comparison diffs.


sebastianbergmann:
Is this related to #2036?

inwaar:
Yeah, it solves the same issue.

sebastianbergmann:
Merged manually, thanks.

PR #2047, opened 2016-01-23
Add 'it_' and 'its_' as test method prefixes

jpkleemans:
I like to write the name of my test methods in a BDD way:

``` php
function it_sends_messages()
{
    // ...
}
```

Instead of:

``` php
function testSendMessage()
{
    // ...
}
```

Currently I have to add the `@test` annotation to all my test methods.
This PR adds support for methods starting with  `it_` and `its_` to be recognized as test methods.

### DIFF
--- a/src/Framework/TestSuite.php
+++ b/src/Framework/TestSuite.php
@@ -899,8 +899,12 @@ class PHPUnit_Framework_TestSuite implements PHPUnit_Framework_Test, PHPUnit_Fra
      */
     public static function isTestMethod(ReflectionMethod $method)
     {
-        if (strpos($method->name, 'test') === 0) {
-            return true;
+        $testMethodPrefixes = ['test', 'it_', 'its_'];
+
+        foreach ($testMethodPrefixes as $prefix) {
+            if (strpos($method->name, $prefix) === 0) {
+                return true;
+            }
         }
 
         // @scenario on TestCase::testMethod()


jpkleemans:
This naming style is common in BDD frameworks like PHPSpec, RSpec, Jasmine, etc.

sebastianbergmann:
This would have to wait for PHPUnit 6.

jpkleemans:
@sebastianbergmann thanks for your response! Could you explain why? It doesn't break backwards compatibility, does it?

sebastianbergmann:
It does, actually, since it adds "special meaning" to certain methods.

jpkleemans:
Ah, I understand. Didn't thought about that... Thanks!

unfunco:
Whilst the change is small and wouldn't require much future maintenance, this change would likely only benefit @jpkleemans – if one's preference for syntax is BDD over TDD, then one should use BDD. The change also supports only those using underscored method names, which is a group whose numbers are decaying following the march towards community standards.

unfunco:
@jpkleemans I had a play around, and came up with this: https://github.com/unfunco/phpunit-prefix – this allows prefixes to be specified within the phpunit.xml configuration file, and might be better suited.

jpkleemans:
@unfunco thanks for your thoughts!

> The change also supports only those using underscored method names, **which is a group whose numbers are decaying** following the march towards community standards.

I'm not sure if that's true, in the BDD/DDD world many are using the underscore convention for their test methods. Also many prominent people are using it, like @mathiasverraes in his (excellent) videos.
To be clear: I only use underscores for my test methods, my source code follows the PSR-2 standards (camelCase method names).

In my opinion PHPunit is perfectly suited for BDD. This PR would make it just a bit more convenient. Your package looks great, I'll give it a try!

sebastianbergmann:
I think that using `@test` to alternatively mark a test method is sufficient.

PR #2048, opened 2016-01-23
Feature suggestion: callNonPublicMethod()

blockjon:
I realize you probably don't want to add this to the framework... but in my projects, I use this function _all the time_. 

In fact, I suspect that a lot of developers end up reluctantly setting functions to public because they don't have an easier way to test them from PHPUnit. In my opinion, this would be useful to a lot of other devs.


sebastianbergmann:
I will not add this, sorry. https://github.com/sebastianbergmann/peek-and-poke exists, though.

PR #2055, opened 2016-02-03
Add support for HHVM and `!PHP` in `@requires`

dshafik:
You can now make tests require specific versions of `HHVM`:

``` php
/**
 * @requires HHVM 3.1
 */
```

As well as using all the regular operators:

```
/**
 * @requires HHVM < 3.2
 */
```

Furthermore, this adds support for a negation operator, which supports:

``` php
/**
 * Only run on alternative PHP runtimes
 *
 * @requires !PHP
 */
```

and

``` php
/**
 * Only run on HHVM
 *
 * @requires HHVM
 */
```

### DIFF
--- a/src/Util/Test.php
+++ b/src/Util/Test.php
@@ -18,7 +18,7 @@ class PHPUnit_Util_Test
     const REGEX_DATA_PROVIDER      = '/@dataProvider\s+([a-zA-Z0-9._:-\\\\x7f-\xff]+)/';
     const REGEX_TEST_WITH          = '/@testWith\s+/';
     const REGEX_EXPECTED_EXCEPTION = '(@expectedException\s+([:.\w\\\\x7f-\xff]+)(?:[\t ]+(\S*))?(?:[\t ]+(\S*))?\s*$)m';
-    const REGEX_REQUIRES_VERSION   = '/@requires\s+(?P<name>PHP(?:Unit)?)\s+(?P<operator>[<>=!]{0,2})\s*(?P<version>[\d\.-]+(dev|(RC|alpha|beta)[\d\.])?)[ \t]*\r?$/m';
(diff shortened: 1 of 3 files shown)


dshafik:
One thing to note is that because this introduces engine-specific functionality, this means there has to be some skipped tests on the opposite engines and I'm not entirely happy with the tests I've added. I definitely feel like I missed something :P

I did add `@group $ENGINE` to each of the tests so you can `--exclude-group`, which you should be able to work into the travis config somehow, to do the right thing on the right platform.

sebastianbergmann:
I am reluctant to merge this as I am wondering whether I want to support HHVM at all going forward.

sebastianbergmann:
Do note that PHPUnit uses the https://github.com/sebastianbergmann/environment library to deal with engine-specific things. Please use its `Runtime::isHHVM()` method instead of `defined('HHVM_VERSION')`, for instance.

dshafik:
@sebastianbergmann I'm reluctant too; TBH. I ran into an issue where I had to rewrite a test away from annotations on warnings-become-exceptions to catching and asserting because HHVM doesn't support the feature (e.g. `mysql_db_query()`) and throws an exception. See [this test case](https://github.com/dshafik/php7-mysql-shim/blob/master/tests/MySqlShimTest.php#L323)

This also plays into not being able to skip a test that has `expected*` annotations (another patch I submitted).

sebastianbergmann:
The list of incompatibilities between PHP and HHVM keeps on growing, exactly. I do not think that I will invest any more time in trying to make PHPUnit compatible with HHVM.

spriebsch:
In my humble opinion, it is Facebook's responsibility to keep HHVM compatible with PHP and/or PHPUnit, or to maintain an HHVM fork of PHPUnit. Given the amount of trouble you have supporting old-ish versions of PHP itself, I'm not sure it serves the *PHP* community to invest more time in HHVM compatibility.

HHVM community, this may be your call.

sebastianbergmann:
I agree with what @spriebsch said.

PR #2061, opened 2016-02-06
Place assertJsonFileEqualsJsonFile() before assertJsonFileNotEqualsJsonFile()

jpkleemans:
All 'negative' ('not') assert methods are placed after the 'positive' assert methods, but `assertJsonFileNotEqualsJsonFile()` was placed before `assertJsonFileEqualsJsonFile()`.
This PR fixes this minor imperfection.


sebastianbergmann:
Merged manually, thanks.

PR #2062, opened 2016-02-07
actualClassOrObject parameter type should be "mixed"

Oliboy50:
Because the `Assert::readAttribute()` method accept both `string` and `object` type.


GrahamCampbell:
Better to use `string|object`.

Oliboy50:
i agree but `Assert::readAttribute()` hint typing is `mixed`... so I don't want to break anything ;)

PR #2066, opened 2016-02-08
assertNull() before assertNotNull()

jpkleemans:
Same problem as #2061


sebastianbergmann:
Merged manually, thanks.

PR #2083, opened 2016-02-19
compatibility with older version, allow NULL as message

remicollet:
Real use case in zend-mail: https://github.com/zendframework/zend-mail/blob/master/test/Header/SenderTest.php#L50

For now, zend-mail test suite pass with 4.8 but fails with 5.2
This PR restore compatibility (even if this method is deprecated)


remicollet:
The failed test doesn't seem related to this change (already failed before)

remicollet:
Thanks

sebastianbergmann:
No, thank you :)

Issue #1993, opened 2015-12-08
Add --list-tests option for listing all available tests (without executing them)


loostro:
Hi,

could someone briefly explain to me how PHPUnit works internally? I want to create a simple class, that takes standard PHPUnit input and returns an array of `testcasePath => [ testOneName, testTwoName ]`

For example, given I'm in `/app_root` and I call `myscript -c phpunit.xml.dist --filter "::testAction" --group api` with `phpunit.xml.dist` configured to look for tests in `src/tests` and given following files exist:

`/app_root/src/tests/Controller/FirstControllerTest.php`

``` php
<?php

class FirstControllerTest extends \PHPUnit_Framework_TestCase
{
    /* @group api */
    public function testActionOne() { /* the test body is not relevant */}

    /* @group api */
    public function testActionTwo() { /* the test body is not relevant */}

    /* @group functional */
    public function testActionThree() { /* the test body is not relevant */}
}
```

`/app_root/src/tests/Service/FirstServiceTest.php`

``` php
<?php

class FirstServiceTest extends \PHPUnit_Framework_TestCase
{
    /* @group api */
    public function testActionList() { /* the test body is not relevant */}

    /* @group api */
    public function testActionFilter() { /* the test body is not relevant */}

    /* @group api */
    public function testElse() { /* the test body is not relevant */}
}
```

I would like my script to return:

``` php
<?php

/* @var array */
$exampleReturnData = [
//  relative path to testsuite                            array of matching test names
    'src/tests/Controller/FirstControllerTest.php'   => ['testActionOne', 'testActionTwo'],
    'src/tests/Service/FirstServiceTest.php'         => ['testActionList', 'testActionFilter'],
];
```

Where do I start? Is there some already existing functionality I could reuse or extend? @sebastianbergmann if you could give me some hints to get me started, please do

SvenRtbg:
I'd say that the XML result file of a test run will give you that kind of information. Probably also `phpunit --testdox` will output some parseable text that may give you the same.

If you don't want to run PHPUnit at all, you'd have to scan the test directory, tokenize or regex the PHP code, and grab what is interesting to you.

loostro:
thanks for reply, however I'd like to get a list of tests without running them, 
I wasn't aware of the `--testdox` parameter, however it does not solve my problem still

I need a list of tests to run them in parallell (1 process per code) becouse my intergration tests take too long. Running a test suite (1 hour) to get a list of avaliable tests is not an option.

[...]

PR #2093, opened 2016-02-23
Implementation of --only-print-tests

bor0:
Let's say I'm running phpunit locally, and lines like these appear:

```
...............I.....IIIIIIIIIIIIIIII.......SSS........I...  1770 / 10237 ( 17%)
.IFF............III..SSSSSSSSSSSSSSSSSSSS..................  1829 / 10237 ( 17%)
```

This parameter offers a good way to find the failing tests in the range 1770±5. This would allow the developer to address the UT failures in advance without waiting for phpunit to finish and show the report of the failing tests.

Example usage:

```
boro@bor0:~/dev/phpunit$ ./phpunit --only-print-tests | head -n 123 | tail -n 3
121. Framework_AssertTest::testAssertEqualsFails with data set #30
122. Framework_AssertTest::testAssertEqualsFails with data set #31
123. Framework_AssertTest::testAssertEqualsFails with data set #32
boro@bor0:~/dev/phpunit$ 
```

In general, to display the nth test ±x, the formula would be `x + n` for head, and `2x + 1` for tail.

Also possibly addresses https://github.com/sebastianbergmann/phpunit/issues/1993.

### DIFF
--- a/src/Framework/TestSuite.php
+++ b/src/Framework/TestSuite.php
@@ -1015,6 +1015,28 @@ class PHPUnit_Framework_TestSuite implements PHPUnit_Framework_Test, PHPUnit_Fra
     }
 
     /**
+     * Returns a (recursively) calculated array of tests for this test suite.
+     *
+     * @return array
+     *
+     * @since  Method available since Release ...
+     */
+    public function asTestArray()
+    {
+        $list = array();
+
+        foreach ($this->tests() as $test) {
+            if ($test instanceof PHPUnit_Framework_TestSuite) {
+                $list = array_merge($list, $test->asTestArray());
+            } else {
+                $list[] = $test;
+            }
+        }
+
+        return $list;
+    }
+
+    /**
      * Template Method that is called before the tests
      * of this test suite are run.
      *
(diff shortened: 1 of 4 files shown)


giorgiosironi:
Seems complicated to calculate a number every time you need to find a failing test, shouldn't you just run the suite with `phpunit --debug` or `phpunit --testdox` which shows all the executed tests?

bor0:
@giorgiosironi, that's the point. With this approach, you don't need to execute the tests in order to find in what order they run.

sebastianbergmann:
Thank you for your contribution. However, I do not think this is useful and decided not to merge it.

lox:
This would have been really nice for listing tests to use with tools that run tests in parallel.

lox:
See https://github.com/sebastianbergmann/phpunit/issues/1993#issuecomment-164310042.

boscowitch:
Its very usefull... atm I'm trying to find out how 2 tests are missing when i split up the testing with the filter parameter in 2 regex cases... but there is no way to tell ...

bor0:
@sebastianbergmann can you re-consider this feature request?

boscowitch:
I managed to kinda get an list of the tests run with the `--debug` flag however there where those 2 misterious tests:
Starting test 'Warning'.
W

With this feature I would have stopped searching earlier, so actually the count and output of debug are wrong since these  2 "Warning" tests are not tests we have written or actually exists in out codebase....

PR #2103, opened 2016-03-03
fix prettifier for trivial test name: test

vkunz:
hi,

phpunit 5.2.10 (7467a57b055985529d356746fc03c8ed3c3be520) introduced a side-effect which could lead to a crash.

In case a test method name equals 'test', php crashes with an "Uninitialized string offset" notice.

Feel free to ignore/close this in case this behaviour is discouraged.

BR
Valentin


vkunz:
thank you

PR #1959, opened 2015-11-18
Fix throwing Prophet error

radmen:
This simple fix allows to throw Prophet predictions exceptions.

Btw, very similar error is in current dev version https://github.com/sebastianbergmann/phpunit/blob/master/src/Framework/TestCase.php#L1002

PR #2115, opened 2016-03-13
Fix Prophecy prediction checks

ddelbondio:
Backport of fix from #1959 for 4.8


sebastianbergmann:
Merged manually, thanks.

PR #2116, opened 2016-03-14
Add ability to set custom filename rewrite callbacks

YuriyNasretdinov:
We implemented "soft mocks" project that is intended to be able to replace runkit and uopz. We do it by rewriting code "on the fly".

Project: https://github.com/badoo/soft-mocks

We used Soft Mocks to rewrite our tests (60k tests) so that PHP7 is supported on our website.

Here is the suggested patch to be able to use Soft Mocks with phpunit.

### DIFF
--- a/src/Framework/TestSuite.php
+++ b/src/Framework/TestSuite.php
@@ -356,8 +356,13 @@ class PHPUnit_Framework_TestSuite implements PHPUnit_Framework_Test, PHPUnit_Fra
         foreach ($this->foundClasses as $i => $className) {
             if (preg_match($shortnameRegEx, $className)) {
                 $class = new ReflectionClass($className);
+                $classFile = $class->getFileName();
 
-                if ($class->getFileName() == $filename) {
+                if ($restoreCallback = PHPUnit_Util_Fileloader::getFilenameRestoreCallback()) {
+                    $classFile = $restoreCallback($classFile);
+                }
+
+                if ($classFile == $filename) {
                     $newClasses = [$className];
                     unset($this->foundClasses[$i]);
                     break;
--- a/src/Runner/StandardTestSuiteLoader.php
+++ b/src/Runner/StandardTestSuiteLoader.php
@@ -63,6 +63,10 @@ class PHPUnit_Runner_StandardTestSuiteLoader implements PHPUnit_Runner_TestSuite
                 $class     = new ReflectionClass($loadedClass);
                 $classFile = $class->getFileName();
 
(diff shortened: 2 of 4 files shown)


YuriyNasretdinov:
Here is an example of test that passes with our patches:

https://github.com/badoo/soft-mocks/blob/master/example-phpunit/ExampleTest.php

YuriyNasretdinov:
Here is an article that talks a little bit about history of Soft Mocks as well:

https://techblog.badoo.com/blog/2016/03/14/how-badoo-saved-one-million-dollars-switching-to-php7/

sebastianbergmann:
Thank you for your contribution. It looks interesting but at first glance I do not fully understand why it does what and how and what the consequences of merging it may be. I will try to look into this ASAP but it might take a while.

sebastianbergmann:
By any chance, is there English documentation available for Soft Mocks?

YuriyNasretdinov:
English documentation for Soft Mocks is on the way. It should be there in
several hours. Patch is needed because phpunit "sniffs" too much about file
names that are loaded :)

YuriyNasretdinov:
English README is complete: https://github.com/badoo/soft-mocks/blob/master/README.md

sebastianbergmann:
Thank you, Yuriy, for providing English documentation for Soft Mocks.

After reading "Soft Mocks will work even without any phpunit patches" I have to say, though, that my motivation to add additional complexity to PHPUnit for this is rather limited. The two limitations listed, "unreadable" stack traces and not being able to redefine code units declared in test code, don't sound too bad, to be honest.

YuriyNasretdinov:
You haven't seen this ugly stack trace yet. It contains paths to
autogenerated files and has additional layers of soft mocks wrappers which
make understanding the trace much more difficult. Actually you modify stack
trace for the same reasons, I guess ;), in order to make stacktrace
readable and not containing trace lines from the framework itself.

I agree that modification is done in a very "straightforward" fashion and
there are probably better ways to be able to hook into phpunit code. I
could not find better solution after spending some time with phpunit
codebase, sadly.

Soft Mocks is intended to be runkit and uopz-compatible and a lot of tests
that people wrote using these extensions are unusable now. At least in our
repository several hundred tests were assuming that you can redefine code
in test files so it was worth it for us to add this functionality than to
rewrite a lot of tests.
So, basically, these patches are just intended to provide more consistency
and usability for our framework which usually is not a bad thing :).

sebastianbergmann:
Thank you for your contribution but I decided not to merge this.

PR #1609, opened 2015-02-06
Improve Composer's settings on Travis

henriquemoody:
Related to #1607.

PR #1608, opened 2015-02-06
Improve Composer's settings on Travis

henriquemoody:
Related to #1607.

PR #2139, opened 2016-04-13
Cache dependencies, resulting in faster builds

wouterj:
This way, dependencies will be loaded from cache all next builds against the same branch (unless there is a newer version, it will be loaded and cached again). This will result in faster builds, as cloning is quite slow.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,11 +12,15 @@ matrix:
 
 sudo: false
 
+cache:
+  directories:
+    - $HOME/.composer/cache/files
+
 before_install:
   - composer self-update
 
 install:
-  - travis_retry composer install --no-interaction --prefer-source
+  - travis_retry composer install --no-interaction --prefer-dist
 
 script:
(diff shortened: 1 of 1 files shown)


WouterJ:
Hmm, something is going wrong here. Seems like github auth is needed, although this mail from github says it isn't: https://github.com/composer/composer/issues/4884#issuecomment-195229989 I'll investigate something further, this PR is on hold for the moment.

sebastianbergmann:
This was attempted before, see #1606, #1607, #1608, and #1609.

WouterJ:
@sebastianbergmann the difference is that by now (at least, from what I've seen in other packages), authentication is no longer needed for downloading the zip achieves.

WouterJ:
@sebastianbergmann according to some people on Twitter, this happens randomly as some GitHub servers aren't yet updated. Restarting the build will probably fix it as another server is used (which is updated and doesn't have this rate limit anymore).

PR #2144, opened 2016-04-18
Log skipped and incomplete tests as skipped tests in junit output file

lstrojny:
According to http://llg.cubic.org/docs/junit/, a skipped test should be tracked in the skipped attribute. If we like that change we could also make PHPUnit’s output more standard conform by using the `<skipped/>`-tag for skipped tests, not error.
### Update
- Turns out we have to use the `<skipped/>`-tag, otherwise Jenkins believes it’s still an error


sebastianbergmann:
IIRC we only added the `$this->logIncompleteSkipped` option because JUnit's XML logfile at the time (or rather the XML produed by Apache Ant's JUnit task) did not support skipped / incomplete tests. We can and should remove it now.

I think the JUnit XML logfile generated by PHPUnit should be as close as possible to what the actual JUnit produced. The question is whether any change we introduced in PHPUnit related to this will result in BC problems.

lstrojny:
Fair enough, let’s remove the conditionals then. I am already using this version in our CI setup and the only configuration change that is needed is to use JUnit instead of PHPUnit publisher, as the PHPUnit publisher does not support the `<skipped/>` tag. But that’s certainly something that we can get fixed in the jenkins-xunit plugin.

lstrojny:
I just wonder now what to do with the `logIncompleteSkipped` as it becomes superfluous and is only in `addRiskyTest`. What about removing `logIncompleteSkipped` and treating a risky test as an error if strict mode is enabled?

sebastianbergmann:
Sounds good to me.

lstrojny:
Done, let me know what you think!

I removed the internal `$attachCurrentTestCase` property that was only used for skipped, incomplete and risky. As far as I understand, the behavior was to not at all include skipped/incomplete and risky tests. For skipped and incomplete I can see why. For risky tests I would expect a risky test to appear just as a regular successful test if I don't want the report to treat it as an error. Makes sense?

sebastianbergmann:
Makes sense.

lstrojny:
Alright, after the build is green I believe we are ready to merge. Should I squash before?

lstrojny:
Also: we should mention somewhere in the changelog, that people should change their Jenkins XUnit formatter from PHPUnit to JUnit.

sebastianbergmann:
Feel free to squash, if you want. I am fine either way. Not sure yet, though, whether I want to merge this into `master` right now because that will become PHPUnit 5.4 in June. An argument could be made that this has to wait until February and PHPUnit 6.

lstrojny:
I am of the opinion that it is a bug fix 😉 
But anyway, I am not a squashing fan either so I am happy to leave it like that.

sebastianbergmann:
Merged into `master`.

PR #2150, opened 2016-05-03
Assertion support within outputCallback

schnittstabil:
I need to test the output of my sut, unfortunately, however, `setOutputCallback` is quite limited.
Therefore, the following doesn't work as expected:

``` php
class UseCaseTest extends \PHPUnit_Framework_TestCase
{
    protected $directory = false;

    protected function setUp()
    {
        // create a temp directory
        $this->directory = ...;
    }

    protected function tearDown()
    {
        // remove the temp directory
        ...
        $this->directory = false;
    }

    public function testFileCreated()
    {
        $this->setOutputCallback(function ($output) {
            // assert file creation
            preg_match( '/Created: (?P<path>.*)/', $output, $matches);
            $this->assertFileExists($this->directory.'/'.$matches['path']);
        });

        // sut
        touch($this->directory.'/foo.txt');
        print 'Created: foo.txt';
    }
}
```

The main issues are:
1. the outputCallback is called after `tearDown`, thus `foo.txt` is already removed
2. the outputCallback is called within `stopOutputBuffering`, thus assertions aren't handled the right way

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -944,10 +944,7 @@ abstract class PHPUnit_Framework_TestCase extends PHPUnit_Framework_Assert imple
         if (!isset($e) || $e instanceof PHPUnit_Framework_RiskyTestError) {
             try {
                 if ($this->outputCallback !== false) {
-                    $this->output = call_user_func_array(
-                        $this->outputCallback,
-                        [$this->output]
-                    );
+                    $this->output = ($this->outputCallback)($this->output);
                 }
 
                 $this->outputCallback = false;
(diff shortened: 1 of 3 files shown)


schnittstabil:
Rebased and PHP7 CS fixed.

@sebastianbergmann Tests are included, do you see any chance this PR get landed at some point?

sebastianbergmann:
Thank you for your contribution. However, I decided not to merge this pull request.

I think what you want to do can be achieved using the existing `getActualOutput()` method. Simply use this to retrieve the output and then perform assertions on it.

schnittstabil:
@sebastianbergmann Many thanks, that sounds reasonable. Unfortunately, `getActualOutput()` is not documented; hence, I've created an issue at https://github.com/sebastianbergmann/phpunit-documentation/issues/409.

Issue #2158, opened 2016-05-10
Failure to run tests in separate processes if a file included into main process contains constant definition


vbashkirtsev:
If any test running in main process includes a file which contains constant definition then all further tests which should be run in separate processes fail.

The reason is that TestCaseMethod.tpl.dist contains following Smarty statements:

{constants}
{included_files}

Any constants defined by included files will appear in {constants} . Then attempt to re-include {included_files} causes "PHPUnit_Framework_Exception: Notice: Constant CONSTANT already defined in" error.

I consider this to be an error on part of PHPUnit as it appears in ordinary execution of tests.

I propose to suppress all errors generated during loading of all constants and included files as they should not happen otherwise. If any included file contains an error it will manifest itself on the first include attempt. So only error which may happen in this case is collision on constant definitions.

PR #2159, opened 2016-05-10
Fix for Issue #2158

vbashkirtsev:
Constant redefinition errors should be ignored when new separate process initialized by TestCaseMethod.dist.tpl . Any errors which may happen in included files are reported at the time of first include attempt so only errors which may be reported in TestCaseMethod.dist.tpl during constants and included files loading are those of global constants redefinitions and they may be safely ignored.

### DIFF
--- a/src/Util/PHP/Template/TestCaseMethod.tpl.dist
+++ b/src/Util/PHP/Template/TestCaseMethod.tpl.dist
@@ -1,4 +1,6 @@
 <?php
+use SebastianBergmann\CodeCoverage\CodeCoverage;
+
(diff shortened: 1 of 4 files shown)


sebastianbergmann:
Thanks. I would like to see this bug also fixed on the PHPUnit 4.8 branch. The patch applies fine there but a test fails.

sebastianbergmann:
Merged this into PHPUnit 5.3 and PHPUnit 5.4. If possible, please send a pull request for PHPUnit 4.8. Thanks!

PR #2160, opened 2016-05-10
PHPUnit should not fail if a constant defined in included file

vbashkirtsev:
As requested: original test fails because of percentage counter in the end of the line. Changed to %s in EXPECTF section to avoid failures in the future.


sebastianbergmann:
Thanks!

PR #2181, opened 2016-05-30
Use setlocale with 0 as second argument to not affect current setting

lolli42:


lolli42:
More explanation on this issue can be found here: https://review.typo3.org/#/c/48416/

sebastianbergmann:
Sorry, but I do not understand this pull request. The information I find on your Gerrit confuses me. Can you please explain the problem here? Thanks!

lolli42:
You code line $this->locale[$category] = setlocale($category, null); tries to backup the current state of locale settings before it sets the locale to the requested value in order to reset to that state after a test has been executed.
The problem is that the second function argument 'null' does not return the current locale untouched, but instead it also sets the locale according to given environment variables. So, the return value is the result of that environment fallback after the locale has been changed.
To retrieve the current locale without changing the state, the argument 0 instead of null has to be given.

Test script:

```
<?php
$untouchedLocale = setlocale(LC_NUMERIC, 0);
$touchedLocale = setlocale(LC_NUMERIC, null);
var_dump($untouchedLocale);
var_dump($touchedLocale);
```

```
output:
~$php foo.php
/home/lolli/foo.php:4:
string(1) "C"
/home/lolli/foo.php:5:
string(11) "de_DE.UTF-8"
```

php documentation: http://php.net/setlocale

> string setlocale ( int $category , array $locale )
> If locale is NULL or the empty string "", the locale names will be set from the values of environment variables with the same names as the above categories, or from "LANG".
> If locale is "0", the locale setting is not affected, only the current setting is returned.


--> The patch uses 0 instead of null to retrieve the current setting without changing state.

sebastianbergmann:
Merged manually, thanks.

PR #2191, opened 2016-06-04
Add tests for TestCase::createMock()

mnapoli:
I added a few tests for the `createMock()` method to ensure its behavior is tested and preserved in the future.


sebastianbergmann:
Merged manually, thanks.

PR #2192, opened 2016-06-04
Allow specifying methods to mock using createMock()

mnapoli:
Pull request based on #2191 (simple tests additions)

---

With this addition it is now possible to quickly mock methods when creating a mock without having to go through the mock builder. This is intended to provide a simpler solution for 80% of the use cases.

Users can provide an array indexed by the method name. Values of the array can be:
- anything other than a `Closure` and an `Exception`, in which case the value will be returned when the method is called:

``` php
$mock = $this->createMock(MyClass::class, [
    'sayHello' => 'Hello',
]);

echo $mock->sayHello(); // Hello world!

// Shortcut for
$mock = $this->createMock(MyClass::class);
$mock->expect($this->any())
    ->method('sayHello')
    ->willReturn('Hello');
```
- a `Closure`, in which case the value returned by the closure will be returned when the method is called:

``` php
$mock = $this->createMock(MyClass::class, [
    'sayHello' => function ($name) {
        return 'Hello ' . $name;
    },
]);
echo $mock->sayHello('Bob'); // Hello Bob
```
- an `Exception`, in which case the exception will be thrown when the method is called:

``` php
$mock = $this->createMock(MyClass::class, [
    'sayHello' => new RuntimeException('Ouch'),
]);
$mock->sayHello(); // RuntimeException is thrown
```

It's important to note these helpers are completely optional and do not prevent using the more advanced API for mocks. For example it's possible to combine simple method mocks plus advanced mocking on other methods:

``` php
$mock = $this->createMock('MyClass', [
    'sayHello' => 'hello',
]);

$mock->expects($this->once())
    ->method('sayGoodbye')
    ->willReturn('Goodbye');
```

This feature is a port of the [mnapoli/phpunit-easymock](https://github.com/mnapoli/phpunit-easymock) package. I wrote and used this package for a year and it's very helpful. The introduction of the new `createMock()` method in PHPUnit 5.4 made this possible because of the lack of extra arguments to `createMock()` (contrary to `getMock()`), hopefully this can be merged and help future users.

PS: I'm more than willing to contribute documentation if this gets accepted.

PR #2193, opened 2016-06-04
Allow specifying methods to mock using createMock()

mnapoli:
Reopening #2192 because I deleted the branch by mistake -_-

---

With this addition it is now possible to quickly mock methods when creating a mock without having to go through the mock builder. This is intended to provide a simpler solution for 80% of the use cases.

Users can provide an array indexed by the method name. Values of the array can be:
- anything other than a `Closure` and an `Exception`, in which case the value will be returned when the method is called:

``` php
$mock = $this->createMock(MyClass::class, [
    'sayHello' => 'Hello',
]);

echo $mock->sayHello(); // Hello world!

// Shortcut for
$mock = $this->createMock(MyClass::class);
$mock->expect($this->any())
    ->method('sayHello')
    ->willReturn('Hello');
```
- a `Closure`, in which case the value returned by the closure will be returned when the method is called:

``` php
$mock = $this->createMock(MyClass::class, [
    'sayHello' => function ($name) {
        return 'Hello ' . $name;
    },
]);
echo $mock->sayHello('Bob'); // Hello Bob
```
- an `Exception`, in which case the exception will be thrown when the method is called:

``` php
$mock = $this->createMock(MyClass::class, [
    'sayHello' => new RuntimeException('Ouch'),
]);
$mock->sayHello(); // RuntimeException is thrown
```

It's important to note these helpers are completely optional and do not prevent using the more advanced API for mocks. For example it's possible to combine simple method mocks plus advanced mocking on other methods:

``` php
$mock = $this->createMock('MyClass', [
    'sayHello' => 'hello',
]);

$mock->expects($this->once())
    ->method('sayGoodbye')
    ->willReturn('Goodbye');
```

This feature is a port of the [mnapoli/phpunit-easymock](https://github.com/mnapoli/phpunit-easymock) package. I wrote and used this package for a year and it's very helpful. The introduction of the new `createMock()` method in PHPUnit 5.4 made this possible because of the lack of extra arguments to `createMock()` (contrary to `getMock()`), hopefully this can be merged and help future users.

PS: I'm more than willing to contribute documentation if this gets accepted.

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -1510,7 +1510,15 @@ abstract class PHPUnit_Framework_TestCase extends PHPUnit_Framework_Assert imple
     /**
      * Returns a test double for the specified class.
      *
+     * You can optionally provide a list of methods to mock as a second argument:
+     *
+     *     $mock = $this->createMock(MyClass::class, [
+     *         'sayHello' => 'Hello world!',
+     *     ]);
+     *     echo $mock->sayHello(); // Hello world!
+     *
      * @param string $originalClassName
+     * @param array $methods            Methods to mock. Array of values to return, indexed by the method name.
      *
      * @return PHPUnit_Framework_MockObject_MockObject
      *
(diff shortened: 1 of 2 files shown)


sebastianbergmann:
If I merge this for PHPUnit 5.5 (and I am inclined to do so) will you provide documentation (pull request for https://github.com/sebastianbergmann/phpunit-documentation)?

spriebsch:
While I understand that the suggested syntax is shorter, personally, I am not convinced that a data structure (array) is better than method calls. I'm still hoping for better IDE support for PHPUnit - moving from explicit method calls to implicit data structures (throwing an exception vs. returning a value must be detected my "magic") will probably not make things easier.

sebastianbergmann:
I agree with @spriebsch.

mnapoli:
> throwing an exception vs. returning a value must be detected my "magic"

@spriebsch we could remove the "exception" part. We could also, if that helps, strip all magic and only allow to define closures? I.e. only allow:

```php
$mock = $this->createMock(MyClass::class, [
    'sayHello' => function ($name) {
        return 'Hello ' . $name;
    },
]);
```

That removes all magic, and still leaves the door open to later add more behaviors in the future if necessary. I understand your point about implicit data structure though, but by forcing the Closure signature maybe that would make the whole thing more explicit (thanks to the `function` keyword)?

PR #2211, opened 2016-06-10
Test & Fix for Windows Spaces

jmauerhan:
I noticed when running a test as an isolated process on Windows 10, there is an error: "'C:\Program' is not recognized as an internal or external command, operable program or batch file." I researched this and found it is an issue with proc_open and spaces, and the solution for Windows is to add a set of double quotes around the entire command. 

Source: https://bugs.php.net/bug.php?id=49139

I manually tested this, and created a unit test as well.

### DIFF
--- a/src/Util/PHP/Windows.php
+++ b/src/Util/PHP/Windows.php
@@ -33,4 +33,9 @@ class PHPUnit_Util_PHP_Windows extends PHPUnit_Util_PHP_Default
             1 => $stdout_handle
         ];
     }
+
+    public function getCommand(array $settings, $file = null)
+    {
+        return '"' . parent::getCommand($settings, $file) . '"';
+    }
 }
--- a/tests/Util/WindowsTest.php
+++ b/tests/Util/WindowsTest.php
@@ -1,0 +1,40 @@
+<?php
+/*
+ * This file is part of PHPUnit.
+ *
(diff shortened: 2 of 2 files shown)


jmauerhan:
Updated test to use stub data, so test can work on Linux CI.

sebastianbergmann:
I have merged the fix from your pull request in ade4756f6cf9d986f3716d55a85e2f662ac5ff8c, thank you for your contribution.

I have verified that your patch works by running PHPUnit's own test suite on Windows 10 with PHP 7 at `C:\Program Files\php\php`. Without the patch there are a lot more test failures (basically all PHPT tests fail) than with the patch.

I have not merged your tests because the PHPUnit test suite currently cannot be successfully run on Windows anyway.

jmauerhan:
@sebastianbergmann First of all, thank you!
Regarding the test, the test is actually agnostic of the actual system it runs on, and passes on Linux - this is the reason I stubbed the the Runtime class. I think the test is still useful to ensure the logic of escaping the string in the extra quotes remains if someone else edits the files.

jmauerhan:
PS: Would you be interested in contributions to make the test suite work on Windows? I have contributed to a few other projects soley in adding/improving tests. I'm a test engineer so it's kind of my thing ;)

PR #2223, opened 2016-06-18
Add support for multiple data providers

borela:
This patch will make phpunit merge the data from multiple data providers and use them in the test. 

example

### DIFF
--- a/src/Util/Test.php
+++ b/src/Util/Test.php
@@ -434,41 +434,48 @@ class PHPUnit_Util_Test
      */
     private static function getDataFromDataProviderAnnotation($docComment, $className, $methodName)
     {
-        if (preg_match(self::REGEX_DATA_PROVIDER, $docComment, $matches)) {
-            $dataProviderMethodNameNamespace = explode('\\', $matches[1]);
-            $leaf                            = explode('::', array_pop($dataProviderMethodNameNamespace));
-            $dataProviderMethodName          = array_pop($leaf);
+        if (preg_match_all(self::REGEX_DATA_PROVIDER, $docComment, $matches)) {
+            $result = array();
+            foreach ($matches[1] as $match) {
+                $dataProviderMethodNameNamespace = explode('\\', $match);
+                $leaf                            = explode('::', array_pop($dataProviderMethodNameNamespace));
+                $dataProviderMethodName          = array_pop($leaf);
 
-            if (!empty($dataProviderMethodNameNamespace)) {
-                $dataProviderMethodNameNamespace = implode('\\', $dataProviderMethodNameNamespace) . '\\';
-            } else {
-                $dataProviderMethodNameNamespace = '';
+                if (!empty($dataProviderMethodNameNamespace)) {
+                    $dataProviderMethodNameNamespace = implode('\\', $dataProviderMethodNameNamespace) . '\\';
+                } else {
+                    $dataProviderMethodNameNamespace = '';
+                }
+
+                if (!empty($leaf)) {
+                    $dataProviderClassName = $dataProviderMethodNameNamespace . array_pop($leaf);
+                } else {
+                    $dataProviderClassName = $className;
+                }
+
+                $dataProviderClass  = new ReflectionClass($dataProviderClassName);
+                $dataProviderMethod = $dataProviderClass->getMethod(
+                    $dataProviderMethodName
+                );
+
+                if ($dataProviderMethod->isStatic()) {
+                    $object = null;
+                } else {
+                    $object = $dataProviderClass->newInstance();
+                }
+
+                if ($dataProviderMethod->getNumberOfParameters() == 0) {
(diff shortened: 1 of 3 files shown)


localheinz:
This is awesome, and we want this!

sebastianbergmann:
@localheinz Can you share a use case?

localheinz:
@sebastianbergmann 

How about https://github.com/refinery29/test-util/pull/84?

We have a bunch of generic data providers which we use to assert that methods receiving parameters guard against invalid parameters being passed in. For this we sometimes mix different data providers, for example

* `providerInvalidString`
* `providerBlankString`

etc.

Would be nice to be able to just use annotations instead of creating a provider like this

```php
use Refinery29\Test\Util\DataProvider;

class FooTest extends \PHPUnit_Framework_TestCase 
{
    use DataProviderTrait;

    /**
     * @dataProvider providerInvalidIdentifier
     *
     * @param mixed $identifier
     */
    public function testConstructorRejectsInvalidIdentifier($identifier)
    {
        $this->expectException(\InvalidArgumentException::class);

        new Foo($identifier);
    }

    public function providerInvalidIdentifier()
    {
        return $this->provideDataFrom(
            new DataProvider\InvalidString(),
            new DataProvider\BlankString()
        );
    }
}
```

vs.

```php
class FooTest extends \PHPUnit_Framework_TestCase 
{
    /**
     * @dataProvider \Refinery29\Test\Util\DataProvider\InvalidString::data()
     * @dataProvider \Refinery29\Test\Util\DataProvider\BlankString::data()
     *
     * @param mixed $identifier
     */
    public function testConstructorRejectsInvalidIdentifier($identifier)
    {
        $this->expectException(\InvalidArgumentException::class);

        new Foo($identifier);
    }
}
```

Does that make sense?

m157y:
@sebastianbergmann I have another use case. At one of the projects where I'm working, we have reports with lot of parameters for filters, all filters is sets via report instance. So we actually have two type of tests, first for testing of correct handling of filters (dataProvider with filters information) and second for testing of correct report generation (same dataProvider with filters information). But we have two type of reports scheduled (created by cron) and custom (created by user via UI). So we have 2 different tests:
1. for filters setup (scheduled and non-scheduled), because parameters for those reports not the same
2. for report generation (and result data validation), because report generator is one for both cases.
So we have structure like
```php
class ReportTest
{
    public function scheduledFiltersProvider() {}

    public function customFiltersProvider() {}

    /** @dataProvider scheduledFiltersProvider */
    public function testScheduledFilters() {}

    /** @dataProvider customFiltersProvider */
    public function testCustomFilters() {}

    /**
     * @dataProvider scheduledFiltersProvider
     * @dataProvider customFiltersProvider
     */
    public function testReports() {}
}
```
now we just using third data provider which calls array_merge() for two other dataProviders and this is just a trick, but two annotations will be more comfortable and obvious

sebastianbergmann:
Thank you for your use cases.

sebastianbergmann:
@borela Can you please add a test for this new functionality? Thanks!

borela:
Done.

sebastianbergmann:
Merged manually, thanks.

localheinz:
Thank you, @borela and @sebastianbergmann!

willemstuursma:
This seems to break data providers that provide data via yield.

sebastianbergmann:
@willemstuursma That is a known issue that is the topic of #2380.

CC007:
Would it be possible to support dimensionality with dataProviders? As in 

```
/**
 * @dataProvider 1 aProvider
 * @dataProvider 2 bProvider
 */
function testSomething($a, $b)
{
     assertTrue(something($a, $b));
}
```

In this case it would be nice if the cross join of the providers was used, eg if `aProvider` had 3 datapoints and `bProvider` had 4, that the test would be run 12 times: once with each combination of `$a` and `$b`.

CC007:
maybe the next notation would be better:
```
/**
 * @dataProvider {aProvider, bProvider}
 */
 function ...
```
As then you can still have multiple separate datasets:

```
/**
 * @dataProvider {aProvider, bProvider}
 * @dataProvider {aProvider2, bProvider2}
 */
 function ...
```

Wirone:
@sebastianbergmann would be even better with ability to include some providers only if conditions are met:

```
/**
 * @dataProvider provideExamples
 * @dataProvider provideExamplesOnPhp82 if PHP 8.2
 */
 function ...
```

because right now whole test must be duplicated with `@requires` annotation.

mfn:
> only if conditions are met:

But you can add any kind of logic check you want _in_ your data provider and simply return nothing if the PHP version does not match?

Wirone:
@mfn yes, but it requires custom logic that could be added multiple times in multiple projects while it could be supported on the tool's level (and it's already there for `@requires`, so it probably could be reused easily in `@dataProvider`). So maybe it's worth adding 🙂

PR #2230, opened 2016-06-28
Add accessors for expectedExceptionCode and expectedExceptionMessage in TestCase.php

slaughter550:


sebastianbergmann:
Can you explain why you need this?

slaughter550:
I'm in the process of writing an auto-gen test framework for our internal API that creates individual tests and executes them using the `__call` method and as a result because of the current implementation using ReflectionMethod, the tests were failing to run because Reflection does not support the magical `__call` as being a catch all. So this forced me to override the `runTest` method in TestCase and implement the error handling myself which without these accessors would be impossible since the error handling for the runTest method relies on private class attributes.

The tests are now generated like so
```
class DynamicTest extends PHPUnit_Framework_TestSuite
{
    public static function suite()
    {
        $suite = new self(__NAMESPACE__);
        foreach (glob(__DIR__.'/models/*.php') as $filename) {
            $suite->addAutoClass(str_replace('.php', '', basename($filename)));
        }

        return $suite;
    }

    public function addAutoClass($class)
    {
        $test = new $class();
        $properties = array_keys($test->getNullModel()->getRestProperties(false));
        foreach ($properties as $scenario => $property) {
            $this->addGeneratedTest(new ReflectionClass($class), $scenario.'-testProperty-'.$property.'-validValues');
            $this->addGeneratedTest(new ReflectionClass($class), $scenario.'-testProperty-'.$property.'-invalidValues');
        }
    }
```

This means I can now run a unique test like `closed-testProperty-score-validValues` without defining that method for each model in our API and it will test all valid values and assert both 200's and 422's correctly. This also means that error reports on bamboo and travis for us are much more descriptive than an entire suite failing.

In order to make the above part work, I took this part of runTest 
```
        try {
            $class  = new ReflectionClass($this);
            $method = $class->getMethod($this->name);
        } catch (ReflectionException $e) {
            $this->fail($e->getMessage());
        }
        $testArguments = array_merge($this->data, $this->dependencyInput);
        $this->registerMockObjectsFromTestArguments($testArguments);
        try {
            $testResult = $method->invokeArgs($this, $testArguments);
        }
```  
and changed it into 
```
$testArguments = array_merge($this->data, $this->dependencyInput);
$this->registerMockObjectsFromTestArguments($testArguments);
try {
            $testResult = $this->$method(...$testArguments);
        }
```

Thus allowing for the calling of a method that is not defined before runtime. I was also thinking about moving the error handler into its own function since its so lengthy and would allow for easier mods to runTest in the future without dealing with merge conflicts.

I'm happy to answer any questions or show you a demo if that would help!

slaughter550:
@sebastianbergmann thoughts on above?

slaughter550:
@sebastianbergmann thoughts on the above?

slaughter550:
@GrahamCampbell or @sebastianbergmann thoughts on the above?

sebastianbergmann:
Merged manually, thanks.

PR #1629, opened 2015-02-27
Make json assertions stricter

kalenp:
While writing a regression test for our API output, we discovered that the assertJson* methods allow some strings to pass which are not strictly equal.  These cases are covered particularly in the `string type not equals number` and `string type not equals boolean` test cases which have been added.  The additional tests help to validate behavior for any Json comparison method and in particular the proposed change.

PR #2241, opened 2016-07-12
Make json assertions stricter

Sazpaimon:
This is #1629 rebased onto the latest master as I ran into the same issues that @kalenp did. I'm hoping no additional work would be required but if so let me know.


sebastianbergmann:
Merged manually, thanks.

Issue #2015, opened 2015-12-23
Prefix all code bundled in PHAR distribution with random/unique namespace


sebastianbergmann:
Issues such as #2014 happen when the code under test shares dependencies with PHPUnit but requires different versions than the ones bundled in the PHAR.

Depends on https://github.com/webmozart/php-scoper being ready for real world usage.

Soullivaneuh:
I was working on it but php-scoper is not on packagist yet. See https://github.com/webmozart/php-scoper/issues/12.

I'm not sure of the `theseer/autoload` work here, can you please elaborate? :-)

Soullivaneuh:
AFAIK, this will partially resolve #2014.

The .phar will be fixed, but not global composer install, right?

lastzero:
Excellent idea! We just had trouble testing our Symfony 2.8 project. PHPUnit's phar (tested with 5.1.4) comes with the Yaml component of Symfony 3.0. Since Yaml::parse() does not support file names anymore, important dependencies such as Doctrine ORM are broken with this version (YAML config files can not be read anymore and you see misleading error messages). I can imagine, many developers are facing similar problems without understanding the reason. They might stop writing unit tests right in this moment. You should really fix this.

sebastianbergmann:
@lastzero In a scenario such as the one you describe you have to install PHPUnit via Composer for the time being.

lastzero:
@sebastianbergmann Sure, but my developers didn't figure this out on their own. It's just plain dangerous to bundle a common standard component such as the Symfony YAML parser in a version with all backwards compatibility removed. Actually, I think there was a previous release of phpunit(.phar), that checked if a vendor directory exists and only used the bundled libraries if not.

*Hint for other developers: If you're facing issues like this, you can use http://php.net/manual/en/reflectionclass.getfilename.php to get the filename of a class to see if it's coming from vendor or from a phar file.*

ssnikiforov:
We have stucked with the same problem. phpunit v3.2 (master) has a dependency "use Symfony\Component\Yaml\Yaml;" in "/Loader/YamlFileLoader.php". And, also, phpunit has an implementation (hard-coded) of Yaml class in phpunit's phploc.php file.

Solution is to set up in your composer.json requirement of "3.0" version of symphony/dependency-injection. You can type smth like this:
"symfony/dependency-injection": "3.0".

COil:
[...]

PR #2264, opened 2016-08-09
Set prepend-autoloader to false in composer.json

nicolas-grekas:
This should allow e.g. symfony/yaml users to test their code with the latest version of the component, whereas today the phar users are forced to whatever version is bundled inside it.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -47,6 +47,7 @@
         "phpdocumentor/reflection-docblock": "3.0.2"
     },
     "config": {
+        "prepend-autoloader": false,
         "platform": {
             "php": "5.6.0"
         }


sebastianbergmann:
What does `composer.json` have to do with the PHAR version of PHPUnit?

nicolas-grekas:
Right, too quick :) So, the phar version is not autoloaded but "require" all files at bootstrap? Can't we use an autoloader here (a custom one to now collide with composer's)?

theseer:
PHPUnit does use a map based autoloader in the PHAR distributed verison, afaik generated by phpab.

But that aside, how would changing the order of autoloading fix the underlying issue you describe?
(Assuming, I understood the problem correctly..)

Your application under test and PHPUnit share the same process space, so you'll always end up with a conflict in case the application and PHPUnit use the same library but in different versions.

Changing the order of loading them won't fix that.

sebastianbergmann:
@nicolas-grekas is correct that the PHAR distribution of PHPUnit does not use an autoloader but instead loads all the sourcecode files from the PHAR on startup. This is done to make sure that no other autoloader prevents PHPUnit from loading its classes from the PHAR.

I will not change the way class loading works for the PHAR. If the PHAR does not work for you then do not use the PHAR.

The real problem that @nicolas-grekas wants to solve, I think, is related to #2015. Somebody really needs to make https://github.com/webmozart/php-scoper usable.

nicolas-grekas:
:+1: with #2015 thanks

PR #2270, opened 2016-08-17
Allow `createPartialMock()` to not mock any methods.

corphi:
Right now, `createPartialMock($class, [])` does the same thing as `createMock($class)`; it mocks all methods. This is somewhat counter-intuitive and caused by the API of the mock builder, where calling `setMethods(null)` has the expected effect.

As a first idea, I allowed passing `null` as a parameter value for consistency. But there is no need for `createPartialMock` to return a complete one - `createMock` already does that. `[]` also expresses the meaning better than `null` (“exactly these methods”). As a bonus, the method signature doesn’t change.

Test included.

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -1574,7 +1574,7 @@ abstract class PHPUnit_Framework_TestCase extends PHPUnit_Framework_Assert imple
                     ->disableOriginalClone()
                     ->disableArgumentCloning()
                     ->disallowMockingUnknownTypes()
-                    ->setMethods($methods)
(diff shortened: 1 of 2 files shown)


sebastianbergmann:
Why would you want to not stub/mock any method?

corphi:
I use such mock objects in tests for setters and getters; mostly it’s about avoiding a call to the constructor. The code to create such objects is the same as in `createPartialMock()` – except for the methods mocked. These tests could pass a random untested method name instead (in fact, that’s the way the tests work right now), but that’s a silly workaround and will likely break.

sebastianbergmann:
Merged manually, thanks.

PR #2279, opened 2016-08-28
Test with HHVM LTS, HHVM Current and in HHVM php7 mode

photodude:
New Testing options for HHVM
- [specific versions of HHVM LTS and Current](https://docs.travis-ci.com/user/languages/php#HHVM-versions)
- [HHVM in PHP7 mode](https://docs.hhvm.com/hhvm/configuration/INI-settings#php-7-settings)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,22 +1,46 @@
+# Use new Travis-CI container Infrastructure for faster builds
+sudo: false
+
 language: php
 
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
Thank you for your effort but I would rather remove HHVM from the Travis CI build than add more HHVM to it.

PR #2280, opened 2016-08-28
Test 4.8 with HHVM LTS, HHVM Current and in HHVM php7 mode

photodude:
Same as #2279 but for PHPUnit 4.8 branch


sebastianbergmann:
Thank you for your effort but I would rather remove HHVM from the Travis CI build than add more HHVM to it.

Issue #2273, opened 2016-08-19
--testsuite <pattern> is not an actual pattern filter


gerardroche:
`--testsuite <pattern> Filter which testsuite to run.`

`<pattern>` is not actually matched as a pattern by PHPUnit.

For example:

```
<?xml version="1.0" encoding="UTF-8"?>
<phpunit>

    <testsuites>
        <testsuite name="unit">
            <directory>test/unit_tests</directory>
        </testsuite>
        <testsuite name="integration">
            <directory>test/integration_tests</directory>
        </testsuite>
    </testsuites>

</phpunit>
```

```
$ phpunit --testsuite=unit
```

Will run `test/unit_tests/*`.

```
$ phpunit --testsuite=integration
```

Will run `test/integration_tests/*`.

However.

```
$ phpunit --testsuite="unit|integration"
$ phpunit --testsuite="(unit|integration)"
$ phpunit --testsuite="/(unit|integration)/"
```

No variation of the above works.

sdudziak:
@sebastianbergmann I would love to take care about this issue.

sdudziak:
@gerardroche which version of PHPUnit have you used for that ?

gerardroche:
@sdudziak Version ~4.0. I think it's the same for ~5.0 too.

sdudziak:
@gerardroche  Yes, I agree.
There is just a simple comparison to check if the testsuite match. However I think it will be achievable to change this behavior to `<pattern>` .

sdudziak:
@gerardroche I think I solved it. Can you check the code under the #2281 MR and give me some feedback?

stepashka69:
I spent a lot of time to understand what is wrong with testsuite pattern in CLI before I found this issue. That's nice that someone fixed it. 
Are there any plans to include it to PHPUnit any soon?

sebastianbergmann:
Closing after 80754cf323fe96003a2567f5e57404fddecff3bf and e60ba3b25f597a97cb26723ce36950c5c7002d2e.

PR #2281, opened 2016-08-28
Issue #2273 option --testsuite <pattern> accepts regex to match testsuites

sdudziak:
Fix/new feature of Issue #2273 option --testsuite <pattern> accepts regex to match testsuites:
Added functional tests to check if solution is working
Added unit tests for new methods
Changed the `PHPUnit_Util_Configuration` class to handle regex match for testsuites.


Jean85:
Maye #2351 is simpler and more lightweight?

sebastianbergmann:
Closing in favor of #2351.

Issue #2288, opened 2016-09-03
Missing extension dependencies in composer.json


vbashkirtsev:
Recently I have created the tools which checks list of extensions in composer.json against actually used extensions in code. This tool is available on packagist.org as **logics/extcheck** . I have run it against PHPUnit and it gave me the following:

```
Following extensions are missing in require section of composer.json:

  ext-SPL
    Used RuntimeException in /home/mega/PHPUnit/phpunit/src/Runner/Exception.php:14
    Used SplObjectStorage in /home/mega/PHPUnit/phpunit/tests/_files/DataProviderDebugTest.php:33

  ext-date
    Used time in /home/mega/PHPUnit/phpunit/src/Util/PHP.php:328
    Used DateTime in /home/mega/PHPUnit/phpunit/tests/Framework/ConstraintTest.php:813
    Used DateTimeZone in /home/mega/PHPUnit/phpunit/tests/Framework/ConstraintTest.php:813
    Used DateTime in /home/mega/PHPUnit/phpunit/tests/Framework/ConstraintTest.php:814
    Used DateTimeZone in /home/mega/PHPUnit/phpunit/tests/Framework/ConstraintTest.php:814

  ext-Reflection
    Used ReflectionClass in /home/mega/PHPUnit/phpunit/src/TextUI/TestRunner.php:91
    Used ReflectionClass in /home/mega/PHPUnit/phpunit/src/TextUI/TestRunner.php:129
    Used ReflectionClass in /home/mega/PHPUnit/phpunit/src/TextUI/TestRunner.php:136
    Used ReflectionClass in /home/mega/PHPUnit/phpunit/src/TextUI/TestRunner.php:143
    Used ReflectionClass in /home/mega/PHPUnit/phpunit/src/TextUI/TestRunner.php:246
    Used ReflectionClass in /home/mega/PHPUnit/phpunit/src/TextUI/TestRunner.php:889

  ext-mbstring
    Used mb_convert_encoding in /home/mega/PHPUnit/phpunit/src/Util/String.php:29

  ext-xml
    Used utf8_encode in /home/mega/PHPUnit/phpunit/src/Util/String.php:31

  ext-tidy
    Used tidy_repair_file in /home/mega/PHPUnit/phpunit/src/Util/Printer.php:91

  ext-libxml
    Used libxml_use_internal_errors in /home/mega/PHPUnit/phpunit/src/Util/XML.php:65
[...]

PR #2289, opened 2016-09-03
Extensions used by PHPUnit must be present in composer.json (#2288)

vbashkirtsev:
I have added PHP extensions check in .travis.yml . This way if any PHP extension which is not mentioned in composer.json will be used by any future code contribution then Travis build will fail with appropriate message.

Also I have added already used PHP extensions to composer.json

No tests since there no change to code base itself


vbashkirtsev:
Built-in extensions no longer listed in composer.json, absence of Phar is ignored

composer.json
```diff
@@ -39,9 +39,13 @@
         "myclabs/deep-copy": "~1.3",
         "ext-dom": "*",
         "ext-json": "*",
-        "ext-pcre": "*",
-        "ext-reflection": "*",
```
GrahamCampbell: ?

vbashkirtsev: As noted by @sebastianbergmann in #2288 these extensions always present in PHP and he does not see any reason to check for them in composer.json. I agree with him,

PR #2290, opened 2016-09-04
Backport expectException() to 4.8

zonuexe:
Libraries and products that support PHP 5.5 has a dependency on PHPUnit 4.

This patch will help migrating to latest version PHPUnit.


sebastianbergmann:
PHPUnit 4.8 will not be changed except for bugfixes.

PR #2312, opened 2016-10-03
Fix docs and message because assertArraySubset() does not support ArrayAccess

jeroennoten:
Exception message and docblocks state that `assertArraySubset()` accepts objects implementing `ArrayAccess` as well, however it actually does not. This PR only updates the docblocks, exception messages and corresponding test to match the actual behavior.


sebastianbergmann:
Closing this because `assertArraySubset()` now supports `ArrayAccess`.

PR #2333, opened 2016-10-16
Fix UTF-8 strings support in assertContains and assertNotContains

mnabialek:
The problem is that PHPUnit_Framework_Constraint_StringContains uses `stripos` and `strpos` functions instead of multibyte counterparts so before this change running:

``` php
$this->assertContains('Oryginał', 'ORYGINAŁ', '', true);
```

will assert `false` instead of `true` so you might get into trouble.

This is especially dangerous when you use `assertNotContains` because:

``` php
$this->assertNotContains('Oryginał', 'ORYGINAŁ', '', true);
```

will assert to `true` instead of `false` and your test will pass although your application might be broken.

### DIFF
--- a/src/Framework/Constraint/StringContains.php
+++ b/src/Framework/Constraint/StringContains.php
@@ -12,8 +12,8 @@
  * Constraint that asserts that the string it is evaluated for contains
  * a given string.
  *
- * Uses strpos() to find the position of the string in the input, if not found
- * the evaluation fails.
+ * Uses mb_strpos() to find the position of the string in the input, if not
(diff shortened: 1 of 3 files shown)


GrahamCampbell:
This breaks support for systems without the mb string extension.

mnabialek:
Yes, that's true, but in `composer.json` multibyte extension is required for using PhpUnit:

```
"ext-mbstring": "*",
```

so I believe when you want to use latest PhpUnit you need to have it enabled.

sebastianbergmann:
Merged manually, thanks.

PR #2340, opened 2016-10-26
Fix multi data generators when using iterator or yield

hervems:
When we use "yield" or "iterator" into data generator, for the moment we have a problem for the actual implementation of multiple data generators.
The proposal fix this problem.


sebastianbergmann:
This should go into `5.7` because that has 4aab496468a415fe687306cfab32ae89cf1a0705 already.

hervems:
It's based on 5.7 now

sebastianbergmann:
Merged manually, thanks.

Issue #2315, opened 2016-10-05
Mandatory listener class and dev dependencies


alskold:
After the update (5.5.6), with the listener class being mandatory (#2261), our production build is failing. The reason being that we have listeners which rely on dev-dependencies. 

Maybe have them mandatory by default, but add an "optional"-flag?

thePanz:
In our case, the failure of a misconfigured listener lead to no tests being executed and phpunit returning 0 as status code. Tests were silently skipped and the 0 return code wrongly interpreted as "everything is Ok".

sebastianbergmann:
Closing after #2349 was merged.

PR #2349, opened 2016-11-07
Respect $exit parameter in command run

thePanz:
See https://github.com/sebastianbergmann/phpunit/issues/2315#issuecomment-258809959


sebastianbergmann:
Cherry-picked into `5.7`, thanks.

PR #2356, opened 2016-11-13
Correct fix for #1956

bwoebi:
Currently tests in separate processes are just not run...

Using php://stdin fixes this.


sebastianbergmann:
Thanks!

PR #2357, opened 2016-11-15
#2246 --fail-on-warning doesn't change status code

### LINKED ISSUE #2246: --fail-on-warning doesn't change status code
```
$ bin/phpunit --fail-on-warning tests/ExampleTest.php
PHPUnit 5.4.6 by Sebastian Bergmann and contributors.

W.                                                                  2 / 2 (100%)

Time: 2.95 seconds, Memory: 30.00MB

There was 1 warning:

1) Warning
The data provider specified for ExampleTest::testFoo is invalid.
Call to undefined method Foo::bar()

$ echo $?
0
```

RolfBabijn:
The conditional was unreachable due to the `$result->wasSuccessful()` conditional already checked internally if there were any warnings.

Original report: #2246

### DIFF
--- a/src/TextUI/TestRunner.php
+++ b/src/TextUI/TestRunner.php
@@ -598,10 +598,6 @@ class PHPUnit_TextUI_TestRunner extends PHPUnit_Runner_BaseTestRunner
                     exit(self::FAILURE_EXIT);
                 }
 
-                if ($arguments['failOnWarning'] && $result->warningCount() > 0) {
-                    exit(self::FAILURE_EXIT);
-                }
-
                 exit(self::SUCCESS_EXIT);
             }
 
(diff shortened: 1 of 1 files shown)


Gadgetdude:
I'm really surprised to see this fail. So I've checked out the branch and ran the tests on both my branch and the `5.6` branch. Both result in the same failures. Maybe it has something to do with a recent PHP update (5.6.27)?

sergeyz:
@Gadgetdude it looks like `$result->wasSuccessful()` should not take into account `$this->warnings` and treat test with warnings as successful and do failure only if option was enabled

sebastianbergmann:
Thank you for your contribution. However, I will not merge your pull request as I think the problem should be solved somewhere else.

Issue #2317, opened 2016-10-05
Locally defined constant use in @expectedExceptionCode


vbashkirtsev:
Please consider the following code which currently fails with "Failed asserting that 1 is equal to expected exception code 'LOCALLY_DEFINED_CONSTANT'":

```
    /**
     * @expectedException Exception
     * @expectedExceptionCode LOCALLY_DEFINED_CONSTANT
     */
    public function testWorks()
    {
        define("LOCALLY_DEFINED_CONSTANT", 1);
        throw new Exception("Message", LOCALLY_DEFINED_CONSTANT);
    }
```

The reason is that evaluation of @expectedExceptionCode annotation happens straight away before test execution in PHPUnit_Util_Test::getExpectedException()  through parseAnnotationContent() (which BTW does not allow use of global constants). In my view evaluation of constants should happen in PHPUnit_Frameworl_TestCase::runTest() at time of assertions after test execution.

I was in the process of making new PR addressing this issue when I found that Util_TestTest has assertions which actually demand evaluation of @expectedExceptionCode, @expectedExceptionMessage and @expectedExceptionMessageRegExp in PHPUnit_Util_Test::getExpectedException() . I believe these assertions should be removed but considering policy of minimum breaks to backward compatibility (ie non-modification of existing tests) I have a question for @sebastianbergmann:

What is your view in this case: should I amend existing tests and move evaluation of these annotation from getExpectedException() to runTest() or just add re-evaluation of these annotations to runTest() ? I am leaning towards former solution as it is cleaner.

sebastianbergmann:
My take on this is that I do not want PHPUnit to resolve constants that are used in annotations.

You may want to read this [article](https://thephp.cc/news/2016/02/questioning-phpunit-best-practices) on why using annotations for exception expectations is a bad idea.

vbashkirtsev:
I've seen that article. And while I agree with you that use of annotations for expected exceptions is bad idea I note that PHPUnit at present does constant resolution in annotations. But such resolution is flawed and does such resolution only for class constants. I thought that this behaviour is actually should be considered as a bug, not a new feature (feature is already there but rather incomplete). Hence my PR.

On other hand not merging this PR will force new test instances to avoid bad practice.

PR #2369, opened 2016-11-23
Constants in expectedException annotations should be evaluated at assertion time, not at annotation parsing time, Issue #2317

vbashkirtsev:
Bug fix for PHPUnit 4.8


sebastianbergmann:
Forgot to close this when I closed #2317.

Issue #2382, opened 2016-12-02
Data Providers throw error with uncloneable object


kokspflanze:
```
<?php
use PHPUnit\Framework\TestCase;

class Test extends TestCase
{

    /**
     * @dataProvider dataProvider
     */
    public function testOne($test)
    {
        $this->assertInstanceOf(\Exception::class, $test);
    }

    public function dataProvider()
    {
        return [
            [
                $this->getMockBuilder(\Exception::class)->getMock()
            ]
        ];
    }
}
```

# Actual

```
PHPUnit 5.7.1 by Sebastian Bergmann and contributors.


Trying to clone an uncloneable object of class Mock_Exception_9ad7aaab


Time: 55 ms, Memory: 4.00MB


ERRORS!
Tests: 1, Assertions: 0, Errors: 1.
```

# Expected

```
PHPUnit 5.7.1 by Sebastian Bergmann and contributors.

.                                                                   1 / 1 (100%)

Time: 48 ms, Memory: 4.00MB

OK (1 test, 1 assertion)
```

PR #2383, opened 2016-12-02
Bugfix/dataprovider uncloneable object #2382

kokspflanze:
added a check, if its allowed to clone the object

https://github.com/sebastianbergmann/phpunit/issues/2382


src/Framework/TestCase.php
```diff
@@ -2625,4 +2631,24 @@ private function setDoesNotPerformAssertionsFromAnnotation()
...
         }
     }
+
+    /**
+     * @param ReflectionObject $reflection
+     * @return bool
+     */
+    private function isSaveCloneForMockObjectsFromTestArguments(ReflectionObject $reflection)
+    {
+        if (method_exists($reflection, 'isCloneable') && !$reflection->isCloneable()) {
```
sebastianbergmann: This is not needed. `ReflectionClass::isCloneable()` was introduced in PHP 5.4 and PHPUnit 5 requires PHP 5.6.

PR #2384, opened 2016-12-03
TestCase - fix handling PHPUnit_Framework_Exception expectedException

keradus:
So, one may expect an exception.
Examples:

- `@expectedException Foo\Bar\Baz`
- `@expectedException \Foo\Bar\Baz`
which are the same. And both are handled the same way.

Currently, there is some movement to prefer later one, as it's problematic when one is requiring `LogicException` thinking he expects one from his imports, while in fact he is requiring one from global namespace.

For that, in some edge cases, one may do this:
`@expectedException \PHPUnit_Framework_Exception`
but it doesn't work.
This would:
`@expectedException PHPUnit_Framework_Exception`

------------------

This PR fixes that:

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -925,7 +925,8 @@ abstract class PHPUnit_Framework_TestCase extends PHPUnit_Framework_Assert imple
 
                 $reflector = new ReflectionClass($this->expectedException);
 
-                if ($this->expectedException == 'PHPUnit_Framework_Exception' ||
+                if ($this->expectedException === 'PHPUnit_Framework_Exception' ||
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
TBH, I would rather remove `@expectedException` (and related annotations) in PHPUnit 6 than make changes to them.

keradus:
still, 4.x is supposed to be maintained.
shortly looking at whole method it doesn't matter would the string come from annotation or `->expectException` method...

keradus:
cool ;)

SpacePossum:
nice, thanks @keradus :)

PR #2385, opened 2016-12-03
Make consistent equal to

exussum12:
The methods greaterThanOrEqual and lessThanOrEqual had a different
definition of equal. This brings the signatures of the methods in line
and allows the deltas to be checked too.

### DIFF
--- a/src/Framework/Assert.php
+++ b/src/Framework/Assert.php
@@ -692,14 +692,18 @@ abstract class PHPUnit_Framework_Assert
      * @param mixed  $expected
      * @param mixed  $actual
      * @param string $message
+     * @param float  $delta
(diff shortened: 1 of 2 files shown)


sebastianbergmann:
Thank you for your contribution. However, I decided not to merge this pull request.

IMO, the `$delta`, `$maxDepth`, `$canonicalize`, and `$ignoreCase` parameters do not make sense for `greaterThanOrEqual()` and `lessThanOrEqual()`.

exussum12:
It was more the delta. A floating point number can fail this test when with equals it would be classed as the same. The others were added in for completeness with the test 

Can I remake with just the delta ? As that one was really the reason this pr was added

PR #2387, opened 2016-12-05
update phpspec prophecy to ^1.6.2

prolic:
This is the very first release of prophecy that also requires "sebastian/recursion-context": "^1.0|^2.0", every other known release only requires ^1.0, so there is a conflict and only 1.6.2 of prophecy and up can be installed anyway.

Additionally: prophecy is not yet 100% compatible with PHP 7.1. And there is a problem with current 1.6.2 release of prophecy, see: https://github.com/phpspec/prophecy/pull/287


sebastianbergmann:
Your patch changes a lot more than the pull request's description suggests. Also note that this should go into `5.7` instead of `master`.

prolic:
Ups sorry. Was a mistake :-)

PR #2390, opened 2016-12-08
Fixed incompatibility with HHVM in PHP7 mode.

ezzatron:
See also facebook/hhvm#7537.

HHVM has a "PHP7 mode" that can be enabled with an INI setting. Currently PHPUnit skipped tests (and probably other features) do not work in this mode because of an HHVM incompatibility involving `instanceof` used with a variable containing `'Throwable'` as the type.

I realize that you could simply wait until HHVM addresses this issue, but hopefully you will consider accepting this PR in any case, as it will allow PHPUnit to function correctly on older versions of HHVM that are unlikely to be patched.

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -2187,9 +2187,7 @@ abstract class PHPUnit_Framework_TestCase extends PHPUnit_Framework_Assert imple
      */
     protected function onNotSuccessfulTest($e)
     {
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
Should be taken care of as part of 0d1447bac826943931174befb12410143d2b73cb.

ezzatron:
Awesome, thanks!

Issue #2353, opened 2016-11-11
Provide a way to force set an environment variable from phpunit.xml


julianxhokaxhiu:
As already mentioned in https://github.com/sebastianbergmann/phpunit/issues/1276 there should be one way to override already defined ENV variabiles.

Looking at this line ( https://github.com/sebastianbergmann/phpunit/blob/5.6/src/Util/Configuration.php#L542 ) there's still a check that defines it, only if undefined. But there are some other cases where should be possible to override them.

My proposal:
```xml
<env name="FOO" value="bar" force="true" />
```

if `force` is set to `true` the env will be always set, no matter what.

Thanks in advance,
Julian

FlashBlack:
+1

There is no need to prohibit the overriding of environment variables.

cxj:
I just spent several hours debugging my unit tests because of this behavior.  PHPUnit will _not set_ an existing environment variable, even if it is **empty**.  That is, this will prevent that carefully crafted `phpunit.xml` clause from setting an environment variable named `APP_HOME`:

```sh
# sh syntax
export APP_HOME=

# Note:
~$ echo $APP_HOME

~$ echo $DOES_NOT_EXIST_AND_IS_NOT_SET

~$ set | egrep 'APP_HOME|DOES_NOT_EXIST'
APP_HOME=
```

Or from PHP (file bar.php):
```php
<?php
echo 'APP_HOME = ' . getenv('APP_HOME') . PHP_EOL;
echo 'DOES_NOT_EXIST_AND_IS_NOT_SET = ' . getenv('DOES_NOT_EXIST_AND_IS_NOT_SET') . PHP_EOL;
?>
```

Output:
```sh
~$ php bar.php
APP_HOME =
DOES_NOT_EXIST_AND_IS_NOT_SET =
```

Thus, gyrations are needed within the test classes themselves:
```php
        if (strlen(getenv('APP_HOME')) < 1) {
            echo PHP_EOL . 'Error:  APP_HOME set but empty' . PHP_EOL;
            $this->stopFlag = true;
        }
```
```php
    public function run(\PHPUnit_Framework_TestResult $result = null)
    {
        $result = parent::run($result);

        if ($this->stopFlag === true)
        {
            exit(1);    // Actually does exit here.
        }
        return $result; // Return as normal.
    }
```

My tests once again work after this.  But it would be nice to have a feature to allow PHPUnit to forcibly set required environment variables to known, sane starting values!

[...]

PR #2401, opened 2016-12-22
2353 force set envirronment variable from phpunit

miguel5fv:


sebastianbergmann:
I think this is "too much solution" for the problem described in #2353.

sebastianbergmann:
No feedback, closing.

Issue #1877, opened 2015-09-27
Add minimum coverage option to fail if coverage falls below a threshold


jdufresne:
When looking at coverage tools for [other languages](https://coverage.readthedocs.org/en/latest/index.html) they frequently provide a configuration option to fail if coverage falls below a threshold. In coverage.py, the [option](https://coverage.readthedocs.org/en/latest/cmd.html?highlight=fail#reporting) is `--fail-under`. I think this feature would be useful in PHPUnit as well.

Some additional option to `phpunit.xml` such that coverage below a threshold results in some kind of failure that returns a non-zero exit status.

This feature would help continuous integration systems that want to catch commits that significantly drop coverage.

redaxmedia:
+1

sebastianbergmann:
I do not thinks that functionality like this belongs in PHPUnit.

sinkcup:
I find a package to do this. https://packagist.org/packages/rregeer/phpunit-coverage-check

PR #2402, opened 2016-12-22
1877 test fails when coverage is below we indicate

miguel5fv:


sebastianbergmann:
Thank you for your contribution. However, as I wrote on #1877, I do not think that this belongs in PHPUnit.

molerat619:
@sebastianbergmann can you suggest an alternative solution? We are also looking for a way to do this in our gitlab ci/cd pipeline.

keradus:
report it to services like Coveralls (eg with https://github.com/php-coveralls/php-coveralls ) and set threshold there

PR #2409, opened 2016-12-30
Use error and failure check instead of wasSuccessful call, fixes #2246

### LINKED ISSUE #2246: --fail-on-warning doesn't change status code

dxops:
Second attempt

### DIFF
--- a/src/Framework/TestResult.php
+++ b/src/Framework/TestResult.php
@@ -1231,7 +1231,7 @@ class PHPUnit_Framework_TestResult implements Countable
      */
     public function wasSuccessful()
     {
-        return empty($this->errors) && empty($this->failures) && empty($this->warnings);
+        return empty($this->errors) && empty($this->failures);
     }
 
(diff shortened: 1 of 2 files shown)


aivus:
Related to https://github.com/sebastianbergmann/phpunit/pull/2357

sebastianbergmann:
Thank you for your contribution. However, I am closing this pull request as I think that it solves the problem in a wrong way. IMO, `wasSuccessful()` should be fixed to return the correct result.

PR #2414, opened 2017-01-02
[6.0][BC] Add classes needed by Symfony's PHPUnit Bridge

Jean85:
This PR stems from symfony/symfony#21125 which underlined the need for further BC support in the 6.0 release to avoid issues in their PHPUnit bridge.

This PR adds the needed classes. I had some doubts about the file naming, since there were 2 file with the `Test` name. I chose to use the latest part of the namespace as a distinguer, so:

 * `UtilTest.php` for class `PHPUnit\Util\Test`
 * `FrameworkTest.php` for interface `PHPUnit\Framework\Test`

### DIFF
--- a/src/BackwardCompatibility/AssertionFailedError.php
+++ b/src/BackwardCompatibility/AssertionFailedError.php
@@ -1,0 +1,15 @@
+<?php
+/*
+ * This file is part of PHPUnit.
+ *
+ * (c) Sebastian Bergmann <sebastian@phpunit.de>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use PHPUnit\Framework\AssertionFailedError;
+
+class PHPUnit_Framework_AssertionFailedError extends AssertionFailedError
+{
+}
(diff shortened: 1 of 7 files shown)


sebastianbergmann:
There will not be a backward compatibility layer.

Jean85:
Would you accept any other solutions? 

We are currently discussing other approaches in symfony/symfony#21125, but we haven't found an easy one: Symfony needs to support PHPUnit 4.8 as long as they support PHP 5.5, so util 4.0 is the only maintained version.

MarkMaldaba:
@sebastianbergmann Just so you are aware, this decision seems to have caused a *lot* of pain across a *lot* of projects, with multiple BC shims being created and a lot of time wasted in discussions, hair-pulling and attempting to avoid having to rewrite entire test suites (particularly in cases where older PHP or PHPUnit versions still need to be supported).

Why not just include a BC shim, like https://github.com/bobbingwide/wordpress-develop-tests/blob/master/phpunit/includes/phpunit6-compat.php as part of PHPUnit?

Or do you expect everyone to add that shim to their projects manually?

PR #2416, opened 2017-01-02
Upgrade phpunit-selenium version on 4.8

paulbriton:
We're still using the old version of PHPUnit which integrates phpunit-selenium. 

I need some of the bug fixes that will be included in the future `2.x` branch release.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -121,7 +121,7 @@
         <exec executable="${basedir}/build/tools/composer">
             <arg value="require"/>
             <arg value="phpunit/dbunit:~1.4"/>
-            <arg value="phpunit/phpunit-selenium:~1.4"/>
+            <arg value="phpunit/phpunit-selenium:~2"/>
             <arg value="phpunit/php-invoker:~1.1"/>
         </exec>
 


sebastianbergmann:
Are you sure that phpunit-selenium 2 is 100% backward compatible with phpunit-selenium 1?

paulbriton:
phpunit-selenium 2 dropped support for the old Selenium API, and uses now only the Selenium 2 API (both were present on the phpunit-selenium 1). I don't think there is still people using this. I can ask @giorgiosironi for more information.

sebastianbergmann:
I would rather not change something in PHPUnit 4.8 at this point (few weeks before the versions End of Life).

paulbriton:
I understand your point, but old line 1.x is not maintained anymore, and could be add manually for those who really wants to use it.

By the way, thank's for your time and availability.

giorgiosironi:
To be fair, it's a `2.x` (new major version) precisely because of dropping the SeleniumTestCase support, so I understand PHPUnit 4.8 should not silently remove that support. We shall find a solution with newer versions of PHPUnit.

PR #2417, opened 2017-01-02
Fix warning

exussum12:
When locally installed phpunit and a composer version exist, using the
local installed version gives a warning about the variable not being set

Tested on Ubuntu 16.10 with local version 5.4.6  and composer version
5.7.5


sebastianbergmann:
You must not mix versions of PHPUnit.

PR #2421, opened 2017-01-12
minor fixes to unittests

gemal:


sebastianbergmann:
Thank you for your contribution. However, I am closing this pull request as it caused tests to fail when it was sent and now has conflicts that need to be resolved.

PR #2423, opened 2017-01-13
Making TestCase public functions final

kubawerlos:
That would help to avoid any accidental overwriting


sebastianbergmann:
It would also prevent intentional overwriting (which is required by some extensions).

Issue #525, opened 2012-03-19
phpunit.xml listeners configuration did not accept nested array elements


llaville:
The PHPUnit_Util_Configuration class that parse phpunit.xml file for listeners did not accept nested array elements uses.

I wanted to configure my listener to accept 4 parameters like this  

```
<listeners>
    <listener class="GrowlTestListener"
              file="../examples/phpunitGrowlTestListener.php">
        <arguments>
           <string>file://../data/autotest_images/pass.png</string>
           <string>file://../data/autotest_images/pending.png</string>
           <string>file://../data/autotest_images/fail.png</string>
           <array>
               <element key="gntp">
                   <array>
                       <element key="host">
                           <string>127.0.0.1</string>
                       </element>
                       <element key="auth_password">
                           <string></string>
                       </element>
                       <element key="sticky">
                           <boolean>true</boolean>
                       </element>
                   </array>
               </element>
               <element key="smtp">
                   <array>
                       <element key="host">
                           <string>smtp.example.com</string>
                       </element>
                   </array>
               </element>
           </array>
        </arguments>
    </listener>
</listeners>
```

I'm expected to get for the fourth parameter :

```
Array
(
    [gntp] => Array
        (
            [host] => 127.0.0.1
            [auth_password] => 
            [sticky] => 1
        )

    [smtp] => Array
        (
            [host] => smtp.example.com
        )

)
```

Instead I got this result :

```
Array
(
    [gntp] => Array
        (
            [host] => 127.0.0.1
            [auth_password] => 
            [sticky] => 1
        )

    [host] => smtp.example.com
    [auth_password] => 
    [sticky] => 1
    [smtp] => Array
        (
            [host] => smtp.example.com
        )

)
```

Here is my patch proposal to fix this issue :

```
@@ -208,18 +208,19 @@
      * @param  DOMElement $element
      * @return mixed
      * @since  Method available since Release 3.4.0
      */
-    public static function xmlToVariable(DOMElement $element)
+    public static function xmlToVariable(DOMElement $element, DOMXPath $xpath)
     {
         $variable = NULL;

         switch ($element->tagName) {
             case 'array': {
                 $variable = array();
-
-                foreach ($element->getElementsByTagName('element') as $element) {
-                    $value = self::xmlToVariable($element->childNodes->item(1));
+                $elements = $xpath->query($element->getNodePath() . "/element");
+
+                foreach ($elements as $element) {
[...]

PR #2428, opened 2017-01-19
Fix duplicated keys in nested XML arrays (#525)

z38:
At the moment nested arrays in XML configuration files are parsed incorrectly because `DOMElement::getElementsByTagName` not only returns immediate children but all descendants.

Example:
```
<array>
    <element key="a">
        <array>
            <element key="b">
                <string>foo</string>
            </element>
        </array>
    </element>
    <element key="c">
        <string>bar</string>
    </element>
</array>
```
```php
array (
  'a' => array (
    'b' => 'foo',
  ),
  'b' => 'foo', // duplicated
  'c' => 'bar',
)
```

### DIFF
--- a/src/Util/XML.php
+++ b/src/Util/XML.php
@@ -200,17 +200,20 @@ class PHPUnit_Util_XML
             case 'array':
                 $variable = [];
 
-                foreach ($element->getElementsByTagName('element') as $element) {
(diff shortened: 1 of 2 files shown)


sebastianbergmann:
Cherry-picked into `4.8` and merged from there. Thanks!

PR #2436, opened 2017-01-24
Fix empty @group causing errors on PHP 7.2+

### LINKED ISSUE #2435: Empty @group annotation causes error on PHP 7.2+
Due to changes in how preg_match reports unmatched subpatterns in PHP 7.2+, leaving the `@group` annotation empty causes an `array_flip(): Can only flip STRING and INTEGER values!` error when running the test, as can be seen [here](https://github.com/doctrine/doctrine2/pull/6251).

PR w/ test will follow shortly.

For reference, from PHP's [UPGRADING](https://github.com/php/php-src/commit/c832ab4de68485df67d09ebb8d2bf55b658c5e30):
> preg_match() and other PCRE functions now distinguish between unmatched subpatterns and empty matches by reporting NULL and "" (empty string), respectively. Formerly, either was reported as empty string.

mbeccati:
Closes #2435

### DIFF
--- a/src/Util/Test.php
+++ b/src/Util/Test.php
@@ -606,7 +606,7 @@ class PHPUnit_Util_Test
             $numMatches = count($matches[0]);
 
             for ($i = 0; $i < $numMatches; ++$i) {
-                $annotations[$matches['name'][$i]][] = $matches['value'][$i];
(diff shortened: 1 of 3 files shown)


sebastianbergmann:
Cherry-picked into PHPUnit 5.7, thanks!

PR #2441, opened 2017-01-25
redirect to locally composer installed phpunit version

remicollet:
In some project:

```
$ vendor/bin/phpunit 
PHPUnit 5.3.5 by Sebastian Bergmann and contributors.
...

$ ~/.composer/vendor/bin/phpunit

==== Redirecting to composer installed version in vendor/phpunit ====

PHPUnit 5.3.5 by Sebastian Bergmann and contributors.
...

$ /work/GIT/phpunit/build/phpunit-nightly.phar

==== Redirecting to composer installed version in vendor/phpunit ====

PHPUnit 5.3.5 by Sebastian Bergmann and contributors.
...
```

### DIFF
--- a/build/binary-phar-autoload.php.in
+++ b/build/binary-phar-autoload.php.in
@@ -20,6 +20,15 @@ if (__FILE__ == realpath($GLOBALS['_SERVER']['SCRIPT_NAME'])) {
     $execute = false;
 }
 
+if (file_exists('./vendor/phpunit/phpunit/phpunit') && file_exists($auto='./vendor/autoload.php')) {
+  echo "\n==== Redirecting to composer installed version in vendor/phpunit ====\n\n";
+  define ('PHPUNIT_COMPOSER_INSTALL', realpath($auto));
+  require PHPUNIT_COMPOSER_INSTALL;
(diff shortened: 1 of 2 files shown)


remicollet:
Open against 5.7 (I have some issue to generate the nightly build from master), but probably more suitable for next version 5.8 or 6.0

sebastianbergmann:
Thank you, @remicollet.

I think this should go into `master` (PHPUnit 6). Making a change like this does not feel right in PHPUnit 5.7.

I think that adding this to the PHAR (`build/binary-phar-autoload.php.in`) is enough. I do not see why this change should be made to `phpunit`.

remicollet:
> I think that adding this to the PHAR (build/binary-phar-autoload.php.in) is enough. I do not see why this change should be made to phpunit.

for people installing it using "composer global require phpunit", and having ~/.composer/vendor/bin in their PATH.

sebastianbergmann:
Ah, right. Forgot about global install via Composer. Thanks!

remicollet:
And last notice, as this patch is already in the packaged (RPM) version, will be simpler to have it upstream ;)

Issue #2451, opened 2017-01-27
phpunit.xsd needs to be updated


sebastianbergmann:
`phpunit.xsd` currently specifies that `phpunit/php/ini`, `phpunit/php/const`, `phpunit/php/var`, `phpunit/php/env`, `phpunit/php/post`, `phpunit/php/get`, `phpunit/php/cookie`, `phpunit/php/server`, `phpunit/php/files`, `phpunit/php/ini`, `phpunit/php/request` elements must have a `name` and a `value` attribute.

As of 84f931bf1a4804a8e0e364e6e877e24227c6933c this is no longer the case. The aforementioned elements must have a `name` element and either a `value` or a `verbatimValue` attribute now.

theseer:
Since libXML2 does (still!) not implement XSD 1.1, there is no XSD-way of defining this requirement. 

We could use a Schematron schema rather than XML Schema for validation but despite libxml2 having support for it, PHP lacks the API into libxml2 to run it.  I'm not sure about PHPStorm or other editors either.

Talking to @sebastianbergmann, the better solution we came up with is to avoid the problem by changing the logic: Sticking with ``value`` but allowing an optional ``verbatim="true"`` attribute to achieve the same goal.

PR #2452, opened 2017-01-28
Add optional verbatim attribute to XML Schema

theseer:


theseer:
This should fix #2451

theseer:
I don't think the travis failures are related to my change.

PR #2466, opened 2017-02-03
bugfix for PHPUnit\Framework\Exception call in tests

kokspflanze:
i get with PHPUnit 6.0.2 following error `PHP Fatal error:  Cannot use PHPUnit\Framework\Exception as Exception because the name is already in use`

file 
phpunit\src\Framework\Constraint\IsEqual.php on line 12

so i changed all parts to the alias `PHPUnitException`

i changed also some parts in `PHPUnit\Util\Log\TeamCity` that there should be `\Exception` and not the internal `PHPUnitException`.

### DIFF
--- a/src/Framework/Constraint/Callback.php
+++ b/src/Framework/Constraint/Callback.php
@@ -9,7 +9,7 @@
  */
 namespace PHPUnit\Framework\Constraint;
 
-use PHPUnit\Framework\Exception;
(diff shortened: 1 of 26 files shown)


sebastianbergmann:
Closing this in favour of d109afe3e4bf2f5e176a9baa95f6cf015135071e.

PR #2475, opened 2017-02-06
Allow users to define a test suite with one file, no directories

benmatselby:
testSuiteFilter is an array, but for the file config aspect, it isn't
being treated as such. This means if you define a test suite with one file, it isn't finding any tests, which is incorrect.


sebastianbergmann:
Merged manually into `5.7`.

PR #2476, opened 2017-02-06
All listeners should be called before/after test

marten-cz:
When listener fails with an exception e.g. cannot connect to selenium server, no additional listeners, like HTML report will be triggered. All listeners should be called and when necessary, exception will be
triggered.

This is the fastest solution I came up with. I think better will be to catch exceptions and throw some listener exception with all exceptions that occurred. I can do that change too, let me know if that will be good to implement.

This change should fix this problem in Codeception https://github.com/Codeception/Codeception/issues/3653 . The problem is that the code is dependent on the actual order of registered listeners. When first listener fails and the last is HTML output, for that it will not call startTest event.

### DIFF
--- a/src/Framework/TestResult.php
+++ b/src/Framework/TestResult.php
@@ -363,8 +363,17 @@ class PHPUnit_Framework_TestResult implements Countable
             $this->topTestSuite = $suite;
         }
 
+        $exception = null;
         foreach ($this->listeners as $listener) {
-            $listener->startTestSuite($suite);
+            try {
+                $listener->startTestSuite($suite);
+            } catch (\Exception $e) {
+                $exception = $exception ?: $e;
+            }
+        }
+
+        if ($exception) {
+            throw $exception;
         }
     }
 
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
I do not understand the problem you are trying to solve.

orlangur:
What is the point in other listeners execution if the whole thing already failed?

> When the selenium grid is under load, the tests will sometimes fail with ... or some other exception. It will terminate the process and another tests will not run.  Is there anyway to just mark one test as failed but to continue? Would it be possible to re-run the test automatically when concrete exception was thrown (like timeout exception)?

This is not how automated testing is supposed to work, you know :) Your desired (_pretty odd_) behavior could be achieved with something like https://github.com/Alorel/phpunit-auto-rerun, but invalid listeners which break the process need to be fixed anyway. 

There is nothing to "fix" here from PHPUnit side. You should fix your tests, infrastructure and, probably, fragile tooling built on top of PHPUnit.

PR #2490, opened 2017-02-07
Make same and equals string messages the same

### LINKED ISSUE #2308: Better failure message when asserting that two strings are equal
This is a feature request but I'm not sure if it's about PHPUnit or PHPStorm IDE

When using `$this->assertSame()` with strings, you may get the following failure message:
image

It would be useful to have: 
- string length or quotes like with `var_dump` to spot a trailing space for instance
- var type in case the expected value is an empty string `''` and the actuel one is `false` or `null`

exussum12:
Fixes #2308

### DIFF
--- a/src/Framework/Constraint/IsIdentical.php
+++ b/src/Framework/Constraint/IsIdentical.php
@@ -82,9 +82,10 @@ class PHPUnit_Framework_Constraint_IsIdentical extends PHPUnit_Framework_Constra
                 $f = new SebastianBergmann\Comparator\ComparisonFailure(
                     $this->value,
                     $other,
-                    $this->value,
-                    $other
+                    $this->exporter->export($this->value),
+                    $this->exporter->export($other)
                 );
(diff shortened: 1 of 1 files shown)


exussum12:
The second part of this (showing the type) I think needs to be fixed here https://github.com/sebastianbergmann/comparator

sebastianbergmann:
The breaks PHPUnit's own test suite. Can you please update the tests?

exussum12:
It was already broken :( something about can't auto pad class ? I had to check out an older version of master  to test.

sebastianbergmann:
Without your patch, all tests in `master` pass. Both locally on my machine as well as on Travis.

exussum12:
Running ant fixed the issue, looking at the failures no

exussum12:
Im going to close this PR and there is a fair bit of work to make the two consistent, Will reopen when that is done

PR #2494, opened 2017-02-09
Fixes #2493: consider that argument --testsuite is a string

### LINKED ISSUE #2493: [BC Break] PHP Warning:  in_array() expects parameter 2 to be array, string given in vendor/phpunit/phpunit/src/Util/Configuration.php on
Hi,

With the new tagged release 5.7.12, my testsuite failed now due to a parameter error for in_array().

My testsuites are writing as is:
```xml
<testsuites>
    <testsuite name="main">
        <file>../vendor/NS/Bundle/Tests/Controller/ApplicationControllerTest.php</file>
        [...]
    <testsuite>
<testsuites>
```

and launch with:
```bash
php vendor/phpunit/phpunit/phpunit -c app/ --testsuite=main --debug
```

Any informations about this ?

Regards,
Boris.

rentalhost:
**Version to be fixed:** 5.7.12
**Reference to:** #2493
**New tests:** 1

### DIFF
--- a/src/Util/Configuration.php
+++ b/src/Util/Configuration.php
@@ -143,6 +143,8 @@
  */
 class PHPUnit_Util_Configuration
 {
+    const TEST_SUITE_FILTER_SEPARATOR = ',';
+
     private static $instances = [];
 
     protected $document;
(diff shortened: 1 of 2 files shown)


rentalhost:
**Note:** Travis CI fails is about using phpunit 6.1/master for a code compatible with 5.7 (_class not found_). As this PR is about 5.7, then I guess that we have no problem to accept that.

Issue #2496, opened 2017-02-09
Allow shallow copy for dependencies


ostrolucky:
I need to clone object where one of it's properties is throwing exception in __clone method. Just a normal ```clone $object``` would suffice here, but PhpUnit uses DeepClone instead.

Additionally, there is no proper way to modify this mechanism, because even though `handleDependencies()` is protected, there is no `handleDependency()` method and  `$dependencies` and `$dependencyInput` are private

PR #2498, opened 2017-02-10
Allow to replace cloning mechanism, as per #2496

ostrolucky:
#2496


gadelat:
That's not as flexible. With my solution, user can implement whatever cloning mechanism he needs.

PR #2510, opened 2017-02-20
Enhancement: Keep packages sorted

localheinz:
This PR

* [x] keeps packages in `composer.json` sorted

💁‍♂️ Because who doesn't like sorted lists? 

For reference, see https://getcomposer.org/doc/06-config.md#sort-packages.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,12 +21,18 @@
     },
     "require": {
         "php": "^7.0",
+        "ext-dom": "*",
+        "ext-json": "*",
+        "ext-libxml": "*",
+        "ext-mbstring": "*",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -52,7 +52,8 @@
     "config": {
         "platform": {
             "php": "7.0.0"
-        }
+        },
+        "sort-packages": true
```
localheinz: Changes were applied by

* adding this option
* running
    ```
    $ composer require phpunit/php-text-template:^1.2
    ```

sebastianbergmann:
Merged manually, thanks.

localheinz:

PR #2511, opened 2017-02-20
Fix: Class names do not match actual case

localheinz:
This PR

* [x] fixes issues where the case of a class name does not match the actually referenced class name


sebastianbergmann:
Merged manually, thanks.

localheinz:

PR #2512, opened 2017-02-20
Fix: Remove unused paramater from method signature

localheinz:
This PR

* [x] removes an unused parameter from a method signature

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -1676,13 +1676,12 @@ abstract class TestCase extends Assert implements Test, SelfDescribing
      * @param bool   $callOriginalConstructor
      * @param bool   $callOriginalClone
      * @param bool   $callAutoload
-     * @param bool   $cloneArguments
      *
      * @return object
      *
      * @throws Exception
      */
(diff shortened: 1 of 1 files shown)


src/Framework/TestCase.php
```diff
@@ -1676,22 +1676,20 @@ protected function getMockForTrait($traitName, array $arguments = [], $mockClass
...
-     * @param bool   $cloneArguments
      *
      * @return object
      *
      * @throws Exception
      */
-    protected function getObjectForTrait($traitName, array $arguments = [], $traitClassName = '', $callOriginalConstructor = true, $callOriginalClone = true, $callAutoload = true, $cloneArguments = false)
+    protected function getObjectForTrait($traitName, array $arguments = [], $traitClassName = '', $callOriginalConstructor = true, $callOriginalClone = true, $callAutoload = true)
     {
         return $this->getMockObjectGenerator()->getObjectForTrait(
```
localheinz: For reference, see [`PHPUnit_Framework_MockObject_Generator::getObjectForTrait()`](https://github.com/sebastianbergmann/phpunit-mock-objects/blob/master/src/Generator.php#L405-L444), which accepts 6 parameters only, so `$cloneArguments` is never used.

sebastianbergmann:
Merged manually, thanks.

localheinz:

PR #2515, opened 2017-02-20
Fix: Remove useless return statements

localheinz:
This PR

* [x] removes useless `return` statements


sebastianbergmann:
Merged manually, thanks.

localheinz:

PR #2516, opened 2017-02-20
Fix: Run php-cs-fixer

localheinz:
This PR

* [x] runs `php-cs-fixer`

💁‍♂️ Since there's a configuration for `php-cs-fixer`, I quickly ran

```
$ composer global require friendsofphp/php-cs-fixer:^2.1.0 && php-cs-fixer fix
```

How about 

* making `friendsofphp/php-cs-fixer` a development dependency
* adding a composer script so it can easily be run without a lot of typing, for example
    ```
    $ composer cs
    ```
* running this at least on one PHP version on Travis

?

### DIFF
--- a/src/Framework/Constraint/IsIdentical.php
+++ b/src/Framework/Constraint/IsIdentical.php
@@ -88,7 +88,6 @@ class IsIdentical extends Constraint
                     sprintf("'%s'", $this->value),
                     sprintf("'%s'", $other)
                 );
-
             }
 
             $this->fail($other, $description, $f);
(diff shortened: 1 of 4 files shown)


sebastianbergmann:
Most of these changes made by `php-cs-fixer` are wrong.

And, no, I do not want `php-cs-fixer` as a Composer-managed dependency. I do not think that it is a good idea to install tools used in a project using Composer.

localheinz:
So what's the point of [`.php_cs`](https://github.com/sebastianbergmann/phpunit/blob/90a14bf618e1c8a49a413baca7c9ea6895d5e533/.php_cs) then?

PR #2517, opened 2017-02-20
Fix: Use class keyword instead of string literals

localheinz:
This PR

* [x] makes use of the `class` keyword instead of referencing class names via string literals


sebastianbergmann:
Merged manually, thanks.

localheinz:

PR #2509, opened 2017-02-20
Enhancement: Use colors for output

localheinz:
This PR

* [x] uses colors for output

💁‍♂️ Because everyone loves colors, don't they?

PR #2518, opened 2017-02-20
Fix: Adjust phpunit.xsd to allow specifying max columns

localheinz:
This PR

* [x] adjusts `phpunit.xsd` to allow specifying `max` columns

Somewhat related to #2509.

💁‍♂️ Not sure, do you like all the columns, or do you just prefer 80? Otherwise we could adjust `phpunit.xml` as well, hehe!


sebastianbergmann:
Merged manually, thanks.

localheinz:

PR #2522, opened 2017-02-20
Fix: Return early

localheinz:
This PR

* [x] removes useless `else` and `elseif` statements where we could `return` or `throw` early instead (or already are)


localheinz:

Issue #2149, opened 2016-04-29
PHPUnit and Generators 


live627:
#1174 is still an issue. The offending code seems to be in 4.xx, 5.xx, and master.

The example test in that thread still applies.

sebastianbergmann:
Copied here from #1174:
```php
<?php
class GeneratorTest extends PHPUnit_Framework_TestCase
{
    public static function generator($range)
    {
        foreach (range(0, $range) as $x) {
            yield $x;
        }
    }

    public function testCountingAGeneratorFailsWithPHPWarning()
    {
        $this->assertCount(10, self::generator(10));
    }
}
```
```
$ phpunit GeneratorTest
PHPUnit 5.3.2-7-g3e6974d by Sebastian Bergmann and contributors.

E                                                                   1 / 1 (100%)

Time: 18 ms, Memory: 4.00Mb

There was 1 error:

1) GeneratorTest::testCountingAGeneratorFailsWithPHPWarning
Exception: Cannot rewind a generator that was already run

/home/sb/GeneratorTest.php:13

FAILURES!
Tests: 1, Assertions: 1, Errors: 1.
```

Keanor:
and problem with mocked methods: http://stackoverflow.com/questions/37118022/phpunit-mock-and-php-generators

keradus:
I encounter this again.
With the following change: https://github.com/keradus/PHP-CS-Fixer/commit/fbbc8c1
the `$this->assertSame('somefile.php', $iterator->current()->getFilename());` fails on HHVM.
If I use `assertSame` instead of `assertCount` everything is working on HHVM.

Nessworthy:
Cause is here:
https://github.com/sebastianbergmann/phpunit/blob/master/src/Framework/Constraint/Count.php#L69
Generators cannot be rewound, but assertCount will try and rewind it to its original state if it can.

Sooo, possible solutions are to either add in a case for generators so the rewind call is skipped or add new assertions which specifically target generators.

Thoughts?

sebastianbergmann:
Closed via 4123b6a1032503f8ed6baa753c61e0ae459a272b.

PR #2523, opened 2017-02-20
Fix: assertCount will no longer cause a rewind error when given a generator (as per #2149)

Nessworthy:
Hi, first attempt at a pull request, attempting to tackle #2149!

The issue was found to be in two areas in Framework\Constraint\Count::getCountOf:

1. Firstly, where a generator had already been iterated at least once, `iterator_count()` attempted to rewind the generator object which fired off the error.
2. Secondly, where a generator had not been iterated through at all, after the `iterator_count()` call exhausted it, the method would then attempt to rewind the generator object to set it back to the original position, also causing the error.

The fix attempts to get around this by looping through the generator itself and counting up the times it was iterated. Non-generators are not affected by this change.

A new test was added to cover both a fresh generator and a generator which had been partially exhausted already.

The caveats from this are that the generator will always be fully exhausted as a result of the assertion, and that counting a partially exhausted generator will only be able to return the count from its current position to the end.

Feedback is appreciated!

### DIFF
--- a/src/Framework/Constraint/Count.php
+++ b/src/Framework/Constraint/Count.php
@@ -1,4 +1,5 @@
 <?php
+
 /*
  * This file is part of PHPUnit.
  *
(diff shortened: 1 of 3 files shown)


sebastianbergmann:
Merged manually into `5.7`, `6.0`, and `master`.

PR #2531, opened 2017-02-23
Improve TeamCity output for PHPStorm

mdwheele:
Hello!

This is my first contribution to PHPUnit so please place this under higher scrutiny and I am ready to accept constructive criticism if things aren't done correctly.

I am contributing an arguably more verbose `TeamCity` result printer implementation that provides more contextual information in PHPStorm when things go wrong. I am unsure of what **other** products use `--teamcity` so I'm unsure if this work is *too* specific to PHPStorm reporting needs (or if the work should be done in the IDE instead). That said, my changes simply add more information to the TeamCity build log. I tried to keep the current test-cases passing while also supporting my feature.

Picture's worth a thousand words, so here is a before and after of a [test project I put together](https://github.com/mdwheele/phpunit/tree/test-project) that included:

- Several test methods with uncaught exceptions (with and without messages)
- Tests marked skipped, incomplete, risky, etc.
- Logical failures

image

My goals for this improvement were to:

* Bring more contextual information to the PHPStorm PHPUnit console log when things go wrong
* Improve interactive navigability of error messages
* Remain as true to the TeamCity build log specifications as I could

Some open questions:

- [ ] As currently specified in PHPUnit test cases for teamcity (`teamcity.phpt` and `log-teamcity.phpt`) there are differences in output. This was unexpected for me, but my implementation reproduces this behaviour. Basically `--log-teamcity php://stdout` includes progress output while `--teamcity` does not. Is this intentional? If not, what should the behaviour be?

- [ ] Would you like me to include the examples from my test project as part of the suite here?

- [ ] I currently use PHPStorm 2016.2. Is there need to test on other versions of PHPStorm or is generating a "correct" TeamCity build log good enough?


mdwheele:
Currently, the top-level test suite is placed in the build log. Is this necessary? It creates the following output given following configuration:

**Output**
image

**Config**
``` xml
<testsuites>
    <testsuite name="Feature Tests">
        <directory suffix="Test.php">./tests/Feature</directory>
    </testsuite>

    <testsuite name="Unit Tests">
        <directory suffix="Test.php">./tests/Unit</directory>
    </testsuite>
</testsuites>
```

sebastianbergmann:
@SvetlanaZem is responsible for the TeamCity logger.

mdwheele:
@SvetlanaZem, do you need any more information from me? Do I need to change anything?

SvetlanaZem:
@mdwheele Thank you for your contribution!

Please avoid commits which remove the whole file and create improved version, i.e. https://github.com/sebastianbergmann/phpunit/pull/2531/commits/460c98abeeafbd3ccd341c32d8df374db22fd873
https://github.com/sebastianbergmann/phpunit/pull/2531/commits/2886aee8221b8645ba0a81146249e18846b96155. It makes review very difficult because you need to compare files manually and git history become messy.

It would be better to handle each improvement separately because some of them are great suggestions and others are controversial. For instance, I like the idea to print "test is marked as risky with a message" for risky tests, but I don't understand why do you want to have a leading space at the beginning of stack trace.

Would you please create a pull request for the first logical improvement?

mdwheele:
> Please avoid commits which remove the whole file and create improved version

I agree! I can rebase that out for sure. TBH the only reason I did that was because the commit that made so many line changes was quite unreadable. I'll take care of that and push.

> I don't understand why do you want to have a leading space at the beginning of stack trace.

This was to reproduce the [current functionality in `master`](https://github.com/sebastianbergmann/phpunit/blob/master/src/Util/Log/TeamCity.php#L370). I don't see any reason to have that leading space in there other than for that reason. If we want to make the current behaviour to leave out the leading space, I can make that change!

> Would you please create a pull request for the first logical improvement?

Would it be acceptable to rebase this PR and better split up commits to be logical?

SvetlanaZem:
> Would it be acceptable to rebase this PR and better split up commits to be logical?

I'd prefer to have separate pull requests because we might want to merge only some of them.
Let's start with the first logical change, push it or revert, move to the next one and so on.

sebastianbergmann:
I agree with @SvetlanaZem and close this pull request now.

@mdwheele Please send individual pull requests, one for each intended change. Thanks!

mdwheele:
Sounds good! I'll send a few PRs shortly.

mdwheele:
I have created an issue to track changes and will be sending the first PR in that bunch after lunch. Thanks again.

Issue #2472, opened 2017-02-05
PHPUnit\Util\Getopt uses deprecated each() function


sebastianbergmann:
The deprecation error is suppressed since 20c70e0b72eb92367d00f7e70f1ff40d12976fd8.

PHPUnit\Util\Getopt should be refactored to not use the `each()` function.

SpacePossum:
Is anyone working on this?
I can pick it up as I would like this out of my white list of known issues on my CI env.

sebastianbergmann:
Go for it.

harikt:
Hey @sebastianbergmann can we back port this to 5.x ?

I was looking at the website and looks it is still supported : `Support ends on February 2, 2018` .

Let me know if I can help with or you can cherry pick and merge ?

Thank you.

harikt:
Oh sorry about this, I noticed it was an error on my part. It was using the flag `composer update --prefer-lowest --prefer-stable` https://travis-ci.org/harikt/psr7-asset/jobs/302542260#L476 .

Thank you.

PR #2536, opened 2017-02-24
Feature Getopt refactoring each

eclipxe13:
This solves PHPUnit\Util\Getopt uses deprecated each() function #2472.

It is a simple refactory to avoid using `each()` function based on `current()` and `next()`.
It includes more tests to verify how `Getopt::getopt` works.

Thanks for your great work and reviewing this PR.

### DIFF
--- a/src/Util/Getopt.php
+++ b/src/Util/Getopt.php
@@ -90,20 +90,17 @@ class Getopt
             }
 
             if (strlen($spec) > 1 && $spec[1] == ':') {
-                if (strlen($spec) > 2 && $spec[2] == ':') {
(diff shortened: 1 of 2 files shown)


sebastianbergmann:
Thank you!

SpacePossum:
thanks @eclipxe13 ! :+1:  :)

leofeyer:
I think we should backport this to PHPUnit 5.7, because the [symfony/phpunit-bridge](https://github.com/symfony/phpunit-bridge) will find the silenced function call (see e.g. https://travis-ci.org/contao/core-bundle/jobs/311481433). What do you think?

Issue #2537, opened 2017-02-27
Improve TeamCity output quality for PHPStorm


mdwheele:
Related: #2531 

In #2531, I proposed a monolith of changes that improved PHPUnit output for PHPStorm (which uses the TeamCity (`--teamcity`) build log format for rendering in the GUI. Understandably, it was a lot to review. 

This issue tracks the intended logical changes that were represented in my original PR and will be split over a few requests. 

**Cross-cutting concerns**

- Improve test coverage around TeamCity output, in general
- Improve docblocks to better explain what's going on

**Logical changes**

- [ ] **Refactor current printer to separate TeamCity-specific implementation from result printer interface methods.** I will be adding tests to improve coverage while extracting methods from the current TeamCity implementation into a `trait`
- [ ] Add `testStdOut` entries to build log to meet the following expectations:
  - Failures from exceptions should display the exception name (and message, if applicable)
  - Comparison failures should display expected / actual as well as the interactive "Click to see difference"
- [ ] Duration should be included in log service messages, where applicable.

SvetlanaZem:
> Failures from exceptions should display the exception name (and message, if applicable)

This one looks legit.  Execute following test to reproduce: 
```
<?php

use PHPUnit\Framework\TestCase;

class ExceptionClassAndMessageTest extends TestCase
{
    public function testGeneralException()
    {
        throw new InvalidArgumentException('Error Message', 10);
    }
}
```

*Expected:*
```
InvalidArgumentException: Error Message
 /path/to/test/ExceptionClassAndMessageTest.php:9
```

*Actual:*
```
Error Message
 /path/to/test/ExceptionClassAndMessageTest.php:9
```

> Comparison failures should display expected / actual as well as the interactive "Click to see difference"

I can't reproduce this one, it's either outdated or is specific for some comparison method.
@mdwheele Would you please provide code fragment to reproduce it?

> Duration should be included in log service messages, where applicable.

Looks like rounding error in TeamCity printer. There's similar issue in PhpStorm issue tracker: https://youtrack.jetbrains.com/issue/WI-38026

sebastianbergmann:
Thank you for your report.

Please provide a **minimal**, **self-contained**, reproducing test case that shows the problem you are reporting.

Without such a minimal, self-contained, reproducing test case I will not be able to investigate this issue.

SvetlanaZem:
[...]

PR #2538, opened 2017-02-27
[refactor] Separate TeamCity specific operations from Printer

mdwheele:
Related: #2537 

This work refactors the current `TeamCity` result printer to distinguish methods used for writing the build log from the interface inherited from `Printer` and `TestListener` through extending `ResultPrinter`. A test was added to cover the current behaviour of `master` when a failure occurs during a test-suite. I used the existing `FailureTest` and also created a new `ExceptionInTest` test case.

This work is in progress towards the goals of #2537 and is mostly structural in nature. No logical changes are intended in this work. I am simply separating concerns and renaming a few methods to better match TeamCity documentation.

### DIFF
--- a/src/Util/Log/TeamCity.php
+++ b/src/Util/Log/TeamCity.php
@@ -11,17 +11,13 @@
 namespace PHPUnit\Util\Log;
 
 use PHPUnit\Framework\AssertionFailedError;
-use PHPUnit\Framework\Exception;
 use PHPUnit\Framework\ExpectationFailedException;
 use PHPUnit\Framework\Warning;
 use PHPUnit\Framework\TestSuite;
 use PHPUnit\Framework\TestResult;
-use PHPUnit\Framework\TestFailure;
 use PHPUnit\Framework\TestCase;
 use PHPUnit\Framework\Test;
 use PHPUnit\TextUI\ResultPrinter;
-use PHPUnit\Util\Filter;
-use ReflectionClass;
 use SebastianBergmann\Comparator\ComparisonFailure;
 
 /**
@@ -30,6 +26,8 @@ use SebastianBergmann\Comparator\ComparisonFailure;
  */
 class TeamCity extends ResultPrinter
 {
+    use TeamCityUtils;
+
     /**
      * @var bool
      */
@@ -39,11 +37,6 @@ class TeamCity extends ResultPrinter
      * @var string
      */
     private $startedTestName;
-
-    /**
-     * @var string
-     */
-    private $flowId;
 
     /**
      * @param string $progress
@@ -184,18 +177,6 @@ class TeamCity extends ResultPrinter
         }
     }
 
-    public function printIgnoredTest($testName, Exception $e)
-    {
-        $this->printEvent(
-            'testIgnored',
-            [
-                'name'    => $testName,
-                'message' => self::getMessage($e),
-                'details' => self::getDetails($e),
-            ]
-        );
-    }
-
     /**
      * A testsuite started.
      *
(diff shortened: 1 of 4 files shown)


mdwheele:
@SvetlanaZem Pending tests passing, this PR is ready for your review. I tried to split out each refactor I made along the way in the hopes that review commit-by-commit is easier. There should be no logical changes or difference in output from `master`.

SvetlanaZem:
@mdwheele Thank you, I'll check it out. But since it's not a trivial change it might take some time.

SvetlanaZem:
@mdwheele Why do you want to extract several methods from TeamCity printer into a trait? 
From my point of view, traits can be useful to extract static methods, but you've also extracted `printEvent` method (new name is `messages`) which is not static. 
As the result, we have two classes with cross references, i.e. `TeamCity` uses `TeamCityUtils` to create a message and `TeamCityUtils` uses `TeamCity` to print messages.

Am I missing something and this kind of trait usage is a good practice?

mdwheele:
@SvetlanaZem When I started to approach a solution to PHPStorm's lacking reporting when test failures occurred, I had a hard time navigating the TeamCity result printer. I had a hard time separating what was responsible for responding to PHPUnit messages and what was responsible for TeamCity-specific concerns. I'm not necessarily tied to separating the interface to more clearly express responsibilities, but I think the name-changes toward what is documented by TeamCity is reasonable.

> Why do you want to extract several methods from TeamCity printer into a trait?

I believed that it more clearly explicated the separation of concerns between the listener/printer, `TeamCity`, and the technical implementation of the build log language, `TeamCityUtils`. Another design might inject a dependency responsible for writing the log into `TeamCity` rather than using `trait`. `trait` is effectively compiler-driven copy paste so I don't have much of an argument of keeping it separated other than it reduces cognitive load when processing the code, for me.

> As the result, we have two classes with cross references, i.e. TeamCity uses TeamCityUtils to create a message and TeamCityUtils uses TeamCity to print messages.

`TeamCityUtils#message` uses `TeamCity#write` (which is inherited from `Printer`), which isn't ideal. If we collapsed the interface back to all-in-one class (while keeping name changes), would that be better?

> Am I missing something and this kind of trait usage is a good practice?

I don't know if I'm best to speak on best practices for traits, but I'm also not necessarily tied to it as an implementation detail. What's more important, to me, is the problem I perceived; which was that it was somewhat straining for me to approach the printer (and also learn the TeamCity build log format) at the same time. 

Would you like me to collapse `TeamCityUtils` back into `TeamCity`?

SvetlanaZem:
> I think the name-changes toward what is documented by TeamCity is reasonable.

Do you mean renaming 'printEvent' to 'messages'? From my point of view, name 'messages' is not the best choice for a method because it's a plural noun instead of a verb. It would be better to go with 'printMessage' then.

> What's more important, to me, is the problem I perceived; which was that it was somewhat straining for me to approach the printer (and also learn the TeamCity build log format) at the same time.

Unfortunately, I don't see how your solution solves this problem. There are still a lot of logic related to TeamCity format, e.g. messages with 'testFinished', 'testSuiteFinished', etc.

mdwheele:
> Do you mean renaming 'printEvent' to 'messages'? From my point of view, name 'messages' is not the best choice for a method because it's a plural noun instead of a verb. It would be better to go with 'printMessage' then.

The method name is `message`. The methods in `TeamCityUtils` are named after their system message type equivalents from the referenced documentation. I am amenable to prefixing all with `print`. Would you like me to do that?

> Unfortunately, I don't see how your solution solves this problem. There are still a lot of logic related to TeamCity format, e.g. messages with 'testFinished', 'testSuiteFinished', etc.

The changes in this PR are not a comprehensive solution to the goals outlined in #2537. In a comprehensive solution, I would wrap pretty much everything in `TeamCity` that calls `TeamCityUtils#message` in a helper specific to a message type. I tried to limit how much I was doing structurally because my last PR was closed for doing too much. 

Would you like me to collapse the implementation back into a single class?

Ultimately, I just want PHPStorm not to give me garbage failures when a test fails. If the changes I'm proposing are fundamentally flawed, I believe I can contribute the necessary changes without "improving" the structure of this part of the code-base. Would you prefer I do that instead?

SvetlanaZem:
I agree that name `printMessage` is better than `printEvent`, so let's rename it.

Other changes don't look like improvements to me because I'm not sure code became more readable after extracting `TeamCityUtils` class.
Changes with`flowId` will change test output if tests are executed with `@runInSeparateProcess`. It won't affect PhpStorm output, but it can mess output for TeamCity itself, for the whole discussion see: https://github.com/sebastianbergmann/phpunit/pull/1909

I suggest to go with a simple rename refactoring for `printEvent` method if you don't mind.

mdwheele:
Sounds good. I appreciate the review. I will send that refactor as part of #2537 in another PR.

PR #2543, opened 2017-03-02
Fix: Merge isset() calls

localheinz:
This PR

* [x] merges `isset()` calls where possible

💁‍♂️ Yay, performance!


sebastianbergmann:
Than you for your contribution but I prefer individual `isset()` statements.

PR #2545, opened 2017-03-02
Fix: Remove import for undefined class

localheinz:
This PR

* [x] removes an import for an undefined class and fixes a related docblock

### DIFF
--- a/src/Framework/Constraint/Constraint.php
+++ b/src/Framework/Constraint/Constraint.php
@@ -12,7 +12,6 @@ namespace PHPUnit\Framework\Constraint;
 
 use Countable;
 use PHPUnit\Framework\ExpectationFailedException;
-use PHPUnit\Framework\SebastianBergmann;
 use PHPUnit\Framework\SelfDescribing;
 use SebastianBergmann\Comparator\ComparisonFailure;
 use SebastianBergmann\Exporter\Exporter;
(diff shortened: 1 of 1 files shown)


src/Framework/Constraint/Constraint.php
```diff
@@ -92,9 +91,9 @@ public function count()
     /**
      * Throws an exception for the given compared value and test description
      *
-     * @param mixed                                          $other             Evaluated value or object.
-     * @param string                                         $description       Additional information about the test
-     * @param SebastianBergmann\Comparator\ComparisonFailure $comparisonFailure
+     * @param mixed             $other             Evaluated value or object.
+     * @param string            $description       Additional information about the test
+     * @param ComparisonFailure $comparisonFailure
```
localheinz: Class is already imported here, see [`Constraint.php#L17`](https://github.com/sebastianbergmann/phpunit/blob/b2e74111300b09f5a23e6213e8abbeac6536bb85/src/Framework/Constraint/Constraint.php#L17).

src/Framework/Constraint/Constraint.php
```diff
@@ -12,7 +12,6 @@
 
 use Countable;
 use PHPUnit\Framework\ExpectationFailedException;
-use PHPUnit\Framework\SebastianBergmann;
```
localheinz: Class doesn't exist.

sebastianbergmann:
Merged manually into `6.0` and `master`.

localheinz:

PR #2546, opened 2017-03-02
Render __FILE__ and __DIR__ in SKIPIF section

c960657:
This patch adds support for `__DIR__` and `__FILE__` in the `SKIPIF` section of phpt tests.

One use-case for this is including Composer's autoload.php in case you want to use `class_exists()` to check whether some specific package is installed.


sebastianbergmann:
Thank you for your contribution.

PR #2008, opened 2015-12-15
Always write the template in order to make it easier to debug isolation mode

dawehner:
Rebase of https://github.com/sebastianbergmann/phpunit/pull/1942

 Debugging the isolation mode can be tricky as xdebug doesn't help you
with the initial $job piped into the php process.

As solution this PR uses the same technique as the windows util,
which is dumping the template into a temporary file and include it from there.

PR #2548, opened 2017-03-03
Always use a template file for in process tests for much better DX

dawehner:
When you have test failures in subprocess tests, they can be hard to debug, as phpunit
uses a piped in php file by default. This makes it impossible for xdebug to inspect the file
and also requires phpunit to ship with a php-eval.php file.

As suggested by @sebastianbergmann in https://github.com/sebastianbergmann/phpunit/pull/2008#issuecomment-283671104
we can drop some subclassing, when we require the file parameter.

Note: I removed the windows class, as the php bug mentioned in it got solved in 2014, so it should be part of php7.

As suggested by sebastian this also moves the class, removes the factory etc.


sebastianbergmann:
A lot of tests fail. Can you have a look? Thanks.

dawehner:
Sure, I just didn't thought about running them :)

dawehner:
I was able to fix a good amount of the failures. Let's see I can figure out the others.

src/Util/PHP.php
```diff
@@ -63,7 +66,7 @@ public function __construct()
         $this->runtime = new Runtime();
     }
 
-    /**
+        /**
```
Bilge: Why?

dawehner: Thank you for the pointer! No reason of course. I fixed that.

sebastianbergmann:
@dawehner Any progress on those remaining test failures? Thanks!

dawehner:
@sebastianbergmann Not yet sadly. I'll try to work on it maybe tonight or more realistic on the weekend. To be honest some of them at least seem to be non trivial for me.

sebastianbergmann:
@dawehner Thank you for the update.

sebastianbergmann:
Sorry to bother you again: any progress? Thanks!

dawehner:
Sorry :( I hope I have some time tomorrow or tonight to debug what is going on. I had hopes to pull some data out of sun when I saw him recently, but you know, sometimes that's hard.

sebastianbergmann:
Any update?

dawehner:
Nope, sorry :(

Sebastian Bergmann <notifications@github.com> schrieb am Di., 4. Juli 2017
um 10:08 Uhr:

sebastianbergmann:
@dawehner Damn! I totally forgot to talk to you about this in Vienna.

src/Util/PHP.php
```diff
@@ -7,25 +7,28 @@
...
 use PHPUnit\Framework\SyntheticError;
-use PHPUnit\Util\InvalidArgumentHelper;
-use SebastianBergmann\Environment\Runtime;
 
 /**
- * Utility methods for PHP sub-processes.
+ * Default utility for PHP sub-processes.
  */
-abstract class AbstractPhpProcess
+class PHP
```
keradus: Name is too generic, when you give only a name to someone, he will have on idea what to expect.
for logic in this class - maybe replace most of it with symfony/process ?

sebastianbergmann: See #1342. But that should not happen before #2015.

keradus: :+1: for switching to process component in separated PR/issue.
Yet, still, `PHP` is way too generic and meaningless name, it shall be modified

src/Runner/PhptTestCase.php
```diff
@@ -66,7 +66,7 @@ class PhptTestCase implements Test, SelfDescribing
      * Constructs a test case with the given filename.
      *
      * @param string             $filename
-     * @param AbstractPhpProcess $phpUtil
+     * @param \PHPUnit\Util\PHP  $phpUtil
```
keradus: changing param type for regular method is BC breaker.
@sebastianbergmann , do you treat changing `__construct` params as BC breaker ? I know some do, some don;t

src/Util/PHP/WindowsPhpProcess.php
```diff
@@ -1,43 +0,0 @@
...
-
-/**
- * Windows utility for PHP sub-processes.
- *
- * Reading from STDOUT or STDERR hangs forever on Windows if the output is
- * too large.
- *
- * @see https://bugs.php.net/bug.php?id=51800
- */
-class WindowsPhpProcess extends DefaultPhpProcess
```
keradus: now my code depending on this public class will stop working :(

sebastianbergmann: Just because a class is `public` does not mean that is part of the public API of a package.

keradus: damn, without any indicator that it's not part of public API, it's hard to know that :(
so one could rely on it

sebastianbergmann: True. But should I really have to add `@internal` to almost all classes?

keradus: There are 2 ways, either add `@internal` to non-public, or `@api` to public.
Without this info, user cannot know that it's internal, so he assume it's open to use and follows SEMVER.

note: Usually, I mark most classes with `@internal` and I'm making most classes final (so non-internal got final), leaving only few of them open to extend via inheritance.

src/Runner/PhptTestCase.php
dawehner: This is a good question!

dawehner:
@sebastianbergmann 
Oh damnit indeed. I cannot promise anything to be honest, there are just so many other things.
I think in terms of debuggability it would be still a great improvement, but I understand if you would close the issue as noone actively works on it.

sebastianbergmann:
This branch has conflicts that must be resolved, please send a new pull request when you're ready.

PR #2550, opened 2017-03-04
Make phpunit/php-code-coverage dependency optional

c960657:
If I check out PHPUnit and run `composer install`, the vendor folder uses 8 MB. If I remove the dependency on *phpunit/php-code-coverage*, the disk usage drops to 5.9 MB.

I propose making the dependency on *phpunit/php-code-coverage* optional, so that users must install it explicitly if they want to generate code coverage reports.

This will save bandwidth and build time, in particular in setups where you download the same files over and over again, e.g. when using PHPUnit to test your PHP project on Travis, or when building Docker images.


sebastianbergmann:
Not going to happen, sorry.

PR #2551, opened 2017-03-05
Allow directory traversal in FILE_EXTERNAL section

Bilge:
I don't know for what contrived reason relative paths were disabled, but if it's for "security" reasons, that's clearly complete nonsense.

### DIFF
--- a/src/Runner/PhptTestCase.php
+++ b/src/Runner/PhptTestCase.php
@@ -330,10 +330,8 @@ class PhptTestCase implements Test, SelfDescribing
 
         foreach ($allowExternalSections as $section) {
             if (isset($sections[$section . '_EXTERNAL'])) {
-                // do not allow directory traversal
-                $externalFilename = str_replace('..', '', trim($sections[$section . '_EXTERNAL']));
+                $externalFilename = trim($sections[$section . '_EXTERNAL']);
 
-                // only allow files from the test directory
                 if (!is_file($testDirectory . $externalFilename) || !is_readable($testDirectory . $externalFilename)) {
                     throw new Exception(
                         sprintf(


sebastianbergmann:
That would be a question for @kabel.

kabel:
Generally speaking, I'd say it's caution for those that might see it as a vector for exposing file content outside a _reasonable_ test base. It's not outside the realm of possibilities that a hostile test file gets added to a remote system (say during a pull request) and that test then exposes the content of a password file on that remote system. Sure, this could be achieved in any number of ways when a hostile php test is run on a remote system, so perhaps this point is moot.

The actual reason the code was added to phpunit was to match the reference implementation of phpt test runners from core php at https://github.com/php/php-src/commit/f61ba6295aa9445982874a055ff8e9a0f9ecc60b#diff-b6bbbff6a2dad1c77f0f1680692cd1beR1073 and is still used to this day. That is, you cannot do directory traversal in this section using the `pear run-phpt` (aka `make test`) runner. From the [phpt spec](https://qa.php.net/phpt_details.php#file_external_section):

>  The file must be in the same directory as the test file, or in a subdirectory.

Bilge:
That's interesting that it's mimicking a restriction of the reference implementation, but as you point out, its security benefits for PHPUnit are moot at best.

In my case, I had to write a [separate test runner file](https://github.com/ScriptFUSION/PHPUnit-Immediate-Exception-Printer/blob/master/test/functional/PHPUnit%20runner.php) that emulates the `phpunit` binary instead of just traversing back to `../vendor/phpunit/phpunit/phpunit` due to this arbitrary restriction, which is unfortunate.

sebastianbergmann:
IMO, the PHPT implementation in PHPUnit should stay as close as possible to the reference implementation (http://git.php.net/?p=php-src.git;a=blob_plain;f=run-tests.php;hb=refs/heads/master).

Bilge:
@sebastianbergmann For what possible benefit, and in particular, in regards to this specific issue?

Issue #1873, opened 2015-09-26
Arguments for an object within a listener cannot accept multiple same argument type since version 4.8


llaville:
I used XML configuration file with listener declared as follow 

``` xml
    <listeners>
        <listener class="Bartlett\LoggerTestListener">
            <arguments>
                <object class="YourMonolog">
                    <arguments>
                        <string>PHPUnit</string>
                        <string>debug</string>
                    </arguments>
                </object>
            </arguments>
        </listener>
    </listeners>
```

My object `YourMonolog` class constructor (accept two arguments) signature is as follow : 

``` php
<?php

use Monolog\Logger;

class YourMonolog extends Logger
{
    public function __construct($name = 'PHPUnit', $level = 'debug')
    {
    }
}
```

With PHP 4.6.10 and 4.7.7 (latest versions or previous branches) I got no issues, and the XML conf was correctly interpreted, and transmit two arguments as wished.

With PHP 4.8.9 (no tried other 4.8.x), the XML is readed and interpreted as only one argument transmitted to listener with walue `PHPUnitdebug`

I got following errors

```
PHP Notice:  Trying to get property of non-object in phar://C:/UwAmp/bin/php/phar/phpunit-4.8.9.phar/phar/phpunit/Util/XML.php on line 224
PHP Stack trace:
PHP   1. {main}() C:\UwAmp\bin\php\phar\phpunit-4.8.9.phar:0
PHP   2. PHPUnit_TextUI_Command::main($exit = *uninitialized*) C:\UwAmp\bin\php\phar\phpunit-4.8.9.phar:547
PHP   3. PHPUnit_TextUI_Command->run($argv = array (0 => '\\UwAmp\\bin\\php\\phar\\phpunit-4.8.9.phar', 1 => '-c', 2 => 'examples\\phpunit.monolog.xml'), $exit = TRUE) phar://C:/UwAmp/bin/php/phar/phpunit-4.8.9.phar/phar/phpunit/TextUI/Command.php:100
[...]

PR #2565, opened 2017-03-10
Fix handling of Listener arguments (#1873)

imjoehaines:
This is a proposed fix for #1873, it's technically a BC break but `PHPUnit\Util\Xml` is an internal PHPUnit class so I'm not sure how much that matters. I think backwards compatibility can be maintained by checking the length of `childNodes` and using the appropriate index, but didn't want to over-complicate the fix if it's not a concern.

The same fix works on 5.7 as well if you want me to open a separate PR for that branch.

### DIFF
--- a/src/Util/XML.php
+++ b/src/Util/XML.php
@@ -221,7 +221,7 @@ class Xml
                 $className = $element->getAttribute('class');
 
                 if ($element->hasChildNodes()) {
-                    $arguments       = $element->childNodes->item(1)->childNodes;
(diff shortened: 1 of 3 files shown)


imjoehaines:
Hi @sebastianbergmann, have you had a chance to review this?

sebastianbergmann:
Not yet, sorry. Hopefully before/during the PHPUnit code sprint in Hamburg next week.

imjoehaines:
No worries, just making sure it wasn't missed :)

sebastianbergmann:
Merged manually, thanks.

PR #2592, opened 2017-03-31
add @requires support for Composer packages (and versions).

abacaphiliac:
i thought something like this might help in testing branches that support multiple versions of dependencies, and maybe suggested/dev dependencies.

i'm not a fan of the way i've implemented discovery and parsing of composer.lock, nor the exception handling. if there is any interest in this, please guide me : )

i branched from `6.0.11`.


sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not you merge it.

PR #2593, opened 2017-03-31
Update TeamCity.php

MaximilianKresse:
The method \Util\Log\TeamCity::addError has a type hint for \Exception but the called getMessage and getDetails only allows \PHPUnit\Framework\Exception\Exception, fixed this by changing the type hint of getMessage and getDetails


sebastianbergmann:
Merged manually, thanks.

PR #2594, opened 2017-03-31
Enhancement: Enable native_function_invocation fixer

localheinz:
This PR

* [x] enables the `native_function_invocation` fixer
* [x] runs `php-cs-fixer`

💁‍♂️ For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/tree/v2.1.2#usage:

> **native_function_invocation**
Add leading `\` before function invocation of internal function to speed
up resolving.
*Rule is: configurable, risky.*

Also see https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2462.

### DIFF
--- a/.php_cs
+++ b/.php_cs
@@ -32,6 +32,7 @@ return PhpCsFixer\Config::create()
             'lowercase_constants' => true,
             'lowercase_keywords' => true,
             'method_argument_space' => true,
+            'native_function_invocation' => true,
             'no_alias_functions' => true,
             'no_blank_lines_after_class_opening' => true,
             'no_blank_lines_after_phpdoc' => true,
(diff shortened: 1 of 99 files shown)


sebastianbergmann:
Not sure I want to merge this. If I do then not because of a (probably not noticeable) speedup but because it would ensure that the internal functions are actually used and are not overridden by namespaced functions of the same name.

localheinz:
No problem, @sebastianbergmann!

sebastianbergmann:
Please send a new pull request that only contains https://github.com/sebastianbergmann/phpunit/pull/2594/commits/929d54f36d6b7c79a0faf0bef9d6c13275c22a38

Issue #2145, opened 2016-04-20
--stop-on-failure fails to stop on PHP 7


dtowell:
PHPUnit 5.3.2 fails to stop when given --stop-on-failure option and error occurs in static function setUpBeforeClass(). Similar results for --stop-on-error.

sebastianbergmann:
I can reproduce this:
```php
<?php
class Text extends PHPUnit_Framework_TestCase
{
    public static function setUpBeforeClass()
    {
        throw new Exception;
    }

    public function testOne()
    {
    }

    public function testTwo()
    {
    }
}
```
```
$ php --version
PHP 7.0.5 (cli) (built: Apr  8 2016 17:03:20) ( NTS )
Copyright (c) 1997-2016 The PHP Group
Zend Engine v3.0.0, Copyright (c) 1998-2016 Zend Technologies
    with Xdebug v2.4.0, Copyright (c) 2002-2016, by Derick Rethans
```
```
$ phpunit Test --stop-on-error
PHPUnit 5.3.2-2-gcd7b836 by Sebastian Bergmann and contributors.

EE                                                                  2 / 2 (100%)

Time: 14 ms, Memory: 4.00Mb

There were 2 errors:

1) Text
Exception in /home/sb/Test.php:6
Stack trace:
#0 /usr/local/src/phpunit/src/Framework/TestSuite.php(697): Text::setUpBeforeClass()
#1 /usr/local/src/phpunit/src/TextUI/TestRunner.php(440): PHPUnit_Framework_TestSuite->run(Object(PHPUnit_Framework_TestResult))
#2 /usr/local/src/phpunit/src/TextUI/Command.php(159): PHPUnit_TextUI_TestRunner->doRun(Object(PHPUnit_Framework_TestSuite), Array, true)
#3 /usr/local/src/phpunit/src/TextUI/Command.php(110): PHPUnit_TextUI_Command->run(Array, true)
#4 /usr/local/src/phpunit/phpunit(47): PHPUnit_TextUI_Command::main()
#5 {main}
2) Text
Exception in /home/sb/Test.php:6
Stack trace:
#0 /usr/local/src/phpunit/src/Framework/TestSuite.php(697): Text::setUpBeforeClass()
#1 /usr/local/src/phpunit/src/TextUI/TestRunner.php(440): PHPUnit_Framework_TestSuite->run(Object(PHPUnit_Framework_TestResult))
[...]

PR #2598, opened 2017-03-31
Fix for #2145

KrunoKnego:
This is a fix for https://github.com/sebastianbergmann/phpunit/issues/2145.


staabm:
should include the test from https://github.com/sebastianbergmann/phpunit/issues/2145#issuecomment-212739254

sebastianbergmann:
Merged manually, thanks!

Issue #2049, opened 2016-01-23
Misleading error message when whitelist configuration contains paths that do not exist


unreal4u:
Hello: 

Suppose the following section in phpunit.xml: 

``` xml
<phpunit colors="true" strict="true">
    <filter>
        <whitelist processUncoveredFilesFromWhitelist="true">
            <directory suffix=".php">./src/lalala/</directory>
        </whitelist>
        <blacklist>
            <directory suffix=".php">./vendor</directory>
        </blacklist>
    </filter>
</phpunit>
```

Which did work fine until I moved all my code from `/src/lalala/` to `/src/`. At that point, when I did run the tests to check everything, phpunit generated the following error: 

`Error:         No whitelist configured, no code coverage will be generated`

It would be much nicer to have another error such as "Whitelist directory doesn't exist" instead so it is easier to spot. 

I'm posting this here instead of php-code-coverage because the error is generated over here: https://github.com/sebastianbergmann/phpunit/blob/master/src/TextUI/TestRunner.php#L341

If however, you need the issue to be created in the php-code-coverage repo let me know and I'll create it over there.

Greetings.

# TL;DR:

Check whether the directory listed in your `<whitelist>` tag exists. If it does not, correct your phpunit.xml so that it points to the correct folder.

diego3:
Yeah, i'am using 5.5.0
and the same problem to me too :(

my phpunit.xml 
```xml

<?xml version="1.0"?>

<!-- see http://www.phpunit.de/wiki/Documentation -->
<phpunit bootstrap="bootstrap.php"
         colors="true"
         convertErrorsToExceptions="true"
         convertNoticesToExceptions="true"
         convertWarningsToExceptions="true"
         stopOnFailure="false"
         forceCoversAnnotation="true">
 	<testsuites>
    	<testsuite name="unit">
      		<directory>tests/src/Alpha</directory>
    	</testsuite>
    	<testsuite name="functional">
      		<directory>tests/src/Functional</directory>
    	</testsuite>
    	<testsuite name="integration">
      		<directory>tests/src/Integration</directory>
    	</testsuite>
  	</testsuites>

  <filter>
  <whitelist processUncoveredFilesFromWhitelist="true">
    <directory suffix="Test.php">/tests/src/</directory>
  </whitelist>
</filter>

</phpunit>
```
then i execute

```shell
phpunit -v -c tests/phpunitxml --coverage-clover tests/reports/coverage.xml tests
```
then i receive
```
[...]

PR #2599, opened 2017-03-31
Added more descriptive error messages for whitelist issues

leedwilkins:
ref: issue #2049 
The "No white list configured" was misleading when you had a whitelist element configured in `phpunit.xml`, but using invalid directories so that no test files were returned. Added a check for the existence of the white list element, and moved the original error to there. Then updated the "empty" error message to more accurately note the problem.


sebastianbergmann:
Merged manually, thanks!

PR #2604, opened 2017-03-31
Fix: Undefined classes

localheinz:
This PR 

* [x] fixes issues where classes are apparently undefined

### DIFF
--- a/src/Framework/Assert/Functions.php
+++ b/src/Framework/Assert/Functions.php
@@ -9,6 +9,7 @@
  */
 use PHPUnit\Framework\AssertionFailedError;
 use PHPUnit\Framework\Constraint\Constraint;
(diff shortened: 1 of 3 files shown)


src/Framework/Assert/Functions.php
```diff
@@ -9,6 +9,7 @@
  */
 use PHPUnit\Framework\AssertionFailedError;
 use PHPUnit\Framework\Constraint\Constraint;
+use PHPUnit\Framework\Constraint\Count;
```
localheinz: For reference, see [`PHPUnit\Framework\Assert\Functions.php`](https://github.com/sebastianbergmann/phpunit/blob/6.0.11/src/Framework/Assert/Functions.php#L1711).

PR #2607, opened 2017-03-31
Fix: Use null coalescing operator

localheinz:
This PR

* [x] makes use of the null coalescing operator where possible

### DIFF
--- a/src/TextUI/TestRunner.php
+++ b/src/TextUI/TestRunner.php
@@ -733,9 +733,9 @@ class TestRunner extends BaseTestRunner
             );
         }
 
-        $arguments['debug']     = isset($arguments['debug'])     ? $arguments['debug']     : false;
-        $arguments['filter']    = isset($arguments['filter'])    ? $arguments['filter']    : false;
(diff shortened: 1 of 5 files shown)


src/TextUI/TestRunner.php
```diff
@@ -733,9 +733,9 @@ protected function handleConfiguration(array &$arguments)
             );
         }
 
-        $arguments['debug']     = isset($arguments['debug'])     ? $arguments['debug']     : false;
-        $arguments['filter']    = isset($arguments['filter'])    ? $arguments['filter']    : false;
-        $arguments['listeners'] = isset($arguments['listeners']) ? $arguments['listeners'] : [];
+        $arguments['debug']     = $arguments['debug'] ?? false;
+        $arguments['filter']    = $arguments['filter'] ?? false;
+        $arguments['listeners'] = $arguments['listeners'] ?? [];
```
localheinz: Not sure, would you prefer alignment at `??` here and elsewhere?

sebastianbergmann: Yes, please :-)

localheinz: Hang on!

PR #2609, opened 2017-03-31
Fix: No need to assign default values

localheinz:
This PR

* [x] removes default value assignments which are immediately overwritten in corresponding constructors

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -151,12 +151,12 @@ abstract class TestCase extends Assert implements Test, SelfDescribing
     /**
      * @var array
      */
-    private $data = [];
+    private $data;
 
     /**
      * @var string
      */
-    private $dataName = '';
+    private $dataName;
 
     /**
      * @var bool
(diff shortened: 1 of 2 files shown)


src/Framework/TestCase.php
```diff
@@ -151,12 +151,12 @@
     /**
      * @var array
      */
-    private $data = [];
+    private $data;
```
localheinz: See [`TestCase:__construct()`](https://github.com/sebastianbergmann/phpunit/blob/6.0.11/src/Framework/TestCase.php#L334).

src/Framework/TestCase.php
```diff
@@ -151,12 +151,12 @@
...
 
     /**
      * @var string
      */
-    private $dataName = '';
+    private $dataName;
```
localheinz: See [`TestCase:__construct()`](https://github.com/sebastianbergmann/phpunit/blob/6.0.11/src/Framework/TestCase.php#L335).

src/TextUI/ResultPrinter.php
```diff
@@ -121,7 +121,7 @@ class ResultPrinter extends Printer implements TestListener
     /**
      * @var bool
      */
-    private $reverse = false;
+    private $reverse;
```
localheinz: See [`ResultPrinter::__construct()`](https://github.com/sebastianbergmann/phpunit/blob/6.0.11/src/TextUI/ResultPrinter.php#L182).

PR #2615, opened 2017-03-31
Use DIRECTORY_SEPARATOR for path comparission

IchHabRecht:
Using the DIRECTORY_SEPARATOR instead of hard coded forward slashes
solves a lot of failing tests on Windows systems.


IchHabRecht:
Before: Tests: 1514, Assertions: 2670, Failures: 136, Skipped: 4
After: Tests: 1513, Assertions: 2665, Failures: 112, Skipped: 4

PR #2619, opened 2017-04-01
2049 Whitelist error messages

leedwilkins:
Configuaration's method getFilterConfiguration is now returning an empty array if the whitelist node is not set in the xml configuration. This allows for short circuiting with an appropriate error message in TestRunner's doRun method.


sebastianbergmann:
Merged manually, thanks!

PR #2620, opened 2017-04-01
Bugfix: #2572 

### LINKED ISSUE #2572: Self referencing array breaks with PHPUnit 5.7 but worked with PHPUnit 4.8
| Q                   | A
| PHPUnit version     | 5.7.16
| PHP version         | 5.6.30

I'm trying to run the following test with Phpunit `5.7.16`. The test does not fail when using PHPUnit `4.8.35`.
I'm using PHP `5.6.30`.

```php
<?php

namespace Puzzle\Pieces;

class Json
{
        public static function encode($value, $option = 0, $depth = 512)
        {
            $json = json_encode($value, $option, $depth);

            if($json === false)
            {
                $jsonLastError = json_last_error();

                throw new Exceptions\JsonEncodeError(json_last_error_msg());
            }

            return $json;
        }
}

```

```php
<?php

namespace Puzzle\Pieces;

class JsonTest extends \PHPUnit_Framework_TestCase
{
    /**
     * @dataProvider providerTestJsonEncodeException
     * @expectedException Puzzle\Pieces\Exceptions\JsonEncodeError
     */
    public function testJsonEncodeException($data)
    {
        Json::encode($data);
    }

    public function providerTestJsonEncodeException()
    {
        $recursionData = array();
        $recursionData[] = &$recursionData;

        return [
            'JSON_ERROR_RECURSION' => [
                'data' => $recursionData,
            ]
        ];
    }
}

```

Here the error.
```
Fatal error: Maximum function nesting level of '256' reached, aborting! in /usr/src/puzzle-pieces/vendor/phpunit/phpunit/src/Framework/TestCase.php on line 2442
```

mcustiel:
Support autoreferenced array as arguments in 5.7: Added a check to avoid infinite recursion on self-referenced arrays. This fixes #2572

### DIFF
--- a/ChangeLog-5.7.md
+++ b/ChangeLog-5.7.md
@@ -2,6 +2,12 @@
 
 All notable changes of the PHPUnit 5.7 release series are documented in this file using the [Keep a CHANGELOG](http://keepachangelog.com/) principles.
 
+## [5.7.18] - 2017-XX-XX
+
+### Fixed
+
+* Fixed [#2572](https://github.com/sebastianbergmann/phpunit/issues/2572): Self referencing array breaks with PHPUnit 5.7 but worked with PHPUnit 4.8
+
 ## [5.7.17] - 2017-03-19
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


mcustiel:
@sebastianbergmann Have you any idea what could be the issue that makes the test fail for php5.6?

sebastianbergmann:
Merged manually, thanks!

Issue #2267, opened 2016-08-14
Passing an integer to the message argument of setExpectedException causes the message matching to always pass


Bilge:
> PHPUnit 4.8.26

``` php
public function testInteger() {
    $this->setExpectedException(Exception::class, 1);

    throw new Exception('2'); // Pass
}

public function testString() {
    $this->setExpectedException(Exception::class, '1');

    throw new Exception('2'); // Failed asserting that exception message '2' contains '1'.
}
```

I believe the message argument should automatically be typecast to string.

stevenbitner:
In PHPUnit 5.4 passing other than a string value for the message will give you the following failure:
`PHPUnit_Framework_Exception: Argument #1 (No Value) of PHPUnit_Framework_TestCase::expectExceptionMessage() must be a string`

leedwilkins:
setExpectedException has been decripcated, in checking expectExceptionCode, expecteExceptionMessage, and expectExceptedMessageRegex I realized that the tests are not run if the string is empty. It was originally using the empty() method, which meant that '0' would also be skipped. I changed it so that only the empty string is skipped, but technically even that should be checked. The problem is that, if it tries to test and empty string many errors/failures occur. This will take some more looking into.

PR #2625, opened 2017-04-01
#2267 Addressing the issue of of passing integers as exception codes/messages

leedwilkins:
Officially, #2267 is referencing the now depricated setExpectedException. But it was necessary to check that the problem did not also exist in expectExceptionCode. Though expectExceptionCode seems to be working as expected, in researching it was discovered that expectExceptionMessage and expectExceptionRegex would skip the test if '0' was set. It turns out that the empty() method was being used to check if the variable was set. Of empty returns true for '0', so the checks were updated to check for empty strings instead.


sebastianbergmann:
Please open a new pull request that only contains the commits relevant to this topic. Thanks!

PR #2637, opened 2017-04-02
Regression when several test arguments are the same

julienfalque:
I think #2620 introduced a regression when several arguments for a test are the same. I added a test that shows it but I don't know how to fix it for now.


julienfalque:
Closing as the test is wrong.

julienfalque:
Correct test with fix in #2638.

PR #2638, opened 2017-04-02
Fix regression with self-referencing arrays handling

julienfalque:
Fixes a regression introduced with #2620: when arrays mix objects with scalars, it produces errors such as:
```
Object of class Foo could not be converted to int
```

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -2441,7 +2441,7 @@ abstract class PHPUnit_Framework_TestCase extends PHPUnit_Framework_Assert imple
                     }
 
                     $this->registerMockObject($testArgument);
-                } elseif (is_array($testArgument) && !in_array($testArgument, $visited)) {
+                } elseif (is_array($testArgument) && !in_array($testArgument, $visited, true)) {
                     $visited[] = $testArgument;
 
                     $this->registerMockObjectsFromTestArguments(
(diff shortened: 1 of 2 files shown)


keradus:
new tests are always nice, not sure about replacing existing test

src/Framework/TestCase.php
```diff
@@ -2441,7 +2441,7 @@ private function registerMockObjectsFromTestArguments(array $testArguments, arra
                     }
 
                     $this->registerMockObject($testArgument);
-                } elseif (is_array($testArgument) && !in_array($testArgument, $visited)) {
+                } elseif (is_array($testArgument) && !in_array($testArgument, $visited, true)) {
```
keradus: nice fix

@sebastianbergmann , would you consider applying strictness for `is_array` globally ?

julienfalque:
Reverted original test and added mine as a new one.

tests/Framework/TestCaseTest.php
```diff
@@ -671,4 +671,20 @@ private function getAutoreferencedArray()
             ]
         ];
     }
+
+    public function testProvidingOfSelfReferencingArray()
```
keradus: what's the difference between autoreferenced array and selfreferencing one?
no need for testing `$arr[] = &arr` here as well, as it's tested in previous test

julienfalque: Renamed the new test and removed self-reference from the array.

sebastianbergmann:
Thanks!

SpacePossum:
thanks @julienfalque , nice fix!

PR #2642, opened 2017-04-04
Support counting non-Iterator Traversable objects

jmikola:
Per the commit message:

> Extension classes may implement Traversable without also implementing Iterator or IteratorAggregate.

I discovered this with [`MongoDB\Driver\Cursor`](http://php.net/manual/en/class.mongodb-driver-cursor.php), which is only Traversable (it implements PHP's internal iteration handlers).

While this does remove the assumption of an Iterator API, it also defeats the purpose of the [original patch](https://github.com/sebastianbergmann/phpunit/commit/36cc20f7f7efa52947a48cd0d11de1742a28231e) from #1125 to undoing the side-effect of `iterator_count()` advancing the object. I have two observations on that:

 * Some Traversables or Iterators may not support rewinding. While [`NoRewindIterator::rewind()`](http://php.net/manual/en/norewinditerator.rewind.php) is implemented as a NOOP, I can imagine some other advance-only iterators throwing exceptions when calling `rewind()`. This is the case with `MongoDB\Driver\Cursor`, as we cannot rewind the result set coming from the database.
 * Traversable objects that are neither an Iterator or IteratorAggregate interface may still support rewinding. In that case, we may want to compose them in an [IteratorIterator](http://php.net/iteratoriterator) to allow the original fix from #1125 to apply.

I'm open to feedback on the above points and can revise the PR as needed.

### DIFF
--- a/src/Framework/Constraint/Count.php
+++ b/src/Framework/Constraint/Count.php
@@ -10,6 +10,7 @@
 namespace PHPUnit\Framework\Constraint;
 
 use Countable;
+use Iterator;
 use IteratorAggregate;
 use Traversable;
 use Generator;
@@ -63,6 +64,10 @@ class Count extends Constraint
 
             if ($iterator instanceof Generator) {
                 return $this->getCountOfGenerator($iterator);
+            }
+
+            if (!$iterator instanceof Iterator) {
+                return \iterator_count($iterator);
             }
 
             $key   = $iterator->key();
--- a/tests/Framework/Constraint/CountTest.php
+++ b/tests/Framework/Constraint/CountTest.php
@@ -93,4 +93,19 @@ class CountTest extends TestCase
         $countConstraint->evaluate($generator, '', true);
         $this->assertEquals(null, $generator->current());
     }
+
+    public function testCountTraversable()
+    {
+        $countConstraint = new Count(5);
+
+        // DatePeriod is used as an object that is Traversable but does not
+        // implement Iterator or IteratorAggregate. The following ISO 8601
+        // recurring time interval will yield five total DateTime objects.
+        $datePeriod = new \DatePeriod('R4/2017-05-01T00:00:00Z/P1D');
+
+        $this->assertInstanceOf(\Traversable::class, $datePeriod);
+        $this->assertNotInstanceOf(\Iterator::class, $datePeriod);
+        $this->assertNotInstanceOf(\IteratorAggregate::class, $datePeriod);
+        $this->assertTrue($countConstraint->evaluate($datePeriod, '', true));
+    }
 }


src/Framework/Constraint/Count.php
```diff
@@ -65,7 +66,7 @@ protected function getCountOf($other)
                 return $this->getCountOfGenerator($iterator);
             }
 
-            $key   = $iterator->key();
+            $key   = ($iterator instanceof Iterator) ? $iterator->key() : null;
```
keradus: test is needed for this change

jmikola: How do you propose testing this in userland if Traversable cannot be implemented directly?

Given the current implementation, I could exercise the `! $iterator instanceof Iterator` condition by having an IteratorAggregate instance return a second IteratorAggregate; however, that makes an assumption that the earlier logic that resolves an IteratorAggregate will always stop after one level of resolution.

I'd also like to discuss the two outstanding questions in the OP before making further changes. It would be premature to invest time in a test case until we establish what the correct fix should be.

keradus: traversable = array or iterable follower, you need to ensure both are covered by tests

jmikola: > traversable = array or iterable follower, you need to ensure both are covered by tests

I don't think we're on the same page. The [iterable](https://wiki.php.net/rfc/iterable) type may be an array or Traversable. A [Traversable](http://php.net/manual/en/class.traversable.php) may be a userland class that implements [Iterator](http://php.net/manual/en/class.iterator.php) or [IteratorAggregate](http://php.net/manual/en/class.iteratoraggregate.php), but it may also be an internal class that implements Traversable directly and the [`get_iterator`](http://www.phpinternalsbook.com/classes_objects/iterators.html#the-get-iterator-handler) Zend object handler. This is touched upon in the following note from the Traversable documentation:

> Internal (built-in) classes that implement this interface can be used in a foreach construct and do not need to implement IteratorAggregate or Iterator.

Such a test case would require a Traversable instance that is neither an Iterator or an IteratorAggregate instance. Short of adding a dev dependency on a third-party PHP extension that provides an internal Traversable class, I am asking how you would propose testing this. Although the approach described in my last comment ("IteratorAggregate returns another IteratorAggregate") would technically exercise the changed LOC, it makes assumptions about the current implementation of `getCountOf()`, which I would argue is broken per my explanation below.

----

The `getCountOf()` function has an early case to handle Countable instances or array types, which is outside the scope of my change. I assume that is already tested.

This PR is modifying logic within the case for a Traversable instance. Within the Traversable case, the function checks for an IteratorAggregate and attempts to unwrap it *once* by calling [`IteratorAggregate::getIterator()`](http://php.net/manual/en/iteratoraggregate.getiterator.php). The function then checks for a Generator instance and handles that separately. After that point, the existing implementation makes two assumptions, which I believe are premature:

 1. It assumes that the object is an Iterator instance when it invokes the `key()` method. This fails to acknowledge that `IteratorAggregate::getIterator()` can return any Traversable instance.
 2. It assumes any Iterator can be rewound.

Ignoring the need for test coverage for a second, it was not my intention that this PR be merged as-is. The existing change is only a suggestion that demonstrates the first problem with the existing implementation. My goal was to have the questions in the OP answered before submitting any changeset for review (hence the "RFC" in the title).

sebastianbergmann:
I understand what you want to achieve, @jmikola, and I am all for it. I am not entirely sure that I understand your patch. I agree with you that this change cannot be tested without adding a development-time dependency on a third-party PHP extension that provides an internal `Traversable` implementation.

jmikola:
> I am not entirely sure that I understand your patch.

The current patch only attempts to avoid calling a nonexistent method, since it is possible at the modified LOC that we have an object that is Traversable but not an Iterator.

> I agree with you that this change cannot be tested without adding a development-time dependency on a third-party PHP extension that provides an internal Traversable implementation.

Searching the PHP sources for `zend_ce_traversable` turned up the [DatePeriod](http://php.net/manual/en/class.dateperiod.php) class, which has existed since PHP 5.3 and implements Traversable without the Iterator or IteratorAggregate interfaces (thanks @derickr). I'll see if this can be used in a test.

In the interest of keeping this PR limited in scope, I will punt on the two questions in my OP and will instead create a separate GitHub issue for future discussion. IMO, these edge cases are the result of an incomplete design in #1125.

Lastly, the issue I described with unwrapping multiple, nested IteratorAggregate objects should be easy to resolve and test. I will create a separate PR for that.

src/Framework/Constraint/Count.php
```diff
@@ -65,6 +66,10 @@ protected function getCountOf($other)
                 return $this->getCountOfGenerator($iterator);
             }
 
+            if (!$iterator instanceof Iterator) {
+                return \iterator_count($iterator);
+            }
```
jmikola: Previously, this PR used a ternary when assigning `$key`. This guard conditional is likely clearer and avoids making an assumption about how a null value of `$key` will be handled later in the code (i.e. deciding whether to attempt rewinding the iterator).

PR #2653, opened 2017-04-20
Fix duplicate requires annotation in test

yonex:
Found this in RequirementsTest. thanks!


sebastianbergmann:
Not sure that this was not intentional to test something weird. Will need to investigate this.

PR #2661, opened 2017-04-26
Consistent JUnit testcase output of test results when using data providers

astepin:
For simple tests, the attributes "class", "classname", "file" and "line" are set in the XML node "testcase"  (--log-junit) for every executed test. 

But if you use the "@dataProvider" annotation, then no information about the class or the file will be added to the XML node.

Apart from the fact that this is not consistent, this also means that the tests generated by data providers are not recognized by SonarQube, for example.

With my change, I have matched the output in both cases.


sebastianbergmann:
Cherry-picked into `6.1`. Thanks!

PR #2663, opened 2017-04-29
Issue#2591

ArturGoldyn:
Added functionality + created tests for it.
Also, php-cs-fixer.phar has been run against all code.
It was suprisingly easy task, because if you skip method name in process isolation routine it runs the whole class in isolation.


src/Util/Test.php
```diff
@@ -967,7 +981,7 @@ private static function getBooleanAnnotationSetting($className, $methodName, $se
             }
         }
 
-        return null;
+        return;
```
GrahamCampbell: this statement is dead code

GrahamCampbell (changes requested):

src/Util/Test.php
ArturGoldyn: And is this the cause of failing Appveyor?

ArturGoldyn:
Closing this for now, I'll try it without this php-cs-fixer run not on my files.

src/Util/Test.php
GrahamCampbell: Doubt it. Not looked though.

ArturGoldyn: I added a new pull request, without this change.
https://github.com/sebastianbergmann/phpunit/pull/2664

PR #2664, opened 2017-04-29
Issue#2591no2

ArturGoldyn:
Use @runClassInSeparateProcess annotation on test class to allow running a test class with process isolation.


src/Util/PHP/Template/TestCaseMethod.tpl.dist
```diff
@@ -47,7 +47,12 @@ function __phpunit_run_isolated_test()
     $result->beStrictAboutTodoAnnotatedTests({isStrictAboutTodoAnnotatedTests});
     $result->beStrictAboutResourceUsageDuringSmallTests({isStrictAboutResourceUsageDuringSmallTests});
 
-    $test = new {className}('{methodName}', unserialize('{data}'), '{dataName}');
```
sebastianbergmann: Would it make sense to use a new, separate template here? Maybe `TestCaseClass.tpl.dist`.

src/Framework/TestCase.php
```diff
@@ -134,6 +134,13 @@
     protected $runTestInSeparateProcess;
 
     /**
+     * Whether or not this class is to be run in a separate PHP process.
+     *
+     * @var bool
+     */
+    protected $runClassInSeparateProcess;
```
sebastianbergmann: Following the minimal visibility principle, this should be `private`.

src/Util/PHP/Template/TestCaseMethod.tpl.dist
ArturGoldyn: Of course, I think so, especially that I've just made changes to TestCaseMethod file.

src/Framework/TestCase.php
ArturGoldyn: I'll try that.

ArturGoldyn:
This fails now, probably due to change in sebastian/diff package from 1.4.1 to 1.4.2 [https://github.com/sebastianbergmann/diff/https://github.com/sebastianbergmann/diff/](url)

ArturGoldyn:
https://github.com/sebastianbergmann/diff/ is probably the reason for failing tests.

ArturGoldyn:
@sebastianbergmann template added, variable access changed, and all tests passing.

sebastianbergmann:
Thanks!

KintradimCrux:
In the current 6.2 release the method TestSuite::createTest() only calls $_test->setRunClassInSeparateProcess() for tests with a data provider.
Can you add this initialization for tests without a data provider, too?

KintradimCrux:
And the @beforeClass, @afterClass, setUp() and tearDown() methods maybe should be called only in the isolated environment running the class.

sebastianbergmann:
@KintradimCrux Can you please open tickets for these bugs? Thanks!

PR #2666, opened 2017-05-02
Allow using a \Traversable as data provider, not only \Iterater

issei-m:
I'm not sure whether it's a bug or not, so I'm targeting `6.1` the current stable branch.

### DIFF
--- a/src/Util/Test.php
+++ b/src/Util/Test.php
@@ -23,6 +23,7 @@ use ReflectionClass;
 use ReflectionException;
 use ReflectionFunction;
 use ReflectionMethod;
+use Traversable;
 
 /**
  * Test helpers.
@@ -517,7 +518,7 @@ class Test
                     $data = $dataProviderMethod->invoke($object, $methodName);
                 }
 
-                if ($data instanceof Iterator) {
+                if ($data instanceof Traversable) {
                     $data = \iterator_to_array($data);
                 }
 
--- a/tests/_files/MultipleDataProviderTest.php
+++ b/tests/_files/MultipleDataProviderTest.php
@@ -57,11 +57,9 @@ class MultipleDataProviderTest extends TestCase
 
     public static function providerE()
     {
-        return new WrapperIteratorAggregate([
-            [null, 'ok', null],
-            [null, 'ok', null],
-            [null, 'ok', null],
-        ]);
+        yield [null, 'ok', null];
+        yield [null, 'ok', null];
+        yield [null, 'ok', null];
     }
 
     public static function providerF()
--- a/tests/_files/MultipleDataProviderTest.php
+++ b/tests/_files/MultipleDataProviderTest.php
@@ -57,9 +57,11 @@ class MultipleDataProviderTest extends TestCase
 
     public static function providerE()
     {
-        yield [null, 'ok', null];
-        yield [null, 'ok', null];
-        yield [null, 'ok', null];
+        return new WrapperIteratorAggregate([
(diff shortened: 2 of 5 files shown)


keradus:
please add a test for that

bkrukowski:
+1

It is a bug, because function `iterator_to_array` expects instance of `\Traversable`, please take a look on result of command `php --rf iterator_to_array`:

```
Function [ <internal:SPL> function iterator_to_array ] {

  - Parameters [2] {
    Parameter #0 [ <required> Traversable $iterator ]
    Parameter #1 [ <optional> $use_keys ]
  }
}
```

So IMO it should be applied also for older versions.

src/Util/Test.php
```diff
@@ -517,7 +518,7 @@ private static function getDataFromDataProviderAnnotation($docComment, $classNam
                     $data = $dataProviderMethod->invoke($object, $methodName);
                 }
 
-                if ($data instanceof Iterator) {
+                if ($data instanceof Traversable) {
                     $data = \iterator_to_array($data);
```
keradus: why not go even forward and use `iterable` instead ?

issei-m: I don't do that because `iterator_to_array` cannot be passed an array.

issei-m:
@keradus Test has been added!

src/Util/Test.php
bkrukowski: @keradus `Iterable` is not not class or interface, it is pseudo-type, `is_iterable` function was introduced in PHP 7.1 and phpunit supports also older versions

keradus: that's why I lowercased it, didn't I?
is_iterable is polyfilled even in 5.x

https://github.com/symfony/polyfill/tree/master/src/Php71

bkrukowski: @keradus You're right, but as @issei-m has mentioned - `iterator_to_array` expects `\Traversable`.

keradus: I referred to main change of this PR.
And my proposal is deeper - to allow dataProvider not only work properly with Traversable, but all iterable.

whith this change, it will probably be like that, good to prove by utests, which are already there

tests/_files/MultipleDataProviderTest.php
```diff
@@ -57,9 +57,11 @@ public static function providerD()
 
     public static function providerE()
     {
-        yield [null, 'ok', null];
-        yield [null, 'ok', null];
-        yield [null, 'ok', null];
+        return new WrapperIteratorAggregate([
```
keradus: please don't modify existing tests. add new one instead.

src/Util/Test.php
bkrukowski: > And my proposal is deeper - to allow dataProvider not only work properly with Traversable, but all iterable.

I think I don't understand. `Traversable` and `array` are all `iterable` possibilities, so IMO this change is good enough. Is there something what is skipped?

keradus: nah, I asked for tests proving that, which @issei-m  done later ;)

issei-m:
@keradus Sorry for the late response. I've fixed test you mentioned.

src/Util/Test.php
issei-m: @keradus I also don't figure it out. Could you tell me your perspective in more detail, and what should I specifically do here?

keradus (approved):

sebastianbergmann:
Merged manually into `master`. Thanks!

PR #2669, opened 2017-05-02
TestCase - clean up handling expected Exception

keradus:


src/Framework/TestCase.php
```diff
@@ -187,7 +187,7 @@
     /**
      * The code of the expected Exception.
      *
-     * @var int|string
+     * @var null|int|string
```
keradus: it's not initialized with any value, if one would not call `expectException`, then this would remain null.

src/Framework/TestCase.php
```diff
@@ -166,7 +166,7 @@
     /**
      * The name of the expected Exception.
      *
-     * @var string
+     * @var null|string
```
keradus: it's not initialized with any value, if one would not call `expectException`, then this would remain null.

src/Framework/TestCase.php
```diff
@@ -525,15 +525,15 @@ public function hasExpectationOnOutput()
     }
 
     /**
-     * @return string
+     * @return null|string
```
keradus: due to updated property typehinting

src/Framework/TestCase.php
```diff
@@ -525,15 +525,15 @@ public function hasExpectationOnOutput()
...
      */
     public function getExpectedException()
     {
         return $this->expectedException;
     }
 
     /**
-     * @return int|string
+     * @return null|int|string
```
keradus: due to updated property typehinting

src/Framework/TestCase.php
```diff
@@ -1070,8 +1070,7 @@ protected function runTest()
                     )
                 );
 
-                if (\is_string($this->expectedExceptionMessage) &&
-                    !empty($this->expectedExceptionMessage)) {
+                if (!empty($this->expectedExceptionMessage)) {
```
keradus: it's always string

src/Framework/TestCase.php
```diff
@@ -1080,8 +1079,7 @@ protected function runTest()
                     );
                 }
 
-                if (\is_string($this->expectedExceptionMessageRegExp) &&
-                    !empty($this->expectedExceptionMessageRegExp)) {
+                if (!empty($this->expectedExceptionMessageRegExp)) {
```
keradus: it's always string

PR #2672, opened 2017-05-04
add --no-bootstrap option

remicollet:
Allow to ignore bootstrap option from configuration file:

```
--no-bootstrap
--bootstrap ""
```

Context, sometime, we may need to ensure some autoloader is load first, thus using:

`php  -d auto_prepend_file=bootstrap.php /usr/bin/phpunit --no-bootstrap`

Not sure if suitable for 6.1 or master.

### DIFF
--- a/src/TextUI/Command.php
+++ b/src/TextUI/Command.php
@@ -90,6 +90,7 @@ class Command
         'loader='                   => null,
         'log-junit='                => null,
         'log-teamcity='             => null,
+        'no-bootstrap'              => null,
         'no-configuration'          => null,
         'no-coverage'               => null,
         'no-extensions'             => null,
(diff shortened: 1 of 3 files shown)


sebastianbergmann:
I do not understand what you are trying to achieve with this.

remicollet:
> I do not understand what you are trying to achieve with this.

When not using composer autoloader and run test suite of some PHPUnit dependency (e.g. phpdocumentor/reflection-docblock required by phpspec/prophecy) when version differs (v2 vs v3), this allow to force use of 1 version (v2) even if phpunit autoloader use another one (v3).

Ok, this is really a edge case.

sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not you merge it.

PR #2680, opened 2017-05-16
Make agument types consistent with implementation and underlying function argument types

morozov:
The types of `$(expected|actual)Xml` aguments in `assertXml*()` methods and corresponding function aliases are declared as `string` however the underlying method `Xml::load()` allows them to be `string|DOMDocument`.

Static code analysis of the tests using these assertions with [phpstan](https://github.com/phpstan/phpstan) produces false positives like:
```
Parameter #2 $actualXml of method PHPUnit\Framework\Assert::assertXmlStringEqualsXmlFile() expects string, DOMDocument given.
```

### DIFF
--- a/src/Framework/Assert.php
+++ b/src/Framework/Assert.php
@@ -1965,9 +1965,9 @@ abstract class Assert
     /**
      * Asserts that two XML documents are equal.
      *
-     * @param string $expectedFile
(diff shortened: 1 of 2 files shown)


sebastianbergmann:
Merged manually, thanks.

PR #2685, opened 2017-05-19
Stop requiring Printer subclassing for test listeners

Jean85:
This PR stops checking if test listeners are subclasses of `PHPUnit\Util\Printer`. It should not be needed, since the check about the `TestListener` interface should be enough.

The rationale behind this is also that the interface has a FC implementation in 5.x, and the `Printer` doesn't.

This PR targets 5.x; if a different PR is needed to port this to 6, I will gladly write that too.

### DIFF
--- a/src/TextUI/Command.php
+++ b/src/TextUI/Command.php
@@ -882,9 +882,7 @@ class PHPUnit_TextUI_Command
         if (class_exists($printerClass)) {
             $class = new ReflectionClass($printerClass);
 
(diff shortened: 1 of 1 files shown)


Jean85:
Test failures seems unrelated, I've tested locally and it fails with and without my PR.

sebastianbergmann:
A `Printer` is something more than just / different than a `TestListener`.

PR #2689, opened 2017-05-22
Unpack nested IteratorAggregate objects for Count

jmikola:
From my previous comment in https://github.com/sebastianbergmann/phpunit/pull/2642#discussion_r117659301:

> Within the Traversable case, the function checks for an IteratorAggregate and attempts to unwrap it *once* by calling [`IteratorAggregate::getIterator()`](http://php.net/manual/en/iteratoraggregate.getiterator.php). The function then checks for a Generator instance and handles that separately. After that point, the existing implementation makes two assumptions, which I believe are premature:
>
> 1. It assumes that the object is an Iterator instance when it invokes the `key()` method. This fails to acknowledge that `IteratorAggregate::getIterator()` can return any Traversable instance.

This PR adds logic to unwrap nested IteratorAggregate objects until we encounter a Traversable that is not an IteratorAggregate instance. This ensures that generator detection (#2149) and internal Traversables (#2642) are properly detected before we attempt to restore the Iterator's position after counting (#1125).

In a separate commit, the PR also adds missing regression tests for Count's handling of non-nested IteratorAggregate objects.

### DIFF
--- a/tests/_files/TestIteratorAggregate2.php
+++ b/tests/_files/TestIteratorAggregate2.php
@@ -1,0 +1,19 @@
+<?php
+
+/* This class is used for testing a chain of IteratorAggregate objects, since
+ * PHP does allow IteratorAggregate::getIterator() to return an instance of the
+ * same class. */
+class TestIteratorAggregate2 implements IteratorAggregate
+{
+    private $traversable;
+
+    public function __construct(\Traversable $traversable)
+    {
+        $this->traversable = $traversable;
+    }
+
+    public function getIterator()
+    {
+        return $this->traversable;
+    }
+}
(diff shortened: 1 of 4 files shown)


tests/_files/TestIteratorAggregate2.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+/* This class is used for testing a chain of IteratorAggregate objects, since
+ * PHP does allow IteratorAggregate::getIterator() to return an instance of the
+ * same class. */
```
jmikola: See: [`zend_user_it_get_new_iterator()`](https://github.com/php/php-src/blob/php-7.1.5/Zend/zend_interfaces.c#L285). If `IteratorAggregate::getIterator()` returns an instance of the same class, PHP throws an exception. This is a basic attempt to prevent infinite recursion, although it is certainly still possible if two different IteratorAggregate implementations return instances of each other.

sebastianbergmann:
Thanks!

PR #2692, opened 2017-05-24
added --no-logging option to disable logging

### LINKED ISSUE #2671: --no-logging option to overwrite config file
| Q                   | A
| PHPUnit version     | 6.x
| PHP version         | 7.x
| Installation Method | n/a

Similarly to #920 it will be nice to see an option to disable external loggers defined in configuration file. During the development my configuration file contains e.g. testdox logger, which on the other hand makes CI logs extensively verbose.

The problem can be currently hacked configuring `testdox-text` to with target of `php://stdout` and running `phpunit` with `--stderr` option + redirection of STDOUT to `/dev/null`, but I believe there should be a nicer solution for such problem.

What do you guys think about such addition?

evgpisarchik:
Resolves #2671

### DIFF
--- a/src/TextUI/Command.php
+++ b/src/TextUI/Command.php
@@ -92,6 +92,7 @@ class Command
         'log-teamcity='             => null,
         'no-configuration'          => null,
         'no-coverage'               => null,
+        'no-logging'                => null,
         'no-extensions'             => null,
         'printer='                  => null,
         'process-isolation'         => null,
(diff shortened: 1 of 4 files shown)


sebastianbergmann:
Thanks!

kiler129:
Thank you!

evgpisarchik:
my pleasure!

PR #2695, opened 2017-05-28
Add new constraint functions

CedX:
- Added the `directoryExists()` function
- Added the `isInfinite()` function
- Added the `isNan()` function
- Added the `isReadable()` function
- Added the `isWritable()` function


cedx:
@sebastianbergmann Now that PHPUnit has switched to namespaces, I think it would be a good idea, but another breaking change, to move the functions into a namespace.

PR #2700, opened 2017-06-01
50% of time spent in array_unique() in coverage validator 

willemstuursma:
Same as #2699 but for the 5.7 branch.


sebastianbergmann:
Thanks!

Issue #2705, opened 2017-06-14
stderr parameter in phpunit.xml always considered true


odoucet:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 6.2.1
| PHP version         | 7.0.20
| Installation Method | Composer

I'm using a phpunit.xml to declare some variables to PHPUnit. It appears that property 'stderr', when defined, is forced to true.

Both code have same behaviour and output all content to stderr
`<phpunit stderr="false">`
`<phpunit stderr="true">`

This outputs content to stdout : 
`<phpunit>`

pull request to fix the bug will follow in a few minutes.

PR #2706, opened 2017-06-14
Bugfix #2705: if 'stderr=false' is set in phpunit.xml, it is ignored and considered true

odoucet:
see bug #2705 

This bug was tested on phpunit 6.2.1, I do not know if it needs to be applied to other branches.


sebastianbergmann:
Cherry-picked into `5.7` and `6.2`.

PR #2708, opened 2017-06-16
fixed assertJsonStringEqualsJsonString diff formatting on test failure

### LINKED ISSUE #2654: Diff problem with assertJsonStringEqualsJsonString in version 6.*
| Q                   | A
| PHPUnit version     | 6.1.0
| PHP version         | 7.1.0

Hello,

I use the _PHPUnit assert functions_ to write feature tests of APIs with `Behat`. So, the assert `assertJsonStringEqualsJsonString` is very useful.

The problem is: since version `6`, the exception throwed when this assert fails do not show the _diff_ anymore like version `5`.

Output using version `5`:
```
Failed asserting that two objects are equal.
--- Expected
+++ Actual
@@ @@
 stdClass Object (
     'message' => 'Validation Failed'
     'details' => Array (
         0 => stdClass Object (
             'field' => 'name'
-            'message' => 'The name must no be blank.'
+            'message' => 'The name must not be blank.'
             'parameters' => stdClass Object (...)
         )
         1 => stdClass Object (
             'field' => 'email'
-            'message' => 'The email must no be blank.'
+            'message' => 'The email must not be blank.'
             'parameters' => stdClass Object (...)
         )
         2 => stdClass Object (
             'field' => 'role'
-            'message' => 'The role must no be blank.'
+            'message' => 'The role must not be blank.'
             'parameters' => stdClass Object (...)
         )
     )
 )
```

Output using version `6`:
```
Failed asserting that '{"message":"Validation Failed","details":[{"field":"name","message":"The name must not be blank.","parameters":{"value":"null"}},{"field":"email","message":"The email must not be blank.","parameters":{"value":"null"}},{"field":"role","message":"The role must not be blank.","parameters":{"value":"null"}}]}' matches JSON string "{
        "message": "Validation Failed",
        "details": [
          {
            "field": "name",
            "message": "The name must no be blank.",
            "parameters": {
              "value": "null"
            }
          },
          {
            "field": "email",
            "message": "The email must no be blank.",
            "parameters": {
              "value": "null"
            }
          },
          {
            "field": "role",
            "message": "The role must no be blank.",
            "parameters": {
              "value": "null"
            }
          }
        ]
      }". (PHPUnit\Framework\ExpectationFailedException)
```

I tested it also using _PHPUnit test case_ and the diff is also not showed.

Is that an intentional change of `assertJsonStringEqualsJsonString` or a bug?

Thank you very much.

evgpisarchik:
Resolves #2654

Current implementation formats json for error output in the following format:

1)
```
--- Expected
+++ Actual
@@ @@
 [
-    "first",
-    "second"
+    "second",
+    "first"
 ]
```
2)
```
--- Expected
+++ Actual
@@ @@
 {
     "message": "Validation Failed",
     "details": [
         {
             "field": "name",
-            "message1": "The name must not be blank.",
+            "message": "The name must not be blank.",
             "parameters": {
                 "value": "null"
             }
         },
         {
             "field": "email",
             "message": "The email must not be blank.",
             "parameters": {
                 "value": "null"
             }
         },
         {
             "field": "role",
             "message": "The role must not be blank.",
             "parameters": {
                 "value": "null"
             }
         }
     ]
 }

```


GrahamCampbell:
> 100644 → 100755

Unintentional?

evgpisarchik:
oh yes, let me fix this

evgpisarchik:
@sebastianbergmann or other repository owners,
did you have a chance to look at this PR and decide whether this can be merged to main branch?.

src/Framework/Constraint/JsonMatches.php
```diff
@@ -45,12 +46,12 @@ public function __construct($value)
      */
     protected function matches($other)
     {
-        list($error, $recodedOther) = $this->canonicalizeJson($other);
+        list($error, $recodedOther) = Json::canonicalize($other);
```
fefas: One question @evgpisarchik: wouldn't it be better to use non-static methods? I don't know if some kind of injection is possible in this context..

src/Util/Json.php
```diff
@@ -0,0 +1,73 @@
...
+     * JSON object keys are unordered while PHP array keys are ordered.
+     * Sort all array keys to ensure both the expected and actual values have
+     * their keys in the same order.
+     */
+    private static function recursiveSort(&$json)
+    {
+        if (\is_array($json)) {
+            \ksort($json);
+            foreach ($json as $key => &$value) {
+                self::recursiveSort($value);
```
fefas: My suggestion here:
```php
if (false === \is_array($json)) { // or just !\is_array($json)
    return;
}

// ...
```

src/Framework/Constraint/JsonMatches.php
evgpisarchik: According to the project's architecture [Utils](https://github.com/sebastianbergmann/phpunit/tree/master/src/Util) rely on invokation of static methods.

I made my solution consistent with current code base

evgpisarchik: but if you have a proposal on how to refactor current implementation of all Utils let me know please

src/Util/Json.php
evgpisarchik: As it seems the changes won't change behavior of the code.
Is there a significant reason to implement this?

Korri: I agree with @fefas cleaner that was, less imbrication levels

fefas: @evgpisarchik .. this kind of change is one of the _Object Calisthenics_ recommendations (_One level of indentation per method_) and the solution I proposed is called _early return_.

It is only about code readability and complexity .. I believe this is also a recommendation from the book Clean Code of Robert Martin :)

src/Framework/Constraint/JsonMatches.php
fefas: It was just a question, because static access is known to hide the existing dependencies between classes/objects and coupling them in an undesirable way.

Anyway, it looks fine to me if it follows the existent Utils architecture and I don't know the code enough to have a refactor suggestion 😝

src/Util/Json.php
evgpisarchik: makes sense,
implemented

PR #2715, opened 2017-06-21
Add forward compatible Test interface

b1rdex:
5.7 is still alive and it misses much of forward compatibility.

### DIFF
--- a/src/ForwardCompatibility/Test.php
+++ b/src/ForwardCompatibility/Test.php
@@ -1,0 +1,13 @@
+<?php
+/*
+ * This file is part of PHPUnit.
+ *
+ * (c) Sebastian Bergmann <sebastian@phpunit.de>
+ *
+ * For the full copyright and license information, please view the LICENSE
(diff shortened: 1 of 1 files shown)


sebastianbergmann:

The forward compatibility layer will not be extended. It is intended to only cover the most common use case, which is basic usage of `PHPUnit\Framework\TestCase`.

b1rdex:
Sad. Let me argue why I added that.

I added this because forward compatibility layer has `TestListener` that requires `PHPUnit_Framework_Test` in it's signature. The only way to solve this — add class alias with such name (class extension will not pass signature checks).

Writing `Listener`s that are compatible with phpunit 5+6 is impossible without that aliasing and this hack is added to client code, which is not correct.

sebastianbergmann:
Thank you for clarifying. I'll extend the forward compatibility layer with `AssertionFailedError`, `Test`, and `TestSuite` to ease enable the development of test listeners that are compatible with PHPUnit 5 and 6.

PR #2722, opened 2017-07-01
introduce 'requires OSFAMILY'

### LINKED ISSUE #2697: RFC: `@requires OSFAMILY`
Currently, `@requires OS` is checking against `PHP_OS` via regex
(`@requires OS Lin` -> `/Lin/i`).

On PHP It's tricky, as `PHP_OS` could be `WIN32`, `WINNT`, `Windows`. It would be nice to add possibility to check against `PHP_OS_FAMILY` (either by dedicated annotation or by extending what `@requires OS` checks)
WDYT?

keradus:
Closes #2697

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -37,7 +37,7 @@
         "phpunit/phpunit-mock-objects": "^4.0",
         "sebastian/comparator": "^2.0",
         "sebastian/diff": "^1.4.3 || ^2.0",
-        "sebastian/environment": "^3.0.2",
+        "sebastian/environment": "^3.1@dev",
         "sebastian/exporter": "^3.1",
         "sebastian/global-state": "^1.1 || ^2.0",
         "sebastian/object-enumerator": "^3.0.2",
--- a/composer.json
+++ b/composer.json
@@ -37,7 +37,7 @@
         "phpunit/phpunit-mock-objects": "^4.0",
         "sebastian/comparator": "^2.0",
         "sebastian/diff": "^1.4.3 || ^2.0",
-        "sebastian/environment": "^3.1@dev",
+        "sebastian/environment": "^3.1",
         "sebastian/exporter": "^3.1",
         "sebastian/global-state": "^1.1 || ^2.0",
         "sebastian/object-enumerator": "^3.0.2",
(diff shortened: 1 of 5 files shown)


composer.json
```diff
@@ -37,7 +37,7 @@
         "phpunit/phpunit-mock-objects": "^4.0",
         "sebastian/comparator": "^2.0",
         "sebastian/diff": "^1.4.3 || ^2.0",
-        "sebastian/environment": "^3.0.2",
+        "sebastian/environment": "^3.1@dev",
```
keradus: Could 3.1 be deployed ?
alternatively, I could use php polyfilled constant from `symfony/polyfill`.

sebastianbergmann:
Yes, I'll release 3.1 shortly. I wanted to wait a while until the constant's value for Apple had settled.

src/Util/Test.php
```diff
@@ -190,10 +191,14 @@ public static function getRequirements($className, $methodName)
...
-            $requires['OS'] = \sprintf(
-                '/%s/i',
-                \addcslashes($matches['value'][$count - 1], '/')
-            );
+            foreach (\range(0, $count - 1) as $i) {
+                $requires[$matches['name'][$i]] = $matches['name'][$i] === 'OSFAMILY'
+                    ? $matches['value'][$i]
+                    : \sprintf(
+                        '/%s/i',
+                        \addcslashes($matches['value'][$i], '/')
```
keradus: logic when to use strictly provided value and when to decorate it for regex matching can be moved direct to place, when we compare the `'OS'` value. I'm for that movement, but with that, we would need to update all related tests (to not contain `/Linux/i` but just `Linux`). WDYT @sebastianbergmann ?

sebastianbergmann:
Makes sense to me.

composer.json
sebastianbergmann: 3.1.0 has been released

src/Util/Test.php
sebastianbergmann: Makes sense to me.

keradus: done

composer.json
keradus: done

PR #2737, opened 2017-07-12
Add package attribute for junit

jskarpe:
When using xUnit plugin within Jenkins context to parse PHPUnit junit.xml report, the test result lists PHPUnit tests under package "(root)". This patch will enable testsuite name to be displayed here instead.

### DIFF
--- a/src/Util/Log/JUnit.php
+++ b/src/Util/Log/JUnit.php
@@ -229,6 +229,7 @@ class JUnit extends Printer implements TestListener
     {
         $testSuite = $this->document->createElement('testsuite');
         $testSuite->setAttribute('name', $suite->getName());
+        $testSuite->setAttribute('package', $suite->getName());
 
         if (\class_exists($suite->getName(), false)) {
             try {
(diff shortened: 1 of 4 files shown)


src/Util/Log/JUnit.php
```diff
@@ -229,6 +229,7 @@ public function startTestSuite(TestSuite $suite)
     {
         $testSuite = $this->document->createElement('testsuite');
         $testSuite->setAttribute('name', $suite->getName());
+        $testSuite->setAttribute('package', $suite->getName());
```
sebastianbergmann: This does not make sense to me. `$suite->getName()` is probably not the package name.

Yuav: Not sure what package name should be in this context though. I took the approach from https://www.npmjs.com/package/karma-junit-reporter which does call it suite

Perhaps better to introduce a new variable for this particular string?

sebastianbergmann: My guess is that JUnit uses the name of the package the class belongs to. In PHP that would be the namespace of the class (or `Vendor_Package` in case the class is not namespaced and its name is `Vendor_Package_Class`).

sebastianbergmann: http://llg.cubic.org/docs/junit/ says "Derived from testsuite/@name in the non-aggregated documents. optional".

sebastianbergmann:

PR #2738, opened 2017-07-13
Issue 2271/allow custom cli options

jscodefix:
Adds new command line option (-g) that enables the user to assign global variables (... well, from the command line).  The highlights are:
- implemented with eval() in the Command class
- includes a revision to the command line help output (and revised .phpt tests of the output)
- includes a unit test of the new Command::handleGlobals() method (using reflection to test protected method)
- the command line option globals are similar to, but **override**, the globals specified in the phpunit.xml configuration file (using < php >< var >)

Example usage:
```$ phpunit -v -g myInt=5 -g myBool=true -g myValue=null -g myStr="good stuff" -g myArr='["foo"=>"bar"]' MyTest```

### DIFF
--- a/src/TextUI/Command.php
+++ b/src/TextUI/Command.php
@@ -378,7 +378,6 @@ class Command
                     break;
 
                 case 'g':
-                    echo "DEBUG: \$option[1] = '$option[1]'\n";
                     $this->arguments['globals'][] = $option[1];
                     break;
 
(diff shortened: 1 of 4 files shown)


sebastianbergmann:

As far as I understand, the name of `-g` is inspired by "global". I find it confusing to use `-g` for setting a global variable and would prefer a name along the lines of `--set-global-variable`.

I also I do not like the magic in https://github.com/sebastianbergmann/phpunit/pull/2738/files#diff-cd488f091cbf795c2aeeb97f5342dec8R1081 to create PHP variables of specific types based on "syntax" passed to `-g`.

PR #2740, opened 2017-07-17
add message into skipped element of jUnit report

SimonHeimberg:


SimonHeimberg:
This information was present in the json report.

Maybe the attribute `type` would be nice as well, to distinguish skipped and ignored.
The stacktrace is probably unnecessary, but would do no harm. So`doAddFault()` could be used instead of `doAddSkipped`.

sebastianbergmann:
But is the information part of http://llg.cubic.org/docs/junit/?

SimonHeimberg:
No, it is not. Thanks for the link, I failed to find the specification.

Is there any other possibility to add this information while remaining in the specification? It was very handy in the json report.
It could be written in `<system-out>` or `<system-error>`, but it does not have anything to do with this. A comment `<!-- -->` could be added, but this is not nice for reading by a parser. 

BTW, while looking at the specification, I see that `<failure>` and `exception` have message **attributes**, but no data. But phpunit writes the message as data. Should there be done anything?

SimonHeimberg:
`<skipped message=.../>` seems to be used by pytest: https://github.com/pytest-dev/pytest/issues/2190
and it is supported by 
* jenkins https://issues.jenkins-ci.org/browse/JENKINS-21883, 
* xUnit.net https://github.com/jenkinsci/xunit-plugin/pull/27/files#diff-a50ab9758dad0381f1a9ca5a72fe62dbR19

sebastianbergmann:

Issue #2237, opened 2016-07-07
assertArraySubset() should provide the diff when assertion fails


klesun:
Greetings, noble developers of phpunit
The talk will go about `PHPUnit_Framework_Assert::assertArraySubset()` introduced in https://github.com/sebastianbergmann/phpunit/pull/1377

Thank you very much for providing this useful assertion function.
It is very nice, but lacks one very important thing - info about _what exactly mismatched_

Consider such code:

``` php
$expected = [
    'a' => 'item a',
    'b' => 'item b',
    'c' => [
        'd' => 'item d',
    ],
];

$this->assertArraySubset($expected, [
    'a' => 'item a',
    'b' => 'item k', // wrong value
    'c' => [
        // 'd' not present
        'g' => 'item g',
    ],
]);
```

With current implementation it would generate such error:

``` php
1) RetrievalCommandManagerTest::checkServicesImportProcess with data set #0 ...
Failed asserting that an array has the subset Array &0 (
    'a' => 'item a'
    'b' => 'item b'
    'c' => Array &1 (
        'd' => 'item d'
    )
).
```

You see what's fishy here? It tells you that subset is incorrect, but it does not tell, where _exactly_ it is incorrect.

How proper error should look like:

``` php
1) RetrievalCommandManagerTest::checkServicesImportProcess with data set #0 ...
Failed asserting that an array has the subset.
--- Expected
+++ Actual
@@ @@
Array &0 (
       'a' => 'item a'
---    'b' => 'item b'
+++    'b' => 'item k'
       'c' => Array &1 (
---        'd' => 'item d'
       )
).
```

I can implement the failure message with diff and do a pull request. Do you like the idea?

klesun:
How I basically override the function it in my top-level code:
```php
public static function assertArraySubset(array $expectation, array $reality, bool $strict = false, string $message = '')
{
//    parent::assertEquals($expectation, $reality, $strict, $message);
    foreach ($expectation as $key => $value) {
        self::assertArrayHasKey($key, $reality, $message);
        if (is_array($value)) {
            self::assertArraySubset($value, $reality[$key], $strict, $message.'['.$key.']');
        } else {
            self::assertEquals($value, $expectation[$key], $message);
        }
    }
}
```

jmauerhan:
Should your diff also include the 'g' key which is present but not expected?

klesun:
@jmauerhan, thanks for reply

I believe it shouldn't.
[...]

PR #2741, opened 2017-07-19
Fix to assertArraySubset() for its lacking error message

rrehbein:
Resolves: #2237

Incomplete: adding unit test for the expected output.


sebastianbergmann:
Merged manually, thanks.

PR #2742, opened 2017-07-22
Allow custom autoloader?

skylerfenn:
Allows custom autoloader outside the vendor folder.

### DIFF
--- a/phpunit
+++ b/phpunit
@@ -27,7 +27,7 @@ if (!ini_get('date.timezone')) {
     ini_set('date.timezone', 'UTC');
 }
 
-foreach (array(__DIR__ . '/../../autoload.php', __DIR__ . '/../vendor/autoload.php', __DIR__ . '/vendor/autoload.php') as $file) {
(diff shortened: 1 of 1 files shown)


skylerfenn:
What do you all think of this?

Also, how can I fix the failed codecov/project check?

sebastianbergmann:
This makes no sense to me.

skylerfenn:
I see. Sorry, I should've put more context to this. 

Laravel has auto-loaded helper functions. To override those functions, we need a way to put define functions before those helper functions are auto-loaded. This is easy enough in the project by defining functions before including the autoload.php file. However, phpunit has no way to define anything before composer's autoload.php file is required. My addition to the array allowed a custom autoload.php to go one level outside of the vendor folder which would give some desired control. Looking back on it, it was a stinky suggestion. 

I've realized that a more appropriate solution would be to modify Laravel by moving the helper functions file from autoloading to a service provider. Then project specific definitions can be allowed to override them.

Until I can't make a pull request for that, I found another way to include files before Laravel's auto-loaded helper functions with this package: https://github.com/funkjedi/composer-include-files.

Thanks again!

Issue #2734, opened 2017-07-11
Compatibility with sebastian/diff ^2.0


sebastianbergmann:

sebastianbergmann:
As of b2ea8d5bf3228a07996057a533cd3880f0e6a4d1 `sebastian/diff ^2.0` is already allowed but the code is not ready for it.

sebastianbergmann:
@SpacePossum It would be great if you could send a pull request that makes PHPUnit 6.2 compatible with `sebastian/diff ^2.0`.

sebastianbergmann:
I have pulled the `2.0.0` release of `sebastian/diff` because of the mistake I mentioned in https://github.com/sebastianbergmann/phpunit/issues/2734#issuecomment-314440796. Not sure why I did that, probably because back in May there was not a BC break in the diff component's code (and the only reason for the bump of the major version was that PHP 5 was no longer supported).

As a result of this mistake, PHPUnit 6.1.4, 6.2.0, 6.2.1, 6.2.2, and 6.2.3 allowed the installation of the `2.0.0` release of `sebastian/diff` but, of course, the code of these releases is not compatible.

The only solution that I saw to fix this mess was to 1) pull the `2.0.0` release of `sebastian/diff` and 2) skip that major version and release the current `master` of diff as `3.0.0` (not done yet).

SpacePossum:
Makes sense, sorry for getting you into this trouble.

We could update `sebastian/diff` to make v2 compat with PHPUnit 6.1.4, 6.2.0, 6.2.1, 6.2.2, and 6.2.3.
It would only require a change in the constructor of the `Differ` class, let me know if you want me to try it out.

keradus:
but that would actually be one-liner change... so not a big thingy, right? :)

keradus:
and with deprecation notice when used with v1-style :D

SpacePossum:
would look something like https://github.com/sebastianbergmann/diff/pull/66/files
(without deprecation notice, as I'm not sure how this done in the PHPUnit ecosystem ATM)

sebastianbergmann:
Here is what I get with current PHPUnit `6.2` branch and current `2.0` branch of `sebastian/diff`:
```
PHPUnit 6.2.3-1-g4bc5697ec by Sebastian Bergmann and contributors.

Runtime:       PHP 7.1.6 with Xdebug 2.5.5
Configuration: /usr/local/src/phpunit/phpunit.xml

[...]

PR #2748, opened 2017-07-28
WIP / POC - start of fixing tests with diff 2.0

SpacePossum:
Changes in `composer.json` should be reverted in the end.... but lets first fix the tests and see if all is OK with diff 2.x

Alright... so... the issue there is now is that `comparator` v.2 is released with a `diff only` output format.
This means we don't get the udiff `@@ @@` header.
Tried to correct here https://github.com/sebastianbergmann/comparator/pull/40

Other way would be to change the `diff only` output builder  to be just like `udiff` in the `diff` package. Than in correct the output builder used by `comparator` in `comparator 3`. This can be done because `diff 2` is not released yet.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -55,7 +55,8 @@
         "platform": {
             "php": "7.0.0"
         },
-        "sort-packages": true
+        "sort-packages": true,
+        "optimize-autoloader": true
     },
     "suggest": {
         "phpunit/php-invoker": "^1.1",
--- a/composer.json
+++ b/composer.json
@@ -36,7 +36,6 @@
         "phpunit/php-timer": "^1.0.6",
         "phpunit/phpunit-mock-objects": "^4.0",
         "sebastian/comparator": "^2.0",
-        "sebastian/diff": "^1.4.3 || ^2.0",
         "sebastian/environment": "^3.0.2",
         "sebastian/exporter": "^3.1",
         "sebastian/global-state": "^1.1 || ^2.0",
@@ -45,7 +44,8 @@
         "sebastian/version": "^2.0"
     },
     "require-dev": {
-        "ext-PDO": "*"
+        "ext-PDO": "*",
+        "sebastian/diff": "^2.0"
     },
     "conflict": {
         "phpunit/dbunit": "<3.0",
@@ -78,6 +78,7 @@
             "tests/_files/CoveredFunction.php"
         ]
     },
+    "minimum-stability": "dev",
     "extra": {
         "branch-alias": {
             "dev-master": "6.2.x-dev"
--- a/tests/Framework/ConstraintTest.php
+++ b/tests/Framework/ConstraintTest.php
@@ -644,10 +644,8 @@ Failed asserting that two strings are equal.
 -b\\n
 +p\\n
 @@ @@
- i\\n
 -j\\n
 +w\\n
- k'
 
 EOF
             ],
@@ -669,7 +667,6 @@ Failed asserting that two arrays are equal.
  Array (
 -    0 => 0
 +    0 => 1
- )
 
 EOF
             ],
@@ -681,7 +678,6 @@ Failed asserting that two arrays are equal.
  Array (
 -    0 => true
 +    0 => 'true'
- )
 
 EOF
             ],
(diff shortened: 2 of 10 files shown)


composer.json
```diff
@@ -45,7 +44,8 @@
         "sebastian/version": "^2.0"
     },
     "require-dev": {
-        "ext-PDO": "*"
+        "ext-PDO": "*",
+        "sebastian/diff": "^2.0"
```
SpacePossum: just to see the impact of updating

tests/Framework/ConstraintTest.php
```diff
@@ -644,10 +644,8 @@ public function isEqualProvider()
 -b\\n
 +p\\n
 @@ @@
- i\\n
 -j\\n
 +w\\n
- k'
```
SpacePossum: new default of `diff` no longer outputs lines that are the same at the end of the diff, and if there are more than X consecutive unchanged lines (5 IIRC)

tests/Regression/GitHub/503.phpt
```diff
@@ -25,8 +25,8 @@ Failed asserting that two strings are identical.
 +++ Actual
 @@ @@
  #Warning: Strings contain different line endings!
- 'foo
- '
+-'foo
++'foo
```
SpacePossum: This change looks good to.
On `diff` v1.4 when using different line endings for a line the line was outputted with any of the two line breaks used.
On `diff` 2.x it lists both lines (with correct `-` or `+` prefix) with their linebreaks, making it possible to "see" the different line breaks.

tests/TextUI/failure-isolation.phpt
```diff
@@ -132,7 +129,6 @@ Failed asserting that string matches format description.
 +++ Actual
 @@ @@
 -FOO
--
 +...BAR...
```
SpacePossum: hmm not sure about this one, need to have a better look

tests/TextUI/teamcity-inner-exceptions.phpt
```diff
@@ -19,7 +19,7 @@ PHPUnit %s by Sebastian Bergmann and contributors.
 
 ##teamcity[testStarted name='testPrintingChildException' locationHint='php_qn://%s%etests%e_files%eExceptionStackTest.php::\ExceptionStackTest::testPrintingChildException' flowId='%d']
 
-##teamcity[testFailed name='testPrintingChildException' message='Child exception|nmessage|nFailed asserting that two arrays are equal.|n--- Expected|n+++ Actual|n@@ @@|n Array (|n-    0 => 1|n+    0 => 2|n )|n' details=' %s%etests%e_files%eExceptionStackTest.php:14|n |n Caused by|n message|n Failed asserting that two arrays are equal.|n --- Expected|n +++ Actual|n @@ @@|n  Array (|n -    0 => 1|n +    0 => 2|n  )|n |n %s%etests%e_files%eExceptionStackTest.php:10|n ' flowId='%d']
+##teamcity[testFailed name='testPrintingChildException' message='Child exception|nmessage|nFailed asserting that two arrays are equal.|n--- Expected|n+++ Actual|n@@ @@|n Array (|n-    0 => 1|n+    0 => 2|n' details=' %s%etests%e_files%eExceptionStackTest.php:14|n |n Caused by|n message|n Failed asserting that two arrays are equal.|n --- Expected|n +++ Actual|n @@ @@|n  Array (|n -    0 => 1|n +    0 => 2|n |n %s%etests%e_files%eExceptionStackTest.php:10|n ' flowId='%d']
```
SpacePossum: I'm not familiar with this format, so cannot state much about this change.
If the format requires that the `(` of the `Array` are closed than this need (re)work in the `diff` package I think

sebastianbergmann:
This is still labeled "WIP / POC" (and has conflicts that must be resolved). Is this ready to be merged? Thanks!

keradus:
Travis is red thus I think it's not.
Please say do you like the direction, so one won't waste time to finish PR and make Travis happy if you would not like the work in the end

sebastianbergmann:
Status Update: https://github.com/sebastianbergmann/phpunit/issues/2734#issuecomment-319891775

sebastianbergmann:
@keradus Why would I not merge this in the end?

sebastianbergmann:
I cherry-picked your changes one by one, @SpacePossum, and committed them in d21aca0876520b67b5ab57ae47c26021997de090. Thank you for your effort!

keradus:
> @keradus Why would I not merge this in the end?

as travis was failing ;)

great to hear to pick it up and finished ;)

Issue #1164, opened 2014-03-07
Fatal error when extenting TestCase without calling parent constructor


voda:
In one of my test I had defined a constructor and forgot to call the parent constructor. When trying to run this test phpunit crashed with `Fatal error: Call to a member function getMock() on a non-object in phar:///usr/bin/phpunit/phpunit/Framework/TestCase.php on line 1301`. It took me some time the find the wrong test, because there is no stacktrace. This is an issue with PHPUnit 4, the test worked in PHPUnit 3.x. It would be nice to print a message saying something like 'Call parent::__construct in class foo.'

Issue #621, opened 2012-08-09
Constructor in subclass of PHPUnit_Framework_TestCase causes @dataProvider to break


nicklevett:
Having a constructor in your test causes the dataProvider annotation to break with an error about missing arguments:

```
PHPUnit 3.6.12 by Sebastian Bergmann.

...E

Time: 0 seconds, Memory: 3.25Mb

There was 1 error:

1) Test::testTheProvider
Missing argument 1 for Test::testTheProvider()

/home/nick/development/PHP-1.php:43

FAILURES!
Tests: 4, Assertions: 5, Errors: 1.
```

Removing the constructor causes the testTheProvider to work again, and does not seem to break @depends annotations.

``` php
<?php
class Test extends PHPUnit_Framework_TestCase {

    public function __construct() {
        parent::__construct();
    }

    public function testEmpty() {
        $stack = array();
        $this->assertEmpty($stack);

        return $stack;
    }

    /**
     * @depends testEmpty
     */
    public function testPush(array $stack) {
        array_push($stack, 'foo');
        $this->assertEquals('foo', $stack[count($stack)-1]);
        $this->assertNotEmpty($stack);

        return $stack;
    }

    /**
     * @depends testPush
     */
    public function testPop(array $stack) {
        $this->assertEquals('foo', array_pop($stack));
        $this->assertEmpty($stack);
    }

    public function theProvider() {
        return array(
            array(true, false)
        );
    }

    /**
     * @dataProvider theProvider
     */
    public function testTheProvider($var1, $var2) {
        //$var1 = true;
        //$var2 = false;

        $this->assertNotEquals($var1, $var2);
    }

}
?>
```

RopoMen:
Hi,
I was lucky to find this answer, saved lot of time. Ty, @foued611 

My case is like this and it fails.
```php
# my base test class
use PHPUnit_Framework_TestCase;

class mytest extends PHPUnit_Framework_TestCase {
   public function __construct() {
     # do some common stuff to my tests...
  }
}


class IntegrationTest extends mytest {
  public function __construct() {
     parent::__construct();
  }
   /**
    * @dataProvider additionProvider
    */
    public function testAdd($a, $b, $expected)
    {
        $this->assertEquals($expected, $a + $b);
    }

    public function additionProvider()
    {
        return [
            'adding zeros'  => [0, 0, 0],
            'zero plus one' => [0, 1, 1],
            'one plus zero' => [1, 0, 1],
[...]

Issue #540, opened 2012-04-09
Surprising behavior when running TestTest.php


taenfu:
On PHPUnit 3.6.10, PHP 5.3.5, running a testcase named TestTest.php and containing 1 false assertion leads to this error :

> phpunit unit\TestTest.php

PHP Fatal error:  Uncaught exception 'PHPUnit_Framework_ExpectationFailedException' with message 'KO1
Failed asserting that 2 matches expected 3.' in C:\Apps\Wampee-2.1-beta-2\bin\php\php5.3.5\includes\PHPUnit-3.6.10\PHPUnit\Framework\Constraint\IsEqual.php:171
Stack trace:
#0 C:\Apps\Wampee-2.1-beta-2\bin\php\php5.3.5\includes\PHPUnit-3.6.10\PHPUnit\Framework\Assert.php(2100): PHPUnit_Framework_Constraint_IsEqual->evaluate(2, 'KO1')
#1 C:\Apps\Wampee-2.1-beta-2\bin\php\php5.3.5\includes\PHPUnit-3.6.10\PHPUnit\Framework\Assert.php(441): PHPUnit_Framework_Assert::assertThat(2, Object(PHPUnit_Framework_Constraint_IsEqual), 'KO1')
#2 C:\Apps\Wampee-2.1-beta-2\www\consultAssur\protected\tests\unit\TestTest.php(7): PHPUnit_Framework_Assert::assertEquals(3, 2, 'KO1')
#3 C:\Apps\Wampee-2.1-beta-2\bin\php\php5.3.5\includes\PHPUnit-3.6.10\PHPUnit\Framework\TestSuite.php(476): TestTest->testTest()
#4 C:\Apps\Wampee-2.1-beta-2\bin\php\php5.3.5\includes\PHPUnit-3.6.10\PHPUnit\Framework\TestSuite.php(831): PHPUnit_Framework_TestSuite::crea in C:\Apps\Wampee-2.1-beta-2\bin\php\php5.3.5\includes\PHPUnit-3.6.10\PHPUnit\Framework\Constraint\IsEqual.php on line 171

Fatal error: Uncaught exception 'PHPUnit_Framework_ExpectationFailedException' with message ' in C:\Apps\Wampee-2.1-beta-2\bin\php\php5.3.5\includes\PHPUnit-3.6.10\PHPUnit\Framework\Constraint\IsEqual.php on line 171

PHPUnit_Framework_ExpectationFailedException: KO1
[...]

PR #2766, opened 2017-08-28
Prevent the TestCase constructor from being overridden

duncan3dc:
The `TestCase` class relies on this constructor being called, by making it final we prevent developers from shooting themselves in the foot by overriding it (#621, #540, #1164)

### DIFF
--- a/src/Framework/TestSuite.php
+++ b/src/Framework/TestSuite.php
@@ -176,6 +176,17 @@ class TestSuite implements Test, SelfDescribing, IteratorAggregate
             return;
         }
 
+        if ($constructor->class !== TestCase::class) {
+            $this->addTest(
+                self::warning(
+                    sprintf(
+                        'Class "%s" overrides the constructor, you should override the `setUp()` method instead.',
+                        $theClass->getName()
+                    )
+                )
+            );
+        }
+
         foreach ($theClass->getMethods() as $method) {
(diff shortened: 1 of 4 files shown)


sebastianbergmann:
This is a breaking change that, if at all, can only be made in a major release.

sebastianbergmann:
Furthermore, as can be glimpsed at in the result of PHPUnit's own test suite (https://travis-ci.org/sebastianbergmann/phpunit/builds/269362111), this change is far from the one-liner that was submitted.

duncan3dc:
Would you be interested in a full patch that targets the 7.0 release?

sebastianbergmann:
Interested: sure. No promises, though, that it will be merged.

The problem you are trying to solve is real. As a first step, a note/warning should be added to the documentation that `TestCase::__construct()` should not be overwritten. When it is overwritten then the original arguments must be accepted and forwarded to the original implementation.

The right solution to this problem is, of course, the declaration of `TestCase::__construct()` as `final`. I cannot, however, foresee all consequences of this change right now.

duncan3dc:
Hi @sebastianbergmann, I've approached this differently now, by adding a warning to the test suite if a class has overridden the constructor (in the same way a private constructor is handled).

This should avoid BC issues, while still making developers aware of the problem

sebastianbergmann:

Issue #2591, opened 2017-03-30
Run a class in separate process.


ArturGoldyn:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 4.0.2
| PHP version         | 5.3.3
| Installation Method | PHAR

We can run individual tests in separate processes, but we cannot run a test class in a separate process.
It would be nice to have **@runTestsClassInSeparateProcess**, not only @runTestsInSeparateProcesses

The goal is to run setUpBeforeClass once (it takes a few seconds), then perform all tests in separate process, and then at the end lose all global variable changes that occured.
Unfortunately some of our objects collide and don't allow us to pass to the next test, but they work with --process-isolation parameter perfectly. Therefore we would like to have a **@runTestsClassInSeparateProcess** annotation to run just few classes in separate process, but use this one process for all tests in this class.

ArturGoldyn:
I've started working on it yesterday, will finish this week.

ArturGoldyn:
For now it fails in AppVeyor. https://github.com/sebastianbergmann/phpunit/pull/2664

ArturGoldyn:
@sebastianbergmann I did it, and fixed that failing merge request, do you consider merging my branch to some release? Or does it need any alterations?

kaphert:
Am I correct that this feature is implemented since 6.2 but is missing in the documentation?

WalterWoshid:
https://github.com/sebastianbergmann/phpunit/pull/5233

PR #2770, opened 2017-09-06
Fixed test failure (ref #2591)

mbeccati:
Test failure across the board with 7.0/7.1/7.2-dev:

https://revive.beccati.com/bamboo/browse/PHP-PHPUN-1221

For future reference (build logs are pruned after some time): 
immagine

PR against master as I'm only running CI for master at this stage. Not sure how far back it should be patched.

### DIFF
--- a/tests/Regression/GitHub/2591-separate-function-no-preserve-no-bootstrap.phpt
+++ b/tests/Regression/GitHub/2591-separate-function-no-preserve-no-bootstrap.phpt
@@ -26,12 +26,10 @@ Time: %s, Memory: %s
 There were 2 errors:
 
 1) Issue2591_SeparateFunctionNoPreserveTest::testChangedGlobalString
-PHPUnit\Framework\Exception: PHP Fatal error:  Class 'PHPUnit\Framework\TestCase' not found %s
-Fatal error: Class 'PHPUnit\Framework\TestCase' not found %s
+PHPUnit\Framework\Exception: Fatal error: Class 'PHPUnit\Framework\TestCase' not found %s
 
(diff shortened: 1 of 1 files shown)


mbeccati:
Oh well, travis complains... not sure what's happening here TBH.

SpacePossum:
The failing `xmllint` cases on Travis are indeed not related to this PR.
It seems it is not (no longer) shipped with Travis env.

mbeccati:
@SpacePossum actually travis behaves like the test did before this PR, while my CI box output is different, i.e. 2 lines vs 1 line and "PHP Fatal Error" vs "Fatal Error".

sebastianbergmann:
Looks to me (at first glance) that this test needs to be split, one for versions of PHP before the change and for versions of PHP after the change.

mbeccati:
@sebastianbergmann Makes sense. However, I haven't been able ot figure out what condition triggers the different output on my CI builds. On the same box I do have other PHP7.0/7.1 installs, built with --disable-debug on Aug 30 2017 (CI builds with --enable-debug) that show the same behaviour. A PHP 7.2-dev built on Jul  5 2017 behaves like Travis CI.

sebastianbergmann:
I cannot reproduce the issue, sorry.

mbeccati:
@sebastianbergmann with help from Dmitry I was able to understand that the different output is caused by the display_errors directive. Depending on the setting (off, STDOUT) the ouput is one or another. I'll create a new PR, hopefully covering both.

sebastianbergmann:
Thanks!

PR #2778, opened 2017-09-19
AssertContains - handle empty string in string.

### LINKED ISSUE #2520: mb_strpos(): Empty delimiter
# Error:
```
PHPUnit 6.0.7 by Sebastian Bergmann and contributors.


mb_strpos(): Empty delimiter
.../vendor/phpunit/phpunit/src/Framework/Constraint/StringContains.php:58 // this line added manualy because error handler catch this and after run debug mode then show line number and correct class
.../vendor/symfony/phpunit-bridge/DeprecationErrorHandler.php:73
.../tests/AppBundle/Controller/LoginControllerTest.php:25
 


Time: 108 ms, Memory: 14.00MB


ERRORS!
Tests: 1, Assertions: 2, Errors: 1.

Process finished with exit code 2
```

# My example test:
```
public function testIndex()
    {
        $client = static::createClient();

        $crawler = $client->request('GET', '/');

        static::assertEquals(200, $client->getResponse()->getStatusCode());
        static::assertContains('', $crawler->filter('body')->text());
    }
```
I think assertContains should return true if expected (needle) value is the same as (haystack) or if some one check if empty string contains another empty string.

# Visualized problem:
```
static::assertContains('', '');
```
or if someone would like to use assertContains for comparing two empty strings, maybe library inform that is not good way - I don't now. Like "assertContains is not for comparing to empty strings!... Please use assertEquals"? - I don't now. I'm not a specialist in XUnit philosophy. Now I fix this by change my testcase. But it is temporary solution.

SpacePossum:
closes https://github.com/sebastianbergmann/phpunit/issues/2520

This PR make the following assert pass:
```
$this->assertContains('', 'test');
```

previous:
```
There was 1 error:

1) myTest::testA
mb_strpos(): Empty delimiter

/home/possum/work/phpunit/src/Framework/Constraint/StringContains.php:59

```

### DIFF
--- a/src/Framework/Constraint/StringContains.php
+++ b/src/Framework/Constraint/StringContains.php
@@ -52,6 +52,10 @@ class StringContains extends Constraint
      */
     protected function matches($other)
     {
+        if ('' === $this->string) {
+            return true;
+        }
+
         if ($this->ignoreCase) {
             return \mb_stripos($other, $this->string) !== false;
         }
--- a/tests/Framework/AssertTest.php
+++ b/tests/Framework/AssertTest.php
@@ -93,6 +93,11 @@ class AssertTest extends TestCase
         $this->assertContainsOnlyInstancesOf(\Book::class, $test2);
     }
 
+    public function testAssertContainsEmptyStringInString()
+    {
+        $this->assertContains('', 'test');
+    }
+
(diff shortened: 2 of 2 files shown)


localheinz (approved):
:+1:

sebastianbergmann:
Merged manually, thanks.

PR #2777, opened 2017-09-19
Travis - add libxml2-utils for xmllint

SpacePossum:
Test to see if this fixes the;

```
The command "xmllint --noout --schema phpunit.xsd phpunit.xml" exited with 127.
0.18s$ xmllint --noout --schema phpunit.xsd tests/_files/configuration.xml
The program 'xmllint' is currently not installed. To run 'xmllint' please ask your administrator to install the package 'libxml2-utils'
The command "xmllint --noout --schema phpunit.xsd tests/_files/configuration.xml" exited with 127.
0.17s$ xmllint --noout --schema phpunit.xsd tests/_files/configuration_empty.xml
The program 'xmllint' is currently not installed. To run 'xmllint' please ask your administrator to install the package 'libxml2-utils'
The command "xmllint --noout --schema phpunit.xsd tests/_files/configuration_empty.xml" exited with 127.
0.17s$ xmllint --noout --schema phpunit.xsd tests/_files/configuration_xinclude.xml -xinclude
The program 'xmllint' is currently not installed. To run 'xmllint' please ask your administrator to install the package 'libxml2-utils'
The command "xmllint --noout --schema phpunit.xsd tests/_files/configuration_xinclude.xml -xinclude" exited with 127.
```

issue people see on Travis here lately.

(ping @ArturGoldyn)

PR #2780, opened 2017-09-20
Add expectExceptionObject that takes \Exception

lorenzleutgeb:
When writing tests I commonly write:

```php
$this->expectException(MyException::class);
$this->expectExceptionMessage(MyException::whatWentWrong($that));
```

I use such static create methods to construct exceptions. This helps with keeping the exception message consistent, refactorable and autocompletable by my IDE. Very similar to `PHPUnit\Util\InvalidArgumentHelper::factory`.

Up to now I carried around a custom implementation:

```
protected function expectExceptionObject(\Exception $e) {
    $this->expectException(get_class($e));
    $this->expectExceptionMessage($e->getMessage());
    $this->expectExceptionCode($e->getCode());
}
```

So ... let's get it upstream :)

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -633,6 +633,22 @@ abstract class TestCase extends Assert implements Test, SelfDescribing
     }
 
     /**
+     * @param $exception \Exception
+     *
+     * @throws Exception
+     */
+    public function expectExceptionObject($exception)
+    {
+        if (!\is_a($exception, \Exception::class)) {
+            throw InvalidArgumentHelper::factory(1, 'exception');
+        }
+
+        $this->expectException(get_class($exception));
+        $this->expectExceptionMessage($exception->getMessage());
+        $this->expectExceptionCode($exception->getCode());
+    }
+
+    /**
      * @param bool $flag
      */
     public function setRegisterMockObjectsFromTestArgumentsRecursively($flag)


lorenzleutgeb:
Build on Travis CI should pass once #2777 is merged.

sebastianbergmann:
I like the idea but I do not think, though, that this new functionality should be added to the existing method.

What do you think about a new method named `expectExceptionObject()`, for instance?

lorenzleutgeb:
Yep, in my workaround I mentioned `expectExceptionObject` :+1: 

Pushed new changes, please give instructions on how I should add tests for the new method.

src/Framework/TestCase.php
```diff
@@ -633,6 +633,22 @@ public function expectExceptionMessageRegExp($messageRegExp)
     }
 
     /**
+     * @param $exception \Exception
```
sebastianbergmann: That should be `@param \Exception $exception`.

src/Framework/TestCase.php
```diff
@@ -633,6 +633,22 @@ public function expectExceptionMessageRegExp($messageRegExp)
...
+     *
+     * @throws Exception
+     */
+    public function expectExceptionObject($exception)
```
sebastianbergmann: That should be `public function expectExceptionObject(Exception $exception)`.

src/Framework/TestCase.php
```diff
@@ -633,6 +633,22 @@ public function expectExceptionMessageRegExp($messageRegExp)
...
+    {
+        if (!\is_a($exception, \Exception::class)) {
```
sebastianbergmann: This "manual type check" can be removed.

PR #2781, opened 2017-09-20
Array subset works with flat arrays;

### LINKED ISSUE #2069: assertArraySubset() bad documented or doesn't work as expected
Hi,

please consider 3 calls:
1) $this->assertArraySubset( ['a', 'b', 'c'], ['a', 'b', 'c', 'd'] );
2) $this->assertArraySubset( ['b', 'c'], ['a', 'b', 'c', 'd'] );
3) $this->assertArraySubset( ['a', 'c', 'b'], ['a', 'b', 'c', 'd'] );

1 - works
2 - doesn't work. Why? Isn't ['b','c'] a subset of ['a', 'b', 'c', 'd']?
3 - doesn't work as well. Elements must be in apriopriate order?

Should it work this way? 
Can information about this be added to docs? (3 little examples above should do)?
It's sort of confusing right now :)

askovron:
Add possibility to check flat arrays within assertArrayContains

Fixes #2069

### DIFF
--- a/src/Framework/Constraint/ArraySubset.php
+++ b/src/Framework/Constraint/ArraySubset.php
@@ -49,22 +49,42 @@ class PHPUnit_Framework_Constraint_ArraySubset extends PHPUnit_Framework_Constra
      */
     protected function matches($other)
     {
-        //type cast $other & $this->subset as an array to allow 
+        //type cast $other & $this->subset as an array to allow
         //support in standard array functions.
-        if($other instanceof ArrayAccess) {
+        if ($other instanceof \ArrayAccess) {
             $other = (array) $other;
         }
 
-        if($this->subset instanceof ArrayAccess) {
+        if ($this->subset instanceof \ArrayAccess) {
             $this->subset = (array) $this->subset;
         }
 
-        $patched = array_replace_recursive($other, $this->subset);
+        if (!$this->isArrayAssociative($this->subset)) {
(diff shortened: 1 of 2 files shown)


sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it because I do not understand the issue (see https://github.com/sebastianbergmann/phpunit/issues/2069#issuecomment-336618421) you're trying to solve here.

skovron:
@sebastianbergmann 

(1) The issue was (and is) that the `arraySubset` assertion passes only when the subset is **located in the very beginning** of the array.
```
Array = ['a', 'b', 'c']
Valid subset = ['a', 'b']
Invalid subset = ['b', 'c']
```

(2) The second issue is that the subset **must be in the same order** in both suubset and the original array.
```
Array = ['a', 'b', 'c']
Valid subset = ['a', 'b']
Invalid subset = ['b', 'a']
```

FYI. During tests had been running on this branch there were some errors not related to this feature.

BonBonSlick:
I am faced issues with auto indexed arrays today, what was solution?

  ```
  "phpunit/phpunit": "^7.1",

    $content  = \json_decode($response->getContent(), true);
    static::assertArraySubset(
      [
        [
          'email' => 'member@email.com',
```

Fails sometimes, because array with email has different int index.

Issue #2688, opened 2017-05-22
assertNotContains interferes with actual string


jpuck:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 6.1.4
| PHP version         | 7.0.15-0ubuntu0.16.04.4
| Installation Method | Composer

Using `assertNotContains` in a failing condition injects `not` into both the expected and actual strings.

Consider this class:

```php
class Foo
{
    public function bar()
    {
        return 'this is foobar';
    }
}
```

When executing this test:

```php
class FooTest extends TestCase
{
    public function test_interferes_with_string()
    {
        $expected = 'this is foobar';

        $actual = (new Foo)->bar();

        $this->assertNotContains($expected, $actual);
    }
}
```

Expected output:
> Failed asserting that 'this is foobar' does not contain "this is foobar".

Actual output:
> Failed asserting that 'this ***not*** is ***not*** foobar' does not contain "this ***not*** is ***not*** foobar".

[Demo Repository](https://github.com/MyTeamName/phpunit-string-interference)

evgpisarchik:
This happens because [assertNotContains ](https://github.com/sebastianbergmann/phpunit/blob/master/src/Framework/Assert.php#L239)invokes [LogicalNot](https://github.com/sebastianbergmann/phpunit/blob/master/src/Framework/Constraint/LogicalNot.php#L17)([StringContains](https://github.com/sebastianbergmann/phpunit/blob/master/src/Framework/Constraint/StringContains.php#L21)(..))
where [LogicalNot](https://github.com/sebastianbergmann/phpunit/blob/master/src/Framework/Constraint/LogicalNot.php#L17) replaces output of  ([StringContains](https://github.com/sebastianbergmann/phpunit/blob/master/src/Framework/Constraint/StringContains.php#L21)(..)) using [negate](https://github.com/sebastianbergmann/phpunit/blob/master/src/Framework/Constraint/LogicalNot.php#L43) method.

This problem happens also if  checked strings have other words from the list
```
[ 'contains ', 'exists', 'has ', 'is ', 'are ', 'matches ', 'starts with ', 'ends with ', 'reference ', 'not not ' ]
```

ryancastle:
Issues #1642 and #764 relate to this bug, but were closed in the 5.0 issue tidy up. I can confirm it's still present in 6.1.3. I doubt there will be any easy fixes that don't require a major refactoring of the way failure descriptions are built.

PR #2801, opened 2017-10-13
Solve: assertNotContains interferes with actual string #2688

fetmo:
Split string messages on regex pattern to just negate the necessary part.


sebastianbergmann:
Merged manually, thanks.

Issue #2675, opened 2017-05-09
PHP Warnings throw exception despite phpunit.xml settings and \PHPUnit\Framework\Error\Warning::$enabled = false;


vedadkajtaz:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 6.1.3 & 6.0
| PHP version         | 7.1.3
| Installation Method | PHAR

Hello,

I have a test case where a legitimate PHP warning raises an exception (and thus fails the test) despite setting `convertWarningsToExceptions="false"` in phpunit.xml and despite the `\PHPUnit\Framework\Error\Warning::$enabled = false;` instruction in test' `::setUp()` method.

Here's the phpunit.xml file:

```
<?xml version="1.0" encoding="UTF-8"?>
<phpunit
	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
	xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/4.5/phpunit.xsd"
	colors="true"
	processIsolation="true"
	convertErrorsToExceptions="true"
	convertNoticesToExceptions="false"
	convertWarningsToExceptions="false"
>
	<testsuites>
		<testsuite name="Bugreport test">
			<directory suffix=".php">/XXX/bug_report/tests</directory>
		</testsuite>
	</testsuites>
</phpunit>

```

The `/XXX/bug_report/tests` folder holds the single `test.php` file:

```
<?php /* test.php */ /* UTF8-Côôkie */
declare(strict_types=1);


class	testClass extends \PHPUnit\Framework\TestCase
{
	protected function	setUp() : void
	{
		// Neither this nor setting 'convertWarningsToExceptions="false"' seem to fix the problem...
		\PHPUnit\Framework\Error\Warning::$enabled = false;
	}

	function	test_php_warning()
	{
		if(($socket = socket_create(AF_UNIX, SOCK_STREAM, 0)) === false)
			throw new Exception('Failed creating socket');

		if(!socket_connect($socket, '/tmp/nosuch'))
		{
			// Some fallback mechanism...
		}
	}
}

```

The test fails with the following output:

> $ ../bin/phpunit-6.1.3.phar --configuration ./phpunit.xml
> PHPUnit 6.1.3 by Sebastian Bergmann and contributors.
> 
> E                                                                   1 / 1 (100%)
> 
> Time: 559 ms, Memory: 6.00MB
> 
> There was 1 error:
> 
> 1) testClass::test_php_warning
[...]

PR #2806, opened 2017-10-14
#2675 PHP Warnings throw exception despite phpunit.xml settings

wesolowski:
[issues-2675](https://github.com/sebastianbergmann/phpunit/issues/2675)
I have start fix that. Now `$enabled`-Variables from class: `PHPUnit\Framework\Error\Deprecated`, `PHPUnit\Framework\Error\Notice` and `PHPUnit\Framework\Error\Warning` are set in `src/Util/PHP/Template/TestCaseMethod.tpl.dist`. 

I do believe there is another bug, the error is still there


sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it because I do not understand your changes and don't believe they solve the problem.

Issue #2802, opened 2017-10-13
assertRessourceType()


jonathanparentlevesque:
By example, I want to make sure that my method (in a trait) returns an FTP resource.

```
public function testGetFTPConnection()
	{
		$errorMessage = "";
		
		$ftp = $this->getFTPConnection($errorMessage);
		
		$this->assertEmpty($errorMessage);
		
		$this->assertTrue(is_resource($ftp));
		
		$this->assertEquals(
			"FTP Buffer",
			get_resource_type($ftp)
		);
		
		ftp_close($ftp);
	}
```

Personally, I would use 2 native functions for the validation:

- `is_resource()` : http://php.net/manual/fr/function.is-resource.php
-  `get_resource_type()` : http://php.net/manual/fr/function.get-resource-type.php

Thank you

sebastianbergmann:
I agree with what @sebastianfeldmann wrote in https://github.com/sebastianbergmann/phpunit/pull/2808#issuecomment-419648431.

PR #2808, opened 2017-10-15
Added assertResourceType assertion

mateuszsip:
Implements #2802.


src/Framework/Assert.php
```diff
@@ -2269,6 +2269,12 @@ public static function assertJsonFileNotEqualsJsonFile($expectedFile, $actualFil
         static::assertThat($actualJson, new LogicalNot($constraintExpected), $message);
     }
 
+    public static function assertResourceType($expectedType, $actual, $message = '')
+    {
+        static::assertThat(is_resource($actual), new IsTrue(), $message);
```
sebastianbergmann: It would be better to use `assertInternalType()` here.

sebastianbergmann: Actually, it would be better to implement a proper `Constraint` object.

kejwmen: Thanks for feedback, `IsResource` constraint implemented.

keradus:
what about not adding new assertion, but change existing one?
eg `assertInternalType('resource/stream', $foo)`

src/Framework/Constraint/IsResource.php
```diff
@@ -0,0 +1,41 @@
...
+ * file that was distributed with this source code.
+ */
+namespace PHPUnit\Framework\Constraint;
+
+/**
+ * Constraint that checks if value is resource
+ *
+ * The file path to check is passed as $other in evaluate().
+ */
+class IsResource extends Constraint
```
sebastianbergmann: `IsResource` is not needed as a `Constraint` object. `IsType('resource')` already exists. What I meant in https://github.com/sebastianbergmann/phpunit/pull/2808/commits/72aae12e7326ef0ae86bc821bfb9f5ce9a47832b#r144721495 was a `Constraint` object that checks a specific resource type.

sebastianbergmann:
@keradus That would also work. However, I think I prefer a specific, new assertion method for this.

keradus:
Then there is an issue that it won't be clear that one shall:
```php
$this->assertInternalType('resource', $foo);
$this->assertResourceType('stream', $foo);
```

or just:
```php
$this->assertResourceType('stream', $foo);
```

And for that, you will also have different console output for that.
I don't like confusing users ;)

sebastianbergmann:
`assertResourceType()`could, and IMO should, internally call `assertInternalType()` first to ensure that it's a resource.

sebastianbergmann:
Wondering if/how this needs to take `gettype() will now return "resource (closed)" instead of "unknown type" for closed resources` (PHP 7.2) into account.

keradus:
> `assertResourceType()`could, and IMO should, internally call assertInternalType() first to ensure that it's a resource.

but that's the thing, it's implementation detail. if so, it shall be described that assertion is checking not only if type is matching, but also if it's a resource in first place

src/Framework/Constraint/IsResource.php
kejwmen: Now I understand, thanks for review.

kejwmen:
> but that's the thing, it's implementation detail. if so, it shall be described that assertion is checking not only if type is matching, but also if it's a resource in first place

@keradus, I renamed assertion to `assertResourceWithType`, what do you think about it?

tests/Framework/AssertTest.php
```diff
@@ -3177,6 +3177,43 @@ public function testAssertStringNotMatchesFormatFile()
         $this->assertStringNotMatchesFormatFile($this->filesDirectory . 'expectedFileFormat.txt', "FOO\n");
     }
 
+    public function testAssertResourceWithTypeForMemoryStream()
+    {
+        // GIVEN
```
keradus: I didn't saw this approach of setting up tests in this codebase,
and test methods are extra small, I don't find any benefit of those comments tbh

src/Framework/Constraint/IsResourceType.php
```diff
@@ -0,0 +1,56 @@
...
+ * The file path to check is passed as $other in evaluate().
+ */
+class IsResourceType extends Constraint
+{
+    /**
+     * @var string
+     */
+    private $type;
+
+    public function __construct($type)
```
keradus: i know it's picky, but I could pass non-string here as well (eg, file descriptor ;P)
then, the behaviour could be funny ;)

tests/Framework/AssertTest.php
kejwmen: Sorry, it became my habit.
I agree, comments removed.

src/Framework/Constraint/IsResourceType.php
kejwmen: You are totally right.
Fixed.

kejwmen:
Tests failure looks unrelated.

sebastianbergmann:
@kejwmen They are, see https://github.com/sebastianbergmann/phpunit/issues/2826. Hopefully I'll have some time soon to look into this.

src/Framework/Constraint/IsResourceType.php
```diff
@@ -0,0 +1,68 @@
...
+ */
+class IsResourceType extends Constraint
+{
+    /**
+     * @var string
+     */
+    private $type;
+
+    /**
+     * IsResourceType constructor.
```
keradus: this comment provides no value

src/Framework/Constraint/IsResourceType.php
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    /**
+     * Evaluates the constraint for parameter $other. Returns true if the
+     * constraint is met, false otherwise.
+     *
+     * @param mixed $other Value or object to evaluate.
+     *
+     * @return bool
+     */
```
keradus: you could replace whole docblock here with inheritDoc from parent class

src/Framework/Constraint/IsResourceType.php
kejwmen: You are right. Removed.

src/Framework/Constraint/IsResourceType.php
kejwmen: Fixed using syntax from [PSR-5 draft](https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc.md#61-making-inheritance-explicit-using-the-inheritdoc-tag).

kejwmen:
Looks like tests are ok, feedback implemented (thanks @keradus). 
Is it mergable?

src/Framework/Assert.php
```diff
@@ -2269,6 +2270,12 @@ public static function assertJsonFileNotEqualsJsonFile($expectedFile, $actualFil
         static::assertThat($actualJson, new LogicalNot($constraintExpected), $message);
     }
 
+    public static function assertResourceWithType($expectedType, $actual, $message = '')
```
keradus:
```diff
-assertResourceWithType
+assertResourceOfType
```

src/Framework/Constraint/IsResourceType.php
```diff
@@ -0,0 +1,60 @@
...
+            throw InvalidArgumentHelper::factory(1, 'string');
+        }
+
+        $this->type = $type;
+    }
+
+    /**
+     * @inheritdoc
+     */
+    protected function matches($other)
```
keradus: actually, my concern is still valid.
once, it was for assertion method, but assertion is only proxy for constraints. Or is it not?
I can use this newly created Constraint for non-resource. Assertion has dedicated support for it, yet constraint doesn't. I believe it shall be moved to constraint.

kejwmen: Thanks again.

> I can use this newly created Constraint for non-resource.

I got the whole point now, you are right.

> Assertion has dedicated support for it, yet constraint doesn't.

right, I missed that types were introduced, great job!

> I believe it shall be moved to constraint.

Could you confirm that:
If constraint does something with specific data type, type must be validated inside.
So `IsResourceType` should be `IsResourceOfType` that uses `IsType` internally?

tests/Framework/AssertTest.php
```diff
@@ -2568,6 +2568,35 @@ public function testAssertStringNotMatchesFormatFile(): void
         $this->assertStringNotMatchesFormatFile($this->filesDirectory . 'expectedFileFormat.txt', "FOO\n");
     }
 
+    public function testAssertResourceWithTypeForMemoryStream()
```
keradus:
```diff
-assertResourceWithType
+assertResourceOfType
```

tests/Framework/AssertTest.php
```diff
@@ -2568,6 +2568,35 @@ public function testAssertStringNotMatchesFormatFile(): void
...
+    {
+        $resource = fopen("php://memory", 'rb');
+
+        $this->assertResourceOfType('stream', $resource);
+
+        fclose($resource);
+    }
+
+    public function testAssertResourceWithTypeThrowForNotAResource()
```
keradus:
```diff
-assertResourceWithType
+assertResourceOfType
```

tests/Framework/AssertTest.php
```diff
@@ -2568,6 +2568,35 @@ public function testAssertStringNotMatchesFormatFile(): void
...
+    {
+        $notAResource = 'test';
+
+        $this->expectException(AssertionFailedError::class);
+
+        $this->assertResourceOfType('stream', $notAResource);
+    }
+
+    public function testAssertResourceWithTypeThrowWhenExpectingOtherType()
```
keradus:
```diff
-assertResourceWithType
+assertResourceOfType
```

src/Framework/Constraint/IsResourceType.php
keradus: > If constraint does something with specific data type, type must be validated inside.

indeed

> So IsResourceType should be IsResourceOfType that uses IsType internally?

commented separatelly

src/Framework/Constraint/IsResourceOfType.php
```diff
@@ -9,48 +9,47 @@
...
 
-use PHPUnit\Util\InvalidArgumentHelper;
-
 /**
- * Constraint that checks if value is resource type
- *
- * The file path to check is passed as $other in evaluate().
+ * Constraint that checks if value is a resource of given type
  */
-class IsResourceType extends Constraint
```
keradus: removal of `IsResourceType` is BC breaker that will break a lot of custom Constraints.
hint: leave `IsResourceType` as is and introduce `IsResourceOfType` as a new class

src/Framework/Constraint/IsResourceOfType.php
```diff
@@ -9,48 +9,47 @@
...
+class IsResourceOfType extends Constraint
```
keradus: removal of `IsResourceType` is BC breaker that will break a lot of custom Constraints.
hint: leave `IsResourceType` as is and introduce `IsResourceOfType` as a new class

keradus: oh, you already dont it, cool ;)

kejwmen:
and thanks again @keradus ;)

keradus (approved):
:+1: looks really nice

src/Framework/Assert.php
```diff
@@ -2524,6 +2525,20 @@ public static function assertJsonFileNotEqualsJsonFile(string $expectedFile, str
...
+    /**
+     * Asserts that given value is resource of specified type
+     *
+     * @param string $expectedType
+     * @param resource $actual
+     * @param string $message
+     *
+     * @throws \Exception
+     */
+    public static function assertResourceOfType(string $expectedType, $actual, string $message = '')
```
sebastianbergmann: `assertResourceNotOfType()` is missing

keradus: maybe there shall be some kind of ~meta test that check if all assertions have theirs negations ?

kejwmen: wasn't aware of that, added.

src/Framework/Assert.php
```diff
@@ -2524,6 +2526,34 @@ public static function assertJsonFileNotEqualsJsonFile(string $expectedFile, str
...
+     *
+     * @param string $expectedType
+     * @param resource $actual
+     * @param string $message
+     *
+     * @throws \Exception
+     */
+    public static function assertResourceNotOfType(string $expectedType, $actual, string $message = '')
+    {
+        static::assertThat($actual, new IsResourceNotOfType($expectedType), $message);
```
keradus: this constraint is highly duplicating positive one.
take a note that there is no negative constraint so far.
drop that extra constraint and instead of it, use `LogicalNot`

keradus (changes requested):

src/Framework/Assert.php
kejwmen: I know, but `IsResourceOfType` uses `IsType` internally.
Because of that I cannot use `LogicalNot` here because in both cases when `$actual` is not a resource match should fail.

Is it ok or should I try with different design?

sebastianbergmann: I do not think that it is a good idea to use `IsType` in `IsResourceOfType`.

sebastianbergmann:
This branch has conflicts that must be resolved

sebastianbergmann:
What is the status of this?

src/Framework/Assert.php
kejwmen: I can stop using `IsType` inside `IsResourceOfType`, but `IsResourceNotOfType` is not an exact opposite of `IsResourceOfType`. `IsNotResourceOfType` would be.

`IsResourceOfType` means that value is a resource of specific type.
`IsResourceNotOfType` means that value is a resource not of specific type.

Both still must be a resource.

`IsNotResourceOfType` sounds like an exact opposite, so it's either not a resource or a resource of different type.

That's why I introduced `IsResourceNotOfType` as a separate constraint.

Should we have 2 separate constraints:
- `IsResourceOfType`
- `IsResourceNotOfType`

and 3 assertions:
- `assertResourceOfType`
- `assertResourceNotOfType`
- `assertNotResourceOfType`

Or having only `IsResourceOfType` with 2 assertions: `assertResourceOfType` and `assertNotResourceOfType` is sufficent? 

Imo `IsResourceNotOfType` is useful.

sebastianbergmann:
Closed in favour of #3298.

sebastianbergmann:
Closed this in favour of #3298 was a bad idea.

sebastianfeldmann:
After fixing the merge commits of this pull request and reviewing the code I think
```php
$this->assertResourceType('stream', $handle);
```
is not much of an enhancement to
```php
$this->assertEquals('stream', get_resource_type($handle));
```
Maybe we should close this pull request and its regarding issue #2802

sebastianbergmann:
I must say that I agree with @sebastianfeldmann.

PR #2821, opened 2017-10-21
Enhancement: Enable ordered_imports fixer

localheinz:
This PR

* [x] enables the `ordered_imports` fixer
* [x] runs `php-cs-fixer fix`

💁‍♂️ For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/tree/v2.7.1#usage:

>**ordered_imports**
>
>Ordering use statements.
>
>Configuration options:
>
>* `importsOrder` (`array`, `null`): defines the order of import types; defaults to `null`
>* `sortAlgorithm` (`'alpha'`, `'length'`): whether the statements should be sorted alphabetically or by length; defaults to `'alpha'`


tests/Util/PHP/AbstractPhpProcessTest.php
```diff
@@ -59,7 +59,7 @@ public function testShouldUseGivenSettingsToCreateCommand()
             'display_errors=1',
         ];
 
-        $expectedCommandFormat  = "%s -d %callow_url_fopen=1%c -d %cauto_append_file=%c -d %cdisplay_errors=1%c";
+        $expectedCommandFormat  = '%s -d %callow_url_fopen=1%c -d %cauto_append_file=%c -d %cdisplay_errors=1%c';
```
localheinz: This change is unrelated to enabling the `ordered_imports` fixer, but follows [`.php_cs#L67`](https://github.com/sebastianbergmann/phpunit/blob/92cba1d0bce5fd91ef74a49c31a8cbf1a42a79ef/.php_cs#L67).

localheinz:

PR #2847, opened 2017-11-09
Add required attrib in config generator template

shoma:
minor bug fix.

Add required attribute `name` at `testsuite` element in configuration generator template.
This is required by xsd.
https://github.com/sebastianbergmann/phpunit/blob/514f081cd66f09aafe8f00d20ac4f91f356690db/phpunit.xsd#L267


sebastianbergmann:
Merged manually, thanks.

Issue #2859, opened 2017-11-14
"PHPUnit_Framework_TestSuite_DataProvider" does not extend PHPUnit_Framework_TestCase


eddmann:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 5.7.24
| PHP version         | 7.1.11
| Installation Method | Composer

Have just noticed this exception being thrown when trying to run a test suite with a `dataProvider` included.

```
Fatal error: Uncaught PHPUnit_Framework_Exception: Class "PHPUnit_Framework_TestSuite_DataProvider" does not extend PHPUnit_Framework_TestCase.
```

Looking at the diff between this version and the previous it seems to be due to the caching of the declared classes. Setting the `declaredClasses` before attempting to load the file (similar to how it worked before) seems to solve the issue.

```php
// TestSuite.php
$this->declaredClasses = get_declared_classes();
$filename   = PHPUnit_Util_Fileloader::checkAndLoad($filename);
$newClasses = array_diff(get_declared_classes(), $this->declaredClasses);
```

sebastianbergmann:
CC @kubawerlos

sebastianbergmann:
Thank you for your report.

Please provide a **minimal**, **self-contained**, reproducing test case that shows the problem you are reporting.

Without such a minimal, self-contained, reproducing test case I will not be able to investigate this issue.

eddmann:
Interestingly it only seems to occur when I include `./tests/*/` within the `phpunit.xml.dist` file

```
<testsuite name="Test Suite">
  <directory>./tests/</directory>
  <directory>./tests/*/</directory>
</testsuite>
```

To be honest, as providing the first directory recurses through child directories, removing the second directory definiton should not be a problem.

moorscode:
When running coverage reports this issue occurs on our builds as well.

`phpunit -c phpunit.xml --coverage-clover build/logs/clover.xml`

```
<phpunit
	backupGlobals="false"
	backupStaticAttributes="false"
	bootstrap="tests/bootstrap.php"
	colors="true"
	convertErrorsToExceptions="true"
	convertNoticesToExceptions="true"
	convertWarningsToExceptions="true"
	processIsolation="false"
	stopOnError="false"
	stopOnFailure="false"
	stopOnIncomplete="false"
	stopOnSkipped="false"
	syntaxCheck="false"
	verbose="true"
	>
	<testsuites>
		<testsuite>
[...]

Issue #2833, opened 2017-10-29
Test class loaded during data provider execution is not discovered


kubawerlos:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 6.4.3
| PHP version         | 7.1.11
| Installation Method | Composer

Having 2 test cases:
```php
<?php

class ATest extends \PHPUnit\Framework\TestCase
{
    /**
     * @dataProvider provide
     */
    public function testNothingA($x)
    {
        $this->assertTrue($x);
    }

    public function provide() : array
    {
        $a = new \ReflectionClass(BTest::class);

        return [[true]];
    }
}
```
and
```php
<?php

class BTest extends \PHPUnit\Framework\TestCase
{
    public function testNothingB()
    {
        $this->assertTrue(false);
    }
}
```
running them will result with executing only `ATest`.

It's not even listed:
```bash
vendor/bin/phpunit --list-tests
PHPUnit 6.4.3 by Sebastian Bergmann and contributors.

Available test(s):
 - Tests\ATest::testNothingA#0
```

After removing `$a = new \ReflectionClass(BTest::class);` or moving this line to `testNothingA` both classes will be executed. The problem is only with data provider.

sebastianbergmann:
#2834 was reverted due to #2859

kubawerlos:
So I use my test case from the first solution plus the test from https://github.com/sebastianbergmann/phpunit/issues/2859#issuecomment-344278138 and try again, right?

sebastianbergmann:
Yes, please.

keradus:
at PHP CS Fixer, we come across this issue as well. it will be great to addition to have this solved!
would be really great to have it solved :+1:
big :+1: for one actually working on it, we are with you @kubawerlos !

PR #2861, opened 2017-11-14
Fix for declaring test case in data provider

kubawerlos:
For https://github.com/sebastianbergmann/phpunit/issues/2833.

I cannot reproduce the problem from https://github.com/sebastianbergmann/phpunit/issues/2859 - as I suspected the minimal, self-contained, reproducing test case from https://github.com/sebastianbergmann/phpunit/issues/2859#issuecomment-344278138 is not triggering the error. _Obviously I have tested my original solution in project with over dozen of data providers and all went fine_.

@sebastianbergmann can you create a branch for the issue and merge this PR on it so @eddmann, @moorscode or @johnbillion will provide a test case for the problem?


tests/_files/DataProviderDeclaringClass/SecondTest.php
```diff
@@ -0,0 +1,15 @@
...
+
+namespace Foo\DataProviderDeclaringClass;
+
+use PHPUnit\Framework\TestCase;
+
+class SecondTest extends TestCase
+{
+    const DUMMY = 'dummy';
+
+    public function testSecond()
```
gavinlove: Change this to be a data provider and it is likely you will see the problem.

kubawerlos:
@gavinlove like this https://github.com/sebastianbergmann/phpunit/pull/2861/commits/dcace3b5b8b71ff559c2ac74f4d4cbfc6478847a ? Of course I tried something like this yesterday, give me some credit - I tried to follow instructions from the https://github.com/sebastianbergmann/phpunit/issues/2859 - none of them work. More, few minutes ago I have run PHPUnit 5.7.24 on master of [PHP-CS-Fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer) which have 495 data providers an (with PHP 7.1.11-1+ubuntu16.04.1+deb.sury.org+1) - all went fine.

TBH it feels little bit irritated - I have discovered the bug - this one that test class loaded during data provider execution is not discovered - provided decent [test case](https://github.com/sebastianbergmann/phpunit/commit/7e8e2625c289a63285625dcdd657b824797a3717) to cover the problem and then made a fix. I have tried the fix with multiple projects I have on my PC and had no error. I'd say I proceeded with the bug quite well.

Then https://github.com/sebastianbergmann/phpunit/issues/2859 was reported with non-working reproductions steps (few snippets, none of them leading to the error). Or maybe I'm wrong here, @sebastianbergmann did you manage to reproduce it, so we can create PHPT to cover it? If not I'd like to fix for this issue be accepted and merged and https://github.com/sebastianbergmann/phpunit/issues/2859 be fixed not by reverting fix of the other bug, but in proper way - PHPT case and then fix (which BTW I'd be happy to implement).

gavinlove:
@kubawerlos Your right the two providers is a not the right way to reproduce it. I looked at @EddMann said I was able to reproduce this.

I used your `FirstTest` class but I then changed my `phpunit.xml.dist` to the following

```

<phpunit bootstrap="vendor/autoload.php">
    <testsuites>
        <testsuite name="Test Suite">
            <directory>./tests/</directory>
            <directory>./tests/*/</directory>
        </testsuite>
    </testsuites>
</phpunit>
```

Note that the same directory is listed twice here which  @eddmann mentioned yesterday. I suspect what is happening is that the directory is parsed twice and then later on phpunit filters it down to a single test.

johnbillion:
Please don't be irritated! We're simply reporting what we're seeing.

AFAICT, the minimal test case I reported in #2859 is enough to reproduce the issue, but there must be something else which causes this, but I don't know what.

I don't think the issue is dependent on PHP version. The issue appears for me when running PHP 7.1 and 7.2.

If it helps, here's a Travis CI job that failed due to this fatal error: https://travis-ci.org/johnbillion/query-monitor/jobs/301942375 .

gavinlove:
@kubawerlos @johnbillion I have pushed up a repo which reproduces it for me, which will hopefully help.

https://github.com/gavinlove/phpunit-bug-2859

kubawerlos:
@gavinlove thanks, your first comment gave me a hint and I just uploaded test case for the problem. Fix in progress.

kubawerlos:
@sebastianbergmann fix might looks as really dummy, but in fact `PHPUnit_Framework_TestSuite_DataProvider` is the only PHPUnit class that implements `PHPUnit_Framework_Test` and would wrongly go into the calling of `addTestSuite` 16 lines below the added `if`.

src/Framework/TestSuite.php
```diff
@@ -87,6 +87,11 @@ class PHPUnit_Framework_TestSuite implements PHPUnit_Framework_Test, PHPUnit_Fra
      */
     private $iteratorFilter = null;
 
+    /**
+     * @var array
```
keradus: array of ? strings? then `string[]`

PR #2862, opened 2017-11-16
Show valid exception message in TeamCity logger in case of ExceptionWrapper

wbars:
Hello! This change is fixing [WI-37135](https://youtrack.jetbrains.com/issue/WI-37135) for PHPUnit 5.0+ since your version of TeamCity printer is used for these versions. We fixed this for our printer already (e.g. for PHPUnit below 5.0) by similar change.


sebastianbergmann:
Thanks! Can you please also send a pull request for the `5.7` branch?

wbars:
Sure!

PR #2866, opened 2017-11-17
Expose TestCase::$dataName

Wes1262:
It would be useful to have this property accessible. I use it to set the breakpoint in PHPStorm:

expla

If this is already possible I am sorry but I couldn't find it.

Thanks.

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -1951,7 +1951,7 @@ abstract class TestCase extends Assert implements Test, SelfDescribing
     /**
      * @return int|string
      */
-    public function dataID()
+    public function dataName()
     {
         return $this->dataName;
     }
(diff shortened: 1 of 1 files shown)


src/Framework/TestCase.php
```diff
@@ -1948,6 +1948,14 @@ public function dataDescription()
         return \is_string($this->dataName) ? $this->dataName : '';
     }
 
+    /**
+     * @return int|string
+     */
+    public function dataID()
```
sebastianbergmann: I am not convinced (yet) that exposing this information is required. In any case, the method should be named `dataName()`.

Netmosfera: I had the impression that "name" would be string only, so I used ID for int|string. I'm ok with any name though.

Issue #2867, opened 2017-11-17
test names show numeric datasource only for iterators as dataprovider


SimonHeimberg:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 6.4.3 / 6.4.4
| PHP version         | 7.y.z
| Installation Method | Composer / PHAR

unaffected PHPUnit version: 6.2.3

bad since revision 71d9a48f7d4a70643d (by bisect), because of fixing #2693 

test script:
```php
<?php

namespace Tests\ExampleTest;

class DatasetTest extends \PHPUnit\Framework\TestCase
{
    /**
     * @dataProvider arrayProvider
     */
    public function testArrayProvider()
    {
        // fine, shows key as datasource
    }

    /**
     * @dataProvider iteratorProvider
     */
    public function testIteratorProvider()
    {
        // wrong, shows numeric datasource
    }

    public static function arrayProvider()
    {
        return array(
            'xyx' => array(11,111),
        );
    }

    public static function iteratorProvider()
    {
        yield 'bla' => array(1, 11);
    }
}
```
phpunit --list-tests
expected: ..., DatasetTest::testIteratorProvider with data set "bla"
result: ..., DatasetTest::testIteratorProvider with data set #0

sebastianbergmann:
Fixed by https://github.com/sebastianbergmann/phpunit/pull/2868.

PR #2868, opened 2017-11-17
string data providers keep their keys #2867

SimonHeimberg:


src/Util/Test.php
```diff
@@ -524,7 +524,15 @@ private static function getDataFromDataProviderAnnotation($docComment, $classNam
                 }
 
                 if ($data instanceof Traversable) {
-                    $data = \iterator_to_array($data, false);
+                    $origData = $data;
```
SimonHeimberg: or should $result be written directly instead of copying data?

sebastianbergmann:
Merged manually, thanks.

PR #2879, opened 2017-11-28
Avoid check name in foreach

bscheshirwork:
`unset` is better IMHO
for  this case


sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it as I do not understand what you are trying to achieve here.

bscheshirwork:
1 operation unset() vs count($newVariableNames) operations if(!=)

nothing matter, yep

PR #2886, opened 2017-11-30
Forced environment variables do not affect getenv()

ralphschindler:
We've noticed that when forcing environment variables from a phpunit.xml file, it will set `$_ENV` appropriately, but will fail to reflect the same change to environment variables that existed in the calling environment when attempting to retrieve them through `getenv()`.

All existing tests pass.


sebastianbergmann:
Merged manually, thanks.

PR #2895, opened 2017-12-04
Update TestTest.php

rquadling:
I'm glad to see I wasn't the only person that had misspelled 'Annotation'.


sebastianbergmann:
Merged manually, thanks.

Issue #2875, opened 2017-11-24
skip test with annotation


mathroc:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 5.7.23
| PHP version         | 5.6.31
| Installation Method | Composer

It would be nice to be able to skip test with an annotation, eg `@skipTest "needs bla bla bla"`
The main advantage with this is that it could skip the `setUp`/`tearDown` too and make the test suite a bit faster

sebastianbergmann:
I do not think that this would be useful.

rpintea:
Hey Sebastian, why'd you put the enhancement in the first place if you didn't think it would be useful you senile old cook?

ob-ivan:
Also, this would allow to skip data provider too.

In my current project I have a piece of massively tested new functionality. The data provider for a single test method returns some hundreds of data sets. The functionality is already merged into master, but the client asked to disable it temporarily for business-wise reasons. If I use `markTestSkipped` inside the test method, I get 373 "test is skipped" messsages, each taking at least 5 lines in terminal, pushing any real error reports out from terminal buffer.

What I want is to be able to skip a test without all the useless verboseness. I tried a workaround prepending test method name with `disabled_` prefix, then it doesn't get executed, but instead I have "No tests found in class" warning, because the test case contains only one test method.

Adding a "disabled" group to it doesn't work for me too, because the test case already has "functional" group assigned, and when I run `phpunit --group=functional` it doesn't exclude "disabled" group.

@sebastianbergmann please give it a second thought.

adrian-enspired:
this - using `markTestSkipped` doesn't actually skip the test. it still invokes the test, it still runs the dataproviders, and if you disable those, you get errors because the test is invoked without args.

epdenouden:
@adrian-enspired thank you for mentioning this, the data providers for skipped tests are on my list for #3736. 

`->markTestSkipped()` follows a different code path and is used for when a developer _wants_ to mark a test skipped from inside the runtime context. The annotation should never invoke the test and its associated fixtures.

epdenouden:
Oops. I just noticed the previous discussion upthread. @sebastianbergmann I will add a @skipTest annotation experimental to the data provider project as a seperate pull request, see if can be useful and clean.

PR #2896, opened 2017-12-05
Skip test from annotation

gregurco:
Issue: #2875
Description: allow to skip test from class or method annotation
Example 1:
```php
class TestSkippedFromAnnotation extends TestCase
{
    /**
     * @testSkipped
     */
    public function testSkipped()
    {
        // test will be skipped
    }

    /**
     * @testSkipped some smart comment
     */
    public function testSkippedWithComment()
    {
        // test will be skipped
    }
}
```

Example 2:
```php
/**
 * @testSkipped
 */
class TestSkippedFromClassAnnotation extends TestCase
{
    public function testSkipped()
    {
        // test will be skipped
    }
}
```
Note 2: I named it `@testSkipped` and not `@skipTest` as was proposed to be more consistent with method `markTestSkipped()`

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -673,6 +673,21 @@ abstract class TestCase extends Assert implements Test, SelfDescribing
         }
     }
 
+    protected function setMarkTestSkippedFromAnnotation()
+    {
+        try {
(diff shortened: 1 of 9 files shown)


sebastianbergmann:

Issue #2907, opened 2017-12-07
StringMatchesFormatDescription Does Not Handle Escaped % Properly


mkasberg:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 6.5.3
| PHP version         | 7.0.22-0ubuntu0.17.04.1
| Installation Method | Git

I would expect the following test case to pass because the format string `'%%'` should produce a literal `'%'` ([sprintf](https://secure.php.net/manual/en/function.sprintf.php)). But it fails in PHPUnit 6.5.3.

**StringMatchesFormatDescriptionTest.php**
```php
<?php

use PHPUnit\Framework\TestCase;

class StringMatchesFormatDescriptionTest extends TestCase {

    public function testEscapedPercent() {
        $this->assertStringMatchesFormat('This %%d is escaped.', 'This %d is escaped.');
    }

}
```

**Sample Output:**
```
$ ./phpunit StringMatchesFormatDescriptionTest.php 
PHPUnit 6.5.3 by Sebastian Bergmann and contributors.

Runtime:       PHP 7.0.22-0ubuntu0.17.04.1 with Xdebug 2.5.0
Configuration: /home/mkasberg/repositories/phpunit/phpunit.xml

F                                                                   1 / 1 (100%)

Time: 436 ms, Memory: 4.00MB

There was 1 failure:

1) StringMatchesFormatDescriptionTest::testEscapedPercent
Failed asserting that string matches format description.
--- Expected
+++ Actual
@@ @@
-This %%d is escaped.
+This %d is escaped.

/home/mkasberg/repositories/phpunit/src/Framework/Constraint/Constraint.php:117
/home/mkasberg/repositories/phpunit/src/Framework/Constraint/Constraint.php:62
/home/mkasberg/repositories/phpunit/src/Framework/Assert.php:2116
/home/mkasberg/repositories/phpunit/src/Framework/Assert.php:1768
/home/mkasberg/repositories/phpunit/StringMatchesFormatDescriptionTest.php:8
/home/mkasberg/repositories/phpunit/src/Framework/TestCase.php:1071
/home/mkasberg/repositories/phpunit/src/Framework/TestCase.php:939
/home/mkasberg/repositories/phpunit/src/Framework/TestResult.php:698
/home/mkasberg/repositories/phpunit/src/Framework/TestCase.php:894
/home/mkasberg/repositories/phpunit/src/Framework/TestSuite.php:755
[...]

PR #2913, opened 2017-12-10
Add Support for Escaped % in StringMatchesFormatDescription

mkasberg:
Add the string format specifier '%%' to be interpreted as a literal '%' character. For example,

            $this->assertStringMatchesFormat('This %%d is escaped.', 'This %d is escaped.');

This addresses #2907. Also, as part of the work for this, I created a new test class for the StringMatchesFormatDescription class.


mkasberg:
This is one of my first pull requests to PHPUnit, so I'm open to criticism or suggestions. Thanks for all your work!

mkasberg:
Seems like I've introduced a bug when the directory separator is a backslash. I'll need to address that before this is ready.

PR #2914, opened 2017-12-10
Allow Escaped % in Format String Matches

mkasberg:
When using the StringMatchesFormatDescription constraint, define the '%%' format specifier to be interpreted as a literal '%', which allows for things like the following:

    $this->assertStringMatchesFormat('This %%d is escaped.', 'This %d is escaped.');

This addresses #2907.


mkasberg:
I'm certainly open to thoughts, criticism, or suggestions :smile:

PR #2917, opened 2017-12-13
Add assertIsCountable assertion

carusogabriel:
With this PR, I want to propose a new assertion: `assertIsCountable`.

### Motivation
In `PHP 7.2`, if the `count` function tries to count something that isn't an `array` or implements the `Countable` interface, it throws a warning. We _could_ check with `PHP` if the variable is countable if there were an `is_countable` function. So, makes perfect sense to test it with an `assertIsCountable` assertion.


sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it as I do believe that the existing `assertInstanceOf()` and `assertInternalType()` assertions are sufficient in case one really needs to assert that a variable contains something that is `count()`able.

PR #2920, opened 2017-12-14
Replace CLI testdox printer with rpkamp/fancy-testdox-printer

rpkamp:
Here is my initial implementation of replacing the current testdox printer with mine (rpkamp/fancy-testdox-printer).

A few remarks:

- My printer is based on the `PHPUnit\TextUI\ResultPrinter` printer, and getting it to work with the `PHPUnit\Util\TestDox\TextResultPrinter` would be near impossible without a big overhaul of how printers are currently set up. So I put mine in the `PHPUnit\TextUI` namespace and load it as default when `--testdox` is provided on the command line. This does mean that:
  - People who have `PHPUnit\Util\TestDox\TextResultPrinter` set as their printer class will _not_ see the new testdox printer after upgrading, but that may be fine.
  - Printing to file/html/xml remains unchanged
- There is one failing test, the test with `Balance cannot become negative` and `Balance cannot become negative 2` which should print only one line. I've tried several different solutions, but somehow I can't seem to get it to fit in the current design. Is this behaviour something you want to keep, or is it fine if it prints as two lines instead of one?

I do like the `\PHPUnit\TextUI\TestDoxPrinterTest` class, which is a class that tests a printer, with itself as input, to test that PHPUnit does it's work correctly, using PHPUnit. Very meta :smiley:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ before_script:
   - echo 'assert.exception=On' >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
 
 script:
-  - ./phpunit --coverage-clover=coverage.xml
+  - ./phpunit --coverage-clover=coverage.xml --testdox --color=auto
   - ./phpunit --configuration ./build/travis-ci-fail.xml > /dev/null; if [ $? -eq 0 ]; then echo "SHOULD FAIL"; false; else echo "fail checked"; fi;
   - xmllint --noout --schema phpunit.xsd phpunit.xml
   - xmllint --noout --schema phpunit.xsd tests/_files/configuration.xml
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ before_script:
   - echo 'assert.exception=On' >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
 
 script:
-  - ./phpunit --coverage-clover=coverage.xml --testdox --color=auto
+  - ./phpunit --coverage-clover=coverage.xml
   - ./phpunit --configuration ./build/travis-ci-fail.xml > /dev/null; if [ $? -eq 0 ]; then echo "SHOULD FAIL"; false; else echo "fail checked"; fi;
   - xmllint --noout --schema phpunit.xsd phpunit.xml
   - xmllint --noout --schema phpunit.xsd tests/_files/configuration.xml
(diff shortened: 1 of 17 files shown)


.travis.yml
```diff
@@ -32,7 +32,7 @@ before_script:
   - echo 'assert.exception=On' >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
 
 script:
-  - ./phpunit --coverage-clover=coverage.xml
+  - ./phpunit --coverage-clover=coverage.xml --testdox --color=auto
```
rpkamp: This is just to demonstrate the output, can be removed later of course

sebastianbergmann: Please do not change `.travis.yml`.

sebastianbergmann:
> There is one failing test, the test with `Balance cannot become negative` and `Balance cannot become negative 2` which should print only one line. I've tried several different solutions, but somehow I can't seem to get it to fit in the current design. Is this behaviour something you want to keep, or is it fine if it prints as two lines instead of one?

I think it's not only fine but correct to print two lines instead of one. For consistency, though, the other TestDox outputs should be changed, too. Can you take care of that?

sebastianbergmann:
> My printer is based on the `PHPUnit\TextUI\ResultPrinter` printer, and getting it to work with the `PHPUnit\Util\TestDox\TextResultPrinter` would be near impossible without a big overhaul of how printers are currently set up. So I put mine in the `PHPUnit\TextUI` namespace and load it as default when `--testdox` is provided on the command line.

Makes sense. However, even though your printer is based on `PHPUnit\TextUI\ResultPrinter` it does not have to be under the `PHPUnit\TextUI` namespace / in the `src/TextUI` directory. Please move it to where the rest of the `TestDox` related code is.

tests/TextUI/dataprovider-testdox.phpt
```diff
@@ -13,7 +13,12 @@ PHPUnit\TextUI\Command::main();
...
- [x] Does something with one
- [x] Does something with two
- [x] Does something else with one
- [x] Does something else with two
+ ✔ Does something with data set "one" [%f ms]
+ ✔ Does something with data set "two" [%f ms]
+ ✔ Does something else with data set "one" [%f ms]
+ ✔ Does something else with data set "two" [%f ms]
+
+
```
sebastianbergmann: Any chance we can get rid of this extra blank line?

sebastianbergmann:
Finally, the most important comment: thank you for taking the time to work on this!

sebastianbergmann:
Thank you, @rpkamp! I'll clean up the CS/WS issues.

rpkamp:
Cool, thanks! Any indication when this will be released? I'm guessing PHPUnit 7.0?

sebastianbergmann:
Yes, this will be in PHPUnit 7.0.

PR #2921, opened 2017-12-14
Tell contributors to use php-cs-fixer-v2

rpkamp:
php-cs-fixer v1 does not work with the current settings because it does not recognise [some of] them


sebastianbergmann:
Merged manually, thanks.

PR #2925, opened 2017-12-16
Fix names of setBeStrictAboutChangesToGlobalState methods

kubawerlos:
so they match this [one](https://github.com/sebastianbergmann/phpunit/blob/5.7/src/Framework/TestCase.php#L1222).


localheinz (commented):
👍

PR #2932, opened 2017-12-18
PHP CS Fixer - update config

keradus:


.php_cs.dist
```diff
@@ -85,20 +85,22 @@ return PhpCsFixer\Config::create()
...
-                'property_protected',
-                'property_private',
-                'construct',
-                'destruct',
-                'magic',
-                'phpunit',
-                'method_public',
-                'method_protected',
-                'method_private'
+                'order' => [
```
keradus: rootless config is deprecated

sebastianbergmann:
The `5.7` and `6.5` branches need the same change.

sebastianbergmann:
Merged manually, thanks.

keradus:
damn, I didn't noticed it was originally applied on 5.7.
I saw only the commit that was targeting master (gh shows it was applied to master, but not prior brnaches, probably due to way how it was obtained from lower branch)

thanks !

sebastianbergmann:
No worries. Thanks!

PR #2935, opened 2017-12-21
Initial try to add phpstan as a build stage

mikeSimonson:


mikeSimonson:
@sebastianbergmann Do you have a phpstan configuration that you use already that could be added to this pull request ?

sebastianbergmann:
I do not have a phpstan configuration yet but I'll create one when I start to use phpstan on Travis.

sebastianbergmann:
I do not have a phpstan configuration yet but I'll create one when I start to use phpstan on Travis.

sebastianbergmann:
If I understand https://travis-ci.org/sebastianbergmann/phpunit/builds/319594903 correctly, then the "Code Quality" stage uses PHP 7.1. Can we configure this to be PHP 7.2?

mikeSimonson:
@sebastianbergmann Are you coming to phpbenelux ? We could discuss the specifics of the phpstan configuration there.

sebastianbergmann:
Please do not mix the addition of phpstan as a build stage with changes to PHPUnit's code.

mikeSimonson:
I made enough change to make the build pass. The other option is that I put all the errors in the ignore config of phpstan and then they can be tackled one by one.

sebastianbergmann:
What I meant was that I need minimal, self-contained pull requests. One for the phpstan build stage and one for each of the code changes.

mikeSimonson:
Ok I will remove the property type change too then

mikeSimonson:
Youhoo and it's green now

phpstan.neon
```diff
@@ -0,0 +1,20 @@
...
+        - '#Caught class SebastianBergmann\\Invoker\\TimeoutException not found.#'
+        - '#Variable \$filename might not be defined.#'
+        - '#Variable \$hasMetRequirements might not be defined.#'
+        - '#Variable \$testResult might not be defined.#'
+        - '#Variable \$color might not be defined.#'
+        - '#Variable \$stdout might not be defined.#'
+        - '#Variable \$stderr might not be defined.#'
+        - '#Variable \$className might not be defined.#'
+        - '#Variable \$testMethod might not be defined.#'
+        - '#Variable \$codeCoverage in isset\(\) always exists and is not nullable.#'
```
ondrejmirtes: Can you please report these `might not be defined` as issues to PHPStan? Why do you think they are not errors worth fixing?

Jean85: I think he's just honoring Sebastian's request:

> What I meant was that I need minimal, self-contained pull requests. One for the phpstan build stage and one for each of the code changes.

So they will tackle errors later.

ondrejmirtes: OK, I recommend starting with level 0 instead.

ondrejmirtes: Rather than ignoring something and then forgetting to fix that.

ondrejmirtes: Errors about `might not be defined` variables are starting with level 1, so if you're ignoring all those, you can start with level 0 😊

mikeSimonson: Well i didn't knew that there was a level 0. But if @sebastianbergmann agree with you I can easily move the PR to that solution.

.travis.yml
```diff
@@ -45,3 +45,12 @@ after_success:
...
   email: false
 
+jobs:
+  include:
+    - stage: Code Quality
+      php: 7.2
+      env: STATIC_ANALYSIS
+      install: travis_retry composer update --prefer-dist --prefer-stable
+      before_script:
+        - travis_retry composer require --dev --prefer-dist --prefer-stable phpstan/phpstan:^0.9
```
mhujer: I suggest using a specific phpstan/phpstan version here (`0.9.1` at the time of writing). If new PHPStan version can detect new types of issues, it would break the build for unrelated PR.

.travis.yml
```diff
@@ -45,3 +45,12 @@ after_success:
...
+      script: vendor/bin/phpstan analyse -l 1 src -c phpstan.neon
```
mhujer: As mentioned in the comments bellow, the level here should be changed to `0` - to start with the basic checks.

.travis.yml
ondrejmirtes: I disagree - by locking to a specific version, you'd deprive yourself of bugfixes and improvements of new versions. My policy for 0.x.y versions is the following:

Bumping number y can bring new tool features (like a new error formatter or a new CLI option) and bugfixes/improvements that lead to better understanding of the code.

Bumping number x usually brings new rules that detect new types of defects in the code and can also introduce BC breaks.

So here I'd require `^0.9`- Although almost any change can "break the build", it's usually welcome by the users that the code is now understood better and that maybe some type of error can no longer need to be ignored etc.

sebastianbergmann:
Replaced by #2980.

PR #2936, opened 2017-12-21
Adding support for partial string matches within TraversableContains 

mmollick:
Currently `assertContains` will behave in the following way:
```
$this->assertContains('foo', 'foo bar'); // Passes, allowing partial match
$this->assertContains('foo', ['foo', 'bar']); // Passes
$this->assertContains('foo', ['foo bar']); // Fails, expects an exact match
```

This pull request will allow the third assertion to pass by enabling the same partial string matching used when both the expected and actual values are both strings available when the actual value is a type of `Traversable` containing strings.

### DIFF
--- a/src/Framework/Assert.php
+++ b/src/Framework/Assert.php
@@ -186,7 +186,8 @@ abstract class Assert
             $constraint = new TraversableContains(
                 $needle,
                 $checkForObjectIdentity,
-                $checkForNonObjectIdentity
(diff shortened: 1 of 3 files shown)


sebastianbergmann:

When both arguments are strings then I expect a substring match. But when the first argument is a string and the second argument is an array, for instance, then I expect exact matches.

Issue #2236, opened 2016-07-06
Exceptions in tearDown should affect $this->getStatus()


rtwo:
Currently if a Tests throws an Exception in its tearDown function $this->getStatus() will still be 0 and i.e. TestListeners would assume the Tests to be successful.

The following Code should be moved down until after the last occurrence of $_e = ...

https://github.com/sebastianbergmann/phpunit/blob/master/src/Framework/TestCase.php#L971

sebastianbergmann:
The link is no longer pointing to the correct line. Can you please send a pull request that implements the change you propose? Thanks!

sebastianbergmann:
@rtwo ping?

rtwo:
Sorry, I missed your first reply. I am a bit busy atm but I will try to look into it.

ekharchenko-avito:
This patch broke our cleanup logic based on knowledge of test status in tearDown method 😞

epdenouden:
I poked around in the tearDown logic, I'll have a look

PR #2940, opened 2017-12-24
ref #2236 - Move status assignment after last occurence of _e AND add…

Morerice:
I think that this is what @rtwo was referring to in #2236

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -925,11 +925,6 @@ abstract class TestCase extends Assert implements Test, SelfDescribing
             $e = $_e;
         }
 
-        if (isset($_e)) {
-            $this->status        = BaseTestRunner::STATUS_ERROR;
-            $this->statusMessage = $_e->getMessage();
-        }
-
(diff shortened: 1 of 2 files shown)


keradus:
as a bug, would be great to fix it on lowest maintained branch, not on not-yet-released phpunit 7 (major, bc breakers)

sebastianbergmann:
Cherry-picked into `6.5` and `master`, backported to `5.7`. Thanks!

CDJojn:
@bigaltum Is it bug or wanted behavior? We are currently facing this issue, all our failing tests are marked as PASSED (internal state) because of TestCase::hasFailed() returns false...

sebastianbergmann:
@bigaltum @CDJojn Please open a new ticket and provide a **minimal**, **self-contained**, reproducing test case that shows your problem.

CDJojn:
@sebastianbergmann Thank you for your reply. It seems my issue is duplicate of this bug https://github.com/sebastianbergmann/phpunit/issues/3031 - comment added.

PR #2946, opened 2018-01-01
Fix property type

mikeSimonson:


sebastianbergmann:
Wasn't this already changed in 7f86e9a03d995472a1c2eaf5a0be67c9aeb63d7a?

mikeSimonson:
I removed that commit from that pull request to put it in a separate one as you requested.

mikeSimonson:
?

Issue #1257, opened 2014-05-10
Split the assertion tool as a separate package


stof:
the PHPUnit assertion system is really useful. Because of that, it is common for Behat extensions to introduce a dependency on PHPUnit to reuse the assertions rather than re-implementing them (Behat wants to get an exception to mark a scenario as failed, so the PHPUnit_Framework_AssertionFailedError works out of the box).
However, depending on `phpunit/phpunit: ~4.0` means depending on at least 12 packages. But many of them are useless when only the assertions are used as they are needed by the test runner, not the assertion system.

Could it be possible to extract the assertion system as a separate package to be able to use it outside PHPUnit ? Thanks to composer, this should make almost no difference for PHPUnit users as the assertion system would be installed automatically when installing PHPUnit.

stof:
any news about this ?

localheinz:
@sebastianbergmann @stof 

Is this still relevant?

Because if it is, I am happy to tackle it!

sebastianbergmann:
@localheinz and others did a lot of work on this. In the end we realized that it 1) requires even more work and 2) is not worth the effort.

PR #2960, opened 2018-01-14
Enhancement: Pull in tests from sebastianbergmann/phpunit-framework-constraint

localheinz:
This PR

* [x] pulls in tests created and extracted as part of the work on `sebastianbergmann/phpunit-framework-constraint`

Related to #1257.
Replaces #2820.

💁‍♂️ During the second PHPUnit code sprint at trivago in October 2017, some work was done towards splitting the constraints into a separate package. However, it seems that while the work @shrikeh, @wesolowski, and I put in was good, maybe splitting wasn't the best idea. Not all is lost, though, since we added and extracted tests, which this pull requests suggests to merge into `6.5`.


localheinz:

PR #2961, opened 2018-01-15
Add PhpStorm meta file

tomasfejfar:
This allows you to do this with code completion in PhpStorm:
```
$mock = $this->createMock(MyClass::class);
$mock->doWhatClassDoes();
$mock->expects($this->any());
```
Class methods are suggested as well as mock methods. 

Unfortunately, it's [not currently possible](https://youtrack.jetbrains.com/issue/WI-37065) to set up this behavior with MockBuilder - the information about the mocked class is lost with the first call. 

Moved from https://github.com/sebastianbergmann/phpunit-mock-objects/pull/399. It makes sense to have it in the main repo as it specifies the behavior of TestCase class.

### DIFF
--- a/.phpstorm.meta.php
+++ b/.phpstorm.meta.php
@@ -1,0 +1,16 @@
+<?php
+
+namespace PHPSTORM_META {
+    override(
+        \PHPUnit\Framework\TestCase::createMock(0),
+        map([
+            '' => '@|\PHPUnit\Framework\MockObject\MockObject',
+        ])
+    );
+    override(
+        \PHPUnit_Framework_TestCase::createMock(0),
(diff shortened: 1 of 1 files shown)


sebastianbergmann:

While I cannot imagine to work without PhpStorm anymore, any type of IDE support is outside the scope of the PHPUnit project itself.

tomasfejfar:
OK, that's a fair point. I'll create a separate package.

tomasfejfar:
I've created https://github.com/tomasfejfar/phpunit-phpstorm-meta instead. Feel free to try. It might become a suggested package later when it proves useful.

PR #2970, opened 2018-01-23
Add ArrayHasKeyValuePair assertion

kwn:
```
In order to assert that an array has a specific key => value pair
As a developer
I want to use nice, simple and clean assertion
```

Currently to check if array contains specific key => value pair we have three options. All of them allow us to avoid raising an error if the key is not present in the array.

1) Make two assertions (making two assertions is annoying)
```php
$response = json_decode($this->httpClient->get('someUrl'), true);
Assert::assertArrayHasKey('errorCode', $response);
Assert::assertEquals(404001, $response['errorCode']);
```

2) Assert subset exists (dangerous as it's recursive, not very clear)
```php
$response = json_decode($this->httpClient->get('someUrl'), true);
Assert::assertArraySubset(['errorCode' => 404001], $response);
```

3) Use null coalescing operator (only in php >=7.0)
```php
$response = json_decode($this->httpClient->get('someUrl'), true);
Assert::assertEquals(404001, $response['errorCode'] ?? null);
```

Perhaps there are some more options but I suggested only the most concise ones. 

My suggestion is to introduce a new assertion called `ArrayHasKeyValuePair` which accepts `key`, `value` and `array` parameters. `ArrayHasKeyValuePair` is not recursive and basically is a combination of `ArrayHasKey` + value identity comparison. It's more semantic and simple to understand than all above options.

To compare with other testing frameworks in Hamcrest you can make the same assertion in 1 line of code: 
```
assertThat($response, hasKeyValuePair('errorCode', 401002));
```

Please let me know your thoughts. Happy to update docs once the PR is accepted.

### DIFF
--- a/src/Framework/Assert.php
+++ b/src/Framework/Assert.php
@@ -15,6 +15,7 @@ use Countable;
 use DOMDocument;
 use DOMElement;
 use PHPUnit\Framework\Constraint\ArrayHasKey;
+use PHPUnit\Framework\Constraint\ArrayHasKeyValuePair;
 use PHPUnit\Framework\Constraint\ArraySubset;
 use PHPUnit\Framework\Constraint\Attribute;
 use PHPUnit\Framework\Constraint\Callback;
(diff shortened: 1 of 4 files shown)


localheinz:
@kwn 

How about implementing custom assertions for your projects via traits? Whether they are domain specific or just slightly more specific than the ones provided here, traits works fine without putting more of a maintenance burden on the project. Maybe https://thephp.cc/dates/2017/10/international-php-conference/domain-specific-assertions helps?

kwn:
@localheinz 

Hi! Thank for the suggestion, however the assertion I suggest to introduce is not domain specific. It's very generic assertion agains an array, and it could be reused in many different cases. It's just about convenience. And there's not much more code to maintain. It's automatically tested anyway 😉

sebastianbergmann:

kwn:
Hi @sebastianbergmann

Can you share some more details about your decision please? Thanks.

PR #2976, opened 2018-01-27
Ensure that assertSame generates diff for arrays

### LINKED ISSUE #2169: Method assertSame doesn't create ComparisonFailure for arrays
```php
<?php
class CompareTest extends PHPUnit_Framework_TestCase
{
    public function testSame()
    {
        $this->assertSame(array('one'), array('two'));
    }

    public function testEquals()
    {
        $this->assertEquals(array('one'), array('two'));
    }
}
```

Method `$this->assertSame(array('one'), array('two'));` will throw `PHPUnit_Framework_ExpectationFailedException` with `ComparisonFailure` equals to `null` while method `$this->assertEquals(...)` will create proper `ComparisonFailure` object.

`$this->assertSame(...)` uses `PHPUnit_Framework_Constraint_IsIdentical.evaluate(...)`
`$this->assertEquals(...)` uses `PHPUnit_Framework_Constraint_IsEqual.evaluate(...)`

The issue was reproduced with PHPUnit 5.3

Corresponding PhpStorm issue: https://youtrack.jetbrains.com/issue/WI-31705

mhujer:
Fixes #2169.

With this patch, the arrays are diffed for `assertSame()` similarly to `assertEquals()`:

```php
<?php declare(strict_types = 1);

class ArrayDiffTest extends TestCase
{
	public function testEquality()
	{
		$this->assertEquals(
			[1, 2, 3, 4, 5, 6],
			[1, 2, 33, 4, 5, 6]
		);
	}

	public function testSame()
	{
		$this->assertSame(
			[1, 2, 3, 4, 5, 6],
			[1, 2, 33, 4, 5, 6]
		);
	}
```

```
1) ArrayDiffTest::testEquality
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
 Array (
     0 => 1
     1 => 2
-    2 => 3
+    2 => 33


2) ArrayDiffTest::testSame
--- Expected
+++ Actual
@@ @@
 Array &0 (
     0 => 1
     1 => 2
-    2 => 3
+    2 => 33

```

I also tested it with PHPStorm and it fixes [the diffing issue](https://youtrack.jetbrains.com/issue/WI-31705) there as well:

image

## Questions:
1) is this a correct approach?
2) Wouldn't is be better to merge it to 6.5 branch?

### DIFF
--- a/src/Framework/Constraint/IsIdentical.php
+++ b/src/Framework/Constraint/IsIdentical.php
@@ -92,6 +92,16 @@ class IsIdentical extends Constraint
                 );
             }
 
+            // if both values are array, make sure a diff is generated
+            if (\is_array($this->value) && \is_array($other)) {
+                $f = new SebastianBergmann\Comparator\ComparisonFailure(
+                    $this->value,
+                    $other,
+                    $this->exporter->export($this->value),
+                    $this->exporter->export($other)
+                );
+            }
+
             $this->fail($other, $description, $f);
         }
     }
@@ -135,6 +145,10 @@ class IsIdentical extends Constraint
 
         if (\is_string($this->value) && \is_string($other)) {
             return 'two strings are identical';
+        }
+
+        if (\is_array($this->value) && \is_array($other)) {
+            return 'two arrays are identical';
         }
 
         return parent::failureDescription($other);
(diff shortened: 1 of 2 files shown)


mhujer:
Appveyor failure seems unrelated. It is caused by http://windows.php.net/downloads/releases/php-7.2.0-Win32-VC15-x86.zip not being available (7.2.1 is available, so probably the version number should be incremented in AppVeyor config).

PR #2977, opened 2018-01-27
Update AppVeyor build PHP versions

mhujer:
Those files are no longer available:
- http://windows.php.net/downloads/releases/php-7.1.12-Win32-VC14-x86.zip
- http://windows.php.net/downloads/releases/php-7.2.0-Win32-VC15-x86.zip

Those are available:
- http://windows.php.net/downloads/releases/php-7.1.13-Win32-VC14-x86.zip
- http://windows.php.net/downloads/releases/php-7.2.1-Win32-VC15-x86.zip

Symfony hosts the binary files itself - see its [appveyor.yml](https://github.com/symfony/symfony/blob/0d975ccdba10387b90ccaa27f8448c3c088e4aba/appveyor.yml#L18). Maybe the similar approach can be taken here to prevent suddenly broken builds.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -33,7 +33,7 @@ install:
   - IF NOT EXIST c:\php mkdir c:\php
   - IF NOT EXIST c:\php\%PHP_VERSION% mkdir c:\php\%PHP_VERSION%
   - cd c:\php\%PHP_VERSION%
-  - IF NOT EXIST php-installed.txt appveyor DownloadFile http://windows.php.net/downloads/releases/php-%PHP_VERSION%-x86.zip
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
This makes me question the usefulness of AppVeyor. Maybe we should just remove it.

mhujer:
Or we can change links to point to http://windows.php.net/downloads/releases/archives/. There are all the older releases, so AppVeyor would run on a little bit older version, which should be fine.

sebastianbergmann:
Using official binaries from `windows.php.net` sounds like a good idea.

mhujer:
I've updated the PR to use the new URL.

PR #2980, opened 2018-01-27
Add PHPStan (level 0) as a build stage

mhujer:
> Depends on https://github.com/sebastianbergmann/phpunit/pull/2978 and https://github.com/sebastianbergmann/phpunit/pull/2979. After they are merged, I will rebase this one.

This is inspired by https://github.com/sebastianbergmann/phpunit/pull/2935 (I hope that @mikeSimonson won't get angry with me :-) ), but some things are done differently:

- ~~I used a specific phpstan/phpstan version to prevent build breakage when new checks are added to PHPStan (It would probably fail on unrelated PRs)~~
- I have enabled only Level 0 checks
- PHPStan is configured to analyze both `src` and `tests` (with different configs to allow separately ignore errors)

PR #2981, opened 2018-01-27
Fix comparison changed in the CS commit

mhujer:
I was trying what happens when I enable higher [PHPStan level](https://github.com/sebastianbergmann/phpunit/pull/2980) and I discovered this bug.

It was broken recently during a CS fix https://github.com/sebastianbergmann/phpunit/commit/3a9494be34135e5d42964ec7f5f17605e736a2c7#diff-29dd5adc8f5a89b298b28f4629b83e79R334

### DIFF
--- a/src/Util/Configuration.php
+++ b/src/Util/Configuration.php
@@ -346,7 +346,7 @@ final class Configuration
                     continue;
                 }
 
-                if (!$node->tagName === 'arguments') {
+                if ($node->tagName !== 'arguments') {
                     continue;
                 }
 


src/Util/Configuration.php
```diff
@@ -346,7 +346,7 @@ public function getListenerConfiguration(): array
                     continue;
                 }
 
-                if (!$node->tagName === 'arguments') {
```
mhujer: PHPStan reported:
`Strict comparison using === between bool and string will always evaluate to false.`

PR #3003, opened 2018-02-12
PHPUnit exits with error status code when php-error was converted to exception

michalbundyra:
By default PHPUnit converts all php errors (notices/warnings/...) to exceptions. As I can guess it is to fail (have exit status !== 0) when we have some notice/warning/... etc. We can disable it in configuration (option: `convert*ToExceptions` -https://phpunit.de/manual/current/en/appendixes.configuration.html) but then we are not going to have exit error code !== 0 on notice/warning/...

Unfortunately it converting php errors to exceptions could cause some fals positive cases.
I've created simple repository to demonstrate one:
https://github.com/webimpress/phpunit-error-conversion

Basically when we have code:
```php
function myFunction() {
  try {
    // some code where we get notice/warning/....
    // it will be converted by PHPUnit to exception
  } catch (Exception $e) {
    // and that exception will be caught here and the result will be 0
    // but in real call execution (without PHPUnit) there is not going to be any exception
    return 0;
  }
  return 1;
}
```

I think then we should always return exit code 1 when we converted something to exceptions and display some information about it.

In this PR I modify only a bit TestRunner to check `ErrorHandler` and converted php-errors.
Probably we need also similar check in `Command`:
https://github.com/sebastianbergmann/phpunit/blob/master/src/TextUI/Command.php#L212-L216

I don't know the whole flow in PHPUnit so I just want to point the issue and I hope you can find the best solution.

Please remember I'm giving very simple example to understand where is the problem. You can argue that this code is wrong and it shouldn't be written like that. But imho the testing tool shouldn't affect the subject of testing.

### DIFF
--- a/src/TextUI/TestRunner.php
+++ b/src/TextUI/TestRunner.php
@@ -27,6 +27,7 @@ use PHPUnit\Runner\StandardTestSuiteLoader;
 use PHPUnit\Runner\TestSuiteLoader;
 use PHPUnit\Runner\Version;
 use PHPUnit\Util\Configuration;
+use PHPUnit\Util\ErrorHandler;
 use PHPUnit\Util\Log\JUnit;
 use PHPUnit\Util\Log\TeamCity;
 use PHPUnit\Util\Printer;
@@ -632,6 +633,14 @@ class TestRunner extends BaseTestRunner
                     );
                 }
             }
+        }
+
+        if ($stack = ErrorHandler::getErrorStack()) {
+            $result->addError(
+                new TestSuite(),
+                new Exception('Some PHP errors were converted to exceptions'),
+                microtime(true)
+            );
         }
 
         if ($exit) {


sebastianbergmann:
I do not understand the problem you are trying to solve with this pull request.

webimpress:
@sebastianbergmann This is more issue report than real problem solution.

Please see my example repository I've created:
- check this test: https://bitbucket.org/purplestudios/location-api/pull-requests/23/mocking-reports-odm-document-persister/diff - false positive
- and code: https://github.com/webimpress/phpunit-error-conversion/blob/master/src/MyClass.php#L10-L19
- and script with the same code as false-positive tests: https://github.com/webimpress/phpunit-error-conversion/blob/master/test.php#L7-L10

So when you run tests all will be green and pass in travis build, the result of the method `method` in false positive test is `0`, but when you run script `php test.php` the result of the method `method` is 0 and we got notice error.

Basically I think in case of converting php-errors to exception PHPUnit should exit with error code.
Hope it's more clear right now.

sebastianbergmann:
When an error, notice, warning, etc. is converted to an exception then that exception will cause a test to fail. Unless the test expects the exception.

webimpress:
@sebastianbergmann not in that case I've presented, please check travis build:
https://travis-ci.org/webimpress/phpunit-error-conversion/builds/340522177?utm_source=github_status&utm_medium=notification

All is green ! All pass ! Test case returns 0, but normal call returns 1. It's false positive, and all is caused by PHPUnit.

weierophinney:
@sebastianbergmann Please look at the code samples he linked. Here's the scenario:

- The SUT has a try/catch block, catching `\Exception`.
- The code in the `try` block attempts to create a `DateTime` instance using an array member from an array provided to the method.
- The array member does not exist, which will trigger an `E_NOTICE` or `E_WARNING` due to the missing member.

Typically, if you do not have an error handler that converts errors to exceptions, this would result in PHP considering the value a `0`, passing that to the `DateTime` constructor, and no exception being raised. The unit test would then fail, as it would receive an unexpected result.

However, because PHPUnit's error handler throws an exception when it handles an error, the `catch` block is triggered, causing a false-positive result.

What @webimpress is suggesting (to my understanding) is that if PHPUnit's error handler catches an error and re-throws it as an exception, it should report a test failure unless that exception is caught by PHPUnit itself. In this particular case, the `DateTime` constructor throws a vanilla `Exception`, which means he's catching the correct thing; it's just not being caught for the correct reason.

That said: @webimpress : you might want to alter the code to validate the input prior to passing it to the `DateTime` constructor, and raise your own exception in such a case. If the array key does not exist, or the value of that key is invalid for purposes of the `DateTime` constructor, raise an `InvalidArgumentException`.

sebastianbergmann:
@weierophinney Thank you for the explanation. However, I will not introduce additional complexity to PHPUnit to "fix" this. The correct solution, as you already wrote, is to not expect / use `Exception`.

dofos:
@sebastianbergmann What if PHPUnit verified whether a notice or a warning was emitted in a similar way in which PHPMocks are verified? I mean the code would run as is while the handler would be only be spying on the notices and warnings it generated. If an assert exception was thrown the user would be faced with it in first place. When s/he fixed the problem with the assertion PHPUnit would verify if any notices/warnings were emitted and it would fail the test just like PHPUnit's mocks.

PR #3004, opened 2018-02-12
Add asserts to test exceptions thrown by functional callable functions

sergiorodenas:
The ```assertException()``` method mentioned on https://thephp.cc/news/2016/02/questioning-phpunit-best-practices

### DIFF
--- a/src/Framework/Assert.php
+++ b/src/Framework/Assert.php
@@ -2525,6 +2525,50 @@ abstract class Assert
     }
 
     /**
+     * Asserts that a given exception is thrown by a callable function
+     *
+     * @param string $expectedException
+     * @param callable $evaluation
+     * @param string $message
+     *
+     * @throws AssertionFailedError
+     * @throws Exception
+     * @throws \Exception
+     * @throws \SebastianBergmann\RecursionContext\InvalidArgumentException
+     */
+    public static function assertException(string $expectedException, callable $evaluation, string $message = ''): void
+    {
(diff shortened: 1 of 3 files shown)


sebastianbergmann:

The reason for not adding this is already in the article you mention:
> "The argument that is made in favor of this approach is that it allows checks for multiple exceptions within a single test. In my opinion, though, this is an argument against it. A unit test should be fine-grained and only test a single aspect of one object. Why would you need to check for different exceptions in a single test?"

Rodenastyle:
I agree, you shouldn't test multiple exceptions at same test, but as we learnt from PHP, who chooses to make things clean or messy is the developer, and some flexibility at this point could be very usable.

One example of usage is testing that despite the database query exception was thrown, database has not commit/persisted the data. Then, you would need to make those asserts at same test and the last one won't be reach even if it do not throws another exception.

The example:
```
/** @test */
    public function locations_must_have_a_customer_id()
    {
        $this->assertException(QueryException::class, function () {
            factory(Location::class)->create([
                'customer_id' => null,
                'name' => 'Location Test Name',
            ]);
        });

        // This will be reached
        $this->assertDatabaseMissing('locations', [
            'name' => 'Location Test Name',
        ]);

        // This will also be reached
        $this->assertCount(0, Location::all());
    }
```

Thank you so much for the quick response.

Rodenastyle:
@sebastianbergmann

PR #3006, opened 2018-02-13
remove redundant '@throws \Exception'

artspb:
All of the declarations are removed except one in `\PHPUnit\Util\ErrorHandler::handleErrorOnce` where it seems to be in place.

The fix introduced a couple more "Missing @throws tag(s)" warnings. I've checked and all such warnings at the moment relate to unhandled `\ReflectionException`. I'm not sure how to properly handle it. The exception doesn't extend `\RuntimeException` thus considered checked. Declaring it with `@throws` will require pulling it through all the callers. It doesn't look to me like a way to go. Probably, it's better to handle the exception where it's thrown and re-throw wrapped in a runtime exception.

### DIFF
--- a/src/Framework/Constraint/ArrayHasKey.php
+++ b/src/Framework/Constraint/ArrayHasKey.php
@@ -38,7 +38,6 @@ class ArrayHasKey extends Constraint
     /**
      * Returns a string representation of the constraint.
      *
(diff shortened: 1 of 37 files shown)


sebastianbergmann:
Merged manually, thanks.

PR #3007, opened 2018-02-13
Performance optimizations.

jwage:
@sebastianbergmann What do you think about these changes? In our test suite with 10000+ tests and 30000+ assertions, it gives us a 10-15% performance boost. If we can get rid of the method_exists call and just let PHP fail, it makes it a bit faster too.

And see my comment in TestSuite.php. We have a base class we extend from that has many non test methods which end up getting passed to `addTestMethod` but I am not sure how we can make that work more generically.

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -1133,19 +1133,28 @@ abstract class TestCase extends Assert implements Test, SelfDescribing
             );
         }
 
-        try {
-            $class  = new ReflectionClass($this);
-            $method = $class->getMethod($this->name);
-        } catch (ReflectionException $e) {
-            $this->fail($e->getMessage());
-        }
-
         $testArguments = \array_merge($this->data, $this->dependencyInput);
 
         $this->registerMockObjectsFromTestArguments($testArguments);
 
         try {
-            $testResult = $method->invokeArgs($this, $testArguments);
+            // only use reflection to invoke the method if we have to
+            if ($testArguments) {
+                try {
+                    $class  = new ReflectionClass($this);
+                    $method = $class->getMethod($this->name);
+                } catch (ReflectionException $e) {
+                    $this->fail($e->getMessage());
+                }
+
+                $testResult = $method->invokeArgs($this, $testArguments);
+            } else {
+                if (!method_exists($this, $this->name)) {
+                    $this->fail(sprintf('Method %s does not exist', $this->name));
+                }
+
+                $testResult =  $this->{$this->name}();
+            }
         } catch (Throwable $t) {
             $exception = $t;
         }
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -1138,23 +1138,7 @@ abstract class TestCase extends Assert implements Test, SelfDescribing
         $this->registerMockObjectsFromTestArguments($testArguments);
 
         try {
-            // only use reflection to invoke the method if we have to
-            if ($testArguments) {
-                try {
-                    $class  = new ReflectionClass($this);
-                    $method = $class->getMethod($this->name);
-                } catch (ReflectionException $e) {
-                    $this->fail($e->getMessage());
-                }
-
-                $testResult = $method->invokeArgs($this, $testArguments);
-            } else {
-                if (!method_exists($this, $this->name)) {
-                    $this->fail(sprintf('Method %s does not exist', $this->name));
-                }
-
-                $testResult =  $this->{$this->name}();
-            }
+            $testResult = $this->{$this->name}(...array_values($testArguments));
         } catch (Throwable $t) {
             $exception = $t;
         }
--- a/src/Framework/TestSuite.php
+++ b/src/Framework/TestSuite.php
@@ -183,12 +183,11 @@ class TestSuite implements Test, SelfDescribing, IteratorAggregate
         }
 
         foreach ($theClass->getMethods() as $method) {
-            // how can we handle this better? In our code base we have a base class we extend with lots of convenience methods that would not be skipped here and eventually get passed to addTestMethod
-            if ($method->getDeclaringClass()->getName() === TestCase::class) {
+            if ($method->getDeclaringClass()->getName() === Assert::class) {
                 continue;
             }
 
-            if ($method->getDeclaringClass()->getName() === Assert::class) {
+            if ($method->getDeclaringClass()->getName() === TestCase::class) {
                 continue;
             }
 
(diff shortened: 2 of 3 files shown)


sebastianbergmann:
I will have a look at this as soon as I have a chance.

src/Framework/TestCase.php
```diff
@@ -1133,19 +1133,28 @@ protected function runTest()
...
-        }
-
         $testArguments = \array_merge($this->data, $this->dependencyInput);
 
         $this->registerMockObjectsFromTestArguments($testArguments);
 
         try {
-            $testResult = $method->invokeArgs($this, $testArguments);
+            // only use reflection to invoke the method if we have to
+            if ($testArguments) {
```
fprochazka: What about using the argument unpacking to remove the reflection altogether?

sebastianbergmann: Definitely worth exploring. @jwage Do you want to look into this and update the pull request accordingly?

src/Framework/TestSuite.php
```diff
@@ -183,10 +183,15 @@ public function __construct($theClass = '', $name = '')
         }
 
         foreach ($theClass->getMethods() as $method) {
+            // how can we handle this better? In our code base we have a base class we extend with lots of convenience methods that would not be skipped here and eventually get passed to addTestMethod
```
docteurklein: Maybe with `is_a`?

sebastianbergmann: We could check whether the class that declares the method is abstract.

sebastianbergmann: Strike that, that does not work.

src/Framework/TestCase.php
sebastianbergmann: Does not work, results in `Cannot unpack array with string keys` error.

sebastianbergmann: Hmkay, using `...array_values($testArguments)` instead of simply `...$testArguments` seems to work.

sebastianbergmann:
Here is my current state:
```patch
diff --git a/src/Framework/TestCase.php b/src/Framework/TestCase.php
index 49ada7945..5d18eacc2 100644
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -1133,19 +1133,12 @@ protected function runTest()
             );
         }
 
-        try {
-            $class  = new ReflectionClass($this);
-            $method = $class->getMethod($this->name);
-        } catch (ReflectionException $e) {
-            $this->fail($e->getMessage());
-        }
-
         $testArguments = \array_merge($this->data, $this->dependencyInput);
 
         $this->registerMockObjectsFromTestArguments($testArguments);
 
         try {
-            $testResult = $method->invokeArgs($this, $testArguments);
+            $testResult = $this->{$this->name}(...array_values($testArguments));
         } catch (Throwable $t) {
             $exception = $t;
         }
diff --git a/src/Framework/TestSuite.php b/src/Framework/TestSuite.php
index 7197f4a00..622e47227 100644
--- a/src/Framework/TestSuite.php
+++ b/src/Framework/TestSuite.php
@@ -183,6 +183,10 @@ public function __construct($theClass = '', $name = '')
         }
 
         foreach ($theClass->getMethods() as $method) {
+            if ($method->getDeclaringClass()->getName() === Assert::class) {
+                continue;
+            }
+
             if ($method->getDeclaringClass()->getName() === TestCase::class) {
                 continue;
             }
diff --git a/tests/Framework/TestSuiteTest.php b/tests/Framework/TestSuiteTest.php
index ded01e946..7bceb56c2 100644
--- a/tests/Framework/TestSuiteTest.php
+++ b/tests/Framework/TestSuiteTest.php
@@ -27,32 +27,6 @@ protected function tearDown(): void
         $this->result = null;
     }
 
-    public static function suite()
-    {
-        $suite = new TestSuite;
-
-        $suite->addTest(new self('testAddTestSuite'));
-        $suite->addTest(new self('testInheritedTests'));
-        $suite->addTest(new self('testNoTestCases'));
-        $suite->addTest(new self('testNoTestCaseClass'));
-        $suite->addTest(new self('testNotExistingTestCase'));
-        $suite->addTest(new self('testNotPublicTestCase'));
-        $suite->addTest(new self('testNotVoidTestCase'));
-        $suite->addTest(new self('testOneTestCase'));
-        $suite->addTest(new self('testShadowedTests'));
-        $suite->addTest(new self('testBeforeClassAndAfterClassAnnotations'));
-        $suite->addTest(new self('testBeforeClassWithDataProviders'));
-        $suite->addTest(new self('testBeforeAnnotation'));
-        $suite->addTest(new self('testTestWithAnnotation'));
-        $suite->addTest(new self('testSkippedTestDataProvider'));
-        $suite->addTest(new self('testTestDataProviderDependency'));
-        $suite->addTest(new self('testIncompleteTestDataProvider'));
-        $suite->addTest(new self('testRequirementsBeforeClassHook'));
-        $suite->addTest(new self('testDoNotSkipInheritedClass'));
-
-        return $suite;
-    }
-
     public function testAddTestSuite(): void
     {
         $suite = new TestSuite(\OneTestCase::class);
@@ -91,17 +65,6 @@ public function testNoTestCaseClass(): void
         new TestSuite(\NoTestCaseClass::class);
     }
 
-    public function testNotExistingTestCase(): void
-    {
-        $suite = new self('notExistingMethod');
-
-        $suite->run($this->result);
-
-        $this->assertEquals(0, $this->result->errorCount());
-        $this->assertEquals(1, $this->result->failureCount());
-        $this->assertCount(1, $this->result);
-    }
-
     public function testNotPublicTestCase(): void
     {
         $suite = new TestSuite(\NotPublicTestCase::class);
```

src/Framework/TestCase.php
jwage: I will test this too and see if it gives us the same level of performance. even when $testArguments is an empty array.

src/Framework/TestSuite.php
jwage: @sebastianbergmann what happened when you continued on classes that are abstract? In my case that would work. We have a base test class that looks like this:

```
<?php

abstract class OpenSkyTestCase extends TestCase
{
    // ...
}

jwage: But now that I am thinking about it, it would break other real scenarios for people where they have a base abstract class with test methods that run as a part of a child class. We do this in our code base too.

sebastianbergmann: Exactly :-)

sebastianbergmann:
Merged manually, thanks.

PR #3010, opened 2018-02-14
PHP errors converted to exception always fail the test

michalbundyra:
As I've tried describe in #3003 there is possibility that php error converted to exception will be caught in user code and then test will passing (false positive).

@sebastianbergmann said:
> When an error, notice, warning, etc. is converted to an exception then that exception will cause a test to fail. Unless the test expects the exception.

But unfortunately it does not happen when user code caught that exception.

This PR adds checks if any error->exception conversion has been made in the test, and if so the test is failing.

Here we have two separate commits - first with failing test and the second with fix.

**Interesting fact:** as the build shows we have FALSE-POSITIVE test cases in PHPUnit !
The problem is in:
https://github.com/sebastianbergmann/phpunit/blob/79d95733a7342f6472957fcc0b96c85fbde086da/tests/Runner/PhptTestCaseTest.php#L127-L144
to specify in mocking `runJob` method. Only first invocation is mocked to return array but unfortunately that method is called twice:
1. https://github.com/sebastianbergmann/phpunit/blob/9c692830988283dcc8e571b6118a012e62bbaedd/src/Runner/PhptTestCase.php#L327
2. https://github.com/sebastianbergmann/phpunit/blob/9c692830988283dcc8e571b6118a012e62bbaedd/src/Runner/PhptTestCase.php#L164
and also after the second execution of the method we try to access `stdout` index of the array (but on the second call we got error, because only first invocation was mocked to return array) - see:             https://github.com/sebastianbergmann/phpunit/blob/9c692830988283dcc8e571b6118a012e62bbaedd/src/Runner/PhptTestCase.php#L172

And there we got converted notice to error, so the result of the test is **wrong**.

/cc @weierophinney @dofos

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -33,6 +33,7 @@ use PHPUnit\Framework\MockObject\Stub\ReturnStub;
 use PHPUnit\Framework\MockObject\Stub\ReturnValueMap as ReturnValueMapStub;
 use PHPUnit\Runner\BaseTestRunner;
 use PHPUnit\Runner\PhptTestCase;
+use PHPUnit\Util\ErrorHandler;
 use PHPUnit\Util\GlobalState;
 use PHPUnit\Util\PHP\AbstractPhpProcess;
 use Prophecy;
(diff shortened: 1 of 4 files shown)


sebastianbergmann:

PR #3012, opened 2018-02-15
Add abbility to getOriginalException() in ExceptionWrapper

denis-chmel:


sebastianbergmann:
IIRC, the whole point of the exception wrapper was/is to have a minimal stacktrace without circular references, etc. Would this not counter that?

denis-chmel:
Hi Sebastian, I may not be fully competent to answer that, but say from Open/closed principle point of view - a wrapper should provide access to a wrapped object. The goal I (and I guess others) are pursuing is to to be to detect instanceOf original exception, and `if ($defect->thrownException()->getOriginalException() instanceOf HasExtraInfo)` be able to render `$defect->thrownException()->getOriginalException()->getExtraInfo()` in TestResultPrinter::printDefectHeader().

Could you elaborate on your concern? My assumtion is that once it's a private var it shouldn't affect any stacktrace or bubble up anywhere elsehow (except when using a Reflection). I guess it's also not affecting memory usage cause it holds a pointer to original object, not a copy/clone.

sebastianbergmann:
If you store a reference to it in the private attribute then that could prevent the stacktrace contents, which can be quite big, from being garbage-collected, thus blocking memory until shutdown.

denis-chmel:
Thank you, seeing the issue more clear now.
I've found pair of workarounds to prevent private from getting into stacktrace:
https://stackoverflow.com/a/43051835/379874 (via __debugInfo magic method)
https://stackoverflow.com/a/13277926/379874 (via static vars in auxiliary method)
Is it reasonable to implement one of these (which one?) or it's an over-engineering?

denis-chmel:
@sebastianbergmann 
I've fixed all properly now, plus added a unittest to check for absence of originalException in a stacktrace. Approach with __debugInfo() does only hide it in var_dump (not in var_export / print_r) so I stick to static property in auxiliary method. What do you think, is it a viable change? Let me know what I can improve further please.

sebastianbergmann:
Thank you for cleaning this up. Before I merge this, though, I would like to understand why you need `getOriginalException()`. Please forgive me if you explained this already in the past and I forgot, I just tried to find this information and could not find it. Thanks!

denis-chmel:
Hi Sebastian, no problem. I'd like to show extra debug info beneath exceptions which has it.
Via registering custom ResultPrinter:
```
<phpunit ...
         printerClass="Tests\CustomResultPrinter"
>
```
with such code:
```
class CustomResultPrinter extends ResultPrinter
{
    protected function printDefectTrace(TestFailure $defect): void
    {
        parent::printDefectTrace($defect);

        $e = $defect->thrownException();
        if ($e instanceof ExceptionWrapper) {
            $original = $e->getOriginalException();
            if ($original instanceof HasContextInterface) {
                $this->write($original->getContextFormatted() . PHP_EOL);
            }
        }
    }
}
```
which will output "Context:" info nicely:
```
There was 1 error:

1) Tests\Unit\TravelStudio\BookingVacationTest::testCreatePassenger
App\Exceptions\TechException: Soap error: customer email format is not valid

/home/vagrant/gvv/app/helpers.php:929
/home/vagrant/gvv/tests/unit/API/BookingTest.php:26

Context:
− bookingId: 1234
− passengerInfo: { name: '', email: '', ... }
− API URL: http://api.com/v2
− Soap Request: <xml>...request..</xml>
− Soap Response: <xml>...response...</xml>
```
And be able to glance the exception payload to understand what could go wrong.

Issue #3018, opened 2018-02-22
PHPStan in CI build


mhujer:
In this issue I will track the progress of fixing issues detected by [PHPStan](https://github.com/phpstan/phpstan).

- [X] PHPStan Level 0:
  -  ~https://github.com/sebastianbergmann/phpunit/pull/2978~
  -  ~https://github.com/sebastianbergmann/phpunit/pull/2979~
  -  ~https://github.com/sebastianbergmann/phpunit/pull/2980~
- [ ]  PHPStan Level 1:
  - `src/`: ~https://github.com/sebastianbergmann/phpunit/pull/3023~ + https://github.com/sebastianbergmann/phpunit/pull/3024
  - `tests/`: ~https://github.com/sebastianbergmann/phpunit/pull/3019~
- [ ]  PHPStan Level 2:
  - ...
  - `tests/`: ~https://github.com/sebastianbergmann/phpunit/pull/3025~
- [ ] PHPStan Level 3:
  - ...
- ...

Note: While developing, I use [this patch](https://gist.github.com/mhujer/c8ea6f882dfbfce183e21ecc8d2041b7) to run the PHPStan locally as a Composer script.

PR #3023, opened 2018-02-26
PHPStan level 1 src/ - preparation

mhujer:
Preparation for PHPStan level 1 in `src/` (https://github.com/sebastianbergmann/phpunit/issues/3018)

Those changes are quite simple so I decided to put them in one PR. Each commit contains one change with an explanation in a commit message.


localheinz (approved):
:+1:

PR #3041, opened 2018-03-15
if data is null set it to empty array

antonydandrea:
I have found a situation where I get an error:

PHPUnit\Framework\TestCase::registerMockObjectsFromTestArguments() must be of the type array, null given, called in phar:///usr/bin/phpunit/phpunit/Framework/TestCase.php on line 1145

It is called by $this->data being NULL instead of an empty array in certain situations. This causes $testArguments to be null. 

This is a little fix that sets it to an empty array in this situation and passes all of the tests.

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -272,7 +272,8 @@ abstract class TestCase extends Assert implements Test, SelfDescribing
             $this->setName($name);
         }
 
-        $this->data     = $data;
+	$this->data     = $data;
+	
         $this->dataName = $dataName;
     }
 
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
Thank you for your pull request.

Please provide a **minimal**, **self-contained**, reproducing test case that shows the problem you are trying to solve.

sebastianbergmann:
No feedback, closing.

antonydandrea:
I'm really sorry, I was finding it difficult to replicate outside of my project (which implies it is a project specific bug).

PR #3043, opened 2018-03-15
Travis: groom SCA jobs

keradus:
as we run stan, why not run PHP CS Fixer, which was already integrated ages ago ;)

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -57,7 +57,7 @@ return PhpCsFixer\Config::create()
             'no_empty_comment' => true,
             'no_empty_phpdoc' => true,
             'no_empty_statement' => true,
-            'no_extra_consecutive_blank_lines' => true,
+            'no_extra_blank_lines' => true,
             'no_homoglyph_names' => true,
             'no_leading_import_slash' => true,
             'no_leading_namespace_whitespace' => true,
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,7 +58,7 @@ jobs:
         - travis_retry composer require --dev --no-update friendsofphp/php-cs-fixer:^2.10 phpstan/phpstan:^0.9
       install: travis_retry composer update --prefer-dist --prefer-stable
       script:
-        - PHP_CS_FIXER_FUTURE_MODE=1 vendor/bin/php-cs-fixer --diff --dry-run -v fix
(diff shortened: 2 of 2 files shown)


.php_cs.dist
```diff
@@ -57,7 +57,7 @@ return PhpCsFixer\Config::create()
             'no_empty_comment' => true,
             'no_empty_phpdoc' => true,
             'no_empty_statement' => true,
-            'no_extra_consecutive_blank_lines' => true,
+            'no_extra_blank_lines' => true,
```
keradus: just a deprecation fix, no logical changes

.travis.yml
```diff
@@ -54,10 +54,11 @@ jobs:
     - stage: Quality
       php: 7.2
       env: PHPStan
+      before_install:
+        - travis_retry composer require --dev --no-update friendsofphp/php-cs-fixer:^2.10 phpstan/phpstan:^0.9
       install: travis_retry composer update --prefer-dist --prefer-stable
-      before_script:
-        - travis_retry composer require --dev --prefer-dist --prefer-stable phpstan/phpstan:^0.9
```
keradus: running `composer update` and `composer require {package}` was causing composer to create deps graph and conflicts twice, with `composer require --no-update {package}`, it only modify json file, and then resolve all deps/conflicts only one in `composer update` run

TIL: no change in logic, only performance grooming

sebastianbergmann:
But `php-cs-fixer` is already part of the build: https://travis-ci.org/sebastianbergmann/phpunit/jobs/350952189

keradus:
I'm so blind ! :dagger: 
thanks, i cleaned up the branch, still some small grooming that could be applied :D

keradus:
@sebastianbergmann , would it make sense to you if we would switch PHP CS Fixer installed via composer ?

keradus:
thanks for merging ;)

sebastianbergmann:
I prefer using tools such as `php-cs-fixer` from a PHAR over installing them via Composer.

keradus:
hm... currently, PHP CS Fixer is installed via phar, and phpstan via composer :D

sebastianbergmann:
Because of phpstan's "weird" mix of static and dynamic analysis.

keradus:
:|

thanks for reasoning ;)

PR #3051, opened 2018-03-23
Fix printer construct

Qrzy:
Check if directory exists BEFORE trying to create it.   

Otherwise `mkdir(): File exists` error arises...

### DIFF
--- a/src/Util/Printer.php
+++ b/src/Util/Printer.php
@@ -53,7 +53,7 @@ class Printer
 
                     $this->out = \fsockopen($out[0], $out[1]);
                 } else {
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
Why do you get an error? `mkdir()` is called with errors being suppressed (`@` operator).

Qrzy:
No idea, to be honest. 
It's on Windows with XAMPP 7.1.15.

sebastianbergmann:
Thank you for your report.

Please provide a **minimal**, **self-contained**, reproducing test case that shows the problem you are reporting.

sebastianbergmann:
No feedback, closing.

Issue #3000, opened 2018-02-09
Directories are not created recursively


kleisauke:
| Q                   | A
| --------------------| ---------------
| PHPUnit version   | 7.0.0
| PHP version          | 7.1.11 & PHP 7.2.1
| Xdebug version    | 2.5.5
| Installation Method | Composer
| CI environment | Travis CI

[This patch](https://github.com/sebastianbergmann/php-code-coverage/commit/583c4e79cb597763b846010da6e423c4cdb2f39e#diff-19b24179073358fac0d4c1ab22d06243R237) should also be applied to https://github.com/sebastianbergmann/phpunit/blob/ef2c7372cabee8d0eee0d60fd2e2cfb2e4b82ea3/src/Util/Printer.php#L56

Relevant issue: https://github.com/sebastianbergmann/php-code-coverage/issues/584.

Travis log:
https://travis-ci.org/andrieslouw/imagesweserv/builds/339592842

```
PHPUnit 7.0.0 by Sebastian Bergmann and contributors.
PHP Fatal error:  Uncaught RuntimeException: Directory "/home/travis/build/andrieslouw/imagesweserv/build/logs" was not created in /home/travis/build/andrieslouw/imagesweserv/vendor/phpunit/phpunit/src/Util/Printer.php:57
Stack trace:
#0 /home/travis/build/andrieslouw/imagesweserv/vendor/phpunit/phpunit/src/Util/Log/JUnit.php(117): PHPUnit\Util\Printer->__construct('/home/travis/bu...')
#1 /home/travis/build/andrieslouw/imagesweserv/vendor/phpunit/phpunit/src/TextUI/TestRunner.php(347): PHPUnit\Util\Log\JUnit->__construct('/home/travis/bu...', true)
#2 /home/travis/build/andrieslouw/imagesweserv/vendor/phpunit/phpunit/src/TextUI/Command.php(205): PHPUnit\TextUI\TestRunner->doRun(Object(PHPUnit\Framework\TestSuite), Array, true)
#3 /home/travis/build/andrieslouw/imagesweserv/vendor/phpunit/phpunit/src/TextUI/Command.php(153): PHPUnit\TextUI\Command->run(Array, true)
#4 /home/travis/build/andrieslouw/imagesweserv/vendor/phpunit/phpunit/phpunit(53): PHPUnit\TextUI\Command::main()
#5 {main}
[...]

PR #3052, opened 2018-03-23
Require code coverage version that fixes directory creation

VasekPurchart:
The problem from issue https://github.com/sebastianbergmann/php-code-coverage/issues/584 has ongoing effect for libraries, if they are tested with lowest dependencies.

The only alternatives I came up with were either creating the folders upfront all locking code coverage package in each individual library.

I hope the pinned version can be pinned here, because otherwise the lowest deps testing will need to be changed at a lot of places.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,7 @@
         "phar-io/manifest": "^1.0.1",
         "phar-io/version": "^1.0",
         "phpspec/prophecy": "^1.7",
-        "phpunit/php-code-coverage": "^6.0",
+        "phpunit/php-code-coverage": "^6.0.1",
         "phpunit/php-file-iterator": "^1.4.3",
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
Merged manually.

VasekPurchart:
Thanks 👍

shadowhand:
Can we get a release for this? This is exactly the same thing I mentioned on #3000.

shadowhand:
See also sebastianbergmann/phpunit/issues/3000#issuecomment-376208533

shadowhand:
Related: https://github.com/sebastianbergmann/phpunit/issues/3000#issuecomment-376208533https://github.com/sebastianbergmann/phpunit/issues/3000#issuecomment-376208533

Issue #3040, opened 2018-03-11
StringMatchesFormatDescription fails when matching multiline with \r\n


jjeising:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 7.0.1
| PHP version         | 7.1.12
| Installation Method | PHAR (via homebrew/php)

`StringMatchesFormatDescription` seems to have issues when matching line breaks with carriage return:

```
$this->assertStringMatchesFormat("\r\n", "\r\n");
```
The diff is not helpful in this case:
```
PHPUnit 7.0.1 by Sebastian Bergmann and contributors.
Runtime:       PHP 7.1.12 with Xdebug 2.5.5

There was 1 failure:

1) Test::test
Failed asserting that string matches format description.
--- Expected
+++ Actual

…/test.php:10
```

This may be related to this replacement:

https://github.com/sebastianbergmann/phpunit/blob/3e2ff3bb45b13f11b8fd958fb30068e6e11a4f7e/src/Framework/Constraint/StringMatchesFormatDescription.php#L32

mkasberg:
I opened up a pull request with a proposed solution.

There are several potential ways to fix this (assuming we regard it as a bug); this should at least get the discussion started...

bkdotcom:
> assuming we regard it as a bug

I just racked my head for quite some time wondering why my assertion was failing..

why wouldn't this be a bug?

PR #3059, opened 2018-04-03
Fix StringMatchesFormatDescription with \r\n

mkasberg:
This is a proposed fix for #3040.

There was a bug in that the following assertion would fail, even though
it should pass:

    $this->assertStringMatchesFormat("\r\n", "\r\n");

Obviously, we would expect a string to match itself. The bug was due to
newline conversion - `\r\n` was being converted to `\n` on the pattern,
but not on the string to be matched against.

In order to be compatible with existing behavior, we need to continue
converting `\r\n` to `\n` in the pattern. (Otherwise, users who expect
this behavior may have tests fail that would have previously passed.)
Therefore, the only option for fixing the bug is to also sanitize the
string to be matched. This commit makes that change, and provides some
additional tests to verify the behavior.

### DIFF
--- a/src/Framework/Constraint/StringMatchesFormatDescription.php
+++ b/src/Framework/Constraint/StringMatchesFormatDescription.php
@@ -29,11 +29,26 @@ class StringMatchesFormatDescription extends RegularExpression
     {
         parent::__construct(
             $this->createPatternFromFormat(
-                \preg_replace('/\r\n/', "\n", $string)
+                $this->convertNewlines($string)
             )
         );
 
         $this->string = $string;
+    }
+
+    /**
+     * Evaluates the constraint for parameter $other. Returns true if the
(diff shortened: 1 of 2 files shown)


mkasberg:
That is correct.

Perhaps it is worth some discussion whether this behavior should be changed or not:

```

sebastianbergmann:
How does the "official" PHPT test runner handle this? Because that is where this comes from.

mkasberg:
Huh, I didn't realize this was based on PHPT. I will see if I can do some experiments and report back in a few days.

PHPT docs: https://qa.php.net/phpt_details.php#expectf_section

jjeising:
`run-tests.php` replaces newlines like the proposed patch:

https://github.com/php/php-src/blob/14de058086d76ac344fde67fc343023fc00279a9/run-tests.php#L2039

PR #3074, opened 2018-04-06
Fix: Do not use deprecated configuration for binary_operator_spaces fixer

localheinz:
This PR

* [x] uses the non-deprecated configuration for `binary_operator_spaces`, as suggested by `php-cs-fixer` itself

h/t @dnaber-de

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -14,8 +14,10 @@ return PhpCsFixer\Config::create()
         [
             'array_syntax' => ['syntax' => 'short'],
             'binary_operator_spaces' => [
-                'align_double_arrow' => true,
-                'align_equals' => true
+                'operators' => [
+                    '=' => 'align',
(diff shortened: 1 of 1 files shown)


localheinz:
@keradus 

Not sure, but @dnaber-de noticed that running `php-cs-fixer` rendered a deprecation warning (with the version of `php-cs-fixer` checked in into https://github.com/sebastianbergmann/phpunit/blob/7.0/build/tools/php-cs-fixer, but neither @sebastianbergmann nor I could reproduce the output. Taking a look later, maybe worth opening an issue at https://github.com/FriendsOfPHP/PHP-CS-Fixer!

sebastianbergmann:
Merged manually, thanks.

localheinz:

keradus:
if we don't have warning content nor way to reproduce, hard to say

PR #3082, opened 2018-04-07
issue/2830: @runClassInSeparateProcess fails for tests with a @dataPr…

mihailogajic:
This will fix the Exception with message "PHPUnit\Framework\TestCase::$name must not be null" that appears as was described in the ticket.


sebastianbergmann:
Merged manually, thanks.

PR #3083, opened 2018-04-07
Issue/2724

mihailogajic:
TestCase with @runClassInSeparateProcess was being executed in main thread, therefore some tests were passing because tests were executing in the same process.

After making the changes in TestSuite, every method will be executed in it's own process when @runClassInSeparateProcess is used, and tests were updated accordingly.

### DIFF
--- a/tests/Regression/GitHub/2591/SeparateClassPreserveTest.php
+++ b/tests/Regression/GitHub/2591/SeparateClassPreserveTest.php
@@ -14,13 +14,15 @@ class Issue2591_SeparateClassPreserveTest extends TestCase
 
     public function testChangedGlobalString()
     {
-        $GLOBALS['globalString'] = 'Hello!';
-        $this->assertEquals('Hello!', $GLOBALS['globalString']);
+        $value = 'Hello! I am changed from inside!';
+
+        $GLOBALS['globalString'] = $value;
+        $this->assertEquals($value, $GLOBALS['globalString']);
     }
 
     public function testGlobalString()
     {
-        $this->assertEquals('Hello!', $GLOBALS['globalString']);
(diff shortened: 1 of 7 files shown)


tests/Regression/GitHub/2591/SeparateClassPreserveTest.php
```diff
@@ -14,13 +14,15 @@ public function testOriginalGlobalString()
...
+        $value = 'Hello! I am changed from inside!';
+
+        $GLOBALS['globalString'] = $value;
+        $this->assertEquals($value, $GLOBALS['globalString']);
     }
 
     public function testGlobalString()
     {
-        $this->assertEquals('Hello!', $GLOBALS['globalString']);
+        $this->assertEquals('Hello', $GLOBALS['globalString']);
```
mihailogajic: $GLOBALS can't be shared between the methods, due to different processes. This was passing before because @runClassInSeparateProcess didn't have any effects, and each test was executed in master thread.

tests/Regression/GitHub/2725/BeforeAfterClassPidTest.php
```diff
@@ -9,29 +9,35 @@
...
      * @beforeClass
      */
     public static function showPidBefore()
     {
-        $GLOBALS['PID_BEFORE'] = getmypid();
+        $GLOBALS[static::PID_VARIABLE] = getmypid();
     }
 
-    public function testComparePids()
+    public function testMethod1WithItsBeforeAndAfter()
```
mihailogajic: This was passing before because everything was running in the master thread.

sebastianbergmann:
Merged manually, thanks.

PR #3086, opened 2018-04-08
Prefix all code bundled in PHAR distribution with unique namesspaces

kambo-1st:
This is a proof of concept for the #2015 it's by no means a final solution. For further detail please look at comments for #2015.


sebastianbergmann:
@kambo-1st @theofidry Has there been progress? Thanks!

theofidry:
I'm hacking on the ongoing issues on PHP-Scoper. However I don't think any of them are blockers to get a working isolated PHAR; just for the merge as you don't want something too ugly/brittle merged.

For the rest sorry I'm not aware of any progress, maybe @kambo-1st was a bit busy last week :)

kambo-1st:
@theofidry is right, there aren’t any blocking issues in PHP-Scoper for building working phar file (at least at this moment), but I indeed not happy about “proof of concept” solution proposed in this pull request. From my point of view  humbug/php-scoper#192 will be need for the final version. In the meantime, I will prepare some e2e tests for the phar file, there is still a huge probability of some other issues.

sebastianbergmann:
@kambo-1st Thank you. I agree that we should wait for a release that includes https://github.com/humbug/php-scoper/issues/192.

theofidry:
Whilst humbug/php-scoper#192 is required for the final version, @kambo-1st found an acceptable workaround for now. So I would suggest to keep going on the progress as I suspect there is a lot more to do still

build/scoper.inc.php
```diff
@@ -0,0 +1,194 @@
...
+$whitelistClasses = [
+    'PHPUnit\Exception',
+    'PHPUnit\Framework\*',
+    'PHPUnit_Framework_MockObject_MockObject',
+    'SebastianBergmann\CodeCoverage\*',
+    'PharIo\*',
+    'PHP_Token',
+    'PHP_TokenWithScope',
+    'PHP_TokenWithScopeAndVisibility',
+    'PHP_Token_ABSTRACT',
```
theofidry: are those classes or constants?

kambo-1st: They are classes without any namespace.

build.xml
```diff
@@ -50,6 +50,11 @@
             <arg value="phpunit/php-invoker:^2.0"/>
         </exec>
 
+        <exec executable="${basedir}/build/tools/composer" taskname="composer">
```
sebastianbergmann: @theofidry Is php-scoper available as a PHAR? If so, @kambo-1st, I would prefer to have that PHAR in `build/tools` and used here.

theofidry: It should. There was a problem with the last version so it didn't get a PHAR but that shouldn't happen anymore.

However I would recommend to switch to a PHAR at the last moment since it makes it much harder to debug/fix in the meantime.

build/scoper.inc.php
sebastianbergmann: `php-token-stream` unfortunately does not use namespaces.

@theofidry Would it be possible to add support for whitelisting classes using `PHP_Token_*` or something along those lines?

theofidry: But do they need to be whitelisted?

theofidry: @sebastianbergmann that could be done. I opened https://github.com/humbug/php-scoper/issues/227 for it

build.xml
sebastianbergmann: Sure, using Composer for the moment is fine until we are sure that 1) php-scoper can handle PHPUnit and 2) PHPUnit uses php-scoper correctly.

kambo-1st: I also prefer PHAR. But as @theofidry  mentioned it wasn't available for latest version. When it will be available, i will use it.

build/scoper.inc.php
kambo-1st: @theofidry PHP_Token_* classes must be whitelisted, because they are instantiated dynamically (https://github.com/sebastianbergmann/php-token-stream/blob/711ca0c13c66f6b66c2ecb586e56415815034330/src/Token/Stream.php#L182). Prefixing break this.

theofidry: Maybe this could be fixed by having a patcher for [this line](https://github.com/sebastianbergmann/php-token-stream/blob/711ca0c13c66f6b66c2ecb586e56415815034330/src/Token/Stream.php#L176) instead,

theofidry:
I'm trying to prepare 0.9 which should provide some nice features:

- Smarter function arguments scoping (https://github.com/humbug/php-scoper/pull/228)
- Whitelist global functions by default (https://github.com/humbug/php-scoper/pull/231)
- Support whitelisting of functions
- Whitelist classes based on a pattern (https://github.com/humbug/php-scoper/issues/227) (depends on the previous point)
- Whitelist files (https://github.com/humbug/php-scoper/issues/196)

Is there anything else that is blocking here or that could help?

sebastianbergmann:
@kambo-1st What is the status here? Can you answer @theofidry's question? Thanks!

sebastianbergmann:
Thank you, @theofidry. I hope to hear from @kambo-1st soon. Otherwise I'll have to look into this myself again.

kambo-1st:
Sorry for keep you waiting. I'll look into it this week.

kambo-1st:
Quick update, new features in scooper look promising, especially wildcards. At this moment I preparing some tests.

sebastianfeldmann:
I tried to follow up on this pull request and encountered an issue with php-scoper.
I opened an issue for it https://github.com/humbug/php-scoper/issues/261

The rest seems to work.

kambo-1st:
@sebastianfeldmann with latest version of the PHP-Scoper 0.9.1 i did not have any problem.

kambo-1st:
@sebastianfeldmann Interesting issue, it seems, that this does not have any effect on the builded phar file, at least on the first sight. This is because there is still a proper alias defininition on the end of the file.

```php
\class_alias('PHPUnit\\PHPUnit_Framework_MockObject_MockObject', 'PHPUnit_Framework_MockObject_MockObject', \false);
```

sebastianfeldmann:
For me executing the PHAR fails instantly because the Mock class can not be found

kambo-1st:
@sebastianfeldmann Interesting, on which project did you try it? In the original pull request one of the problematic projects was Templado (https://github.com/templado/engine), i try it and it works, but i am maybe missing something (i can imagine, that there is maybe issue with autoloader which is loading PHPUnit_Framework_MockObject_MockObject behind my back). By the way, i commit new changes, feel free to them.

sebastianfeldmann:
I’m trying to scope the `PHPUnit` PHAR.
So libs included in PHPUnit do not interfere with local libs while unit testing.

I’ll try to reproduce the problem in a smaller example or even a unit test as you suggested.

kambo-1st:
@sebastianfeldmann It's start to be little confusing here, i presume that your previous comment belongs to issue in php-scoper https://github.com/humbug/php-scoper/issues/261 not here. I was merely asking which unit tests failed, because i don't find out any issue (yet) with changes in this pull request.

I am trying to reproduce your issue with `PHPUnit_Framework_MockObject_MockObject`, but i am not able to do it. I build the phar file `ant phar` and then I try to run unit tests of project https://github.com/templado/engine with result phar file. Everything work as should.

sebastianfeldmann:
Ok, I can verify your current branch is working.
The thing is that meanwhile the `MockObject` stuff has been reintegrated into the PHPUnit main repository.
It seems, that `php-scoper` is doing the right thing in case of the mock stuff being an "external" package, and is handling it wrong if the Mock stuff is integrated in PHPUnit.

In order to get this merged you can try to rebase your changes on top of the current master branch.
If you get it working with the reintegrated version we can merge this and bring this massive enhancement to everybody.

sebastianbergmann:
@Any update on this? Thanks!

sebastianfeldmann:
I will test the new version and provide feedback of my findings

sebastianfeldmann:
I still get this error executing the scoped PHAR 

```
PHP Fatal error:  Interface 'PHPUnit_Framework_MockObject_MockObject' not found in phar:///phpunit/build/phpunit-7.5-g28af7d442.phar/phpunit/Framework/MockObject/ForwardCompatibility/MockObject.php on line 14
```

Now that I dug into this a bit deeper `php-scorer` seems to work fine with the latest bugfix.
What I am seeing now is a problem in `phpab`s PHAR creation.

The reason why the PHAR file crashes is `phpab` PHAR stub that just requires all files that should be included in the PHAR.
And there seems to be an issue with the order those files get required.

Have a look at `/phpunit/Framework/MockObject/MockObject.php` / `PHPUnit_Framework_MockObjeckt_MockObject`

In the current master version it gets required before the `/phpunit/Framework/MockObject/ForwardCompatibility/MockObject.php`

```php
...
require 'phar://phpunit-7.5-ga406c85c5.phar' . '/phpunit/Framework/MockObject/MockBuilder.php';
require 'phar://phpunit-7.5-ga406c85c5.phar' . '/phpunit/Framework/MockObject/MockMethod.php';
require 'phar://phpunit-7.5-ga406c85c5.phar' . '/phpunit/Framework/MockObject/MockMethodSet.php';
require 'phar://phpunit-7.5-ga406c85c5.phar' . '/phpunit/Framework/MockObject/MockObject.php';
require 'phar://phpunit-7.5-ga406c85c5.phar' . '/phpunit/Framework/MockObject/ForwardCompatibility/MockObject.php';
require 'phar://phpunit-7.5-ga406c85c5.phar' . '/phpunit/Framework/MockObject/Exception/RuntimeException.php';
...
```

If I create the PHAR with the scoped files the order is different
The legacy `MockObject` class is required after the compatibility class which causes the not found error.

```php
...
require 'phar://phpunit-7.5-g28af7d442.phar' . '/phpunit/Framework/MockObject/ForwardCompatibility/MockObject.php';
require 'phar://phpunit-7.5-g28af7d442.phar' . '/phpunit/Framework/MockObject/Exception/RuntimeException.php';
require 'phar://phpunit-7.5-g28af7d442.phar' . '/phpunit/Framework/MockObject/Stub.php';
...
require 'phar://phpunit-7.5-g28af7d442.phar' . '/phpunit/Framework/Warning.php';
require 'phar://phpunit-7.5-g28af7d442.phar' . '/phpunit/Framework/WarningTestCase.php';
require 'phar://phpunit-7.5-g28af7d442.phar' . '/phpunit/Framework/MockObject/MockObject.php';
...
```

Right now I have no clue why the order is changing or if there is a way to manipulate the order in any way. Or if this is just an error on my machine.

I pushed my latest changes to my fork to the `feature/php-scoper` branch so you can check it yourself.

theofidry:
@sebastianfeldmann could you confirm that the scoped code (i.e. the code not bundled in the PHAR) is working fine?

Making the scoped work in the PHAR is another problem, but it's best to treat them separately to pinpoint what the issue is. If it's just the PHAR, rather than the order of files I would suspect the autoload file used. PHPScoper generates that `scoper-autoload.php` file which needs to be used and maybe the stub used uses the old Composer `autoload.php`.

sebastianbergmann:
@theseer should be able to shed some light on the issue at hand.

sebastianbergmann:
It might be worth mentioning that the interface in question, `PHPUnit_Framework_MockObject_MockObject`, is scheduled for removal. This is tracked in #3123.

sebastianbergmann:
@sebastianfeldmann Thank you for your ongoing work on this important topic, it is much appreciated.

sebastianfeldmann:
So far it looks like `php-scoper` works fine now. I can't be a 100% sure because I still can not run the PHAR file, but so far the scoped classes are looking good and the whitelisting works as expected.

To the PHAR autoloading problem.

I tried to change the stub from
```php
...
___FILELIST___
...
```
to a more dynamic version I copied from a default template provided by @theseer
```php
...
spl_autoload_register(
    function($class) {
        static $classes = null;
        if ($classes === null) {
            $classes = array(
                ___CLASSLIST___
            );
        }
        $cn = strtolower($class);
        if (isset($classes[$cn])) {
            require 'phar://___PHAR___' . $classes[$cn];
        }
    }
);
...
```

But the ___CLASSLIST___ placeholder did not get replaced with a list of all classes as expected.
It still remains in the final PHAR stub as `___CLASSLIST___` which causes an error.

sebastianbergmann:
Just changing the template will not be enough, you also need to remove the `--static` option from the invocation.

sebastianfeldmann:
That did the trick.
I removed the `--static` fom the invocation and now the PHAR and the autoloading works.

And now we are back to square one. 
```
Fatal error: Uncaught Error: Class 'PHP_Token_OPEN_TAG' not found in phar:///phpunit/php-token-stream/Token/Stream.php on line 131
```
In the `php-scoper` configuration I whitelisted `PHP_TOKEN_*`
```php
<?php
declare(strict_types=1);
/*
 *
 *
 */
$whitelistClasses = [
    'PHPUnit\*',
    'SebastianBergmann\CodeCoverage\*',
    'PharIo\*',
    'PHP_Token_*',
];
return [
    'whitelist' => $whitelistClasses,
];
```
The resulting class looks like this.
```php
<?php
namespace PHPUnit;
class PHP_Token_Stream implements \ArrayAccess, \Countable, \SeekableIterator
{
...
}
\class_alias('PHPUnit\\PHP_Token_Stream', 'PHP_Token_Stream', \false);
```
As before I was expecting a not prefixed version because of the whitelist.
Prefixing the `Token` classes does not work because of dynamic object creation that is close to impossible to find and to manipulate for `php-scoper`.

Do I use the whitelist wrong or is there another reason those classes are getting still prefixed.

theseer:
Just for the record:

In `--static'-Mode, `phpab` doesn't have a classlist to replace since it's technically not creating an autoloader but a static require list. Hence, `___FILELIST___` only.


I'm not sure as to why @sebastianbergmann uses static lists for PHPUnit's phar but it might actually be faster to simply load everything known to be required anyway right away.

theseer:
On a second note: I don't understand why having a static list of files should be an issue? Even if the prefixing happens afterwards, that shouldn't affect the sort order or filenames - or shouldn't it?

I have to admit, I didn't actually look into it. Just wondering...

theofidry:
>Prefixing the Token classes does not work because of dynamic object creation that is close to impossible to find and to manipulate for php-scoper.

Unless you are whitelisting a whole namespace, PHPScoper will _always_ prefix it. However it will record that this class has been whitelisted and create an alias to point the non-prefixed class to the prefixed one. But for that mechanism to kick in you either need to load the `scoper-autoload.php` file or load _all_ the files which contains the `class_alias()` statements.

There's bound to be some dynamic code which is not prefixed as it should because PHP is what it is but that's where there is [patchers](https://github.com/humbug/php-scoper#patchers).

That said without a reproducer, I cannot really look at the code and tell if it's wrong or not. Also it's important to determine if the issue is about the scoped code or bundling the scoped code into the PHAR.

sebastianfeldmann:
Just to be clear what is happening

### PHAR creation
- copy all files that should be included in the phar into a `phar` directory
- using `php-scoper` to scope classes except of whitelisted ones into `phar-scoped` directory
- use `phpab` to create the PHAR including autoloader stub for it

Maybe using `scoper-autoload.php` would somehow fix the issue because it knows about all the classes and aliases. But I'm currently not seeing why it should be neccesary.

### PHPScoper 
@theofidry 
The `Token` classes are using pre PHP5 namespaces `PHP_Token_Stream` ... so all classes are in the global namespace.
If I whitelist a PHP5 namespace the like `PHPUnit\` the files don't get touched at all. No prefix no `class_alias`. I would expect the same behavior for pre PHP5 namespaces. So I would not need any aliasing just because the classes get prefixed even though I put them on the whitelist.

The expected behavior of my whiteliste
```php
$whitelistClasses = [
    'PHPUnit\*',
    'PHP_Token_*',
];
```
would be that classes in `PHPUnit\*` and classes like `PHP_Token_*` in the global namespace do not get touched at all.

Is there a problem not prefixing pre PHP5 namespace classes I'm not seeing?

So far I have not created an as small as possible reproducer. But I will do this next if you realy need it.

### PHP Autoloader
@theseer 
The order of the static file list changed if  `phpab` is used with the prefixed classes.
You can check it by running `ant phar` in the `master` and in the `feature/php-scoper` branch of https://github.com/sebastianfeldmann/phpunit

theofidry:
I’m actually not sure the PHP_Token whitelist you have works as expected.
I’ll try to check that ASAP.

However I think you can do without. Indeed by default now PHPScoper no
longer scopes the classes belonging to the global namespace. While some
classes may be unnecessarily whitelisted, given that those classes are
gonna be removed or namespaced in the future, maybe it is good enough.

If that’s still not working I will need a reproducer in any form to be able
to have a look and tell what is going on instead of doing guess work :)

theseer:
@sebastianfeldmann I can reproduce the difference in sort order. While I cannot really explain why they appear, they are logically grouped. Meaning, the blocks are topologically sorted in themselves and their position as a block within the final list should be irrelevant.

(Yes, I'm aware of the fact I wrote _should_ ;-) )

theseer:
One assumption for the difference in sort order is the generated prefix for the fully qualified classnames. Since the list of classes is internally sorted, having a different namespace (prefix) is likely to cause a different overall sort order.

Since that difference is of course not affecting topological sorting (as in, recursively sorting based upon dependencies) the "block" will pretty much look the same.

theofidry:
I think I found something @sebastianfeldmann. I'll formalise this as an e2e test in PHPScoper itself to guarantee its behaviour and keep you in the loop

theofidry:
Should be better with https://github.com/humbug/php-scoper/releases/tag/0.11.1

sebastianfeldmann:
I can't see any difference :(
```php
$whitelistClasses = [
    'PHPUnit\*',
    'SebastianBergmann\CodeCoverage\*',
    'PharIo\*',
    'PHP_Token*',
];
return [
    'whitelist' => $whitelistClasses,
];
```
and `php-scoper` still scopes `PHP_Token` and following.
```php
<?php
namespace PHPUnit;
/**
 * A PHP token.
 */
abstract class PHP_Token
{
...
}
\class_alias('PHPUnit\\PHP_Token', 'PHP_Token', \false);
```

I'm not sure we are looking for the same issue. Can we agree, that with the given whitelist this class should not be prefixed with the `PHPUnit` namespace?

theofidry:
Arg forgot to check if the whildcard works on global classes like that. That said I would argue you should be able to do without: the class aliasing should do the trick and doing without is really tricky

theofidry:
I've had a very quick look and will try to dig into it more, but from what I see so far:

- The whitelisting works including for the `PHP_Token` classes: I see the class alias statements and aliasing, however you were missing the "load `scoper-autoload.php`
- I'm really not sure to understand the current whitelisting configuration: with `'PHPUnit\*'` you are whitelisting the whole PHPUnit namespace including the runner and several utility classes. Shouldn't they be scoped as well and expose only the API used by the user?

theseer:
@theofidry Just out of curiosity: Why would, for the phar which uses a static require list, the `scoper-autoload.php` be needed? And even if it wouldn't be a static require list but an actual autoload being based on a classmap?

theofidry:
For classes if you were to require all the files it wouldn’t be an issue,
but it would still be for functions (cf the PHPScoper doc I can’t link it
atm).

That said for now I’m trying to make it work outside the PHAR first once
it’s done it will be easy to tell where the proble comes from.

theofidry:
@sebastianfeldmann I did a PR with some fixes. There is some stuff to not merge but I would keep them for now and remove them later when everything is working. My process:

- scope everything in a temporary directory (see the `compile`)
- go in that directory, dump the autoloader again and run the tests

I released a new version of PHP-Scoper to fix a bug related to the whitelisting feature. From what I see most of the remaining failures are due to class strings incorrectly being scoped or not scoped, if you check the diff of my PR you will see what I mean. There might be one or two failures which requires a big more of digging than that but that should already be a solid start.

Once this step is done, what is left is bundling it in the PHAR and ensuring the PHAR works as expected.

theofidry:
>Could you explain one more time why it is the case that if I put something on the whitelist it gets prefixed anyway?

I've added that doc entry for it, you might want to take a look: https://github.com/humbug/php-scoper/#implementation-insights

>I know there are class aliases and it should work, but why bother?

A few reasons:

- Autoloading: you cannot easily mix different namespaces unless you do a "classmap": all" kind of thing
- You would need to tell the content of the file beforehand as if there is multiple classes in the file and one of the class should not be prefixed, figuring out how to wrap the two in different namespaces is really hard

Whereas always prefixing and appending class aliases when appropriate is very straightforward.

>The whitelisted namespace PHPUnit doesn't get prefixed into PHPUnit\PHPUnit\* and aliased.

Because whitelisting a whole namespace is easier to achieve without breaking much stuff than selectively whitelisting some elements that way. But note that it has some limitations as well cf. [namespace whitelisting](https://github.com/humbug/php-scoper/#namespaces-whitelisting)

I understand the feeling that always prefixing and adding a class alias feels weird, but whitelisting some elements just didn't work well enough.

sebastianbergmann:
@sebastianfeldmann Not sure it matters, but `PHPUnit_Framework_MockObject_MockObject` is gone in `master` now.

theofidry:
Is it really an issue though? As I said before: before shopping the PHAR
you need to make sure the scoping is done right and from what I see in the
tests there is still 1 or 2 cases that should be inspected.

The scoper-autoload.php file is in vendor next to the Composer autoload.php

theseer:
For the record: `phpab` is - afaik - only used when a phar is being build.

On top of that, the generated "autoload" is a static require list. That means, all files will be required even before one single line of other code is being executed. Any `class_alias` definition in the required file would thus also be executed right away.

Or in other words: There's not autoloading in phar mode. Given that `phpab` is only used when and for building the phar, I fail to see how `phpab` could cause any issues?

Or am I missing the point? ;)

theofidry:
@theseer that was my understanding and it should make it easy. I was just pointing out that getting the scoping working outside the PHAR is needed first as it makes it easier to know where a potential issue might come from as well as allows to easily test the scoped code by running the (non-scoped) test suite against it

sebastianfeldmann:
@theseer the problem is/was, that `phpab` sorted the static file list wrong. In a way that a class that depended on another class was loaded first and the require statement failed because of a missing dependency. That's why I switched to a dynamic autoloading.
The cause for this was the prefixed `PHPUnit_Framework_MockObject_MockObject`, but since it is gone in master I'll go back to the static file list and see if that is working again.

@theofidry currently I'm creating the scoped files with this ant task
```xml
    <exec executable="${basedir}/build/tools/php-scoper" taskname="php-scoper">
        <arg value="add-prefix" />
        <arg value="--no-ansi" />
        <arg value="--force" />
        <arg value="--config" />
        <arg path="${basedir}/build/scoper.inc.php" />
        <arg value="--no-interaction" />
        <arg value="--stop-on-failure" />
        <arg value="--output-dir" />
        <arg path="${basedir}/build/phar-scoped" />
        <arg value="--prefix" />
        <arg value="PHPUnit" />
        <arg path="${basedir}/build/phar" />
    </exec>
```

    ./build/tools/php-scoper add-prefix --no-ansi --force --config ./build/scoper.inc.php --no-interaction --stop-on-failure --output-dir ./build/phar-scoped --prefix PHPUnit ./build/phar

PHAR generation for PHPUnit works like this:
1. Copy stuff that should be included in the PHAR from `vendor` to `build/phar` 
2. Add `manifest` file
3. Create PHAR including autoloading with `phpab` with the temp `build/phar` directory
4. Delete the `build-pear` directory

Now there is a **2.a** where everything in `build/phar` is getting scoped into `build/phar-scoped` by `php-scoper` and `phpab` is then using the `build/phar-scoped` dir to create the PHAR

With the command above no `scoped-autoload.php` seems to be generated.

theseer:
> @theseer the problem is/was, that `phpab` sorted the static file list wrong. In a way that a class that depended on another class was loaded first and the require statement failed because of a missing dependency. That's why I switched to a dynamic autoloading.
> The cause for this was the prefixed `PHPUnit_Framework_MockObject_MockObject`, but since it is gone in master I'll go back to the static file list and see if that is working again.

I'm not sure I understand how that would be logically possible. `phpab` scans a given directory for classes, interfaces and traits. It resolves dependencies and uses that for sorting the list (aka, performs a topological sort) before writing the require file. Unless I have a bug in the sorting code, there is no way processing the generated list top to bottom can cause undefined class errors.

As with autoloading this of course is a all or nothing approach. If not all units are defined and found at build time, the list will be incomplete. 

I guess, I'm just trying to figure out whether or not there's something I can change or even have to fix in `phpab` ;)

glensc:
@theseer I can see one scenario when static sorting fails: in the case of circular dependencies. but that would require multiple classes defined in the same file. not sure if that's the case here.

theseer:
> @theseer I can see one scenario when static sorting fails: in the case of circular dependencies. but that would require multiple classes defined in the same file. not sure if that's the case here.

Afaik, PHP can't do resolve circular dependencies on this level.

theofidry:
>It resolves dependencies and uses that for sorting the list (aka, performs a topological sort) before writing the require file. Unless I have a bug in the sorting code, there is no way processing the generated list top to bottom can cause undefined class errors.

Does this topological sorting handles class aliasing?

theseer:
> > It resolves dependencies and uses that for sorting the list (aka, performs a topological sort) before writing the require file. Unless I have a bug in the sorting code, there is no way processing the generated list top to bottom can cause undefined class errors.
> 
> Does this topological sorting handles class aliasing?

No. Good point.

I guess that would require finding and "understanding" the class_alias function call. Currently the parser looks up class/interface/trait definitions as well as their extends, implements and uses.

That's a problem, or is it?

theofidry:
I think that could be the source of the issue since almost all whitelisted code get aliased.

But since you are scanning the files first and operate a sorting, maybe adding support for the class aliasing wouldn't be too big of a deal?

sebastianfeldmann:
After being called out yesterday by @sebastianbergmann at the php user group Frankfurt I immediately after arriving at home sat down and made some checks I had only layed out in my head so far.

So I rebased to the latest master.
Moved the latest version of `php-scoper` to the tools directory
Now on the `ant phar` build a `build/phar-scoped` directory gets created with all the contents of the previously prepared `build/phar` directory.

I verified it by running the unit tests with a bootstrap file  that loads the PHP files from `build/phar-scoped` instead of `src` and the tests all passed.

I used to created PHAR to run all phpunit 8 ready test suites on my machine (not much but some) and everything worked.

I'm still hesitant to say it works,  but for me it certainly looks that way.

You can check by creating your own scoped phar by:

    $ git clone https://github.com/sebastianfeldmann/phpunit.git
    $ git checkout feature/php-scoper
    $ ant phar

sebastianbergmann:
It was not my intention to "call you out" and "build up pressure". Quite the opposite: I wanted to you credit for the work you did. Sorry if I messed that up.

I think it is time we closed this pull request and @sebastianfeldmann sends a new pull request with the current state of his work. This would make it a lot easier for me to work on this.

sebastianfeldmann:
Don't worry, you messed nothing up :)
I got your intention, but in order to deserve the credit...
I will open a new pull request with the current state.

sebastianbergmann:
Superseded by #3575.

PR #3087, opened 2018-04-09
Allow returning null result

greg0ire:
This can and does happen with Symfony's phpunit-bridge listener. Apart
for that, this property is not guaranteed to be not null (there is a
setter and no constructor requirement), so the return type hint should
reflect that.
See https://github.com/symfony/symfony/issues/26861

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -1033,7 +1033,7 @@ abstract class TestCase extends Assert implements Test, SelfDescribing
         $this->outputCallback = $callback;
     }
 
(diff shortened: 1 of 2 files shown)


greg0ire:
Here is where the call happens: https://github.com/symfony/symfony/blob/9a999553a5495a23ea3be4271ecdbe89a9c64bba/src/Symfony/Bridge/PhpUnit/Legacy/SymfonyTestsListenerTrait.php#L182

sebastianbergmann:
Merged manually, thanks.

Issue #2660, opened 2017-04-26
Request for ability to order/filter test suites and/or tests without changes to underlying test code.


padraic:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 7.0.0 (version designated to restrict Listeners)
| PHP version         | 7.x.x
| Installation Method | Composer / PHAR

In light of #2477 which would remove the current ability to (sneakily, I know!) manipulate test suites within Listeners, this is an issue to request some alternative means of injecting a class with write access. Currently this can be provided within test classes, however there may be some advantage to doing this indirectly as Listeners currently allow.

To explain my own specific needs, a few of us are bringing Humbug back to life, as a mutation testing framework. This involves generating small file changes, and running PHPUnit. We run phpunit for every single change...potentially hundreds of times. To improve performance, we take a number of basic steps described here: https://github.com/humbug/humbug#performance which can be summarised as:

1. Only run tests on covered lines.
2. Eliminate tests not applicable to a line.
3. Order tests fastest-first to hopefully reach a failure quickly.
4. Stop on first failure/error/exception (serves no purpose to continue after one).

Technically, we only order (or filter) test suites - I'll need to double check if that was a limitation of using Listeners, or of PHPUnit in general, or a limitation in our analysis of code coverage data. We do not order individual tests within a parent suite (unless the children are themselves suites) which would actually be the preferred approach.

The performance benefits are central to the framework being useful, as we are not running all tests all the time hundreds of times.

TL;DR - One way of summarising this is getting an array of tests (which would be multidimensional, grouping tests by suite), manipulating that array, injecting a revised array, all while assuming execution order is simply determined by array index. It could perhaps be done by exporting/importing a name map where avoiding any deeper write level interference with actual objects is preferable.

epdenouden:
@padraic working on something that might be of interest to you, too: https://github.com/epdenouden/phpunit/tree/breakfast is a change for PHPUnit that will allow tests to run tests in _defects-first_ mode. It is a followup of the #3092 test reordering feature.

mcfedr:
With the upcoming removal of TestListener it wont be possible to do the filtering that I am currently using `startTestSuite` to do, will there be a replacement for this? I really want to inject a `RecursiveFilterIterator` into the filters that are called

epdenouden:
[...]

PR #3092, opened 2018-04-12
Add functionality to test dependencies by changing test running order

epdenouden:
|Note: updated documentation|
|:---|
|There is a wiki page with more [documentation with examples](https://github.com/epdenouden/phpunit/wiki/PHPUnit-test-running-order-management) while the official docs are updated.|

# Inspiration
Are your unit tests truly independent units?

In the past I have often been bitten by test suites that do not test what they claim to test. A common reason is hidden dependencies in the runtime context. Worst case this has caused applications to fail when deployed in production environments.

Changing the order of unit tests turns out to be a simple and effective way of bringing these assumptions out in the open.

Thanks to @crazybus and @geogdog for providing real-world scenarios and feedback during development.

# Summary of functionality
The test ordering functionality can be configured using command line parameters:
- `--reverse-order` run tests in *reverse order*
- `--random-order` run tests in *random order*
- `--random-order-seed=<number>` fix the random order, allowing random runs to be reproduced
- `--ignore-dependencies` disable the bundled dependency resolver

# Design decisions
## Low impact, high performance
The feature will only wake up when there is work to do. The iterative dependency resolver is designed to do as little work as possible, using few iterations and minimal memory. It uses `@depends` annotations to keep as many dependent tests as possible in the required order.

No unnecessary extra output is generated. Only the randomizer outputs the *random seed* using the default results printer.

## Clean abstraction
The reordering is built into the main test runner and only changes the order of the tests of each test suite. Other parts of PHPUnit are not affected, with the exception of a bit of configuration handling.

## Changes to the built-in testsuite
- added a fourth `MultiDependencyTest` and updated existing tests to match
- added new TextUI tests for `--reverse-order`, `--random-order` and `--ignore-dependencies`
- added `@depends` for testGlobalsBackupPre() to PHPUnit's own testsuite

# Future work
- update the main PHPUnit documentation to reflect the changes in (optional) order and dependency handling
- add configuration options for `phpunit.xml` in addition to the command line flags
- add options and/or annotations for finer grained reordering config (e.g. "never reorder this suite")


tests/Util/ConfigurationTest.php
```diff
@@ -330,11 +329,12 @@ public function testPHPConfigurationIsHandledCorrectly(): void
...
     {
-        $_ENV['foo'] = false;
+        $fooValue    = \uniqid();
+        $_ENV['foo'] = $fooValue;
         $this->configuration->handlePHPConfiguration();
 
-        $this->assertFalse($_ENV['foo']);
-        $this->assertEquals(1, \getenv('foo'));
+        $this->assertEquals($fooValue, $_ENV['foo']);
+//        $this->assertEquals(false, \getenv('foo'));
```
sebastianbergmann: Is this commented out intentionally?

src/Framework/TestCase.php
```diff
@@ -1002,6 +1017,49 @@ public function setPreserveGlobalState(bool $preserveGlobalState): void
         $this->preserveGlobalState = $preserveGlobalState;
     }
 
+    /**
+     * @param string $order
+     */
+    public function setTestRunningOrder($order): void
```
sebastianbergmann: Please use scalar type declarations when possible (like here). Also please do not use `@param` annotations etc. when they provide no additional information.

src/Framework/TestSuiteIterator.php
```diff
@@ -91,4 +123,72 @@ public function hasChildren(): bool
     {
         return $this->tests[$this->position] instanceof TestSuite;
     }
+
+    /**
+     * Reorder Tests within a TestCase in such a way as to resolve as many dependencies as possible.
+     * The algorithm will leave the tests in original running order when it can.
```
sebastianbergmann: What about cross-class dependencies? They are possible, though I rarely see them "in the wild".

sebastianbergmann:
Thank you, @epdenouden, for working on this. I will look into this more deeply as soon as I can. I hope we can get this into PHPUnit 7.2.

src/Framework/TestCase.php
epdenouden: Fixed

sebastianbergmann:
@epdenouden I have moved the current state of this to a new branch: https://github.com/sebastianbergmann/phpunit/tree/test-reorder.

Please rebase this pull request onto that branch and target that branch for future changes. Thanks!

sebastianbergmann:
I am wondering whether it would make sense to move the reorder logic from `TestSuiteIterator` into separate classes, one for each type of reorder (random, dependencies, ...).

Pseudo Code:
```php
<?php
class DependencyResolver extends ArrayIterator
{
    public function __construct(TestSuiteIterator $tests)
    {
        $tests = iterator_to_array($tests);

        // reorder $tests

        parent::__construct($tests);
    }
}
```

sebastianbergmann:
I am wondering whether it would make sense to add an extension hook for test reordering. This would make @padraic and everyone interested in #2660 happy.

sebastianbergmann:
@lstrojny This is probably interesting to you, too, considering "clever and smart".

epdenouden:
@sebastianbergmann Thanks for the quick replies! I will rebase and continue making the basic version clean and stable for 7.2. Good ideas for making this cleaner and more extensible. I am up for doing more coding for that.

Re: the clever and fast logic: this project is actually the groundwork for something that we call the _Breakfast_ extension, it automatically sorts the tests to run skipped, failed, etc first, but keeps their dependencies working. It also measures test and suite timings to make better sorting predictions.

sebastianbergmann:
The "breakfast" thing sounds interesting. Lets get this sorted first, then have a look at that. Maybe it can also be integrated into the "core", maybe we can put it into an extension that uses the new, yet-to-be-added hook.

sebastianbergmann:
@epdenouden When tests are executed in {random|reverse} order, are dependencies expressed using `@depends` taken into account? My guess is they are not (but should be) and this is why those two tests of PHPUnit's own test suite cause problems when run in reverse order.

epdenouden:
Moinmoin @sebastianbergmann ! Yes, the `@depends` annotation should be kept in mind, they are 'food' for the dependency resolver. I am looking at the getenv() vs $_ENV issue at the moment, there the resolver becomes very visible by reordering `TestCaseTest::testStaticAttributesBackupPre()` to run before `testStaticAttributesBackupPost` and `testGlobalsBackupPost`

epdenouden:
For example, running it like

```
./phpunit --reverse-order
[...output snipped...]
4) PHPUnit\Framework\TestCaseTest::testStaticAttributesBackupPost
This test depends on "PHPUnit\Framework\TestCaseTest::testStaticAttributesBackupPre" to pass.

/Users/ewout/proj/phpunit-new-order/src/Framework/TestCase.php:1795
/Users/ewout/proj/phpunit-new-order/src/Framework/TestCase.php:695
/Users/ewout/proj/phpunit-new-order/src/Framework/TestSuite.php:788
/Users/ewout/proj/phpunit-new-order/src/Framework/TestSuite.php:788
/Users/ewout/proj/phpunit-new-order/src/Framework/TestSuite.php:788
/Users/ewout/proj/phpunit-new-order/src/TextUI/TestRunner.php:565
/Users/ewout/proj/phpunit-new-order/src/TextUI/Command.php:202
/Users/ewout/proj/phpunit-new-order/src/TextUI/Command.php:155

5) PHPUnit\Framework\TestCaseTest::testGlobalsBackupPost
This test depends on "PHPUnit\Framework\TestCaseTest::testGlobalsBackupPre" to pass.

/Users/ewout/proj/phpunit-new-order/src/Framework/TestCase.php:1795
/Users/ewout/proj/phpunit-new-order/src/Framework/TestCase.php:695
/Users/ewout/proj/phpunit-new-order/src/Framework/TestSuite.php:788
/Users/ewout/proj/phpunit-new-order/src/Framework/TestSuite.php:788
/Users/ewout/proj/phpunit-new-order/src/Framework/TestSuite.php:788
/Users/ewout/proj/phpunit-new-order/src/TextUI/TestRunner.php:565
/Users/ewout/proj/phpunit-new-order/src/TextUI/Command.php:202
/Users/ewout/proj/phpunit-new-order/src/TextUI/Command.php:155

FAILURES!
Tests: 1657, Assertions: 2898, Failures: 2, Skipped: 5.
```

whereas

```
./phpunit --reverse-order --reorder-dependencies
[...output snipped...]
FAILURES!
Tests: 1657, Assertions: 2912, Failures: 2, Skipped: 3.
```

epdenouden:
@sebastianbergmann The TestSuiteSorter looks nice and tidy like this :)

sebastianbergmann:
* `TestSuiteIterator` must not know anything about sorting/reordering, therefore it must not know `TestSuiteSorter`
* `TestSuite` does not need to know anything about sorting/reordering once `TestSuiteIterator` does not know `TestSuiteSorter` anymore
* Tests are sorted (in the runner) before the first test is executed
* This is also the point where we can let extensions hook into the sorting/reordering

epdenouden:
@sebastianbergmann Excellent points, thanks for taking the time to review and even cleanup the state of things overnight. I'll rebase and continue where I left off yesterday evening.

epdenouden:
@sebastianbergmann Fresh look this morning, bit of clean up and done some more testing and performance timings. Looks stable and efficient. Anything missing to target [milestone 7.2](https://github.com/sebastianbergmann/phpunit/milestone/24)?

It has been a good learning experiment. Thanks for the all the time you spend on building and maintaining one of my main tools. :-)

epdenouden:
I'll focus on the original run-failed-and-skipped-first project for which this was the first building block. For Breakfast I'll look through my notes for ideas that have been left by the wayside, such as config file parameters, optional extra output (for running order, timings, random seed in other formats, etc etc). It can now simply become an extra sorting-method.

sebastianbergmann:
@epdenouden Thank you for your ongoing work on this. I plan to look at (and hopefully merge) this over the next weekend.

epdenouden:
@sebastianbergmann Thanks! Will check my notes for any open issues before then. I've put together a description and examples to try to see the feature at work: https://github.com/epdenouden/phpunit/wiki/PHPUnit-test-running-order-management

sebastianbergmann:
This branch cannot be rebased due to conflicts. Can you please fix the conflicts? Thanks!

epdenouden:
Yes, will do that when I get back from lunch

epdenouden:
Github's error message isn't very helpful unfortunately, what conflicts? :) When comparing my `new-order` branch to both your `test-reorder` and `master` Github seems happy to do it. For `test-reorder` I got this message:

image

I did see two conflicts when merging `7.1` in outside this PR it seems, one in the PHPStan config and one `Runner/Version.php`. Hope I have fixed them correctly; running a few test scenarios and then I'll push the merge changes from 7.1.5 here.

epdenouden:
Okidoki. I'll be heading for some wifi and coffee while the checks are running :)

sebastianbergmann:
Strange, GitHub still tells me "This branch cannot be rebased due to conflicts". I was able to rebase it manually locally, though.

epdenouden:
Not sure what to do about the Travis failing build. The change to PHPStan level 2 in f59852f593cdcce49e4b799c2a5b73233dacfc01 causes it to fail.

epdenouden:
Found the missing piece of PHPStan configuration that went missing while merging a25126a852f8476b5674e417bb2bd1e013d4b4af. PHPStan works as before.

src/Framework/TestSuiteIterator.php
```diff
@@ -14,21 +14,18 @@
 /**
  * Iterator for test suites.
  */
-class TestSuiteIterator implements RecursiveIterator
+final class TestSuiteIterator implements RecursiveIterator
```
keradus: btw, this is a  BC breaker for all of us that were using this class. (class is not marked as `@internal`)

src/Util/TestDox/TestResult.php
```diff
@@ -9,147 +9,19 @@
  */
 namespace PHPUnit\Util\TestDox;
 
-final class TestResult
```
keradus: btw, this is a BC breaker for all of us that were using this class. (class is not marked as `@internal`)
making it non-class (->interface) is even bigger change

keradus (commented):
great work in general!

let me just point out 2 bc breakes for edge-case users

sebastianbergmann:
@keradus PHPUnit still lacks a clear statement of what is considered part of its public API. I hope to have this in place "soon", at the latest in time for PHPUnit 8.

Off the top of my head, I would say that everything internal with the exception of `Test`, `TestCase`, `TestListener`, etc.

keradus:
I know there is no clear statement on that, but especially as it's not documented, all is public and non-internal :(

src/Framework/TestSuiteIterator.php
epdenouden: In the end I didn't need to change many of the existing internals as the code moved to TestSuiteSorter. Extensions support is coming, I tried that first :)

keradus:
just wondering... can we make `random-order` a default one please?
or at least, allow to set it via config file

sebastianbergmann:
-1 for default (because of `--testdox`, for instance)
+0 for configuration

Come to think of it, we should probably disable the dependency resolution by default in PHPUnit 7.2 and turn it on by default in PHPUnit 8.

keradus:
value of this PR is nice, yet requires manual work that one always remember to use that flag.
with config file, it would be way easier

sebastianbergmann:
@keradus I have opened #3131 for this and will (hopefully) implement it tomorrow.

keradus:
Thanks for accepting the request :+1:
no need to treat this one as prio, no rush on it ;)

epdenouden:
Indeed thanks for merging it in. I will continue on the followup feature for rerunning failures first now

epdenouden:
@keradus having reordering as a default would be fun, but it would spook a lot of people that run phpunit in "any 7.x" version will do. Since testing frameworks are important for creating trust that is not so great :)

keradus:
> Indeed thanks for merging it in. I will continue on the followup feature for rerunning failures first now

Sounds like https://github.com/fiunchinho/phpunit-randomizer is obsolete, now it's time for
https://github.com/lstrojny/phpunit-clever-and-smart ? :)

> having reordering as a default ...

That's why I suggested to have at least config option, thus one don't need to remember about option switch every single time.
On the other hand, if I have tests that are always passing, and then we would detect that `assertSame` doesnt work for `string` for some weird reason, I do expect `assertSame` to be fixed and yes, my tests would start failing. It is not that my test was working and suddenly it stopped. It's that my tests were not able to detect real issue in my application, and after update of testing framework they are. And it's super cool to know that my src has bugs.

PR #3098, opened 2018-04-17
DX: PHP CS Fixer default finder is looking for .php by default, no need to claim it again

keradus:
As we are using default finder and then revoke .phpt, and not using raw finder and then filter to .php,
we don't need to filter to .php again, as default finder already does that


sebastianbergmann:
Merged manually, thanks.

PR #3110, opened 2018-04-29
avoid web execute

melbahja:
You maybe see this:
https://pbs.twimg.com/media/DZayD9LW4AAC6kB.jpg
same websites have vendor at document root and autoload with dev packages
this PR to make this file for CLI only


sebastianbergmann:

If you upload PHPUnit onto a webserver then your deployment process is broken.

PR #3113, opened 2018-05-05
Add the ability to specify environment variable requirements

maciejtarnowski:
Hello!

In one of the projects, I'm using PHPUnit to run tests that are partially configured using environment variables. Due to that, some of them need to be skipped based on certain variable values. In this PR I've added a new feature to the `@requires` annotation that allows you to specify environment variable name and regular expression that it has to match. For example:
```php
/**
 * @requires env SOME_ENV_VARIABLE some_value
 */
```
or:
```php
/**
 * @requires env STAGE ^dev|ci$
 */
```

The additional `@requires env` annotation would save us the need to add if statement in the test body and skip tests "manually".

Please let me know what you think.

### DIFF
--- a/tests/_files/RequirementsClassDocBlockTest.php
+++ b/tests/_files/RequirementsClassDocBlockTest.php
@@ -14,6 +14,7 @@
  * @requires OS Linux
  * @requires function testFuncClass
  * @requires extension testExtClass
+ * @requires env TEST_ENV dev|ci
  */
 class RequirementsClassDocBlockTest
 {
--- a/src/Util/Test.php
+++ b/src/Util/Test.php
@@ -88,6 +88,11 @@ final class Test
     /**
      * @var string
      */
+    private const REGEX_REQUIRES_ENV = '/@requires\s+(?P<name>env)\s+(?P<var>([^ ]+)){1}\s*(?P<value>([^ ]+?))[ \t]*\r?$/m';
+
+    /**
+     * @var string
+     */
     private const REGEX_REQUIRES = '/@requires\s+(?P<name>function|extension)\s+(?P<value>([^ ]+?))\s*(?P<operator>[<>=!]{0,2})\s*(?P<version>[\d\.-]+[\d\.]?)?[ \t]*\r?$/m';
 
     /**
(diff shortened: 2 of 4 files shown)


epdenouden:
Ping @maciejtarnowski, could you look into the conflict please?

This looks like a useful feature. Using ENV variable to control deployments is quite common, I think this would make a nice addition.

tests/_files/RequirementsClassDocBlockTest.php
```diff
@@ -14,6 +14,7 @@
  * @requires OS Linux
  * @requires function testFuncClass
  * @requires extension testExtClass
+ * @requires env TEST_ENV dev|ci
```
epdenouden: I would much prefer to see the common regex slashes so it is more obvious what it is. Now this is hidden in the `/%s/i` in `src/Util/Test.php`

maciejtarnowski:
@epdenouden Happily resolved conflicts and rebased the branch, thank you for looking into it.

tests/_files/RequirementsClassDocBlockTest.php
maciejtarnowski: I agree that it would be better, but OS requirement is also a regex and it hides it the same way. IMHO it's better to keep it consistent and maybe change both of them in some future major release. Let me know what you think.

epdenouden:
Goodmorning @maciejtarnowski! Great to see conflicts fixed so quickly. :)

I had not realized the OS is also a regex, thanks for pointing that out. Functionality and code look fine, I'd be in favour of having this included in PHPUnit. Many infra and ops setups use environment variables in their deployments.

@sebastianbergmann What's your opinion on including this?

sebastianbergmann:
I am not convinced that this adds value (or rather, that the problem it solves is one that should be solved).

epdenouden:
Hm, yes. After some more pondering I would not include it in the core of PHPUnit. I suspect that there would be lots of quickfixes done by users to simple skip tests based on ENV.

Your latter argument resonates with my experience debugging systems. Relying on `$ENV` makes the dependencies so hard to spot. An added complication is the way `putenv`/`getenv` are not a mirror of `$_ENV` in PHP, but rather a one-way street to `$ENV`.

@maciejtarnowski for some context, yesterday I (finally!) plugged a environment-dependency that had been bugging me for months: https://github.com/sebastianbergmann/phpunit/commit/2bf74c9b5641f50c010fdfc7282e8b65e750bbee

sebastianbergmann:

PR #3115, opened 2018-05-07
Feature/method covers overrides class covers nothing

Danack:
Implements https://github.com/sebastianbergmann/php-code-coverage/issues/611 aka allows a 'method covers' annotation to override the 'class coversNothing' annotation.

Reasonably chance this might need to wait until PHPUnit 7.2.0


keradus:
Really glad to see this, thanks @Danack, nice work here !

Issue #3107, opened 2018-04-27
CliTestDoxPrinter::addError() cannot handle errors in setUpBeforeClass()


stanolacko:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 7.1.2
| PHP version         | 7.2.2
| Installation Method | PHAR

If you made error in TestCase, like calling non_existing_function(), you get error, which tells you nothing. In this case, I'm not able to find, that problem is non_existing_function().

class CombiPneuTestSnazzyBee extends PHPUnit\Framework\TestCase {
	
	public static function setUpBeforeClass() {
		non_existing_function();
	}
}

Produce:
PHP Fatal error:  Uncaught Error: Call to a member function fail() on null in phar://REAL_PATH/phpunit.phar/phpunit/Util/TestDox/CliTestDoxPrinter.php:116
Stack trace:
#0 phar://REAL_PATH/phpunit.phar/phpunit/Framework/TestResult.php(265): PHPUnit\Util\TestDox\CliTestDoxPrinter->addError(Object(PHPUnit\Framework\TestSuite), Object(PHPUnit\Framework\ExceptionWrapper), 0)
#1 phar://REAL_PATH/phpunit.phar/phpunit/Framework/TestSuite.php(754): PHPUnit\Framework\TestResult->addError(Object(PHPUnit\Framework\TestSuite), Object(PHPUnit\Framework\ExceptionWrapper), 0)
#2 phar://REAL_PATH/phpunit.phar/phpunit/TextUI/TestRunner.php(529): PHPUnit\Framework\TestSuite->run(Object(PHPUnit\Framework\TestResult))
#3 phar://REAL_PATH/phpunit.phar/phpunit/TextUI/Command.php(198): PHPUnit\TextUI\TestRunner->doRun(Object(PHPUnit\Framework\TestSuite), Array, true)
#4 phar://REAL_PATH/phpunit.phar/phpunit/TextUI/Command.php( in phar://REAL_PATH/phpunit.phar/phpunit/Util/TestDox/CliTestDoxPrinter.php on line 116

Fatal error: Uncaught Error: Call to a member function fail() on null in phar://REAL_PATH/phpunit.phar/phpunit/Util/TestDox/CliTestDoxPrinter.php:116
Stack trace:
[...]

PR #3116, opened 2018-05-09
Re-solve Regression 3107 in a much cleaner way

rpkamp:
Thanks to @samlitowitz who created a PR for rpkamp/fancy-testdox-printer and took this approach to solving regression #3107, which is much simpler than the approach I originally took :+1:


sebastianbergmann:
Should this go into PHPUnit 7.1.Z or into PHPUnit 7.2.0?

rpkamp:
7.1.Z is fine; it's not a new feature, just a refactoring of how a problem was solved.
Output is a bit different, but not changes enough to warrant a minor bump imo.

sebastianbergmann:
Cherry-picked into `7.1`.

PR #3117, opened 2018-05-09
Use symfony/process to run isolated tests

### LINKED ISSUE #1342: Replace PHP process isolation code with Symfony Process component

rpkamp:
Fixes #1342

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -43,7 +43,8 @@
         "sebastian/global-state": "^2.0",
         "sebastian/object-enumerator": "^3.0.3",
         "sebastian/resource-operations": "^1.0",
-        "sebastian/version": "^2.0.1"
+        "sebastian/version": "^2.0.1",
+        "symfony/process": "~3.4|~4.0"
     },
     "require-dev": {
         "ext-PDO": "*"
--- a/src/Util/PHP/AbstractPhpProcess.php
+++ b/src/Util/PHP/AbstractPhpProcess.php
@@ -174,10 +174,15 @@ abstract class AbstractPhpProcess
     /**
      * Returns the command based into the configurations.
      */
-    public function getCommand(array $settings, string $file = null): string
+    public function getCommand(array $settings, string $file = null): array
     {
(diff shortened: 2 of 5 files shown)


sebastianbergmann:
Thank you for working on this, @rpkamp. However, this will have to wait until #2015 is implemented.

rpkamp:
That's okay. Tests for windows are still failing anyway. Will look at that later.

rpkamp:
I've checked to see if I could fix the errors in AppVeyor and ran into the following:

`C:\phpunit\tests\TextUI\phpt-args.phpt`:

The problem here is that the argument supplied in `ARGS` in the `.phpt` file is passed with a trailing `\n`, which doesn't seem to matter on *nix, but on windows, the supplied value is literally `help\n`, which isn't equal to `help`, so the test fails. I could change the test to `if ($argc > 0 && strpos($argv[1], 'help') === 0)` and then it would succeed, but it is a break in functionality that others might run into as well. Not really sure how to correctly solve this without any BC breaks. Input would be appreciated.

`C:\phpunit\tests\Regression\GitHub\1348.phpt`

This one is even tougher, as `symfony/process` doesn't use regular STDOUT/STDERR on windows as there are several problematic bugs around it. Instead it redirects all OUT/ERR to files. This means that writing to STDOUT from the test file will _not_ be sent to the terminal, but to the file that STDOUT is redirected to instead. This also a BC break that I don't see a nice way of working around. Again, input would be appreciated.

rpkamp:
Replaced by https://github.com/sebastianbergmann/phpunit/pull/3117

Issue #3129, opened 2018-05-12
phpstan fails after merge of phpunit-mock-objects


sebastianbergmann:
```
 ------ ------------------------------------------------------------- 
  Line   tests/Framework/MockObject/Builder/InvocationMockerTest.php  
 ------ ------------------------------------------------------------- 
  25     Call to an undefined method                                  
         PHPUnit\Framework\MockObject\MockObject::foo().              
  38     Call to an undefined method                                  
         PHPUnit\Framework\MockObject\MockObject::foo().              
  39     Call to an undefined method                                  
         PHPUnit\Framework\MockObject\MockObject::foo().              
  40     Call to an undefined method                                  
         PHPUnit\Framework\MockObject\MockObject::foo().              
  53     Call to an undefined method                                  
         PHPUnit\Framework\MockObject\MockObject::foo().              
  54     Call to an undefined method                                  
         PHPUnit\Framework\MockObject\MockObject::foo().              
  55     Call to an undefined method                                  
         PHPUnit\Framework\MockObject\MockObject::foo().              
  68     Call to an undefined method                                  
         PHPUnit\Framework\MockObject\MockObject::foo().              
  70     Call to an undefined method                                  
         PHPUnit\Framework\MockObject\MockObject::foo().              
  72     Call to an undefined method                                  
         PHPUnit\Framework\MockObject\MockObject::foo().              
 ------ ------------------------------------------------------------- 
 ------ ------------------------------------------------------------------- 
  Line   tests/Framework/MockObject/GeneratorTest.php                       
 ------ ------------------------------------------------------------------- 
  108    Call to an undefined method                                        
         PHPUnit\Framework\MockObject\MockObject::doSomething().            
  109    Call to an undefined method                                        
         PHPUnit\Framework\MockObject\MockObject::returnAnything().         
  154    Call to an undefined method                                        
         PHPUnit\Framework\MockObject\MockObject::mockableMethod().         
  155    Call to an undefined method                                        
         PHPUnit\Framework\MockObject\MockObject::anotherMockableMethod().  
  205    Call to an undefined method                                        
         PHPUnit\Framework\MockObject\MockObject::someMethod().             
 ------ ------------------------------------------------------------------- 
 ------ ------------------------------------------------------------------ 
  Line   tests/Framework/MockObject/Matcher/ConsecutiveParametersTest.php  
 ------ ------------------------------------------------------------------ 
  29     Call to an undefined method                                       
[...]

PR #3130, opened 2018-05-12
Enable phpstan/phpstan-phpunit

mhujer:
Improves fix for #3129 from 2ef1018ee0f4a5d44c399604b0ed7defe3c01cdc.

[phpstan-phpunit](https://github.com/phpstan/phpstan-phpunit) extension helps with analysis of the mocks (e.g. when you mock a real class with methods). It resolves some of the ignored errors, but the others are caused by new methods being defined while mocking, which I think are fine to ignore.


sebastianbergmann:

I decided to not use PHPStan and will use Phan (and/or Psalm) in the future.

PR #3134, opened 2018-05-14
Malformed withConsecutive doesn't raise errors

Slamdunk:
References:

1. https://github.com/sebastianbergmann/phpunit-mock-objects/pull/359
1. https://github.com/sebastianbergmann/phpunit-mock-objects/pull/393

Each `withConsecutive` parameter should be an array (or Traversable) in order to be used by [./src/Framework/MockObject/Matcher/ConsecutiveParameters.php](https://github.com/sebastianbergmann/phpunit/blob/8fa2aff0793e896be20f3e85585c827bc5e60530/src/Framework/MockObject/Matcher/ConsecutiveParameters.php#L47) and I would expect an error if the expectation is malformed.

Instead the test passes, but should fail.

### DIFF
--- a/tests/Framework/MockObject/Matcher/ConsecutiveParametersTest.php
+++ b/tests/Framework/MockObject/Matcher/ConsecutiveParametersTest.php
@@ -65,4 +65,22 @@ class ConsecutiveParametersTest extends TestCase
 
         $mock->foo('invalid');
     }
+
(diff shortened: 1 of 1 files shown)


willemstuursma:
Could easily be fixed by changing the definition to:

`public function withConsecutive(iterable ...$arguments)`

sebastianbergmann:
Superseded by #3798.

PR #3145, opened 2018-05-27
Some minor performance optimizations

willemstuursma:
* Use strict `in_array()` so that the new `ZEND_IN_ARRAY ` opcode can be used
* Use splat operator instead of `func_get_args()` where applicable.

### DIFF
--- a/src/Framework/Assert.php
+++ b/src/Framework/Assert.php
@@ -2375,20 +2375,16 @@ abstract class Assert
         static::assertThat($actualJson, new LogicalNot($constraintExpected), $message);
     }
 
-    public static function logicalAnd(): LogicalAnd
+    public static function logicalAnd(Constraint ...$constraints): LogicalAnd
     {
-        $constraints = \func_get_args();
-
         $constraint = new LogicalAnd;
         $constraint->setConstraints($constraints);
 
         return $constraint;
     }
 
-    public static function logicalOr(): LogicalOr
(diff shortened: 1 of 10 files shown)


sebastianbergmann:
Please send two separate pull requests. The refactoring to use the splat operator needs to wait until PHPUnit 8 due to BC considerations.

willemstuursma:
OK, will get to it later this week. 

>  The refactoring to use the splat operator needs to wait until PHPUnit 8 due to BC considerations.

What breaks because of this change? The `Constaint` type is checked a bit later anyway. Just curious.

sebastianbergmann:
There is no semantic change, but a syntactic one as the signature of the methods changes.

Whether or not it is a good idea to overwrite PHPUnit's "built-in" assertions (it's not) in your test case classes, changing their signature is a BC break and existing tests will cease to work due to signature mismatches.

sebastianbergmann:
Cherry-picked https://github.com/sebastianbergmann/phpunit/pull/3145/commits/6b04697d2b1800f6f48b07a9124bc5888d26c511 into `master`.

willemstuursma:
> Whether or not it is a good idea to overwrite PHPUnit's "built-in" assertions (it's not) in your test case classes. 

They could be made `final` in PHPUnit 8.

sebastianbergmann:
Yes, that is an idea I would like to pursue.

Issue #3002, opened 2018-02-12
Towards a better plugin infrastructure


sebastianbergmann:
It should be easier to extend PHPUnit's test runner. More importantly, it should be possible to do so without abusing the `TestListener` interface.

In a first step, two hooks should be implemented: one that is triggered before the first test is executed and one that is triggered after the last test is executed. This is simple enough to implement and has been requested a couple of times over the years.

Another possible hook could be test filtering and ordering for #2660.

I currently envision the following:

```php
namespace PHPUnit\Runner;

interface Hook
{
}
```
```php
namespace PHPUnit\Runner;

interface BeforeFirstTestHook extends Hook
{
    public function executeBeforeFirstTest(): void;
}
```
```php
namespace PHPUnit\Runner;

interface AfterLastTestHook extends Hook
{
    public function executeAfterLastTest(): void;
}
```
Following the Interface Segregation Principle, the idea here is to avoid one big interface and instead have one interface per extension point at which extensions can hook into PHPUnit's test runner.

Here is an example of an extension that uses the two hooks mentioned above:
```php
namespace vendor;

use PHPUnit\Runner\AfterLastTestHook;
use PHPUnit\Runner\BeforeFirstTestHook;

final class MyExtension implements BeforeFirstTestHook, AfterLastTestHook
{
    public function executeAfterLastTest(): void
    {
        // ...
    }

    public function executeBeforeFirstTest(): void
    {
        // ...
    }
}
```
Here is how you configure PHPUnit's test runner to use that extension:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/7.1/phpunit.xsd">
    <extensions>
        <extension class="vendor\MyExtension"/>
    </extensions>
</phpunit>
```

PR #3147, opened 2018-05-28
Breakfast: rerun defects first

epdenouden:
## Use case
Rerun defective unit tests first to speed up the red-green-refactor cycle!

Test results and timings are shared between tests runs using a cache. This allows for sorting defective tests to the front of the execution order using an additional sorter. This new sort option is compatible with existing test ordering options.

This project is a followup to my previous work on [reordering test execution](https://github.com/epdenouden/phpunit/wiki/PHPUnit-test-running-order-management). The code is available in my [Breakfast development branch](https://github.com/epdenouden/phpunit/tree/breakfast).

## Summary of changes
- the `TestSuiteSorter` gains the ability to sort suites and tests by priority of defects and execution time
- added a `ResultCacheExtension` to gather result and timing information during test runs
- added a `TestResultCache` for sharing of result state between runs
- added caching of run results using CLI flag `--cache-result` and configuration attribute `cacheResult='true'`
- added sorting defects to run as quick as possible with CLI flag `--order-by=defects` and configuration attribute `executionOrder='defects'`
- ability to specifiy the cache filename using CLI flag `--cache-result-file` and configuration attribute `cacheResultFile`
- more extensive testing of sorting scenarios

## How to test
1. To activate storing the results of a run use `--cache-result`:

   ```
   phpunit --cache-result
   ```
2. For this example we switch the cache on for every run by adding the attribute `cacheResult="true"` to the `<phpunit>` element in `phpunit.xml`.

   ```xml
   <phpunit cacheResult="true"> <-- no other changes required --></phpunit>
   ```
3. Break `TestCaseTest` by running it backwards:

   ```
   phpunit --order-by=reverse tests/Framework/TestCaseTest.php
   ```
4. Look at the cache file. By default this is `.phpunit.result.cache` in the PHPUnit working directory.
5. Break `TestCaseTest` again with the sorting feature turning on:

   ```
   phpunit --order-by=defects tests/Framework/TestCaseTest.php
   ```
5. The skipped tests are still there but are run immediately now, but still fail. This is easily avoided by asking the sorter to keep dependencies in mind:

   ```
   phpunit --order-by=depends,defects tests/Framework/TestCaseTest.php
   ```

## Design considerations
Like the _New Order_ feature the aim is for a robust system with sane defaults that silently does its work and is easily maintained by other developers:
- Require as little new configuration as possible. Just add `cacheResult="true"` to your configuration and it silently does its work.
- Add no extra output as PHPUnit often lives in automated pipelines and dev scripts.
- Provide robust and maintainable code. This feature also provides a lot of new test coverage for functionality from #3092.
- Implemented a simplistic cache on purpose. Obviously people might want more complex caches based on users, groups, configuration, git hash, etc etc. Let's see what people need.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,4 +17,4 @@
 /tests/TextUI/*.out
 /tests/TextUI/*.php
 /vendor
-
+/.phpunit.result.cache
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -2,6 +2,7 @@
 <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xsi:noNamespaceSchemaLocation="phpunit.xsd"
          bootstrap="tests/bootstrap.php"
+         cacheResult="true"
          verbose="true">
     <testsuites>
         <testsuite name="small">
--- a/phpunit.xsd
+++ b/phpunit.xsd
@@ -167,8 +167,13 @@
     <xs:simpleType name="executionOrderType">
         <xs:restriction base="xs:string">
             <xs:enumeration value="default"/>
-            <xs:enumeration value="reverse"/>
+            <xs:enumeration value="defects"/>
+            <xs:enumeration value="depends"/>
+            <xs:enumeration value="depends,defects"/>
             <xs:enumeration value="random"/>
+            <xs:enumeration value="reverse"/>
+            <xs:enumeration value="depends,random"/>
+            <xs:enumeration value="depends,reverse"/>
         </xs:restriction>
     </xs:simpleType>
     <xs:complexType name="fileFilterType">
@@ -216,6 +221,8 @@
         <xs:attribute name="backupGlobals" type="xs:boolean" default="false"/>
         <xs:attribute name="backupStaticAttributes" type="xs:boolean" default="false"/>
         <xs:attribute name="bootstrap" type="xs:anyURI"/>
+        <xs:attribute name="cacheResult" type="xs:boolean"/>
+        <xs:attribute name="cacheResultFile" type="xs:anyURI"/>
         <xs:attribute name="cacheTokens" type="xs:boolean"/>
         <xs:attribute name="colors" type="xs:boolean" default="false"/>
         <xs:attribute name="columns" type="columnsType" default="80"/>
@@ -228,6 +235,7 @@
         <xs:attribute name="printerClass" type="xs:string" default="PHPUnit\TextUI\ResultPrinter"/>
         <xs:attribute name="printerFile" type="xs:anyURI"/>
         <xs:attribute name="processIsolation" type="xs:boolean" default="false"/>
+        <xs:attribute name="stopOnDefect" type="xs:boolean" default="false"/>
         <xs:attribute name="stopOnError" type="xs:boolean" default="false"/>
         <xs:attribute name="stopOnFailure" type="xs:boolean" default="false"/>
         <xs:attribute name="stopOnWarning" type="xs:boolean" default="false"/>
--- a/src/Framework/TestResult.php
+++ b/src/Framework/TestResult.php
@@ -177,6 +177,11 @@ class TestResult implements Countable
     /**
      * @var bool
      */
+    protected $stopOnDefect = false;
+
+    /**
+     * @var bool
+     */
     protected $lastTestFailed = false;
 
     /**
@@ -229,7 +234,7 @@ class TestResult implements Countable
                 $test->markAsRisky();
             }
 
-            if ($this->stopOnRisky) {
+            if ($this->stopOnRisky || $this->stopOnDefect) {
                 $this->stop();
             }
         } elseif ($t instanceof IncompleteTest) {
@@ -274,7 +279,7 @@ class TestResult implements Countable
      */
     public function addWarning(Test $test, Warning $e, float $time): void
     {
-        if ($this->stopOnWarning) {
+        if ($this->stopOnWarning || $this->stopOnDefect) {
             $this->stop();
         }
 
@@ -301,7 +306,7 @@ class TestResult implements Countable
                 $test->markAsRisky();
             }
 
-            if ($this->stopOnRisky) {
+            if ($this->stopOnRisky || $this->stopOnDefect) {
                 $this->stop();
             }
         } elseif ($e instanceof IncompleteTest) {
@@ -322,7 +327,7 @@ class TestResult implements Countable
             $this->failures[] = new TestFailure($test, $e);
             $notifyMethod     = 'addFailure';
 
-            if ($this->stopOnFailure) {
+            if ($this->stopOnFailure || $this->stopOnDefect) {
                 $this->stop();
             }
         }
@@ -1009,6 +1014,14 @@ class TestResult implements Countable
     public function stopOnSkipped(bool $flag): void
     {
         $this->stopOnSkipped = $flag;
+    }
+
+    /**
+     * Enables or disables the stopping for defects: error, failure, warning
+     */
+    public function stopOnDefect(bool $flag): void
+    {
+        $this->stopOnDefect = $flag;
     }
 
     /**
--- a/src/Runner/TestSuiteSorter.php
+++ b/src/Runner/TestSuiteSorter.php
@@ -32,9 +32,42 @@ final class TestSuiteSorter
     public const ORDER_REVERSED = 2;
 
     /**
+     * @var int
+     */
+    public const ORDER_DEFECTS_FIRST = 3;
+
+    /**
+     * List of sorting weights for all test result codes. A higher number gives higher priority.
+     */
+    private const DEFECT_SORT_WEIGHT = [
+        BaseTestRunner::STATUS_ERROR      => 6,
+        BaseTestRunner::STATUS_FAILURE    => 5,
+        BaseTestRunner::STATUS_WARNING    => 4,
+        BaseTestRunner::STATUS_INCOMPLETE => 3,
+        BaseTestRunner::STATUS_RISKY      => 2,
+        BaseTestRunner::STATUS_SKIPPED    => 1,
+        BaseTestRunner::STATUS_UNKNOWN    => 0
+    ];
+
+    /**
+     * @var array<string, int> Associative array of (string => DEFECT_SORT_WEIGHT) elements
+     */
+    private $defectSortOrder = [];
+
+    /**
+     * @var TestResultCacheInterface
+     */
+    private $cache;
+
+    public function __construct(?TestResultCacheInterface $cache = null)
+    {
+        $this->cache = $cache ?? new NullTestResultCache;
+    }
+
+    /**
      * @throws Exception
      */
-    public function reorderTestsInSuite(Test $suite, int $order, bool $resolveDependencies): void
+    public function reorderTestsInSuite(Test $suite, int $order, bool $resolveDependencies, int $orderDefects): void
     {
         if ($order !== self::ORDER_DEFAULT && $order !== self::ORDER_REVERSED && $order !== self::ORDER_RANDOMIZED) {
             throw new Exception(
@@ -42,16 +75,25 @@ final class TestSuiteSorter
             );
         }
 
-        if ($suite instanceof TestSuite && !empty($suite->tests())) {
+        if ($orderDefects !== self::ORDER_DEFAULT && $orderDefects !== self::ORDER_DEFECTS_FIRST) {
+            throw new Exception(
+                '$orderDefects must be one of TestSuiteSorter::ORDER_DEFAULT, TestSuiteSorter::ORDER_DEFECTS_FIRST'
+            );
+        }
+
+        if ($suite instanceof TestSuite) {
             foreach ($suite as $_suite) {
-                $this->reorderTestsInSuite($_suite, $order, $resolveDependencies);
+                $this->reorderTestsInSuite($_suite, $order, $resolveDependencies, $orderDefects);
             }
 
-            $this->sort($suite, $order, $resolveDependencies);
+            if ($orderDefects === self::ORDER_DEFECTS_FIRST) {
+                $this->addSuiteToDefectSortOrder($suite);
+            }
+            $this->sort($suite, $order, $resolveDependencies, $orderDefects);
         }
     }
 
-    private function sort(TestSuite $suite, int $order, bool $resolveDependencies): void
+    private function sort(TestSuite $suite, int $order, bool $resolveDependencies, int $orderDefects): void
     {
         if (empty($suite->tests())) {
             return;
@@ -63,9 +105,27 @@ final class TestSuiteSorter
             $suite->setTests($this->randomize($suite->tests()));
         }
 
+        if ($orderDefects === self::ORDER_DEFECTS_FIRST && $this->cache !== null) {
+            $suite->setTests($this->sortDefectsFirst($suite->tests()));
+        }
+
         if ($resolveDependencies && !($suite instanceof DataProviderTestSuite) && $this->suiteOnlyContainsTests($suite)) {
             $suite->setTests($this->resolveDependencies($suite->tests()));
         }
+    }
+
+    private function addSuiteToDefectSortOrder(TestSuite $suite): void
+    {
+        $max = 0;
+
+        foreach ($suite->tests() as $test) {
+            if (!isset($this->defectSortOrder[$test->getName()])) {
+                $this->defectSortOrder[$test->getName()] = self::DEFECT_SORT_WEIGHT[$this->cache->getState($test->getName())];
+                $max                                     = \max($max, $this->defectSortOrder[$test->getName()]);
+            }
+        }
+
+        $this->defectSortOrder[$suite->getName()] = $max;
     }
 
     private function suiteOnlyContainsTests(TestSuite $suite): bool
@@ -85,6 +145,40 @@ final class TestSuiteSorter
         \shuffle($tests);
 
         return $tests;
+    }
+
+    private function sortDefectsFirst(array $tests): array
+    {
+        \usort($tests, function ($left, $right) {
+            return $this->cmpDefectPriorityAndTime($left, $right);
+        });
+
+        return $tests;
+    }
+
+    /**
+     * Comparator callback function to sort tests for "reach failure as fast as possible":
+     * 1. sort tests by defect weight defined in self::DEFECT_SORT_WEIGHT
+     * 2. when tests are equally defective, sort the fastest to the front
+     * 3. do not reorder successful tests
+     */
+    private function cmpDefectPriorityAndTime(Test $a, Test $b): int
+    {
+        $priorityA = $this->defectSortOrder[$a->getName()] ?? 0;
+        $priorityB = $this->defectSortOrder[$b->getName()] ?? 0;
+
+        if ($priorityB <=> $priorityA) {
+            // Sort defect weight descending
+            return $priorityB <=> $priorityA;
+        }
+
+        if ($priorityA || $priorityB) {
+            // Sort test duration ascending
+            return $this->cache->getTime($a->getName()) <=> $this->cache->getTime($b->getName());
+        }
+
+        // do not change execution order
+        return 0;
     }
 
     /**
--- a/src/TextUI/Command.php
+++ b/src/TextUI/Command.php
@@ -68,6 +68,8 @@ class Command
     protected $longOptions = [
         'atleast-version='          => null,
         'bootstrap='                => null,
+        'cache-result'              => null,
+        'cache-result-file='        => null,
         'check-version'             => null,
         'colors=='                  => null,
         'columns='                  => null,
@@ -103,6 +105,7 @@ class Command
         'no-coverage'               => null,
         'no-logging'                => null,
         'no-extensions'             => null,
+        'order-by='                 => null,
         'printer='                  => null,
         'process-isolation'         => null,
         'repeat='                   => null,
@@ -113,6 +116,7 @@ class Command
         'reverse-list'              => null,
         'static-backup'             => null,
         'stderr'                    => null,
+        'stop-on-defect'            => null,
         'stop-on-error'             => null,
         'stop-on-failure'           => null,
         'stop-on-warning'           => null,
@@ -291,6 +295,16 @@ class Command
 
                 case '--bootstrap':
                     $this->arguments['bootstrap'] = $option[1];
+
+                    break;
+
+                case '--cache-result':
+                    $this->arguments['cacheResult'] = true;
+
+                    break;
+
+                case '--cache-result-file':
+                    $this->arguments['cacheResultFile'] = $option[1];
 
                     break;
 
@@ -490,6 +504,11 @@ class Command
 
                     break;
 
+                case '--order-by':
+                    $this->handleOrderByOption($option[1]);
+
+                    break;
+
                 case '--process-isolation':
                     $this->arguments['processIsolation'] = true;
 
@@ -502,6 +521,11 @@ class Command
 
                 case '--stderr':
                     $this->arguments['stderr'] = true;
+
+                    break;
+
+                case '--stop-on-defect':
+                    $this->arguments['stopOnDefect'] = true;
 
                     break;
 
@@ -693,7 +717,7 @@ class Command
                     break;
 
                 case '--random-order':
-                    $this->arguments['executionOrder'] = TestSuiteSorter::ORDER_RANDOMIZED;
+                    $this->handleOrderByOption('random');
 
                     break;
 
@@ -703,7 +727,7 @@ class Command
                     break;
 
                 case '--resolve-dependencies':
-                    $this->arguments['resolveDependencies'] = true;
+                    $this->handleOrderByOption('depends');
 
                     break;
 
@@ -713,7 +737,7 @@ class Command
                     break;
 
                 case '--reverse-order':
-                    $this->arguments['executionOrder'] = TestSuiteSorter::ORDER_REVERSED;
+                    $this->handleOrderByOption('reverse');
 
                     break;
 
@@ -1098,6 +1122,7 @@ Test Execution Options:
   --columns <n>               Number of columns to use for progress output
   --columns max               Use maximum number of columns for progress output
   --stderr                    Write to STDERR instead of STDOUT
+  --stop-on-defect            Stop execution upon first not-passed test
   --stop-on-error             Stop execution upon first error
   --stop-on-failure           Stop execution upon first error or failure
   --stop-on-warning           Stop execution upon first warning
@@ -1118,9 +1143,9 @@ Test Execution Options:
   --printer <printer>         TestListener implementation to use
 
   --resolve-dependencies      Resolve dependencies between tests
-  --random-order              Run tests in random order
+  --order-by=<order>          Run tests in order: default|reverse|random|defects|depends
   --random-order-seed=<N>     Use a specific random seed <N> for random order
-  --reverse-order             Run tests last-to-first
+  --cache-result              Write run result to cache to enable ordering tests defects-first
 
 Configuration Options:
 
@@ -1133,6 +1158,7 @@ Configuration Options:
   --include-path <path(s)>    Prepend PHP's include_path with given path(s)
   -d key[=value]              Sets a php.ini value
   --generate-configuration    Generate configuration file with suggested settings
+  --cache-result-file==<FILE> Specify result cache path and filename
 
 Miscellaneous Options:
 
@@ -1292,5 +1318,37 @@ EOT;
         }
 
         return TestRunner::SUCCESS_EXIT;
+    }
+
+    private function handleOrderByOption(string $value): void
+    {
+        foreach (\explode(',', $value) as $order) {
+            switch ($order) {
+                case 'default':
+                    $this->arguments['executionOrder']        = TestSuiteSorter::ORDER_DEFAULT;
+                    $this->arguments['executionOrderDefects'] = TestSuiteSorter::ORDER_DEFAULT;
+                    $this->arguments['resolveDependencies']   = false;
+
+                    break;
+                case 'reverse':
+                    $this->arguments['executionOrder'] = TestSuiteSorter::ORDER_REVERSED;
+
+                    break;
+                case 'random':
+                    $this->arguments['executionOrder'] = TestSuiteSorter::ORDER_RANDOMIZED;
+
+                    break;
+                case 'defects':
+                    $this->arguments['executionOrderDefects'] = TestSuiteSorter::ORDER_DEFECTS_FIRST;
+
+                    break;
+                case 'depends':
+                    $this->arguments['resolveDependencies'] = true;
+
+                    break;
+                default:
+                    $this->exitWithErrorMessage("unrecognized --order-by option: $order");
+            }
+        }
     }
 }
--- a/src/TextUI/TestRunner.php
+++ b/src/TextUI/TestRunner.php
@@ -25,9 +25,12 @@ use PHPUnit\Runner\Filter\Factory;
 use PHPUnit\Runner\Filter\IncludeGroupFilterIterator;
 use PHPUnit\Runner\Filter\NameFilterIterator;
 use PHPUnit\Runner\Hook;
+use PHPUnit\Runner\NullTestResultCache;
+use PHPUnit\Runner\ResultCacheExtension;
 use PHPUnit\Runner\StandardTestSuiteLoader;
 use PHPUnit\Runner\TestHook;
 use PHPUnit\Runner\TestListenerAdapter;
+use PHPUnit\Runner\TestResultCache;
 use PHPUnit\Runner\TestSuiteLoader;
 use PHPUnit\Runner\TestSuiteSorter;
 use PHPUnit\Runner\Version;
@@ -174,10 +177,23 @@ class TestRunner extends BaseTestRunner
             \mt_srand($arguments['randomOrderSeed']);
         }
 
-        if ($arguments['executionOrder'] !== TestSuiteSorter::ORDER_DEFAULT || $arguments['resolveDependencies']) {
-            $sorter = new TestSuiteSorter;
+        if ($arguments['cacheResult']) {
+            if (isset($arguments['cacheResultFile'])) {
+                $cache = new TestResultCache($arguments['cacheResultFile']);
+            } else {
+                $cache = new TestResultCache;
+            }
+            $this->extensions[] = new ResultCacheExtension($cache);
+        }
 
-            $sorter->reorderTestsInSuite($suite, $arguments['executionOrder'], $arguments['resolveDependencies']);
+        if ($arguments['executionOrder'] !== TestSuiteSorter::ORDER_DEFAULT || $arguments['executionOrderDefects'] !== TestSuiteSorter::ORDER_DEFAULT || $arguments['resolveDependencies']) {
+            if (!isset($cache)) {
+                $cache = new NullTestResultCache;
+            }
+            $cache->load();
+            $sorter = new TestSuiteSorter($cache);
+
+            $sorter->reorderTestsInSuite($suite, $arguments['executionOrder'], $arguments['resolveDependencies'], $arguments['executionOrderDefects']);
 
             unset($sorter);
         }
@@ -251,6 +267,10 @@ class TestRunner extends BaseTestRunner
 
         if ($arguments['stopOnSkipped']) {
             $result->stopOnSkipped(true);
+        }
+
+        if ($arguments['stopOnDefect']) {
+            $result->stopOnDefect(true);
         }
 
         if ($arguments['registerMockObjectsFromTestArgumentsRecursively']) {
@@ -806,6 +826,18 @@ class TestRunner extends BaseTestRunner
                 $arguments['bootstrap'] = $phpunitConfiguration['bootstrap'];
             }
 
+            if (isset($phpunitConfiguration['cacheResult']) && !isset($arguments['cacheResult'])) {
+                $arguments['cacheResult'] = $phpunitConfiguration['cacheResult'];
+            }
+
+            if (isset($phpunitConfiguration['cacheResultFile']) && !isset($arguments['cacheResultFile'])) {
+                $arguments['cacheResultFile'] = $phpunitConfiguration['cacheResultFile'];
+            }
+
+            if (isset($phpunitConfiguration['cacheTokens']) && !isset($arguments['cacheTokens'])) {
+                $arguments['cacheTokens'] = $phpunitConfiguration['cacheTokens'];
+            }
+
             if (isset($phpunitConfiguration['cacheTokens']) && !isset($arguments['cacheTokens'])) {
                 $arguments['cacheTokens'] = $phpunitConfiguration['cacheTokens'];
             }
@@ -832,6 +864,10 @@ class TestRunner extends BaseTestRunner
 
             if (isset($phpunitConfiguration['processIsolation']) && !isset($arguments['processIsolation'])) {
                 $arguments['processIsolation'] = $phpunitConfiguration['processIsolation'];
+            }
+
+            if (isset($phpunitConfiguration['stopOnDefect']) && !isset($arguments['stopOnDefect'])) {
+                $arguments['stopOnDefect'] = $phpunitConfiguration['stopOnDefect'];
             }
 
             if (isset($phpunitConfiguration['stopOnError']) && !isset($arguments['stopOnError'])) {
@@ -928,6 +964,10 @@ class TestRunner extends BaseTestRunner
 
             if (isset($phpunitConfiguration['executionOrder']) && !isset($arguments['executionOrder'])) {
                 $arguments['executionOrder'] = $phpunitConfiguration['executionOrder'];
+            }
+
+            if (isset($phpunitConfiguration['executionOrderDefects']) && !isset($arguments['executionOrderDefects'])) {
+                $arguments['executionOrderDefects'] = $phpunitConfiguration['executionOrderDefects'];
             }
 
             if (isset($phpunitConfiguration['resolveDependencies']) && !isset($arguments['resolveDependencies'])) {
@@ -1116,6 +1156,7 @@ class TestRunner extends BaseTestRunner
         $arguments['backupStaticAttributes']                          = $arguments['backupStaticAttributes'] ?? null;
         $arguments['beStrictAboutChangesToGlobalState']               = $arguments['beStrictAboutChangesToGlobalState'] ?? null;
         $arguments['beStrictAboutResourceUsageDuringSmallTests']      = $arguments['beStrictAboutResourceUsageDuringSmallTests'] ?? false;
+        $arguments['cacheResult']                                     = $arguments['cacheResult'] ?? true;
         $arguments['cacheTokens']                                     = $arguments['cacheTokens'] ?? false;
         $arguments['colors']                                          = $arguments['colors'] ?? ResultPrinter::COLOR_DEFAULT;
         $arguments['columns']                                         = $arguments['columns'] ?? 80;
@@ -1130,6 +1171,7 @@ class TestRunner extends BaseTestRunner
         $arguments['excludeGroups']                                   = $arguments['excludeGroups'] ?? [];
         $arguments['failOnRisky']                                     = $arguments['failOnRisky'] ?? false;
         $arguments['failOnWarning']                                   = $arguments['failOnWarning'] ?? false;
+        $arguments['executionOrderDefects']                           = $arguments['executionOrderDefects'] ?? TestSuiteSorter::ORDER_DEFAULT;
         $arguments['groups']                                          = $arguments['groups'] ?? [];
         $arguments['processIsolation']                                = $arguments['processIsolation'] ?? false;
         $arguments['processUncoveredFilesFromWhitelist']              = $arguments['processUncoveredFilesFromWhitelist'] ?? false;
@@ -1148,6 +1190,7 @@ class TestRunner extends BaseTestRunner
         $arguments['stopOnRisky']                                     = $arguments['stopOnRisky'] ?? false;
         $arguments['stopOnSkipped']                                   = $arguments['stopOnSkipped'] ?? false;
         $arguments['stopOnWarning']                                   = $arguments['stopOnWarning'] ?? false;
+        $arguments['stopOnDefect']                                    = $arguments['stopOnDefect'] ?? false;
         $arguments['strictCoverage']                                  = $arguments['strictCoverage'] ?? false;
         $arguments['testdoxExcludeGroups']                            = $arguments['testdoxExcludeGroups'] ?? [];
         $arguments['testdoxGroups']                                   = $arguments['testdoxGroups'] ?? [];
--- a/src/Util/Configuration.php
+++ b/src/Util/Configuration.php
@@ -27,6 +27,8 @@ use SebastianBergmann\FileIterator\Facade as FileIteratorFacade;
  * <phpunit backupGlobals="false"
  *          backupStaticAttributes="false"
  *          bootstrap="/path/to/bootstrap.php"
+ *          cacheResult="false"
+ *          cacheResultFile=".phpunit.result.cache"
  *          cacheTokens="false"
  *          columns="80"
  *          colors="false"
@@ -38,6 +40,7 @@ use SebastianBergmann\FileIterator\Facade as FileIteratorFacade;
  *          disableCodeCoverageIgnore="false"
  *          forceCoversAnnotation="false"
  *          processIsolation="false"
+ *          stopOnDefect="false"
  *          stopOnError="false"
  *          stopOnFailure="false"
  *          stopOnWarning="false"
@@ -65,6 +68,7 @@ use SebastianBergmann\FileIterator\Facade as FileIteratorFacade;
  *          reverseDefectList="false"
  *          registerMockObjectsFromTestArgumentsRecursively="false"
  *          executionOrder="default"
+ *          executionOrderDefects="default"
  *          resolveDependencies="false">
  *   <testsuites>
  *     <testsuite name="My Test Suite">
@@ -716,6 +720,13 @@ final class Configuration
             );
         }
 
+        if ($root->hasAttribute('stopOnDefect')) {
+            $result['stopOnDefect'] = $this->getBoolean(
+                (string) $root->getAttribute('stopOnDefect'),
+                false
+            );
+        }
+
         if ($root->hasAttribute('stopOnError')) {
             $result['stopOnError'] = $this->getBoolean(
                 (string) $root->getAttribute('stopOnError'),
@@ -908,21 +919,45 @@ final class Configuration
             );
         }
 
+        if ($root->hasAttribute('cacheResult')) {
+            $result['cacheResult'] = $this->getBoolean(
+                (string) $root->getAttribute('cacheResult'),
+                false
+            );
+        }
+
+        if ($root->hasAttribute('cacheResultFile')) {
+            $result['cacheResultFile'] = $this->toAbsolutePath(
+                (string) $root->getAttribute('cacheResultFile')
+            );
+        }
+
         if ($root->hasAttribute('executionOrder')) {
-            switch ((string) $root->getAttribute('executionOrder')) {
-                case 'random':
-                    $result['executionOrder'] = TestSuiteSorter::ORDER_RANDOMIZED;
+            foreach (\explode(',', $root->getAttribute('executionOrder')) as $order) {
+                switch ($order) {
+                    case 'default':
+                        $result['executionOrder']        = TestSuiteSorter::ORDER_DEFAULT;
+                        $result['executionOrderDefects'] = TestSuiteSorter::ORDER_DEFAULT;
+                        $result['resolveDependencies']   = false;
 
-                    break;
+                        break;
+                    case 'reverse':
+                        $result['executionOrder'] = TestSuiteSorter::ORDER_REVERSED;
 
-                case 'reverse':
-                    $result['executionOrder'] = TestSuiteSorter::ORDER_REVERSED;
+                        break;
+                    case 'random':
+                        $result['executionOrder'] = TestSuiteSorter::ORDER_RANDOMIZED;
 
-                    break;
+                        break;
+                    case 'defects':
+                        $result['executionOrderDefects'] = TestSuiteSorter::ORDER_DEFECTS_FIRST;
 
-                default:
-                    $result['executionOrder'] = TestSuiteSorter::ORDER_DEFAULT;
+                        break;
+                    case 'depends':
+                        $result['resolveDependencies'] = true;
 
+                        break;
+                }
             }
         }
 
--- a/src/Util/TestResultCache.php
+++ b/src/Util/TestResultCache.php
@@ -1,0 +1,179 @@
+<?php
+/*
+ * This file is part of PHPUnit.
+ *
+ * (c) Sebastian Bergmann <sebastian@phpunit.de>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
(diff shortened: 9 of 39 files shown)


epdenouden:
Oh no not again. Is there a way to have Github show _what_ conflicts there are? Up until recently it told me "this branch can be merged into master" when on the page for a new pull request. :-(

sebastianbergmann:
* This branch cannot be rebased due to conflicts.
* Why is the `ResultCacheListener` needed, or rather why is it needed to be known to the user? As this pull request is about adding functionality to the "core" of PHPUnit, there is no need to expose the usage of an extension point like that. We can simply attach the listener when we need it.
* I think we should add a CLI switch and configuration option to record test results instead.
* I would prefer not to use the old, bloated `TestListener` interface for this but rather the new hook interfaces (introduced in PHPUnit 7.1 with #3002), if possible.

epdenouden:
Thanks for the quick review, again!
- Branch is fixed. The culprit was a backported `TestSuiteSorterTest`. Doing a `git rebase` instead of `merge` even suggested just removing a conflicting commit altogether.
- I extended the public `TestListener` based on the example in the manual. I would also prefer to keep this internal. Will refactor into an extension that can be switched on with a command line flag, also makes testing cleaner.

src/Util/TestResultCache.php
```diff
@@ -0,0 +1,127 @@
...
+
+    public function loadFromJSON(string $json)
+    {
+        $cacheData = \json_decode($json, true);
+
+        if ($cacheData === null) {
+            return;
+        }
+
+        $this->defects = $cacheData['defects'];
```
keradus: we have no clue that decoded data contain such key...
and even if - what is the content of it.

it's optimistic approach here.

cache file could be treated as-is, one is not supposed to modify it manually.
maybe serialize with `unserialize(..., ['allowed_classes' => CacheData::class])` ?

src/Util/TestResultCache.php
```diff
@@ -0,0 +1,127 @@
...
+
+    private function clearCache(): void
+    {
+        $this->defects = [];
+        $this->times   = [];
+    }
+
+    private function formatJSONOutputStruct(): string
+    {
+        return \json_encode([
```
keradus: `json_encode` can fail as well (eg on non utf-8 input)

src/Util/TestResultCache.php
```diff
@@ -0,0 +1,127 @@
...
+
+    /**
+     * @var array
+     */
+    private $defects = [];
+
+    /**
+     * @var array
+     */
+    private $times = [];
```
keradus: what is the content of those collections? array of.... ints? is key meaningful or numeric?
how single defect looks like ?

phpunit.xsd
```diff
@@ -227,6 +233,7 @@
     <xs:attribute name="printerClass" type="xs:string" default="PHPUnit\TextUI\ResultPrinter"/>
     <xs:attribute name="printerFile" type="xs:anyURI"/>
     <xs:attribute name="processIsolation" type="xs:boolean" default="false"/>
+    <xs:attribute name="stopOnDefect" type="xs:boolean" default="false"/>
```
keradus: would be nice to document somewhere what is defect

src/Framework/ResultCacheListener.php
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    private function fullTestName(\PHPUnit\Framework\Test $test)
+    {
+        return $test->getName();
+    }
+
+    private function updateRunState(\PHPUnit\Framework\Test $test, string $state): void
+    {
+        $testName                  = $this->fullTestName($test);
```
keradus: too many spaces

src/Runner/TestSuiteSorter.php
```diff
@@ -31,27 +32,74 @@
...
+        ResultCacheListener::TEST_ERROR      => 6,
+        ResultCacheListener::TEST_FAILURE    => 5,
+        ResultCacheListener::TEST_WARNING    => 4,
+        ResultCacheListener::TEST_INCOMPLETE => 3,
+        ResultCacheListener::TEST_RISKY      => 2,
+        ResultCacheListener::TEST_UNKNOWN    => 0
+    ];
+
+    /**
+     * @var array
```
keradus: array of ?

src/Runner/TestSuiteSorter.php
```diff
@@ -31,27 +32,74 @@
...
 
+    /**
+     * @var int
+     */
+    public const ORDER_DEFECTS_FIRST = 3;
+
+    /**
+     * @var array
+     */
+    private $defectSortWeight = [
```
keradus: const

src/Runner/TestSuiteSorter.php
```diff
@@ -31,27 +32,74 @@
...
-            $this->sort($suite, $order, $resolveDependencies);
+            if ($orderDefects === self::ORDER_DEFECTS_FIRST) {
+                $this->addSuiteToDefectSortOrder($suite);
+            }
+            $this->sort($suite, $order, $resolveDependencies, $orderDefects);
         }
     }
 
-    private function sort(TestSuite $suite, int $order, bool $resolveDependencies): void
+    public function setCache(TestResultCache $cache): void
```
keradus: I believe this property shall not be exposed

tests/Runner/TestSuiteSorterEmptyTestCaseTest.php
```diff
@@ -0,0 +1,53 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace PHPUnit\Runner;
+
+use PHPUnit\Framework\TestCase;
+use PHPUnit\Framework\TestSuite;
+
+class TestSuiteSorterEmptyTestCaseTest extends TestCase
```
keradus: there is no `TestSuiteSorterEmptyTestCase` class, how we can have test for it ?

test here shall be part of `TestSuiteSorterTest` test class, it's one concrete, edge scenario to be tested, not completely other class

src/Runner/TestSuiteSorter.php
epdenouden: Fixed, was a precalculated array before.

src/Runner/TestSuiteSorter.php
epdenouden: You're right, the method is only used in the test. Will clean it up :)

tests/Runner/TestSuiteSorterTest.php
```diff
@@ -101,10 +101,15 @@ public function testSuiteSorterDefectsOptions(int $order, bool $resolveDependenc
     {
         $suite = new TestSuite;
         $suite->addTestSuite(\MultiDependencyTest::class);
+
         $cache = new TestResultCache();
         $cache->loadFromListenerRunstate($runState);
+
         $sorter = new TestSuiteSorter();
-        $sorter->setCache($cache);
+        $_sorter = new \ReflectionClass($sorter);
```
keradus: _? leading _?

what about `$sorterReflection` ?

also, having a need to access class' internals in test indicate bad class structure. maybe Cache shall be injected to Sorter via constructor ?

src/Runner/TestSuiteSorter.php
```diff
@@ -51,7 +51,7 @@
     ];
 
     /**
-     * @var array
+     * @var array[string]int Associative array of (string => DEFECT_SORT_WEIGHT) elements
```
keradus: `array[string]int` is not proper type according to phpdoc
array<string, int>`

src/Util/TestResultCache.php
epdenouden: Thanks, all good points regarding the results cache. Started refactoring that part to make it more robust and cleaner.

src/Runner/TestSuiteSorter.php
epdenouden: Thanks! Where did you find this syntax? I saw this recommended a few times, but have found no definitive answer in the manual https://docs.phpdoc.org/guides/types.html

I found some discussions regarding proposals but most of them seem to have gone to sleep, for example: https://github.com/phpDocumentor/phpDocumentor2/issues/650

keradus: psr-5 always-proposal, sadly. but it's widely suported by IDEs and SCAs

src/Util/TestResultCache.php
```diff
@@ -0,0 +1,127 @@
...
+namespace PHPUnit\Runner;
+
+use PHPUnit\Framework\ResultCacheListener;
+
+class TestResultCache
+{
+    /**
+     * @var string
+     */
+    public const DEFAULT_RESULT_CACHE_FILENAME = 'result_cache.json';
```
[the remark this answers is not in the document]
epdenouden: I have no opinion on the filename of the default value. As a developer I want the following:
1. Simple usage that just works in a basic _code, test, repeat_ cycle. A file in the working directory works well.
2. PHPUnit is part of automation: use the environment variable to do your devops and have it store the cache(s) for users/configs/build-unique-ID/git-hashes/etc however en whereever you want.

The default implementation of `TestResultCache` is there so the simplest use cases work out-of-the-box. The current cache is just a simple MVP. I would prefer NOT to use an external dependency like SQLite like _PHPUnit Clever and Smart_ did, but I am starting to see why they went that way.

src/Util/TestResultCache.php
epdenouden: Yes, the earliest versions of Breakfast were just that: patches to the code that quickly (un)serialized a results file to store the state between runs. I'll try some other options.

src/Util/TestResultCache.php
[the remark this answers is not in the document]
keradus: well, here I'm even not talking about using SQLite or anything like that. (while Indeed, i can see value of it - yet in caching of PHP CS Fixer we don't use it as well - to fancy for our simple purpose).

what I'm saying here @epdenouden is that if I will run phpunit in my project, using cache and out-of-the-box approach (so I don't configure anything extra), phpunit will leave `result_cache.json` file there. if phpunit will leave sth, it shall indicate it's his leftover, putting it to the name part.
like
```diff
-result_cache
+phpunit_cache
```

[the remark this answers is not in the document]
epdenouden: @keradus quick reply: that was a good suggestion, thx! Have pushed a change for that.

I mixed two related topics regarding the caching, making my comment confusion. Adding in persistence cleanly has been more of a chore than I expected. Will spend more time tonight on a cleaner approach.

src/Util/TestResultCache.php
```diff
@@ -62,79 +62,72 @@ public function saveToFile(): void
...
-        $json = \file_get_contents($this->getResultCacheFilename());
+        $cacheData = \file_get_contents($this->getResultCacheFilename());
 
-        if ($json === false) {
+        if ($cacheData === false) {
             return;
         }
 
-        $this->loadFromJSON($json);
+        $cache = \unserialize($cacheData);
```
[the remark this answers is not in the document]
epdenouden: Yes thx :) had added that in the meantime. Since you're here, is it very picky about namespacing?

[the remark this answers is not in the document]
epdenouden: Done! :)

src/Util/TestResultCache.php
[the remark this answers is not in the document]
keradus: new name looks shiny, thanks for the change!
then, I will try to review changes tomorrow  ;)

src/Util/TestResultCache.php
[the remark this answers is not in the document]
keradus: `self::class` takes FQCN into consideration, so all good ;)

tests/Util/TestResultCacheTest.php
```diff
@@ -12,14 +12,12 @@
...
-    /**
-     * @preserveGlobalState
-     */
     public function testCacheFilenameViaEnv()
     {
         $_ENV['PHPUNIT_RESULT_CACHE'] = '/some/cache';
         $cache                        = new TestResultCache;
 
         $this->assertEquals('/some/cache', $cache->getResultCacheFilename());
+        unset($_ENV['PHPUNIT_RESULT_CACHE']);
```
keradus: if `assertEquals` would fail,
this line would never be executed

src/Framework/ResultCacheListener.php
```diff
@@ -0,0 +1,138 @@
...
+    public const TEST_WARNING    = 'warning';
+    public const TEST_FAILURE    = 'failure';
+    public const TEST_RISKY      = 'risky';
+    public const TEST_INCOMPLETE = 'incomplete';
+    public const TEST_UNKNOWN    = 'unknown';
+
+    private $runState = [];
+
+    /**
+     * @var TestResultCache
```
keradus:
```
new ResultCacheListener()
```

boom ! this property is `null`, not `TestResultCache` ;)

epdenouden: added some phpdoc ...which I then even had to sort 🤓

src/Runner/TestSuiteSorter.php
```diff
@@ -31,27 +32,76 @@
...
+
+    /**
+     * @var array[string]int Associative array of (string => DEFECT_SORT_WEIGHT) elements
+     */
+    private $defectSortOrder;
+
+    /**
+     * @var TestResultCache
+     */
+    private $cache;
```
keradus: still.
```php
new TestSuiteSorter()
```

and this property is null, not `TestResultCache`

use `?TestResultCache` or `null|TestResultCache`

epdenouden: There might be a few more of those, I will check and change my code style for this.

epdenouden:
@keradus Hope your offline breakfast was good :) as mentioned in the chat, here's the pull request for the defects-first project. Any comments very welcome! Small things I will take care of ASAP, larger changes to functionality for e.g. mutation testing for a next version.

epdenouden:
Goodmorning @sebastianbergmann! The functionality and code for Breakfast are done, so let's stick a fork in it. ;-) I have been using the changes for a few days now and it looks solid.

 I've checked off the list of changes you requested after I made the PR and taken care of @keradus suggestions so far.
- the `ResultCacheListener` is now and extension
- the result cache extension is automatically is added by using a CLI flag or configuration attribute. No more `phpunit.xml` changes required
- branch can be merged and checks are green
- test coverage has improved since the intial PR, both in functionality-covered and percentage

While working on this a lot of usage scenarios have come up and I have plenty of ideas to make this more useful in the future. I have started to look at other PHP quality and testing projects such as Infection and the various scripts, listeners and printers users have created over the years.

Thanks again for all your work on PHPUnit! Hope my added functionality is useful and saves somebody somewhere a bit of time and frustration. :)

keradus:
Sounds great ;)
If possible, please gave me some time to make a proper review @sebastianbergmann , I would like to do it this evening

sebastianbergmann:
@epdenouden Thank you for letting me know, I hope to find the time to look at this ASAP.

In the meantime, I look forward to see @keradus' review :) Thanks, Dariusz!

src/Framework/TestResult.php
```diff
@@ -229,21 +234,21 @@ public function addError(Test $test, Throwable $t, float $time): void
...
-            if ($this->stopOnIncomplete) {
+            if ($this->stopOnIncomplete || $this->stopOnDefect) {
                 $this->stop();
             }
         } elseif ($t instanceof SkippedTest) {
             $this->skipped[] = new TestFailure($test, $t);
             $notifyMethod    = 'addSkippedTest';
 
-            if ($this->stopOnSkipped) {
+            if ($this->stopOnSkipped || $this->stopOnDefect) {
```
keradus: to be honest not sure about this...
basically, it means that I will **never** use `--stopOnDefect`
in bigger projects, there are always some tests depending on environment, eg different version of PHP, OS, ext, ini...

src/TextUI/Command.php
```diff
@@ -1118,9 +1143,11 @@ protected function showHelp(): void
   --printer <printer>         TestListener implementation to use
 
   --resolve-dependencies      Resolve dependencies between tests
+  --defects-first-order       Run previously unsuccessful tests first
   --random-order              Run tests in random order
   --random-order-seed=<N>     Use a specific random seed <N> for random order
   --reverse-order             Run tests last-to-first
```
keradus: what's the outcome of setting up both, `--reverse-order` and `--defects-first-order` ? will it be respected, ignored, crashed ?

src/Runner/TestSuiteSorter.php
```diff
@@ -30,27 +30,71 @@
...
+        BaseTestRunner::STATUS_FAILURE    => 5,
+        BaseTestRunner::STATUS_WARNING    => 4,
+        BaseTestRunner::STATUS_INCOMPLETE => 3,
+        BaseTestRunner::STATUS_RISKY      => 2,
+        BaseTestRunner::STATUS_SKIPPED    => 1,
+        BaseTestRunner::STATUS_UNKNOWN    => 0
+    ];
+
+    /**
+     * @var array[string]int Associative array of (string => DEFECT_SORT_WEIGHT) elements
```
keradus: this syntax is invalid
`array<string, int>` shall be used

src/Runner/TestSuiteSorter.php
```diff
@@ -30,27 +30,71 @@
...
+     */
+    private $defectSortOrder = [];
+
+    /**
+     * @var null|TestResultCache
+     */
+    private $cache;
+
+    public function __construct(TestResultCache $cache = null)
```
keradus:
```diff
-TestResultCache $cache = null
+?TestResultCache $cache
```

src/Runner/TestSuiteSorter.php
```diff
@@ -30,27 +30,71 @@
...
-            $this->sort($suite, $order, $resolveDependencies);
+            if ($orderDefects === self::ORDER_DEFECTS_FIRST && $this->cache !== null) {
+                $this->addSuiteToDefectSortOrder($suite);
+            }
+            $this->sort($suite, $order, $resolveDependencies, $orderDefects);
         }
     }
 
-    private function sort(TestSuite $suite, int $order, bool $resolveDependencies): void
+    private function sort(TestSuite $suite, int $order, bool $resolveDependencies, $orderDefects): void
```
keradus: what's the type of `$orderDefects` ?

src/Runner/TestSuiteSorter.php
```diff
@@ -62,11 +106,29 @@ private function sort(TestSuite $suite, int $order, bool $resolveDependencies):
...
+        if ($orderDefects === self::ORDER_DEFECTS_FIRST && $this->cache !== null) {
+            $suite->setTests($this->sortDefectsFirst($suite->tests()));
+        }
+
         if ($resolveDependencies && $suite->tests()[0] instanceof TestCase) {
             $suite->setTests($this->resolveDependencies($suite->tests()));
         }
     }
 
+    private function addSuiteToDefectSortOrder(TestSuite $suite): void
```
keradus: **MAJOR**
This method can be called even when `$this->cache = null`, and then it gonna crash hardly.
It means whenever one want to call `addSuiteToDefectSortOrder`, he need to check `$this->cache !== null` manually, and he need to be aware about this! This proves that architecture of this class is defective.

Also, having a need to do that check everywhere isn't optimal - I would suggest to remove possibility to have that field nullable - eg, by introducing `TestResultCacheInterface` and `NullTestResultCache`

src/Runner/TestSuiteSorter.php
```diff
@@ -79,6 +141,38 @@ private function randomize(array $tests): array
         return $tests;
     }
 
+    private function sortDefectsFirst(array $tests): array
+    {
+        \usort($tests, [self::class, 'cmpDefectPriorityAndTime']);
```
keradus: please avoid calling methods by their names. Static Code Analysis will not detect that...

src/Runner/TestSuiteSorter.php
```diff
@@ -79,6 +141,38 @@ private function randomize(array $tests): array
...
+
+        return $tests;
```
keradus: this array was already modified by `usort` under the hood, then it's returned...
with that, it's possible that one will write code:
```php
$sortedArray = $this->sortDefectsFirst($array);
```
and think that `$sortedArray !== $array`

src/Runner/TestSuiteSorter.php
```diff
@@ -62,11 +106,29 @@ private function sort(TestSuite $suite, int $order, bool $resolveDependencies):
             $suite->setTests($this->randomize($suite->tests()));
         }
 
+        if ($orderDefects === self::ORDER_DEFECTS_FIRST && $this->cache !== null) {
+            $suite->setTests($this->sortDefectsFirst($suite->tests()));
+        }
+
         if ($resolveDependencies && $suite->tests()[0] instanceof TestCase) {
             $suite->setTests($this->resolveDependencies($suite->tests()));
         }
```
keradus: from this code, it looks like it's possible to have, on single run, sorted by `randomize`, then `sortDefectsfirst`, then `resolveDependencies`. is it by purpose?

src/TextUI/Command.php
keradus: looks like `--defects-first-order` overrides `--random-order` and `--reverse-order`

src/Runner/TestSuiteSorter.php
```diff
@@ -79,6 +141,38 @@ private function randomize(array $tests): array
...
+
+    /**
+     * Comparator callback function to sort tests for "reach failure as fast as possible":
+     * 1. sort tests by defect weight defined in self::DEFECT_SORT_WEIGHT
+     * 2. when tests are equally defective, sort the fastest to the front
+     * 3. do not reorder successful tests
+     */
+    private function cmpDefectPriorityAndTime(Test $a, Test $b): int
+    {
+        $priorityA = $this->defectSortOrder[$a->getName()] ?? 0;
```
keradus: this method is called like that:
`\usort($tests, [self::class, 'cmpDefectPriorityAndTime']);`

meaning, it's called on class, not on instance - like it would be static method. yet, it's not declared as static, and it's accessing `$this->...` inside.
please decide, shall it be static or not, and declare/call it properly

src/TextUI/TestRunner.php
```diff
@@ -174,14 +176,31 @@ public function doRun(Test $suite, array $arguments = [], bool $exit = true): Te
...
 
-        if ($arguments['executionOrder'] !== TestSuiteSorter::ORDER_DEFAULT || $arguments['resolveDependencies']) {
-            $sorter = new TestSuiteSorter;
+        if (isset($arguments['cacheResultFile'])) {
+            $cache = new TestResultCache($arguments['cacheResultFile']);
+        }
+
+        if ($arguments['executionOrder'] !== TestSuiteSorter::ORDER_DEFAULT || $arguments['executionOrderDefects'] !== TestSuiteSorter::ORDER_DEFAULT || $arguments['resolveDependencies']) {
+            if (!isset($cache)) {
+                $cache = new TestResultCache;
```
keradus: why no `cacheResultFile` used here?

src/TextUI/TestRunner.php
```diff
@@ -174,14 +176,31 @@ public function doRun(Test $suite, array $arguments = [], bool $exit = true): Te
...
 
-            $sorter->reorderTestsInSuite($suite, $arguments['executionOrder'], $arguments['resolveDependencies']);
+            $sorter->reorderTestsInSuite($suite, $arguments['executionOrder'], $arguments['resolveDependencies'], $arguments['executionOrderDefects']);
 
             unset($sorter);
         }
 
+        if ($arguments['cacheResult']) {
+            if (isset($arguments['cacheResultFile'])) {
+                $outputCache = new TestResultCache($arguments['cacheResultFile']);
```
keradus: do we really need to distinguish `$cache` from `$outputCache` ?
can't we use same cache instance ?

Also, if one would pass `cacheResultFile` but no `cacheResult`, he will get (input) `$cache`, but no `$outputCache`.
I see this problematic and I would not allow for that

src/TextUI/TestRunner.php
```diff
@@ -1111,50 +1154,53 @@ protected function handleConfiguration(array &$arguments): void
...
-        $arguments['timeoutForMediumTests']                           = $arguments['timeoutForMediumTests'] ?? 10;
-        $arguments['timeoutForSmallTests']                            = $arguments['timeoutForSmallTests'] ?? 1;
-        $arguments['verbose']                                         = $arguments['verbose'] ?? false;
+        $arguments['addUncoveredFilesFromWhitelist']                           = $arguments['addUncoveredFilesFromWhitelist'] ?? true;
+        $arguments['backupGlobals']                                            = $arguments['backupGlobals'] ?? null;
+        $arguments['backupStaticAttributes']                                   = $arguments['backupStaticAttributes'] ?? null;
+        $arguments['beStrictAboutChangesToGlobalState']                        = $arguments['beStrictAboutChangesToGlobalState'] ?? null;
+        $arguments['beStrictAboutResourceUsageDuringSmallTests']               = $arguments['beStrictAboutResourceUsageDuringSmallTests'] ?? false;
+        $arguments['cacheResult']                                              = $arguments['cacheResult'] ?? false;
```
keradus: shouldn't this be enabled by default? this is super useful feature, why to hide it ?

src/Util/Configuration.php
```diff
@@ -924,6 +948,19 @@ public function getPHPUnitConfiguration(): array
...
+        if ($root->hasAttribute('executionOrderDefects')) {
+            switch ((string) $root->getAttribute('executionOrderDefects')) {
+                case 'defectsFirst':
+                    $result['executionOrderDefects'] = TestSuiteSorter::ORDER_DEFECTS_FIRST;
+
+                    break;
+
+                default:
+                    $result['executionOrderDefects'] = TestSuiteSorter::ORDER_DEFAULT;
+
```
keradus: dummy lines doesn't look needed here

src/Util/TestResultCache.php
```diff
@@ -0,0 +1,174 @@
...
+     * @var array array<string, int>
+     */
+    private $defects = [];
+
+    /**
+     * The list of execution duration of suites and tests (in seconds)
+     *
+     * <code>
+     * // Record running time for test
+     * $this->times[$testName] = 1.234;
```
keradus: missing `</code>`

src/Util/TestResultCache.php
```diff
@@ -0,0 +1,174 @@
...
+        BaseTestRunner::STATUS_FAILURE,
+        BaseTestRunner::STATUS_ERROR,
+        BaseTestRunner::STATUS_RISKY,
+        BaseTestRunner::STATUS_WARNING
+    ];
+
+    /**
+     * Alternative path and filename for result cache file
+     *
+     * @var null|string
```
keradus: when could this be null, if it fallbacks to `DEFAULT_RESULT_CACHE_FILENAME` ?

tests/Framework/ResultCacheExtensionTest.php
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    public function testEmptySuite(): void
+    {
+        $suite = new TestSuite;
+        $suite->addTestSuite(\EmptyTestCaseTest::class);
+        $suite->run($this->result);
+
+        $this->assertEquals(BaseTestRunner::STATUS_WARNING, $this->cache->getState('Warning'));
+        $this->assertEquals(0, $this->cache->getTime('test_name'));
```
keradus: `assertSame` please. globally

keradus (changes requested):

src/Framework/TestResult.php
epdenouden: Yes, thanks for noticing! This is also a huge doubt of mine. On the one hand I want to have a shorthand for "I just want a quick edit-test-edit-test cycle right now", but it has also gotten in my way so much I had moved the "not really bad" defects down the sorter list.

For now I will remove the halting part and leave it up to the user to supply explicit `--stop-on-warning`, etc options

src/TextUI/Command.php
epdenouden: I actually combines them. Each of the sorters has a specific preference and some are more important than others:
1. _random_ and _reverse_ go first, they are considered mostly cosmetic. In theory they have no effect on the outcome.
2. More important is sorting the defects to the front. It is an explicit request from the (human) user that indicates "I have seen what happens during the runs and I would like an explicit change to that."
3. The dependency resolver has the last say since it's job is preventing breakages. A side effect of _defects first_ I found is getting unrelated failures as the defects were sometimes sorted the wrong 'side' of their dependants.

epdenouden: I agree I need to explicitly document this behaviour and am working on a few diagrams that show the limits of the sorters when looking at how suites can be inside other suites, what dependencies really do, and so on.

src/Framework/TestResult.php
keradus: please dont
(elaboration soon)

keradus: stopping on given type of result (~defect) shall work the same as putting them to cache to be executed first.

it's great value for me to have `--stop-on-defect` rather than `--stop-on-*` 10 times.

Only thing is to properly specify what is a defect.
For me, it all that "executed and failed" - thus all that need to manual fix from developer.
Thus, not met expectation, warning, ..., but not "skipped" or "incomplete"

src/TextUI/Command.php
keradus: if you need big description and diagrams, it's probably means it's overcomplicated.
maybe let us shrink it down to what would be the most value for user?

I would suggest sth like:
`--order-by=default|reverse|random|defect`
without possibility to mix them, like `reverse + defect`. I don't see much value of that combination, do you ?

src/TextUI/TestRunner.php
epdenouden: The cacheResultFile argument is handled right above. There is an edge case that I sometimes used myself where you want to specifiy a cache file to *read* from, but do not overwrite it. The tests for the project itself use this with an example cache in `tests/_files/...`

Another scenario is when using `--stop-on-*` where the result file will only store the defects (and timings) of the tests of the shortened run.

src/TextUI/TestRunner.php
epdenouden: I agree! It's in my local config file. :) I'll leave that up to Sebastian if he wants it on by default.

src/Framework/ResultCacheExtension.php
```diff
@@ -0,0 +1,132 @@
...
+{
+    /**
+     * @var array<string, array> array to keep track of suite/test state and timing information
+     */
+    private $runState = [];
+
+    /**
+     * @var null|TestResultCache
+     */
+    private $cache;
```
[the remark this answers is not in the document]
epdenouden: fixed

src/Framework/ResultCacheExtension.php
```diff
@@ -0,0 +1,132 @@
...
+use PHPUnit\Runner\AfterTestWarningHook;
+use PHPUnit\Runner\BaseTestRunner;
+use PHPUnit\Runner\BeforeFirstTestHook;
+use PHPUnit\Runner\BeforeTestHook;
+use PHPUnit\Runner\TestResultCache;
+
+final class ResultCacheExtension implements BeforeFirstTestHook, BeforeTestHook, AfterSuccessfulTestHook, AfterSkippedTestHook, AfterRiskyTestHook, AfterIncompleteTestHook, AfterTestErrorHook, AfterTestWarningHook, AfterTestFailureHook, AfterLastTestHook
+{
+    /**
+     * @var array<string, array> array to keep track of suite/test state and timing information
```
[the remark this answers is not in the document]
epdenouden: By using the `$time` provided to the individual tests I was able to refactor the `$runstate` out completely. The previous `ResultCacheListener` also kept track of TestSuite timings, which was done using this internal run state.

src/Framework/ResultCacheExtension.php
```diff
@@ -0,0 +1,132 @@
...
+    {
+        $this->flush();
+    }
+
+    private function updateRunState(string $test, int $state): void
+    {
+        $testname = $this->getTestName($test);
+
+        if (!isset($this->runState[$testname])) {
+            $this->runState[$testname] = [
```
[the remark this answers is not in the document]
epdenouden: I have simplified the `ResultCacheExtension` quite a lot. It directly tells the cache when it has something of interest, bypassing the need for the internal state and a few hooks.

src/Runner/TestSuiteSorter.php
epdenouden: fixed

src/Runner/TestSuiteSorter.php
epdenouden: fixed

src/Runner/TestSuiteSorter.php
epdenouden: oops, fixed!

src/Runner/TestSuiteSorter.php
epdenouden: Yup, I overlooked the check for the cache in `addSuiteToDefectSortOrder` since it would never happen internally. However it is good to check for it inside the method. At least for now I have added that.

The interface suggestion sounds like an interesting approach, I will give it a spin

src/TextUI/TestRunner.php
keradus: > There is an edge case that I sometimes used myself where you want to specifiy a cache file to read from, but do not overwrite it

i would drop that, imo it brings overcomplication and confusion

src/TextUI/TestRunner.php
keradus: maybe, for that, let us enable this for now in PR, and see Sebastian's overall review ?

src/Runner/TestSuiteSorter.php
epdenouden: Fixed, thanks for the syntax suggestion :)

src/Runner/TestSuiteSorter.php
epdenouden: It is meant to be a private utility function used to work on a specific instance / runtime context. It requires knowledge of the state the cache and sort order to do its work.

src/Runner/TestSuiteSorter.php
epdenouden: While it is true that `usort` changes the `$tests` we still need to pass that information back to the caller. We could pass `&$tests` but then the method would have a different signature than the others for random, reverse and resolve.

src/Util/TestResultCache.php
epdenouden: fixed

src/Util/TestResultCache.php
epdenouden: Good catch, cleaned up. In previous versions `null` was used to the default as a fallback in another part of the code.

src/Util/Configuration.php
epdenouden: This supports when somebody uses the config attribute `executionOrderDefects='default'` which can happen when automated pipelines do something like `executionOrderDefects='$value'`. I have added a test for it.

src/TextUI/TestRunner.php
epdenouden: Alright let's go

src/Framework/TestResult.php
epdenouden: I like this suggestion very much: this was the first version I made. Remember that code style fix for the `defectsWeights` constant? That was a leftover of the original configurable version which I thought overcomplicated at the time.

I will look for a nice way to add it back in.

src/TextUI/Command.php
epdenouden: Yes, I think that would be better. My personal use case of 'randomized but with previous defects first' might be too exotic. Let's keep it simple.

src/Framework/TestResult.php
epdenouden: Rereading your comment: things that need a manual fix from the developer in the test or source would be:
- error: usually PHP syntax mistakes
- failure: assert failing
- risky: developer forgot to add assertions to a new test

Am I reading you correctly?

src/Runner/TestSuiteSorter.php
epdenouden: Nevermind the previous comment, I have given the interface a try. Is this the kind of pattern you had in mind?

src/Runner/TestSuiteSorter.php
epdenouden: I have fixed the static call by the `usort()`

src/TextUI/TestRunner.php
epdenouden: I have changed the tests to not cause the edge case anymore. Will simplify the usage of `cacheResult` and `cacheResultFile` tomorrow. :)

src/TextUI/TestRunner.php
epdenouden: Agreed. The edge case in the PHPUnit self-tests has been solved differently. I can now make this cleaner, too.

src/Framework/TestResult.php
keradus: exactly.

to clarify - i'm not thinking about making it configurable again, it shall not.
defects are all that require developer fix, so exactly ones you listed :+1:

src/Runner/TestSuiteSorter.php
keradus: indeed

src/Util/Configuration.php
keradus: I meant the empty ("dummy") line number 960

tests/Util/TestResultCacheTest.php
```diff
@@ -0,0 +1,95 @@
...
+class TestResultCacheTest extends TestCase
+{
+    public function testReadsCacheFromProvidedFilename(): void
+    {
+        $cacheFile = \dirname(__DIR__) . '/_files/MultiDependencyTest_result_cache.txt';
+        $cache     = new TestResultCache($cacheFile);
+        $cache->load();
+
+        $this->assertEquals(BaseTestRunner::STATUS_UNKNOWN, $cache->getState('testOne'));
+        $this->assertEquals(BaseTestRunner::STATUS_SKIPPED, $cache->getState('testFive'));
```
keradus: `assertSame` please, globally

src/Framework/ResultCacheExtension.php
```diff
@@ -18,19 +18,12 @@
...
-use PHPUnit\Runner\BeforeTestHook;
 use PHPUnit\Runner\TestResultCache;
 
-final class ResultCacheExtension implements BeforeFirstTestHook, BeforeTestHook, AfterSuccessfulTestHook, AfterSkippedTestHook, AfterRiskyTestHook, AfterIncompleteTestHook, AfterTestErrorHook, AfterTestWarningHook, AfterTestFailureHook, AfterLastTestHook
+final class ResultCacheExtension implements AfterSuccessfulTestHook, AfterSkippedTestHook, AfterRiskyTestHook, AfterIncompleteTestHook, AfterTestErrorHook, AfterTestWarningHook, AfterTestFailureHook, AfterLastTestHook
 {
     /**
-     * @var array<string, array> array to keep track of suite/test state and timing information
-     */
-    private $runState = [];
```
keradus: ping @epdenouden ,
dropping this property will save a lot of memory for huge testbase (like `15023` of mine)

phpunit.xml
```diff
@@ -2,6 +2,8 @@
 <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xsi:noNamespaceSchemaLocation="phpunit.xsd"
          bootstrap="tests/bootstrap.php"
+         cacheResult="true"
+         executionOrderDefects="defectsFirst"
```
sebastianbergmann: I would prefer not to change the configuration of PHPUnit's own configuration (for now, at least).

.gitignore
```diff
@@ -17,4 +17,4 @@
 /tests/TextUI/*.out
 /tests/TextUI/*.php
 /vendor
-
+/.phpunit.result.cache
```
sebastianbergmann: This is not needed if we do not enable result caching for PHPUnit's own test suite.

src/Framework/ResultCacheExtension.php
```diff
@@ -0,0 +1,112 @@
...
+use PHPUnit\Runner\AfterRiskyTestHook;
+use PHPUnit\Runner\AfterSkippedTestHook;
+use PHPUnit\Runner\AfterSuccessfulTestHook;
+use PHPUnit\Runner\AfterTestErrorHook;
+use PHPUnit\Runner\AfterTestFailureHook;
+use PHPUnit\Runner\AfterTestWarningHook;
+use PHPUnit\Runner\BaseTestRunner;
+use PHPUnit\Runner\TestResultCache;
+
+final class ResultCacheExtension implements AfterSuccessfulTestHook, AfterSkippedTestHook, AfterRiskyTestHook, AfterIncompleteTestHook, AfterTestErrorHook, AfterTestWarningHook, AfterTestFailureHook, AfterLastTestHook
```
sebastianbergmann: The `Framework` directory is about "writing tests with PHPUnit". Classes related to "running tests with PHPUnit" belong in `Runner`.

phpunit.xml
keradus: Please elaborate @sebastianbergmann ,

While I can agree that setting `ignore-dependencies` can bring failures for codebase with hidden dependencies in tests (which, anyway, is great to detect and devs shall fix that in their repos!),
having caching by default is great, running test that failed on previous run as first will save a lot of time to the devs!
And i'm afraid that if it won't be the default, very few of us in the wild will be even aware of this option...

src/TextUI/Command.php
```diff
@@ -68,6 +68,8 @@ class Command
     protected $longOptions = [
         'atleast-version='          => null,
         'bootstrap='                => null,
+        'cache-result'              => null,
+        'cache-result-file='        => null,
```
sebastianbergmann: I am not sure if we really want this to be configurable using CLI options. If caching could only be turned on through `phpunit.xml` then the hidden cache file could be placed next to it.

src/TextUI/Command.php
```diff
@@ -505,6 +519,11 @@ protected function handleArguments(array $argv): void
 
                     break;
 
+                case '--stop-on-defect':
```
sebastianbergmann: I do not understand why `--stop-on-defect` is needed.

src/TextUI/Command.php
keradus: > If caching could only be turned on through `phpunit.xml`

`cache-result` is to enable cache via CLI

src/TextUI/Command.php
```diff
@@ -1133,6 +1160,7 @@ protected function showHelp(): void
   --include-path <path(s)>    Prepend PHP's include_path with given path(s)
   -d key[=value]              Sets a php.ini value
   --generate-configuration    Generate configuration file with suggested settings
+  --cache-result-file==<FILE> Specify result cache path and filename
```
sebastianbergmann: What about "Specify path to result cache file"?

src/TextUI/TestRunner.php
```diff
@@ -1111,50 +1151,53 @@ protected function handleConfiguration(array &$arguments): void
...
+        $arguments['stopOnWarning']                                            = $arguments['stopOnWarning'] ?? false;
+        $arguments['stopOnDefect']                                             = $arguments['stopOnDefect'] ?? false;
+        $arguments['strictCoverage']                                           = $arguments['strictCoverage'] ?? false;
+        $arguments['testdoxExcludeGroups']                                     = $arguments['testdoxExcludeGroups'] ?? [];
+        $arguments['testdoxGroups']                                            = $arguments['testdoxGroups'] ?? [];
+        $arguments['timeoutForLargeTests']                                     = $arguments['timeoutForLargeTests'] ?? 60;
+        $arguments['timeoutForMediumTests']                                    = $arguments['timeoutForMediumTests'] ?? 10;
+        $arguments['timeoutForSmallTests']                                     = $arguments['timeoutForSmallTests'] ?? 1;
+        $arguments['verbose']                                                  = $arguments['verbose'] ?? false;
```
sebastianbergmann: I do not understand this whitespace change.

src/TextUI/Command.php
sebastianbergmann: I understand that. What I am not sure about is the value of being able to turn on caching on the commandline.

src/Util/Configuration.php
epdenouden: I thought that was way too simple! Even your comments on whitespace scare me into making more tests haha

tests/Util/TestResultCacheTest.php
epdenouden: thx, got 'm

phpunit.xml
epdenouden: Oops. This actually was an accidental commit-push-zzzzz from me yesterday evening, but these are the standard settings to speed up development.

Most developers would never find out about the option, I have to agree with @keradus. Not that they would not _want_ it, yet many end users simply never read the changelog and docs in that amount of detail. For context, I still see test suites that do not even use fixtures or dataproviders, handling those things themselves manually.

.gitignore
epdenouden: True! However, I can also imagine a user being surprised by a change in `git status` when they turn on the feature and all of a sudden an untracked file appears.

src/Framework/ResultCacheExtension.php
epdenouden: Just to confirm, you would like `ResultCacheExtension` to be moved to `Runner`, right? No problem, I will move it around!

sebastianbergmann: Yes, please. Thanks!

src/TextUI/Command.php
epdenouden: Hmmm yes. I haven't really turned it off anymore since it became stable to use. 🤔

src/TextUI/Command.php
epdenouden: Strictly it isn't. I added it as a shorthand for the a combination of `--stop-on-$status` flags. @keradus do you have a more compelling explanation?

src/TextUI/Command.php
epdenouden: It contains a the full `$path . $filename` not just a path. The filename itself doesn't have to be `.phpunit.result.cache`

src/TextUI/TestRunner.php
epdenouden: Me neither, to be honest. I have just removed all whitespace between `]` and `=` and had it reformatted. Looks like the assignment alignment doesn't remove excess whitespace once it has been introduced. Will commit a WS fix

src/TextUI/Command.php
epdenouden: @sebastianbergmann do you have a preference. Guessing from your other comments, you prefer the simpler UX as well?

tests/Framework/ResultCacheExtensionTest.php
epdenouden: Missed a few, fixed!

.gitignore
keradus: I would also keep it, I suggest to always have `.php_cs.cache` in .gitignore even if `.php_cs.dist` is not having cache enabled (but `.php_cs` can have, right ?)

src/Framework/ResultCacheExtension.php
epdenouden: Done!

src/TextUI/Command.php
keradus: With current definition of "defect", it's nothing more than a shortcut for putting down `--stop-on-*` multiple times.
It's not mandatory to have, but it's super nice shortcut for fast fail, eg for CLI.
(a lot of users runs all tests in local env, but want to fail on first issue on CI level)

src/Framework/ResultCacheExtension.php
keradus: @epdenouden , `ResultCacheExtensionTest.php` shall also change a folder ;)

epdenouden:
@keradus @sebastianbergmann In https://github.com/sebastianbergmann/phpunit/pull/3147/commits/11360b3f0dabdaf6ddd022e7df68ad79541df988 I have added a suggestion for what a cleaner `--order-by` CLI flag could look like. Having tried it out yesterday evening while working it feels easy to use. Much better than the current clutter of `--defects-first-order`, `--resolve-dependencies`, and so on.

image

Notes:
- I have kept the existing flags introduced in 7.2 to not break current installations.
- `Command` is not easy to test without doing functional tests or reflection

src/TextUI/Command.php
```diff
@@ -1143,10 +1143,9 @@ protected function showHelp(): void
   --printer <printer>         TestListener implementation to use
 
   --resolve-dependencies      Resolve dependencies between tests
-  --defects-first-order       Run previously unsuccessful tests first
-  --random-order              Run tests in random order
+  --order-by=<order>          Run tests in 'default', 'random' or 'reverse' order
+  --order-by=defects          Run previously failed tests first
```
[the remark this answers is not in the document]
epdenouden: Good points, yes. Wanted to give it a quick try using it myself for a while first. If this is something that @sebastianbergmann also likes I will make it production-ready.

src/TextUI/Command.php
```diff
@@ -1321,4 +1320,34 @@ private function handleListTestsXml(TestSuite $suite, string $target, bool $exit
 
         return TestRunner::SUCCESS_EXIT;
     }
+
+    private function handleOrderByOption(string $value): void
+    {
+        foreach (\explode(',', $value) as $order) {
```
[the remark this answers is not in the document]
epdenouden: Yes. I didn't see other `--options` throw nice error messages yesterday. Will have a look at how to do this nicely. Silent failures are a big no-no here, I have typed `defecst` waaaaay too often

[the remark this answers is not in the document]
epdenouden: Added a simple error message with failure exit code.

src/TextUI/Command.php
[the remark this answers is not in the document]
epdenouden: It barely fits on one line now, but everybody reads the manual... right? :)

epdenouden:
@keradus The `--order-by=foobar` will now exit with an error message:
```
./phpunit --order-by=defects,foobar,depends
```

```
PHPUnit 7.3-g0a4a8c258 by Sebastian Bergmann and contributors.

unrecognized order-by option -- foobar
```

The help message is now:
```
  --order-by=<order>          Run tests in order: default|reverse|random|defects|depends
```

src/TextUI/Command.php
```diff
@@ -1347,6 +1346,8 @@ private function handleOrderByOption(string $value): void
                     $this->arguments['resolveDependencies'] = true;
 
                     break;
+                default:
+                    $this->exitWithErrorMessage("unrecognized order-by option -- $order");
```
keradus: why double `--` ?

phpunit.xsd
```diff
@@ -171,6 +171,12 @@
       <xs:enumeration value="random"/>
     </xs:restriction>
   </xs:simpleType>
+  <xs:simpleType name="executionOrderDefectsType">
```
keradus: this shall be removed now, and `executionOrderType` shall be extended to contain `defects` and `depends`

src/Framework/TestResult.php
```diff
@@ -1011,6 +1016,14 @@ public function stopOnSkipped(bool $flag): void
         $this->stopOnSkipped = $flag;
     }
 
+    /**
+     * Enables or disables the stopping for any defect: skipped, error, failure, warning, incomplete or risky.
```
keradus: I believe this description is not up to date now

epdenouden: Fixed

src/TextUI/Command.php
epdenouden: Fixed. Was a leftover from where I found this in `GetOpt`.

phpunit.xsd
epdenouden: Removed. I have added a list of the common scenarios that make sense. Couldn't find a nice way to define the list in the XSD to help the tooling with validation and autocomplete.

tests/Runner/TestSuiteSorterTest.php
```diff
@@ -7,78 +7,322 @@
...
+                TestSuiteSorter::ORDER_DEFAULT,
+                self::RESOLVE_DEPENDENCIES,
+                [
+                    'testOne'   => ['state' => BaseTestRunner::STATUS_PASSED, 'time' => 1],
+                    'testTwo'   => ['state' => BaseTestRunner::STATUS_PASSED, 'time' => 1],
+                    'testThree' => ['state' => BaseTestRunner::STATUS_PASSED, 'time' => 1],
+                    'testFour'  => ['state' => BaseTestRunner::STATUS_FAILURE, 'time' => 1],
+                    'testFive'  => ['state' => BaseTestRunner::STATUS_FAILURE, 'time' => 1],
+                ],
+                ['testFive', 'testOne', 'testTwo', 'testThree', 'testFour']],
```
keradus: :-1:
`testFour` and `testFive` failed, so I want to have them as soon as possible, with respect to their dependencies.

I would expect that order, instead:
`['testFive', 'testThree', 'testFour', 'testOne', 'testTwo']`
or
`['testThree', 'testFour', 'testFive', 'testOne', 'testTwo']`

epdenouden: - `testThree` has dependencies on `testOne` and `testTwo`
- `testFour` has a dependency on `testThree`

what happens:
1. after sorting for defects the order is: `4` (failed), `5` (failed), `1`, `2`, `3`
2. running it like this would skip `4` because it needs to run `3` first
3. same story for `3`: it needs both `1` and `2`
4. the dependency resolver fixes this quietly for us: `5` can stay up front (=defects FIRST) but `4` goes behind `3` which goes behind `1,2` (=defects ASAP ;-)

keradus: ha, ok, your first 2 points were not part of description (while dependencies of testThree for TestFour was)

then, all good :+1:

epdenouden: Good but not good enough :) I will document the use of `MultiDependencyTest` in the dataproviders to avoid some of the confusion.

src/TextUI/TestRunner.php
```diff
@@ -1111,50 +1151,53 @@ protected function handleConfiguration(array &$arguments): void
...
+        $arguments['enforceTimeLimit']                               = $arguments['enforceTimeLimit'] ?? false;
+        $arguments['excludeGroups']                                  = $arguments['excludeGroups'] ?? [];
+        $arguments['failOnRisky']                                    = $arguments['failOnRisky'] ?? false;
+        $arguments['failOnWarning']                                  = $arguments['failOnWarning'] ?? false;
+        $arguments['executionOrderDefects']                          = $arguments['executionOrderDefects'] ?? TestSuiteSorter::ORDER_DEFAULT;
+        $arguments['groups']                                         = $arguments['groups'] ?? [];
+        $arguments['processIsolation']                               = $arguments['processIsolation'] ?? false;
+        $arguments['processUncoveredFilesFromWhitelist']             = $arguments['processUncoveredFilesFromWhitelist'] ?? false;
+        $arguments['randomOrderSeed']                                = $arguments['randomOrderSeed'] ?? \time();
+        $arguments['registerMockObjectsFromTestArgumentsRecursively']= $arguments['registerMockObjectsFromTestArgumentsRecursively'] ?? false;
```
keradus: a space is missing before `=`
after solving that, alignment will be back to original one again

keradus (approved):
looks damn good piece of work here !

src/TextUI/TestRunner.php
epdenouden: I tried something like that in the past, let's see if your suggestion works nicer.

epdenouden:
Apologies for the force push @keradus, but resolving text-based-diff 'conflicts' got a bit tiresome.

epdenouden:
@sebastianbergmann Hi Sebastian! @keradus and I think this pull request is done. For new ideas and improvements I will open new PRs in the near future. :)

Squashing the commits before merge seems a good idea, due to the high number of refactoring cycles during development.

keradus (approved):
double checked after squash, looks same as my local copy before squashing

sebastianbergmann:
Thanks!

epdenouden:
🎁 thanks for the merge! :)

keradus:
:fireworks:

PR #3148, opened 2018-05-28
AppVeyor: Workaround for PHP release server limit

mhujer:


mhujer:
@sebastianbergmann https://ci.appveyor.com/project/sebastianbergmann/phpunit/build/790.master-pihkgdkd/job/q255bt1t5f8ai187

The server returns this instead of the zip file:

```
20/feb/2018: Hi! We seem to be receiving high volume requests coming from empty user agents. While this shouldn't be an issue, this unfortunately resulted in bandwidth issues on this server causing all downloads to be unavailable. We've therefore decided to temporarily block empty user agents until we upgraded our server bandwidth.
03/mar/2018: We've upgraded the server bandwidth. This is however still not sufficient to handle all empty user agent connections. Please update the user agent in your scripts accordingly or contact us so we can discuss it.
Thank you for your understanding. 
```

I will try to fix it somehow.

mhujer:
OK, this is ready for review.

Do you have any idea who from php.net we can ping to let them know that is is probably AppVeyor causing the load on the servers?

sebastianbergmann:
My guess would be Anatol Belski (@weltling here on GitHub).

PR #3149, opened 2018-05-30
--list-tests output shows number of available tests

chapeupreto:
Instead of outputing something like 

```plain
Available test(s):
 - PHPUnit\Framework\AssertTest::testFail
 - PHPUnit\Framework\AssertTest::testAssertSplObjectStorageContainsObject
 - PHPUnit\Framework\AssertTest::testAssertArrayContainsObject
 - PHPUnit\Framework\AssertTest::testAssertArrayContainsString
```

change `--list-tests` in order to output something like

```plain
4 available test(s):
 - PHPUnit\Framework\AssertTest::testFail
 - PHPUnit\Framework\AssertTest::testAssertSplObjectStorageContainsObject
 - PHPUnit\Framework\AssertTest::testAssertArrayContainsObject
 - PHPUnit\Framework\AssertTest::testAssertArrayContainsString
```

### DIFF
--- a/src/Util/TextTestListRenderer.php
+++ b/src/Util/TextTestListRenderer.php
@@ -18,7 +18,7 @@ final class TextTestListRenderer
 {
     public function render(TestSuite $suite): string
     {
-        $buffer = 'Available test(s):' . PHP_EOL;
+        $buffer = \sprintf('%d available test(s):%s', $suite->count(), PHP_EOL);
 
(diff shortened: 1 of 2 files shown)


keradus:
:-1: for inconsistency between `--list-groups` and `--list-suites` vs `--list-tests`

Also, I run it on concrete repo - PHP CS Fixer. I got `15052` tests. Not sure what value that number brings...

PR #3161, opened 2018-06-06
Support indexed arrays in ArraySubset

### LINKED ISSUE #3101: Array Subset does not work as expected on indexed / flat arrays
| Q                   | A
| PHPUnit version     | 7.1.4
| PHP version         | 7.1.16

Digging up a older issue which i do not feel is explained properly, see #2781 and #2069.

In regards to the comment you wrote, "is it fixed in  #2237?", no it is not. The the usage of array_replace_recursive and how indexed arrays work in PHP is the problem here. This seem to confuse a lot of people me included. If it works as intended is up for debate, but i will try to explain the confusion.

In set theories the set `{1,2,3}` will have the subset `{2,3}`. If we use `array_replace_recursive` with this. It will use the array `[0 => 1, 1 => 2, 2 => 3]` and replace the keys with the subset `[0 => 2, 1 => 3]`, which for this usage of `array_replace_recursive` should instead be `[1 => 2, 2 => 3]`, to make {2,3} pass as a subset, with the way the current code is written. Yes i agree, that arrays are not sets, but i feel this is how people logically assert these things in their minds. So replacing array_replace_recursive on indexed arrays would solve this as the implementer @marcioAlmada of the subset class has suggested in @2069.

pfrenssen:
This fixes #3101 and earlier efforts #2781 and #2069.

I have added support for indexed arrays, and it is now possible to correctly check multidimensional arrays. It allows to match indexed arrays that are out of order, so for example if you have an array `[0, 1, 2, 3]` you can assert that `[3, 1]` is a valid subset.

Because of this out-of-order matching there is a known limitation, the code loops over the data set and checks every data point and returns on the first match. This means it will generate a false positive if multiple identical values are present in the subset. For example if an array `[0, 1, 2, 3]` is given and you assert the subset `[0, 1, 1, 2]` then it will return `TRUE` even though `1` is present only once in the data set. I wasn't sure if the added complexity for supporting this edge case is worth it.

This is my first contribution to PHPUnit and I am not sure on a few points:

1. I would like to include this in both PHPUnit 6 and 7, but I don't know the procedure to follow. This PR is rolled against the current master. Can this be backported or am I supposed to roll this for PHPUnit 6 first?
2. I am heavily relying on anonymous functions, I don't know if this fits the code style of PHPUnit, or would it be prefered to rework this using private methods?

### DIFF
--- a/src/Framework/Constraint/ArraySubset.php
+++ b/src/Framework/Constraint/ArraySubset.php
@@ -117,6 +117,8 @@ class ArraySubset extends Constraint
 
                                 if (!empty($recursed)) {
                                     $intersect[$key] = $recursed;
+
+                                    break;
                                 }
                             }
                         }
--- a/src/Framework/Constraint/ArraySubset.php
+++ b/src/Framework/Constraint/ArraySubset.php
@@ -59,97 +59,16 @@ class ArraySubset extends Constraint
      */
     public function evaluate($other, $description = '', $returnResult = false)
     {
-        // Anonymous function that checks whether the given array is
-        // associative.
-        $is_associative = function (array $array): bool {
-            return \array_reduce(\array_keys($array), function (bool $carry, $key): bool {
-                return $carry || \is_string($key);
-            }, false);
-        };
-
-        // Anonymous function that compares the two given values using either
-        // strict or loose comparisons.
-        $strict  = $this->strict;
-        $compare = function ($first, $second) use ($strict): bool {
-            return $strict ? $first === $second : $first == $second;
-        };
-
-        // Anonymous function that sorts the given multidimensional array.
-        $deep_sort = function (array &$array) use (&$deep_sort, $is_associative): void {
-            foreach ($array as &$value) {
-                if (\is_array($value)) {
-                    $deep_sort($value);
-                }
-            }
-
-            if ($is_associative($array)) {
-                \ksort($array);
-            } else {
-                \sort($array);
-            }
-        };
-
-        $array_intersect_recursive = function (array $array, array $subset) use (&$array_intersect_recursive, $is_associative, $compare): array {
-            $intersect = [];
-
-            if ($is_associative($subset)) {
-                // If the subset is an associative array, get the intersection
-                // while preserving the keys.
-                foreach ($subset as $key => $subset_value) {
-                    if (\array_key_exists($key, $array)) {
-                        $array_value = $array[$key];
-
-                        if (\is_array($subset_value) && \is_array($array_value)) {
-                            $intersect[$key] = $array_intersect_recursive($array_value, $subset_value);
-                        } elseif ($compare($subset_value, $array_value)) {
-                            $intersect[$key] = $array_value;
-                        }
-                    }
-                }
-            } else {
-                // If the subset is an indexed array, loop over all entries in
-                // the haystack and check if they match the ones in the subset.
-                foreach ($array as $array_value) {
-                    if (\is_array($array_value)) {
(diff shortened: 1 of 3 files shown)


sebastianbergmann:
I prefer private methods over anonymous functions.

sebastianbergmann:
If this is a bug fix and the respective bug is in `6.5` then this should go into `6.5`. It will then be merged from there to `7.2` and `master`.

If this is an enhancement it will go into `master` only.

pfrenssen:
Thanks for the quick reply! I will refactor this using private methods.

Regarding or not whether this is a bug fix, I think this can be debatable. I personally think of this as a bug since the current documentation does not mention that it is only designed to work with associative arrays:

```
/**
 * Asserts that an array has a specified subset.
 *
 * @param array|ArrayAccess $subset
 * @param array|ArrayAccess $array
 * @param bool              $strict  Check for object identity
 * @param string            $message
 *
 * @throws Exception
 */
function assertArraySubset($subset, $array, bool $strict = false, string $message = ''): void;
```

When I read this documentation, then I would expect the following to return `TRUE`, but it doesn't. It feels a lot like a bug:

```
$this->assertArraySubset( ['b', 'c'], ['a', 'b', 'c', 'd'] );
```

However after I started working on this I read in https://github.com/sebastianbergmann/phpunit/issues/2069#issuecomment-260271597 that @marcioAlmada was only targeting associative arrays in the original implementation. So I guess from his point of view this is a new feature.

sebastianbergmann:
I think it should go into `master` then. Unless it changes/breaks current behaviour. Then it would have to wait unt

pfrenssen:
OK! To the best of my knowledge there should not be a B/C break. I took care not to break the existing implementation for associative arrays, and have not touched the original tests, they are still passing.

If people are using the current implementation for indexed arrays then there might be some unexpected failures appearing, but this will actually be because bugs in their code are now being correctly detected.

For example, this will previously have returned `FALSE`, but now it will (correctly) return `TRUE`:

```
$this->assertArraySubset( ['b', 'c'], ['a', 'b', 'c', 'd'] );
```

Is it the policy of PHPUnit to never cause any existing test to change output, even if the behaviour relied on is objectively broken?

pfrenssen:
Thanks!

src/Framework/Constraint/ArraySubset.php
```diff
@@ -130,4 +128,83 @@ private function toArray(iterable $other): array
...
+    private function deepSort(array &$array): void
+    {
+        foreach ($array as &$value) {
+            if (\is_array($value)) {
+                $this->deepSort($value);
+            }
+        }
+
+        if ($this->isAssociative($array)) {
+            \ksort($array);
```
chx: This is weird. Are you sure this shouldn't be `asort`?

pfrenssen: The goal is to be able to compare the two arrays, and so both `ksort()` and `asort()` do the trick for this use case.

chx: Got it, as long as they end up in the same order, you are good. Thanks for the quick reply.

taylorotwell:
This was a breaking change. Broke the existing Laravel framework tests.

sebastianbergmann:
@taylorotwell Please be open a ticket for this breaking change (and be more specific than "Broke the existing Laravel framework tests"). Thanks!

pfrenssen:
@taylorotwell thanks for the report. Can you check if the tests still fail if you run them using PR #3186?

In that PR support is added for checking subsets that contain duplicate values. This might be the cause of the failures.

PR #3167, opened 2018-06-14
fix stdout in testcase

robberphex:
For testcase:
```php
<?php

/**
 * @runTestsInSeparateProcesses
 */
class ATest extends TestCase
{
    public function testA()
    {
        $res = fopen('php://stdout','w');
        fclose($res);
        $this->assertEquals(1, 2);
    }
}
```
Because child php process stdout will closed after testA, phpunit cannot get error in child process, so will mark this case passed.
**Phpunit should open standalone `php://stdout`**.

---
Some logging library, such as [log4php](https://github.com/apache/logging-log4php/blob/master/src/main/php/appenders/LoggerAppenderConsole.php#L44), will open standalone `php://stdout`, And then the subprocess test result will lost.

### DIFF
--- a/src/Util/PHP/Template/TestCaseClass.tpl.dist
+++ b/src/Util/PHP/Template/TestCaseClass.tpl.dist
@@ -47,6 +47,11 @@ function __phpunit_run_isolated_test()
     $result->beStrictAboutTodoAnnotatedTests({isStrictAboutTodoAnnotatedTests});
     $result->beStrictAboutResourceUsageDuringSmallTests({isStrictAboutResourceUsageDuringSmallTests});
 
+    // If testcase open stdout and close it, the output will be lost.
+    // So we open it manually.
+    $stdout = fopen('php://stdout', 'w');
+
+    /** @var TestCase $test */
     $test = new {className}('{name}', unserialize('{data}'), '{dataName}');
     $test->setDependencyInput(unserialize('{dependencyInput}'));
     $test->setInIsolation(TRUE);
@@ -59,8 +64,8 @@ function __phpunit_run_isolated_test()
     }
 
     @rewind(STDOUT); /* @ as not every STDOUT target stream is rewindable */
-    if ($stdout = stream_get_contents(STDOUT)) {
-        $output = $stdout . $output;
+    if ($testcaseOutput = stream_get_contents(STDOUT)) {
+        $output = $testcaseOutput . $output;
         $streamMetaData = stream_get_meta_data(STDOUT);
         if (!empty($streamMetaData['stream_type']) && 'STDIO' === $streamMetaData['stream_type']) {
             @ftruncate(STDOUT, 0);
@@ -76,6 +81,7 @@ function __phpunit_run_isolated_test()
         'output'        => $output
       )
     );
+    fclose($stdout);
 }
 
 $configurationFilePath = '{configurationFilePath}';
--- a/src/Util/PHP/Template/TestCaseMethod.tpl.dist
+++ b/src/Util/PHP/Template/TestCaseMethod.tpl.dist
@@ -48,6 +48,10 @@ function __phpunit_run_isolated_test()
     $result->beStrictAboutTodoAnnotatedTests({isStrictAboutTodoAnnotatedTests});
     $result->beStrictAboutResourceUsageDuringSmallTests({isStrictAboutResourceUsageDuringSmallTests});
 
+    // If testcase open stdout and close it, the output will be lost.
+    // So we open it manually.
+    $stdout = fopen('php://stdout', 'w');
+
     /** @var TestCase $test */
     $test = new {className}('{methodName}', unserialize('{data}'), '{dataName}');
     $test->setDependencyInput(unserialize('{dependencyInput}'));
@@ -61,8 +65,8 @@ function __phpunit_run_isolated_test()
     }
 
     @rewind(STDOUT); /* @ as not every STDOUT target stream is rewindable */
-    if ($stdout = stream_get_contents(STDOUT)) {
-        $output = $stdout . $output;
+    if ($testcaseOutput = stream_get_contents(STDOUT)) {
+        $output = $testcaseOutput . $output;
         $streamMetaData = stream_get_meta_data(STDOUT);
         if (!empty($streamMetaData['stream_type']) && 'STDIO' === $streamMetaData['stream_type']) {
             @ftruncate(STDOUT, 0);
@@ -78,6 +82,7 @@ function __phpunit_run_isolated_test()
         'output'        => $output
       )
     );
+    fclose($stdout);
 }
 
 $configurationFilePath = '{configurationFilePath}';
--- a/tests/Regression/GitHub/2725-separate-class-before-after-pid.phpt
+++ b/tests/Regression/GitHub/2725-separate-class-before-after-pid.phpt
@@ -12,7 +12,6 @@ PHPUnit\TextUI\Command::main();
 PHPUnit %s by Sebastian Bergmann and contributors.
 
 ..                                                                  2 / 2 (100%)
-@afterClass output - PID difference should be zero: 0
 
 Time: %s, Memory: %s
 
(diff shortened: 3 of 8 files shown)


RobberPhex:
@sebastianbergmann Could you review this pr?

tests/Regression/GitHub/3167-class.phpt
```diff
@@ -0,0 +1,32 @@
+--TEST--
+https://github.com/sebastianbergmann/phpunit/pull/3167
+--SKIPIF--
+<?php
+if (defined('HHVM_VERSION')) {
+    print 'skip: PHP runtime required';
```
keradus: why ?

sebastianbergmann: HHVM is not supported.

sebastianbergmann:
I do not understand the problem this wants to solve. Can you elaborate? Thanks!

tests/Regression/GitHub/3167-class.phpt
keradus: I know, especially why I see no point of putting the condition here.
We don't do it in other tests, right?
Because HHVM is not supported by whole project, not just this one, concrete test.

Like... we don't support PHP4, yet we don't check it in each phpt file ;)

composer is perfect place to specify required engine and version

sebastianbergmann: `s/HHVM is not supported/HHVM is not supported, we do not need this check/`

Happy now? :-)

RobberPhex:
Removed hhvm detection in phpt

tests/Regression/GitHub/3167-class.phpt
keradus: haha, yeah, thanks :+1:

epdenouden: A+ nitpicking, would buy again

src/Util/PHP/Template/TestCaseMethod.tpl.dist
```diff
@@ -27,6 +27,10 @@ if ($composerAutoload) {
 
 function __phpunit_run_isolated_test()
 {
+    // if testcase open stdout and close it, the output will lost.
```
keradus: Full dot at the end, but sentence is not started with upper case.
`will be lost`

RobberPhex: Fixed

sebastianbergmann:
I now understand what this issue is about. However, I am not yet sure on potential (adverse) side effects of the proposed change.

@keradus What do you think (apart from test structure)?

keradus:
I do have some remarks, but I need to go deeper into it. Will try to do in this Monday

sebastianbergmann:
@keradus Much appreciated, thanks!

RobberPhex:
Is there anyone can merge this to fix this **serious bug**?

RobberPhex:
@sebastianbergmann Is there anyone working on this?
This is a confirmed bug, and all tests passed. Why are we wasting time on review? What are we waiting for?

tests/Regression/GitHub/2725-separate-class-before-after-pid.phpt
```diff
@@ -12,7 +12,6 @@ PHPUnit\TextUI\Command::main();
 PHPUnit %s by Sebastian Bergmann and contributors.
 
 ..                                                                  2 / 2 (100%)
-@afterClass output - PID difference should be zero: 0
```
sebastianbergmann: Please explain why (the expectation of) an existing test case had to be modified.

tests/Regression/GitHub/2725/BeforeAfterClassPidTest.php
```diff
@@ -32,6 +32,10 @@ public function testThatClassDidNotReload()
      */
     public static function showPidAfter()
     {
-        echo "\n@afterClass output - PID difference should be zero: " . ($GLOBALS['PID_BEFORE'] - getmypid());
```
sebastianbergmann: Please explain why an existing test case had to be modified.

sebastianbergmann:
> This is a confirmed bug

So far the bug has only been confirmed by you.

> all tests passed

This is not correct: the pull request modifies the expectation of an existing test case.

> Why are we wasting time on review

I am sorry if I take too long to review pull requests for your patience.

sebastianbergmann:
@keradus Did you get a chance to look into this?

tests/Regression/GitHub/2725/BeforeAfterClassPidTest.php
keradus: it has not,
logic here was kept the same,
only difference is that originally we saw only difference of PIDs, now we see their values, which helps debuggings

I agree that it's not nice to never display it, as you don't know if PIDs are same or they were not executed at all

keradus:
> Why are we wasting time on review

review is important process, and you found out really edge case IMO.
software that big and complex, heavily used around the world simply need to ensure new changes won't break existing integrations.


----------------

@sebastianbergmann , I can confirm the behaviour and I kind of able to agree this is a bug, but it affects super few users. (Proof? it's there since years, nobody noticed)
Thing is, that is not usual to print sth to stdout during tests, it's considered risky (and simply not clean), for that RobberPhex putted it into separated process.
I don't consider putting sth directly to stdout as good thing personally, when I need to test output of my app, I let it print to abstracted output and then capture it in tests with buffer, eg `Symfony\Component\Console\Output\BufferedOutput`. Yet, if one don't want to do that, eg he has super low-level script, why to forbid him ?

After saying that, I do not see this change breaking sth, I did also run it across my heavier projects, all good. As always, your decision do you want to incorporate the change or not (especially considering amount of affected users).

On a side note, I will raise a minor concern about style of solution soon

src/Util/PHP/Template/TestCaseClass.tpl.dist
```diff
@@ -26,6 +26,10 @@ if ($composerAutoload) {
 
 function __phpunit_run_isolated_test()
 {
+    // If testcase open stdout and close it, the output will be lost.
+    // So we open it manually.
+    $__stdout = fopen('php://stdout', 'w');
```
keradus: 1. I see a confusion between this handler and `STDOUT`, especially as both are used in very same method

2. I saw issues that raised from doing that things - opening a file creates a file handler, which are not free and unlimited - vide `FD_SETSIZE`, please close it as soon as it's not needed anymore

3. btw, it's only variable here with leading `__`, what for?

RobberPhex: 1. As global, `STDOUT` may be closed by a test case. When `STDOUT` closed, the test result in subprocess will be lost.
2. I will rename this variable.
3. If there is a bug, and we can fix it, why not fix it quickly?

keradus: 1. still confusing, you just gave next reason to use `$stdout` over `STDOUT` in `stream_get_contents(STDOUT)`
2. :+1:
3. As Sebastian and me already answered, to ensure this is a bug and doesn't bring new issues

tests/Regression/GitHub/3167/Issue3167ClassTest.php
```diff
@@ -0,0 +1,23 @@
...
+use PHPUnit\Framework\TestCase;
+
+/**
+ * @runClassInSeparateProcess
+ */
+class Issue3167ClassTest extends TestCase
+{
+    public function testSTDOUT(): void
+    {
+        $fp = \fopen('php://stdout', 'w');
```
keradus: would be nice to have sibling case with stderr

RobberPhex: This bug doesn't affect STDERR.
And even user close stderr, the test case works, it means the phpt always passed.

keradus: awesome it works, let us ensure it won't stop working after some refactoring in future

RobberPhex: Hi, @keradus .
1. this bug doesn't affect STDERR, so testcase for STDERR is unnecessary.
2. Currently, this pr cannot be merged because of *some refactoring in future*? Am I misunderstand? *We cannot hold a bugfix just for refactoring.*

---
And, if there is anything I can do, please tell me. Thanks.

PR #3169, opened 2018-06-15
Fix #3093 by improving test name handling in dependency resolver

### LINKED ISSUE #3093: Unable to chain a @dataProvider in method a with a @depends in method b
| Q                   | A
| PHPUnit version     | 7.0.2
| PHP version         | 7.1.10

Here is a code snippet of what I mean in the title

	/**
	 * @return array
	 */
	public function demoDP(): array 
	{
		return [[0], [1], [2], [3]];
	}

	/**
	 * @dataProvider demoDP
	 * @param int $nr
	 * @return int
	 */
	public function testDemoA(int $nr): int
	{
		self::assertTrue($nr >= 0);
		return $nr + 1;
	}

	/**
	 * @depends testDemoA
	 * @param int $nr
	 */
	public function testDemoB(int $nr)
	{
		self::assertTrue($nr > 0);
	}

I would expect `testDemoA` to run 4 times: with `$nr` equal to 0, 1, 2 and 3. I also expect `testDemoB` to run 4 times: with `$nr` equal to 1, 2, 3 and 4.

`testDemoA` works fine, but `testDemoB` gives the error `Argument 1 passed to <testclass>::testDemoB() must be of the type integer, null given

If I don't use the dataprovider for `testDemoA`, but a hardcoded value for `$nr` instead, `testDemoB` works fine, so the problem lies with using `@depends` in one method that refers to a method that uses a dataprovider.

epdenouden:
Fix for #3093. The dependency resolver didn't use the long format for test names causing it to overlook some dependencies when mixing tests with and without `@dataprovider`. Fixed by always normalizing test names.

- format names as `TestClassName::testMethodName` internally for better matching
- the resolver doesn't need to check dependencies between data rows in a `DataProviderTestSuite`
- regression test from @bcremer included

### DIFF
--- a/src/Runner/TestSuiteSorter.php
+++ b/src/Runner/TestSuiteSorter.php
@@ -63,7 +63,7 @@ final class TestSuiteSorter
             $suite->setTests($this->randomize($suite->tests()));
         }
 
-        if ($resolveDependencies && $this->suiteOnlyContainsTests($suite)) {
+        if ($resolveDependencies && !($suite instanceof DataProviderTestSuite) && $this->suiteOnlyContainsTests($suite)) {
             $suite->setTests($this->resolveDependencies($suite->tests()));
         }
     }
--- a/src/Runner/TestSuiteSorter.php
+++ b/src/Runner/TestSuiteSorter.php
@@ -150,7 +150,7 @@ final class TestSuiteSorter
             $testClass = \get_class($test);
         }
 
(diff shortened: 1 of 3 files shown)


sebastianbergmann:
Is this a bugfix that should go into PHPUnit 7.2? If so, please send a pull request against `7.2` if it's not too much trouble. Thanks!

epdenouden:
Yes, it's a small fix. Would be good to have it in 7.2.x :)

PR #3170, opened 2018-06-18
Ensure environment values are stored in $_ENV

IchHabRecht:
in \PHPUnit\Util\Configuration

### DIFF
--- a/src/Util/Configuration.php
+++ b/src/Util/Configuration.php
@@ -568,6 +568,8 @@ class Configuration
                 \putenv("{$name}={$value}");
             }
 
+            $value = \getenv($name);
+
             if (!isset($_ENV[$name])) {
                 $_ENV[$name] = $value;
             }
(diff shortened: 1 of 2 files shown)


sebastianbergmann:
Looks like a bugfix that should go at least into `7.2`. Does it also need to go into `6.5`?

IchHabRecht:
@sebastianbergmann 
I would love to have it in both versions, yes. I didn't found any information in your contribution workflow how to deal with the branches/versions. Do you need pull requests for each branch instead of the one against master?

sebastianbergmann:
One pull request is enough. If it's for a bugfix then it should be made against the oldest branch still supported, which is `6.5` at this point in time.

IchHabRecht:
Thank you for the fast merge!

sebastianbergmann:
Thank you for the pull request :)

PR #3172, opened 2018-06-19
fix stdout in testcase 7.x

robberphex:
fix #3167 at 7.x version


sebastianbergmann:
Duplicate of #3167.

When/If #3167 is merged into `6.5`, the changes will be merged from there to `7.2` and `master`.

PR #3179, opened 2018-06-21
DX: Assert - use assertSame over assertEquals

keradus:


sebastianbergmann:
I am not convinced that this does not have any unexpected side effects.

keradus:
assertEquals -> assertSame is changed only for comparing strings and ints, works the same

about dropping `if (\is_bool($expected) && \is_bool($actual)) {` - I found no reason why it's there, `IsIdentical` works good for comparing booleans. No test failed as well...

keradus:
thanks for merging :+1:

keradus:
@sebastianbergmann , sure I will !
I'm a bit overwhelmed right now, sorry for delays on this on other topics...

PR #3182, opened 2018-06-22
Update .editorconfig to match existing XSD for comfort

epdenouden:
I could also reformat `phpunit.xsd` if you like that better. :)


sebastianbergmann:
I think I would rather have the XSD reformatted, thanks.

epdenouden:
Will do!

PR #3186, opened 2018-06-27
Add support for checking subsets that contain duplicate values to ArraySubset

pfrenssen:
In #3161 support was added for checking indexed arrays in `ArraySubset`. The edge case of checking subsets that contain multiple identical values was left out of the scope of that PR. This means that in the current implementation the following example will generate a false positive:

* Array: `[0, 1, 2, 3]`
* Subset: `[0, 1, 1, 2]`
* Expected result: `FALSE` since the value `1` is only present once in the array, but twice in the subset.
* Actual result: `TRUE`

This PR adds support for this edge case, so that it will become possible to check subsets with duplicate values.

### DIFF
--- a/src/Framework/Constraint/ArraySubset.php
+++ b/src/Framework/Constraint/ArraySubset.php
@@ -149,7 +149,7 @@ class ArraySubset extends Constraint
         foreach ($array as &$value) {
             if (\is_array($value)) {
                 $this->deepSort($value);
-            } elseif (!\is_string($value) || empty($value)) {
+            } elseif (\is_scalar($value) && !\is_string($value) || empty($value)) {
                 // In order to ensure consistent sorting results in data sets
                 // that have a mix of strings and integers we need to perform
                 // string comparisons on all scalar values. Typically this is
--- a/src/Framework/Constraint/ArraySubset.php
+++ b/src/Framework/Constraint/ArraySubset.php
@@ -149,6 +149,13 @@ class ArraySubset extends Constraint
         foreach ($array as &$value) {
             if (\is_array($value)) {
                 $this->deepSort($value);
+            } elseif (!\is_string($value) || empty($value)) {
+                // In order to ensure consistent sorting results in data sets
+                // that have a mix of strings and integers we need to perform
+                // string comparisons on all scalar values. Typically this is
+                // done by passing the SORT_STRING flag to a sorting function
+                // but this doesn't work on multidimensional arrays.
+                $value = (string) (int) $value;
             }
         }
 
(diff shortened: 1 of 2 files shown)


KristopherWindsor:
We have a few tests that started failing when phpunit 7.3 was released.
Not good to have a minor phpunit version break tests. 😭  Anyway, we think this PR will make our tests pass again.

```
07:16:26 There were 4 failures:
07:16:26 
07:16:26 1) CompanyTest\Integration\Lib\Model\Test\ClassNameTest::testHandleFollowUpCacheKeyCallSettingOffAwaySettingOnNoSentReply
07:16:26 Failed asserting that an array has the subset Array &0 (
07:16:26     0 => '30'
07:16:26     1 => '14097220017'
07:16:26     2 => 'A'
07:16:26     3 => 'F'
07:16:26     4 => 'F'
07:16:26     5 => 'N'
07:16:26     6 => 'P'
07:16:26     7 => 'SMS'
07:16:26     8 => 'ProductName'
07:16:26 ).
07:16:26 --- Expected
07:16:26 +++ Actual
07:16:26 @@ @@
07:16:26  Array
07:16:26  (
07:16:26 -    [0] => 30
07:16:26 -    [1] => 14097220017
07:16:26 +    [1] => 30
07:16:26      [2] => A
07:16:26 -    [3] => F
07:16:26 -    [4] => F
07:16:26 -    [5] => N
07:16:26 -    [6] => P
07:16:26 -    [7] => SMS
07:16:26 -    [8] => ProductName
07:16:26 +    [3] => 14097220007
07:16:26 +    [4] => P
07:16:26 +    [5] => F
07:16:26 +    [6] => ProductName
07:16:26 +    [7] => F
07:16:26 +    [8] => N
07:16:26 +    [9] => SMS
07:16:26  )
```

pfrenssen:
@KristopherWindsor did you try it with this PR? I see you have duplicate values in your subset. That's what this PR is fixing. If it is not, could you provide a test case so I can have a look?

pfrenssen:
In #3240 a regression was reported against `assertArraySubset()`, but this problem no longer occurs in this PR. I have included the test case for that issue to prove that this is fixed here.

RikSomers:
@pfrenssen We use the Laravel Codebase, and serveral tests stopped working when upgrading to PHPUnit 7.3.0 due to the ArraySubset change.

I can confirm this PR fixes any failing tests that were introduced with the 7.3.0 changes.

sebastianbergmann:
@pfrenssen Can you target the `7.3` branch, please? Thanks!

sebastianbergmann:
I have reverted #3161 because of #3240.

@pfrenssen Please send a new pull request against `master` that contains both #3161 and #3186. I am sorry for having to ask this of you -- and that you have to wait until PHPUnit 7.4 to get these improvements into a stable release of PHPUnit.

kmbt:
Having some background in mathematics and also in python programming language I find this way of thinking really confusing.

In mathematical set theory, an object just may be or may not be a member of a set. In set theory there is no concept of an object being a member "more than once" of a set.

This reasoning is applied in python, where the following expression evaluates as `True`:
```
set([[0, 1, 1, 2]) == set([0, 1, 2])
```

Please kindly take my remarks into your consideration.

Stadly:
@pfrenssen Did you have a chance to create a new PR containing both https://github.com/sebastianbergmann/phpunit/pull/3161 and https://github.com/sebastianbergmann/phpunit/pull/3186? Would you like me to look into it?

Issue #3111, opened 2018-05-03
Coverage PHPT tests


ABashkirtseva:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 7.1.15
| PHP version         | 7.1.16
| Installation Method | Composer

Run simple PHPT test

```
--FILE--
<?php
  var_dump(1);
?>
--EXPECT--
int(1)
```
by PHPUnit command with phpdbg. But I've got the error

TypeError: Return value of PHPUnit\Runner\PhptTestCase::cleanupForCoverage() must be of the type array, boolean returned in /home/anastasia/foundation-packages/visualxmleditor/trunk/vendor/phpunit/phpunit/src/Runner/PhptTestCase.php:547

> composer info | sort
```
cilex/cilex                         1.1.0   The PHP micro-framework for Command line tools based on the Symfony2 Components
cilex/console-service-provider      1.0.0   Console Service Provider
clue/graph                          v0.9.0  A mathematical graph/network library written in PHP
container-interop/container-interop 1.2.0   Promoting the interoperability of container objects (DIC, SL, etc.)
doctrine/annotations                v1.6.0  Docblock Annotations Parser
doctrine/collections                v1.5.0  Collections Abstraction library
doctrine/instantiator               1.1.0   A small, lightweight utility to instantiate objects in PHP without invoking their constructors
doctrine/lexer                      v1.0.1  Base library for a lexer that can be used in Top-Down, Recursive Descent Parsers.
erusev/parsedown                    1.7.1   Parser for Markdown.
graphp/algorithms                   v0.8.1  Common mathematical graph algorithms
graphp/graphviz                     v0.2.1  GraphViz graph drawing for mathematical graph/network
herrera-io/json                     1.0.3   A library for simplifying JSON linting and validation.
herrera-io/phar-update              1.0.3   A library for self-updating Phars.
jms/metadata                        1.6.0   Class/method/property metadata management in PHP
jms/parser-lib                      1.0.0   A library for easily creating recursive-descent parsers.
jms/serializer                      1.11.0  Library for (de-)serializing data of any complexity; supports XML, JSON, and YAML.
justinrainbow/json-schema           1.6.1   A library to validate a json schema.
kherge/version                      1.0.1   A parsing and comparison library for semantic versioning.
logics/baselib                      0.1.6   Base library
logics/cli                          0.2.2   Command Line Interface base class providing command line parsing facilities
[...]

PR #3200, opened 2018-07-07
Fix test and expectations with phpdbg sapi

kabel:
Fixes the following differences when running with phpdbg:

* phpdbg doesn't ignore the -- argument separator when passed to php
* phpdbg doesn't support display_errors=stderr, so errors go to STDOUT
* Tests that enable xdebug.scream affect process isolation in phpdbg
* Errors/Exceptions/Stacks are formatted slightly different in phpdbg

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,7 @@ before_script:
 
 script:
   - ./phpunit --coverage-clover=coverage.xml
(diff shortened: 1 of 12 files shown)


kabel:
Closing to rework into #3111, which this pretty much depends on.

PR #3205, opened 2018-07-10
Fixing method covers taking priority over class coversNothing.

Danack:
I tried to use the feature sent in the PR https://github.com/sebastianbergmann/phpunit/pull/3115

It didn't work due to the code that this PR fixes.

This PR obviously needs at least one test adding, and probably more. Please could you point me in the right direction for where those tests should be/go?


Idrinth:
would propose adding another file to tests/_files/ with a class and it's test(class:no,method:yes).

Danack:
@Idrinth that was included in the original PR and is present in the [current files](https://github.com/sebastianbergmann/phpunit/blob/master/tests/_files/CoverageCoversOverridesCoversNothingTest.php).

So presumably a different test elsewhere is needed.

src/Framework/TestResult.php
```diff
@@ -573,8 +573,13 @@ public function run(Test $test): void
 
             $annotations = $test->getAnnotations();
 
-            if (isset($annotations['class']['coversNothing']) || isset($annotations['method']['coversNothing'])) {
+            if (isset($annotations['method']['coversNothing'])) {
                 $coversNothing = true;
+            } elseif (isset($annotations['class']['coversNothing'])) {
+                if (!isset($annotations['method']['covers'])) {
```
Idrinth: could be merged into one if with the previous line

Idrinth:
my mistake - let me blame it on the time of day(and just ignore me otherwise ;) ).
https://codecov.io/gh/sebastianbergmann/phpunit/pull/3205/src/src/Util/Test.php#L1123 is missing from the original tests - so a coverless class without method annotations.
I assume that would also cover the result cases.

Danack:
@Idrinth that does need a test adding, but is separate thing to the lines affected by this PR which seem to be the problem.

Idrinth:
@Danack I believe both are the same missing test - just slightly different stages of processing it https://codecov.io/gh/sebastianbergmann/phpunit/pull/3205/src/src/Framework/TestResult.php?before=src/Framework/TestResult.php#L581

Idrinth:
Something along these lines maybe @Danack ?

```
diff --git a/tests/Framework/CodeCoverageStub.php b/tests/Framework/CodeCoverageStub.php
new file mode 100644
index 0000000..9fc3f26
--- /dev/null
+++ b/tests/Framework/CodeCoverageStub.php
@@ -0,0 +1,15 @@
+<?php
+
+namespace PHPUnit\Framework;
+
+class CodeCoverageStub
+{
+    public $started = 0;
+    public $stopped = 0;
+    public function start() {
+        $this->started++;
+    }
+    public function stop() {
+        $this->stopped++;
+    }
+}
\ No newline at end of file
diff --git a/tests/Framework/TestResultTest.php b/tests/Framework/TestResultTest.php
index b88e03f..406cba7 100644
--- a/tests/Framework/TestResultTest.php
+++ b/tests/Framework/TestResultTest.php
@@ -62,4 +62,42 @@
         $this->assertAttributeEquals(true, 'lastTestFailed', $result);
         $this->assertAttributeContainsOnly(TestFailure::class, 'notImplemented', $result);
     }
+
+    private function generateTestWithAnnotations(array $annotations): Test
+    {
+        $test = $this->getMockBuilder(TestCase::class)->disableOriginalConstructor()->getMock();
+        $test->expects($this->once())
+            ->method('getAnnotations')
+            ->with()
+            ->willReturn($annotations);
+        return $test;
+    }
+    public function testCoverageForNoCoverageClass(): void
+    {
+        $test   = $this->generateTestWithAnnotations(['class' => ['coversNothing' => true], 'method' => []]);
+        require_once 'CodeCoverageStub.php';
+        $cover  = new CodeCoverageStub();
+        $result = new TestResult();
+        $coverage = new \ReflectionProperty($result, 'codeCoverage');
+        $coverage->setAccessible(true);
+        $coverage->setValue($result, $cover);
+        $coverage->setAccessible(false);
+        $result->run($test);
+        $this->assertEquals(0, $cover->started);
+        $this->assertEquals(0, $cover->stopped);
+    }
+    public function testCoverage(): void
+    {
+        $test   = $this->generateTestWithAnnotations(['class' => [], 'method' => []]);
+        require_once 'CodeCoverageStub.php';
+        $cover  = new CodeCoverageStub();
+        $result = new TestResult();
+        $coverage = new \ReflectionProperty($result, 'codeCoverage');
+        $coverage->setAccessible(true);
+        $coverage->setValue($result, $cover);
+        $coverage->setAccessible(false);
+        $result->run($test);
+        $this->assertEquals(1, $cover->started);
+        $this->assertEquals(1, $cover->stopped);
+    }
 }
```

Danack:
I moved the logic for testing if any coverage is needed to be it's own function to avoid needing those complex tests.

The remaining codestyle issue reported by Travis are not related to this PR.

sebastianbergmann:
Can you please make this pull request against `7.2`? AFAICS, this is a bugfix.

Danack:
I'll make a new pr after rebasing.

Issue #10, opened 2010-08-20
Refactor how tests are run


sebastianbergmann:
- Move test execution logic out of `TestResult`
  - Use `TestResult` just as a Collecting Parameter to collect test results
- Before the test execution starts
  - Use a `TestSuiteIterator` to get all tests that are to be run from the top-level `TestSuite`
    - Use `FilterIterator` implementations to filter tests based on `--filter`, `--group`, and `--exclude-group`, for instance
    - This allows the number of tests that are to be run to be calculated correctly before the tests are actually run
  - Sort the tests according to `@depends` annotations
- Delay creation of `TestCase` objects and clean them up immediately after a test has been run
  - Allows for more elegant implementation of test execution in isolated PHP processes
  - Allows for running the tests in random order
  - Reduces memory footprint

iltar:
Before I report this as a bug, as it seems to be a design issue. Running tests with `--filter` causes _all_ data providers to run. For some reason a test was failing when running it by file name rather than with a filter, as it was influenced by a data provider.

CC007:
> - Sort the tests according to @depends annotations

What is the status of this?

Idrinth:
wouldn't it make sense to also delay the creation of DataProviderTestSuites when doing this? They'd still need to collect all the data  beforehand with the proposed changes, keeping them rather large.

This would obviously need the data providers to either be run twice(i.e. one for counting, one later for the data) or allow for changes to the testcase maximum count once the provider is being processed.

sebastianbergmann:
Data providers should be run once. If that means that counting no longer works and we need to change how progress is displayed, so be it.

keradus:
data providers can provide also a label for cases, which one can use as a value for tests filter he want to run,
this feature shall not be broken by accident

sebastianbergmann:
Superseded by #3213.

PR #3210, opened 2018-07-14
WIP: initialize data-provided tests late

Idrinth:
Currently very unclean and broken, meant to be a part of #10, so maybe interesting for @epdenouden so we don't duplicate work


tests/TextUI/test-order-randomized-with-dependency-resolution.phpt
```diff
@@ -18,7 +18,7 @@ PHPUnit %s by Sebastian Bergmann and contributors.
 Runtime:       %s
 Random seed:   %d
 
-.....                                                               5 / 5 (100%)
+.....
```
keradus: :-1: for dropping info how far we are with execution

Idrinth: a percentage of an unknown number of testcases it impossible to calculate, so he options were having over 100% tests executed or just showing the number of tests executed.
The only reason there is no number in this specific line is, that it's the last one and directly beneath it the result - including the number of lines - is shown (multiple lines, see: https://github.com/sebastianbergmann/phpunit/pull/3210/files/47fdd0b8a6309e1534190ef8b76a97fff0d919ac#diff-cc5dd71ca2e247cdaa4cde8688137e59R18 ).

I picked the later one for now since I didn't see a way of doing the counting properly, see https://github.com/sebastianbergmann/phpunit/issues/10#issuecomment-404713231

sebastianbergmann: When additional tests are discovered the total number of tests can be increased making the percentage work again.

Idrinth: will also lead to very weird cases, let's take the following:
Suite1: 100 Tests, 20 DataProviderTestsuites -> initial estimation 120 tests
Assuming that doesn't fit into a line we could end up with
```
... 80/120 (66%)
... 160/160 (100%)
... 180/180 (100%)
```
I think this is even more confusing.

Idrinth: Just wondering, would keeping the count itself work for you? That would bypass the issue of having the target number change but at least provide some information on where we are at? @keradus ?

sebastianbergmann:
This branch has conflicts that must be resolved. If the problem you wanted to address with this pull request still exists then please send a new pull request.

PR #3211, opened 2018-07-15
Adding assertArrayStructure

MmdohAhmd:
This assertion makes sure that an array conforms to a given structure.

`$strict` parameters makes the assertion fail if there are one or more elements that are present in `$array` but not present in `$expectedStructure`  

Examples: 

```php
        
        $expected = [
            'a',
            'b',
            'c' => ['d', 'e'],
        ];

        $actual = [
            'a' => 'aval',
            'b' => 'bval',
            'c' => ['d' => 'dval', 'e' => 'eval'],
            'f',
        ];

        $this->assertArrayStructure($expected, $actual); //should pass
        $this->assertArrayStructure($expected, $actual, true); //should NOT pass

```

```php
        
        $expected = [
            'a',
            'b',
            'c'
        ];

        $actual = [
            'a' => 'aval',
            'b' => 'bval',
        ];

        $this->assertArrayStructure($expected, $actual); //should NOT pass

```

### DIFF
--- a/src/Framework/Assert.php
+++ b/src/Framework/Assert.php
@@ -14,6 +14,7 @@ use Countable;
 use DOMDocument;
 use DOMElement;
 use PHPUnit\Framework\Constraint\ArrayHasKey;
+use PHPUnit\Framework\Constraint\ArrayStructure;
 use PHPUnit\Framework\Constraint\ArraySubset;
 use PHPUnit\Framework\Constraint\Attribute;
 use PHPUnit\Framework\Constraint\Callback;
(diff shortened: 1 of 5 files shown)


keradus:
:-1: for this Constraint
it proves one needs to switch to named Value Object rather than using random assoc arrays
this Constraint would promote bad practices

sebastianbergmann:
I am with @keradus on this. For the same reason plus the fact that I am reluctant to accept new assertions into the main PHPUnit distribution.

sebastianbergmann:

MmdohAhmd:
NP

Issue #3213, opened 2018-07-16
New Test Runner


sebastianbergmann:
### Introduction

This ticket supersedes #10.

Thanks to @epdenouden, the reordering of tests is a solved problem. This is something that I did not think possible without rewriting the test runner. And thanks to @Idrinth, it might be possible to delay the execution of data providers without rewriting the test runner.

However, as valuable as these contributions are, I do not think that fundamentally rewriting the test runner is something that can be avoided. The world of PHP has changed a lot in the 18 years since PHPUnit's inception: a lot of work(arounds) that is (are) done while finding and loading tests alone make(s) no more sense. Assumptions have changed (global variables are no longer important, for instance) and components such as [PHP-Parser](https://github.com/nikic/php-parser) and [BetterReflection](https://github.com/Roave/BetterReflection/) are now available and remove the need to really load test classes before they are needed.

At its core, the current test runner has the same architecture and design as the one released as part of PHPUnit 2. Over the years, more and more features were added, layering workaround upon workaround to overcome the test runner's limitations. For almost two decades, this test runner has served us well (at least good enough). It's time to re-think test execution and come up with a new test runner that, hopefully, serves us just as well but without causing so many headaches when it comes to maintenance and the implementation of new features.

### How tests are executed today

Before the first test is executed:

* Test director{y|ies} {is|are} searched recursively for files with names ending in `Test.php` (default)
* Each such file is loaded
* Classes that were not declared before the loading of such a file but are declared the loading are introspected using the Reflection API
* If such a newly found class extends `PHPUnit\Framework\TestCase` then each of its `public` methods are introspected using the Reflection API
* If such a public method has a name that begins with `test` or if the method's docblock contains a `@test` annotation then the method is considered a test method
* If the method's docblock contains a `@dataProvider` annotation then the referenced data provider is executed and the next step is performed for each data set returned
* An object of the test case class is created for the test method (or multiple objects, one for each data set, when `@dataProvider` is used)
* This test object is added to a `PHPUnit\Framework\TestSuite` object

The above means that we have one object for each test that is to be executed before the first test is executed. These objects will remain in memory until the end of the PHP process that executed the test runner. Furthermore, data providers cannot leverage the benefits provided by generators when generators are used to implement data providers.

The above also means that data providers are executed and test objects are created even for tests that will not be executed later on because they are filtered, for instance using `--filter`, `--group`, or `--exclude-group`.

[...]

PR #3215, opened 2018-07-16
Add tests for missing options and arguments

yvoyer:
This PR adds coverage to all uncovered options of the `Command` class.

This is to make sure that no options are missed when changing the runner (ie. if we change to `symfony/console`, or #3213).

This is a migration of the tests found in #1868


sebastianbergmann:

PR #3217, opened 2018-07-18
Fix fatal syntax error "unexpected '-', expecting '{'"

Hecke29:
caused by special chars in wsdl file name.

**This is the fix, you can find the tests here**: https://github.com/sebastianbergmann/phpunit-mock-objects/pull/424
Im sorry but the fix might be best placed here while the tests for it are in another project :(


sebastianbergmann:
The tests are in this project, just in a different branch. phpunit-mock-objects was merged into this project starting with PHPUnit 7.2.

Hecke29:
will fix this in phpunit 7.2 (mock-object project was merged into main project)

Issue #2573, opened 2017-03-17
getMockFromWsdl does not handle urls with query params


gadelat:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 6.0.9
| PHP version         | 7.1.2
| Installation Method | any

```
$ ./phpunit.phar src/AppBundle/Tests/DataProvider/JobBoard/BugTest.php 
PHPUnit 6.0.9 by Sebastian Bergmann and contributors.

E                                                                   1 / 1 (100%)

Time: 925 ms, Memory: 6.00MB

There was 1 error:

1) BugTest::testWsdlBug
ParseError: syntax error, unexpected '.', expecting '{'

/var/www/src/AppBundle/Tests/DataProvider/JobBoard/BugTest.php:9

ERRORS!
Tests: 1, Assertions: 0, Errors: 1.
```

Reproducible example:
```php
use PHPUnit\Framework\TestCase;

class BugTest extends TestCase
{
    public function testWsdlBug()
    {
        $this->getMockFromWsdl('https://soapapi.findapprenticeship.service.gov.uk/Services/VacancyManagement/VacancyManagement51.svc?singleWsdl');
    }
}
```

PR #3219, opened 2018-07-19
Fix fatal syntax error "unexpected '-', expecting '{'"

Hecke29:
caused by special chars in wsdl filename

This might also fix https://github.com/sebastianbergmann/phpunit-mock-objects/issues/357 and fix https://github.com/sebastianbergmann/phpunit/issues/2573

Stumbled over that bug yesterday. Remote file could be load via http (`http://example.org/soap.wsdl`) but saved as local file with name (`/home/user/my-dir/acme-soap.wsdl`) lead to that error.


sebastianbergmann:
Cherry-picked into `7.3` and merged to `master` from there.

Issue #2085, opened 2016-02-19
Configurable default for --enforce-time-limit


ztec:
~~Currently by default all tests without large or medium annotation are considered as small for the timeout.~~

Currently by default, all tests without a tag among small/medium/large are not considered limited at all and do not fall into the time limitation set for those tags. 

I would like to either configure a custom timelimit for them, or at least a configurable default behavoir so I can say all tests are large by default so i can only define some specific ones as small (Or the other way around to be more strict)

Thanks @Flyingmana for the original issue : https://github.com/sebastianbergmann/phpunit/issues/1799

ztec:
related : 
https://github.com/sebastianbergmann/phpunit-documentation/issues/311

sebastianbergmann:
I do not understand what you mean by "Currently by default all tests without large or medium annotation are considered as small for the timeout". For a test without a `@small`, `@medium`, or `@large` annotation no timeout is enforced.

ztec:
I copied the original Bug report. You are right, you changed the behavior in the 4.7.

The idea stay the same : Be able to set a default value for untagged tests.

sebastianbergmann:
A default for "untagged tests" makes no sense as no time limit is not (and should not) be enforced for them.

reinholdfuereder:
@sebastianbergmann Sorry for revisiting this old issue and also #1799:

Frankly I also really like the idea behind which would allow enabling this test timeout checking for "legacy" projects containing loads of PHPUnit tests with just a single switch also for tests without `@small` annotation. The benefit and comfort and default maintenance effort would increase, as otherwise one would have to do (a) a big regex-search-replace initially, maybe combined with some manual work for "legacy" tests that e.g. do not use `@test`; and (b) for all newly written tests developer might forget to add the annotation again...

So it would allow for the defensiveness that I really like in the PHPUnit tool and its IMHO very sane and defensive defaults (e.g. being strict about useless tests)...

=> May I open a new (duplicate) issue of this one here (or could you re-open this one)? Or would that be insulting/stalking/... ;-)

sebastianbergmann:
@reinholdfuereder I will not implement this myself but I might consider a pull request.

reinholdfuereder:
Thanks for your fast and encouraging answer.

Frankly, I cannot exactly say when I will have time for this pull request and especially how long it will take me to be able to understand the existing code and prepare the contribution.

However, we could nonetheless try to discuss the requirements and API/usage:
[...]

PR #3224, opened 2018-07-26
Add --default-time-limit

epdenouden:
Implementation for #2085 

This is a convenience feature for people managing large(r) and potentially slow legacy test collections. Classic example: tests for database integration that have huge `@dataprovider` sets or when using PHPUnit to test webservices/APIs.

Note: I tried running around with my laptop, however I did not manage to validate the timer at relativistic speeds.

### DIFF
--- a/src/Framework/TestResult.php
+++ b/src/Framework/TestResult.php
@@ -1106,14 +1106,6 @@ class TestResult implements Countable
     /**
      * Returns the set timeout for large tests.
      */
-    public function getDefaultTimeoutTests(): int
-    {
-        return $this->defaultTimeLimit;
-    }
-
-    /**
-     * Returns the set timeout for large tests.
-     */
     public function getTimeoutForLargeTests(): int
     {
         return $this->timeoutForLargeTests;
--- a/src/Framework/TestResult.php
+++ b/src/Framework/TestResult.php
@@ -645,7 +645,8 @@ class TestResult implements Countable
         try {
             if (!$test instanceof WarningTestCase &&
                 $this->enforceTimeLimit &&
-                ($this->defaultTimeLimit || $test->getSize() != \PHPUnit\Util\Test::UNKNOWN)) {
+                ($this->defaultTimeLimit || $test->getSize() != \PHPUnit\Util\Test::UNKNOWN) &&
+                \extension_loaded('pcntl') && \class_exists(Invoker::class)) {
                 switch ($test->getSize()) {
                     case \PHPUnit\Util\Test::SMALL:
                         $_timeout = $this->timeoutForSmallTests;
--- a/src/Framework/TestResult.php
+++ b/src/Framework/TestResult.php
@@ -142,7 +142,7 @@ class TestResult implements Countable
     /**
      * @var int
      */
-    protected $defaultTimeLimit = 1;
+    protected $defaultTimeLimit = 0;
 
     /**
      * @var bool
@@ -644,8 +644,8 @@ class TestResult implements Countable
 
         try {
             if (!$test instanceof WarningTestCase &&
-                $test->getSize() != \PHPUnit\Util\Test::UNKNOWN &&
                 $this->enforceTimeLimit &&
+                ($this->defaultTimeLimit || $test->getSize() != \PHPUnit\Util\Test::UNKNOWN) &&
                 \extension_loaded('pcntl') && \class_exists(Invoker::class)) {
                 switch ($test->getSize()) {
                     case \PHPUnit\Util\Test::SMALL:
@@ -660,6 +660,11 @@ class TestResult implements Countable
 
                     case \PHPUnit\Util\Test::LARGE:
                         $_timeout = $this->timeoutForLargeTests;
+
+                        break;
+
+                    case \PHPUnit\Util\Test::UNKNOWN:
+                        $_timeout = $this->defaultTimeLimit;
 
                         break;
                 }
--- a/src/Framework/TestResult.php
+++ b/src/Framework/TestResult.php
@@ -140,6 +140,11 @@ class TestResult implements Countable
     protected $beStrictAboutResourceUsageDuringSmallTests = false;
 
     /**
+     * @var int
+     */
+    protected $defaultTimeLimit = 1;
+
+    /**
      * @var bool
      */
     protected $enforceTimeLimit = false;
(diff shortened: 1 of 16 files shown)


epdenouden:
Ah yes, I was afraid it would not work on Windows yet. Will fix that later today.

epdenouden:
@sebastianbergmann I'd like to fix covecov before merging, too

src/TextUI/Command.php
```diff
@@ -1111,6 +1117,7 @@ protected function showHelp(): void
   --disallow-test-output      Be strict about output during tests
   --disallow-resource-usage   Be strict about resource usage during small tests
   --enforce-time-limit        Enforce time limit based on test size
+  --default-time-limit=<sec>  Maximum running time in seconds for unsized tests
```
reinholdfuereder: I am not perfectly happy with that help text: (a) I personally think something like "Timeout in seconds ..." is more common? (b) Is "unsized tests" also understandable for others? How short/compact must the help text be: is "... for tests without either `@small`, `@medium` or `@large`" too long?
(Sorry, if this is too picky)

epdenouden: First of all: feel free to be as picky as you want, it was your idea after all!  There is no deadline pressure from management, so we can take all the time we need to get it from 'good enough' to 'exactly what I needed'. :-)

I agree with your criticism of the help text. It isn't great as I tried to get all the relevant information in less than 80 characters. Mentioning the three sizes of tests might not fit. I do agree that using _timeout_ would be nice. The config flag is already called `*-time-limit` and adding `timeout` will make it extra clear what it does. I'll commit an improved help text, let me know what you think.

epdenouden:
@sebastianbergmann Getting the diff coverage above the threshold would require `php-invoker` or adding a fake `Invoker`. For such a small feature in the `TestRunner`, are you okay with it as-is?

src/TextUI/Command.php
reinholdfuereder: Thanks, I like this!

karmakoder:
@epdenouden @sebastianbergmann @reinholdfuereder any update here? This change makes a lot of sense for our legacy phpunit suites and a merge will be much appreciated.

epdenouden:
@karmakoder No news from my side. I agree it would be nice to see it merged.

bc-garybottom:
@sebastianbergmann huge +1 here as well, working with our php code base for functional and unit tests, particularly for UI issues having this functionality would be huge.  Running in remote CI/CD SaaS solutions we have the need to gracefully timeout more and more to achieve automation autonomy without chasing down red herrings in our tests or CI/CD tool.  We're currently using phpunit 6.0.13 for our test suite, for reference, which works great except having the ability to skip tests that are idle to rerun them later with this functionality.

epdenouden:
@bc-garybottom This pull request is for version 7. If it gets merged I'd be happy to backport it to 6.x for you if @sebastianbergmann would accept a PR.

sebastianbergmann:
Sorry, but no new features for PHPUnit 6.

karmakoder:
NM. Figured what was wrong with my local setup. All good.

epdenouden:
@karmakoder No problem. Let me know if you run into bugs or edge cases I didn't think of yet.

reinholdfuereder:
@karmakoder Ad "Any update here?": Not really

- there is still my small pending review request (also mentioned/outlined in https://github.com/sebastianbergmann/phpunit/issues/2085#issuecomment-409466372)
- and actually based on your experience mentioned in https://github.com/sebastianbergmann/phpunit/pull/3224#issuecomment-418922883 one suggestion: what about replacing the `writeMessage` with a more defensive failing of the test?

epdenouden:
@reinholdfuereder Apologies for missing your recent comment, missed that completely! Let me give that a think.

epdenouden:
@sebastianbergmann if you have time to look at the PR during the Code Sprint, that'd be lovely 🔬

src/Framework/TestResult.php
```diff
@@ -139,6 +139,11 @@ class TestResult implements Countable
      */
     protected $beStrictAboutResourceUsageDuringSmallTests = false;
 
+    /**
+     * @var int
+     */
+    protected $defaultTimeLimit = 0;
```
sebastianbergmann: New attributes should be `private`.

src/TextUI/Command.php
```diff
@@ -84,6 +84,7 @@ class Command
         'disallow-test-output'      => null,
         'disallow-resource-usage'   => null,
         'disallow-todo-tests'       => null,
+        'default-time-limit='       => null,
```
sebastianbergmann: For consistency, I would rather not add an option like `--default-time-limit` to set the default time limit on the CLI. The timeouts for `@small`, `@medium`, and `@large` can also only be configured in the configuration file.

src/Framework/TestResult.php
epdenouden: Fixing

src/TextUI/Command.php
epdenouden: Yes, that is true. I asked @reinholdfuereder what the exact use case would be and this time limit is meant to be for tests that are _not marked_ with a `@size`: https://github.com/sebastianbergmann/phpunit/issues/2085#issuecomment-405539959

I know a bunch of older test collections with legacy would benefit from this.

localheinz: Not sure about the trailing`=`, what do you think?

epdenouden: The `=` is required for `--key=value`, right? I have not found a clean way of doing `--key[=value]`.

sebastianbergmann: The `=` means that the option requires a value.

localheinz: Thank you for clarifying, @sebastianbergmann!

karmakoder:
@epdenouden After doing several rounds of testing, am learning that this default-time-limit doesnt always take effect for our tests. I basically backported your changes in to phpunit-6.0.13 but the behavior of tests getting marked as failed doesnt seem to be working 100% of the time. Any pointers on what could be the reason? Unfortunately, we cannot upgrade to latest phpunit version as several of our dependencies don't yet support higher versions of php and phpunit. If it helps, we can connect offline to discuss this. Thx in advance.

PR #3226, opened 2018-07-26
Fixed TypeHinting for JUnit::setWriteDocument

siad007:
By invoking `JUnit::setWriteDocument` the result is:
```
Return value of PHPUnit\Util\Log\JUnit::setWriteDocument() must be of the type string or null, none returned
```

### DIFF
--- a/src/Util/Log/JUnit.php
+++ b/src/Util/Log/JUnit.php
@@ -367,9 +367,11 @@ class JUnit extends Printer implements TestListener
      * @param bool $flag Enables or disables the writing of the document.
      *                   default: `true`
      */
-    public function setWriteDocument(bool $flag): void
+    public function setWriteDocument($flag): void
     {
-        $this->writeDocument = $flag;
+        if (\is_bool($flag)) {
+            $this->writeDocument = $flag;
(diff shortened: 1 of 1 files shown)


src/Util/Log/JUnit.php
```diff
@@ -363,12 +363,13 @@ public function getXML(): string
      *
      * This is a "hack" needed for the integration of
      * PHPUnit with Phing.
+     *
+     * @param bool $flag Enables or disables the writing of the document.
+     *                   default: `true`
      */
-    public function setWriteDocument($flag): ?string
+    public function setWriteDocument(bool $flag): void
```
keradus: say hello to BC breaker

sebastianbergmann:
This will be properly addressed in time for PHPUnit 8.

siad007:
> 
> 
> This means that phing can't properly support phpunit7. Any chance of this then being reevaluated?

@kenguest this is solved in https://github.com/phingofficial/phing/pull/962 already.

kenguest:
@siad007 For once I'm delighted to be wrong! :-)

Issue #3237, opened 2018-08-03
result cache file created by default


kubawerlos:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 7.3.0
| PHP version         | 7.2.8
| Installation Method | Composer

After upgrading from PHPUnit 7.2 when running tests the `.phpunit.result.cache` was created. Shouldn't be check [here](https://github.com/sebastianbergmann/phpunit/blob/7.3.0/src/Util/TestResultCache.php#L76) for not defined?

sebastianbergmann:
@epdenouden *ping*

epdenouden:
@sebastianbergmann @kubawerlos will look into it this weekend!

keradus:
the variable is hack/trick to NOT save cache file when running tests for PHPUnit itself during PHPUnit development (maybe a comment there would be nice, for now it's hidden knowledge, and each hack deserves a comment)

By default, the new caching feature is enabled: https://github.com/sebastianbergmann/phpunit/blob/00d26bb6c5ada299716fe85163b2be39100c8af1/src/TextUI/TestRunner.php#L1160 , as it's hard to discover the feature by users if you don't track the changes, while starting taking benefits from it is super nice

kubawerlos:
[Changelog](https://github.com/sebastianbergmann/phpunit/blob/7.3.0/ChangeLog-7.3.md) claims it's disabled by default.

epdenouden:
@kubawerlos Quick question: would you be fine with a revised changelog or you have a use case where the creation of the cachefile causes trouble? We would like to have the feature enabled if it doesn't interfere with use cases I didn't think of during development.

kubawerlos:
@epdenouden it's totally fine to have it enabled by default, only it came as surprise after reading Changelog.

dmitry-ivanov:
+1

What is the reason to have that `.phpunit.result.cache` file not deleted by default?
I think that for most of users - we don't need it, as it was before. No?

epdenouden:
@dmitry-ivanov The answer to your question has two parts:
1. the file is _not deleted_ as it is a cache to store information about test results between runs.
2. it is currently _created_ by default as it takes up little space and the cache needs to be active to support the 'run failed tests first' (`--order-by=defects`) feature. See #3147 for more details

dmitry-ivanov:
@keradus "possibility to execute tests faster"? How do results from the previous run can help to execute tests faster?

[...]

PR #3241, opened 2018-08-04
Do not create cache file by default #3237

Douglasdc3:
The documentation read this feature is disabled by default it was enabled anyway.

Disable it by default to maintain backwards compatibility and allow tests to run on read only filesystem.

ref: #3237

### DIFF
--- a/phpunit.xsd
+++ b/phpunit.xsd
@@ -223,7 +223,7 @@
         <xs:attribute name="backupGlobals" type="xs:boolean" default="false"/>
         <xs:attribute name="backupStaticAttributes" type="xs:boolean" default="false"/>
         <xs:attribute name="bootstrap" type="xs:anyURI"/>
-        <xs:attribute name="cacheResult" type="xs:boolean"/>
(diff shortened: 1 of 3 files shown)


kubawerlos:
Can we have the default also in `phpunit.xsd`?

kubawerlos (approved):

sebastianbergmann:
Merged manually, thanks.

keradus:
:(

is there a plan to enable feature by default for 8.0 ?

sebastianbergmann:
https://github.com/sebastianbergmann/phpunit/issues/3244 :)

epdenouden:
Nice one, thanks for quickly resolving this.

@keradus Yes it will become a default. Before that let's have a quick think about how to work with readonly filesystems gently. I might take a look at the PHP-CS-Fixer thing, I heard they use a cachefile too

keradus:
on readonly fs: if cache file was requested, but saving to it failed, it shall trigger the warning message to stderr, but continue with regular flow of application

PR #3242, opened 2018-08-05
Support @...Order annotations to set order of @before, @beforeClass, …

simPod:
Support @...Order annotations to set order of @before, @beforeClass, @after and @afterClass

This adds support for annotations that help prioritize execution of hook methods. If this sounds reasonable, I can add update for documentation.

Basically, we're using @before/after etc. to initialize kernel, databases, kafka and other services in tests. However sometimes dependencies require other dependencies (eg. our Database-injecting trait requires kernel trait to be injected before).

This helps to sort those hooks between each other when for example `@beforeClassOrder 10` is added.

`@before` with `@beforeClassOrder 10` is executed sooner than `@before` with `@beforeClassOrder 11` and `@before` with no _order_ annotation.

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -829,7 +829,8 @@ abstract class TestCase extends Assert implements Test, SelfDescribing
 
             if ($this->inIsolation) {
                 foreach ($hookMethods['beforeClass'] as $method) {
-                    $this->$method();
+                    $methodName = $method['name'];
+                    $this->$methodName();
                 }
             }
 
@@ -837,7 +838,8 @@ abstract class TestCase extends Assert implements Test, SelfDescribing
             $this->setDoesNotPerformAssertionsFromAnnotation();
 
             foreach ($hookMethods['before'] as $method) {
-                $this->$method();
+                $methodName = $method['name'];
+                $this->$methodName();
             }
 
             $this->assertPreConditions();
@@ -884,12 +886,14 @@ abstract class TestCase extends Assert implements Test, SelfDescribing
         try {
             if ($hasMetRequirements) {
                 foreach ($hookMethods['after'] as $method) {
-                    $this->$method();
+                    $methodName = $method['name'];
+                    $this->$methodName();
                 }
 
                 if ($this->inIsolation) {
                     foreach ($hookMethods['afterClass'] as $method) {
-                        $this->$method();
+                        $methodName = $method['name'];
+                        $this->$methodName();
                     }
                 }
             }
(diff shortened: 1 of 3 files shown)


epdenouden:
@simPod Hi there! Speaking purely for myself, I dislike this idea very much, based on your description. It sounds like you are configuring your infrastructure that the tests are running on from within your tests? Maybe I am misunderstanding you; if you could link to a repository or detailed description of use case that'd be helpful.

Context from my end: working with larger applications we always use 'software defined infrastructure' tools to setup and teardown our environments. All the usual tools for managing fleets of containers/vms, stock them with the software we need and then _inside_ those environments run our development tools - of which PHPUnit is only one.

Could you provide a bit more context?

simPod:
@epdenouden sure, I'll try to explain further:

Each test runs in it's separate database.  
So before the actual test is run, the database needs to be created (`CREATE DATABASE blabla`).   
For that however, I need kernel from my framework to load configured connections. 

So I need two traits, one setuping db and adding db functionality to test class and second one setuping kernel.

```php
trait WithKernel
{
    /**
     * @internal
     * @var Kernel
     */
    private $kernel;

    /**
     * @internal
     * @before
     * @beforeOrder 10
     */
    protected function setupKernel() : void
    {
        ...
        $this->kernel->boot();
    }

    protected function getContainer() : Container
    {
        return $this->kernel->getContainer();
    }
}
```

```php
trait WithDatabase {

    abstract protected function getContainer();

	/**
     * @internal
     * @before
     * @beforeOrder 50
     */
    public static function setupDatabaseTestCase() : void
    {
        $connection = $this->getContainer()->get(Connection::class);
       // CREATE DB etc.
        ...
    }
```

The usage is then following:
```php
class MyTest extends TestCase {

    use WithDatabase;
    use WithKernel;

    public function testSomethingHere() {
    ...
```

That works because kernel is booted and container is provided.

```php
class MyTest extends TestCase {

    use WithKernel;
    use WithDatabase;

    public function testSomethingHere() {
    ...
```
This won't work because database is trying to be setup before kernel is booted.

The order of traits is easy to remember when there are only two. However, some traits for tests might eventually look like:

```php
use WithGraphQL;
use WithClickHouse;
use WithKernel;
use WithRedis;
use WithPostgres;
use WithKafka;
```

Without implicit ordering it causes higher congnitive load on developer to properly sort traits in each test class.

sebastianbergmann:
I am not in favor of adding additional complexity to running code before/after tests.

simPod:
I'm not the one who decides I guess 🤔 .

Issue #3240, opened 2018-08-03
assertArraySubset() does not work as expected


rafrsr:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 7.3.0
| PHP version         | 7.1
| Installation Method | Composer

After upgrade to phpunit `7.3` the `assertArraySubset` does not works as expected. I see that in this version great changes was made in this part; check the following example; works well with `7.2.7`.

````php
<?php

// the following lines emulate a response
// from other library that return a array of stdClass
$field1 = new \stdClass();
$field1->name = 'amount';
$field1->required = true;

$field2 = new \stdClass();
$field2->name = 'accountNumber';
$field2->required = true;

$dataArray = [$field1, $field2];
// end of emulation

//-> TEST

// convert the response in stdClass into associative array
$array = json_decode(json_encode($dataArray), true);

$subset = [
    [
        'name' => 'amount',
        'required' => true,
    ],
];

Assert::assertArraySubset($subset, $array);
`````
The first part of the example is only for demonstration, really I use another library to get a JSON response (the library return that response using stdClass). But the point is, for some reason after convert this data to array the `assertArraySubset` does not work as expected; and the problem only happens with the `boolean` field, otherwise pass.

sebastianbergmann:
This might be what was mentioned in https://github.com/sebastianbergmann/phpunit/pull/3161#issuecomment-410364871.

To be honest, I do not know how this is expected to work as I never used `assertArraySubset()`.

@marcioAlmada originally implemented `assertArraySubset()` (and most recently @pfrenssen did some work on it). I am not writing this to assign blame, simply to explain why I do not know how `assertArraySubset()` is expected to work.

If @pfrenssen really introduced a BC break with https://github.com/sebastianbergmann/phpunit/pull/3161 then, of course, this needs to be fixed (or reverted if it cannot be fixed).

pfrenssen:
@rafrsr thanks for the report. I have made a test case out of your code: #3243. I was able to replicate the issue without having to use `json_decode(json_encode($array))`.

The good news is that there is already a PR that fixes this: #3186. I will add the test for this case to that PR to prove that this is conclusively fixed.

sebastianbergmann:
I have reverted #3240 to fix this issue.

PR #3243, opened 2018-08-05
Regression test for #3240

pfrenssen:
This is the test case from #3240 rolled in a test.


pfrenssen:
This test fails against 7.3.0. But this bug does not occur any more in PR #3186 which is a follow up that fixes edge cases in `assertArraySubset()`. I will add this test case to that PR to provide that it is fixed there.

PR #3250, opened 2018-08-12
Add missing xs:choice 

### LINKED ISSUE #3248: Warning in console when phpunit.xml exclusions are not sorted properly
| Q                   | A
| PHPUnit version     | 7.3.1
| PHP version         | 7.2.8

I recently updated my version of PHPUnit. I don't know exactly which version I was running before. Now that I have updated it, I get this warning

```
PHPUnit 7.3.1 by Sebastian Bergmann and contributors.

  Warning - The configuration file did not pass validation!
  The following problems have been detected:

  Line 21:
  - Element 'directory': This element is not expected. Expected is ( file ).

  Test results may not be as expected.
```
With my long unchanged phpunit.xml. (From standard Laravel 5.6)

```
<filter>
        <whitelist processUncoveredFilesFromWhitelist="true">
            <directory suffix=".php">./app</directory>
            <exclude>
                <file>./app/Http/routes.php</file>
                <directory>./app/Exceptions</directory>  <-- This is line 21 in my phpunit.xml
                <directory>./app/Providers</directory>
            </exclude>
        </whitelist>
    </filter>
```

Everything works fine, but the warning is quite annoying. So I tried to fix it. I checked if the way to exclude directory was different. Apparently, from the latest doc, it hasn't changed. So I started to mess around with my config. It ended up that moving the nodes around made the warning disappear.

```
<whitelist processUncoveredFilesFromWhitelist="true">
            <directory suffix=".php">./app</directory>
            <exclude>
                <directory>./app/Exceptions</directory>
                <directory>./app/Providers</directory>
                <file>./app/Http/routes.php</file>
            </exclude>
        </whitelist>
```

I don't really know if this is bug or a validation rule. But as far as I know, usually the order of nodes in an XML file should not be relevant. So I am posting this here to validate if it is a bug or validation rule?

theseer:
Some xs:sequence nodes of the xsd need xs:choice to make the sort order become irrelevant as PHPUnit doesn't care and the config is perfectly valid either way.

Fixes #3248

### DIFF
--- a/phpunit.xsd
+++ b/phpunit.xsd
@@ -18,12 +18,14 @@
     </xs:complexType>
     <xs:complexType name="filterType">
         <xs:sequence>
-            <xs:group ref="pathGroup"/>
-            <xs:element name="exclude" maxOccurs="unbounded" minOccurs="0">
-                <xs:complexType>
-                    <xs:group ref="pathGroup"/>
-                </xs:complexType>
-            </xs:element>
+            <xs:choice maxOccurs="unbounded" minOccurs="0">
+                <xs:group ref="pathGroup"/>
+                <xs:element name="exclude">
+                    <xs:complexType>
+                        <xs:group ref="pathGroup"/>
+                    </xs:complexType>
+                </xs:element>
+            </xs:choice>
         </xs:sequence>
     </xs:complexType>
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
Merged manually, thanks.

PR #3251, opened 2018-08-13
Print duration of the tests events in the TeamCity logger

wbars:
fixes https://youtrack.jetbrains.com/issue/WI-36291 for PHPUnit >= 5.0

### DIFF
--- a/src/Util/Log/TeamCity.php
+++ b/src/Util/Log/TeamCity.php
@@ -63,6 +63,7 @@ class TeamCity extends ResultPrinter
                 'name'    => $test->getName(),
                 'message' => self::getMessage($t),
                 'details' => self::getDetails($t),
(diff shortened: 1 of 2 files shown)


sebastianbergmann:
I manually merged this into `7.3` and `master`.

If possible/applicable, please also send a pull request for `6.5`. Thanks!

wbars:
@sebastianbergmann Sure, here it is #3252. Could you please give an insight in what PHPUnit releases these patches will be shipped so I can inform the users?

sebastianbergmann:
PHPUnit 6.5.12 and PHPUnit 7.3.2 will have the fix. As bugfix releases do not follow a fixed schedule, I cannot give you a date. But it should be either this week or next week.

wbars:
@sebastianbergmann Perfect, thanks!

PR #3253, opened 2018-08-13
Add Exception callable

Garrcomm:
I've added the option to configure a callable for exceptions. The goal is to test custom Exception properties, like [ErrorException::getSeverity](http://php.net/manual/en/errorexception.getseverity.php).

A test could look like this:
```$this->expectException(\ErrorException::class);
$this->expectExceptionMessage($errstr);
$this->expectExceptionCustom(function (\ErrorException $ex) use ($errno) {
    $this->assertEquals($errno, $ex->getSeverity());
});
```


sebastianbergmann:

PR #3255, opened 2018-08-14
ArgumentGroup needs choice as well. Fixes #3233

### LINKED ISSUE #3233: Unable to send array arg to listeners
| Q                   | A
| PHPUnit version     | 7.2.7
| PHP version         | 7.2

I have a listener that requires an array, the [docs for 7.1](https://phpunit.readthedocs.io/en/7.1/configuration.html) state that `array` is a legal arg type. However, in 7.2 (which I can't find the docs for) I get this error:

`Element 'array': This element is not expected. Expected is one of ( string, double, null, object, file, directory, boolean ).`

My config inside the `<arguments>`
```
                <string>foo</string>
                <string>bar</string>
                <array> <!-- includes -->
                    <element key="0">
                        <string>string value</string>
                    </element>
                </array>
```

After some experimentation, it would appear that array elements must appear first in the list of `arguments` though the documentation makes no mention of this. 

Is this a bug?

theseer:
The argument list should be order-independent. This change should ensure that.

On a side note:
I'm not convinced the duplication `argumentGroup` and `argumentChoice` is actually required but I refrained from changing it as it's not related to the problem at hand.

### DIFF
--- a/phpunit.xsd
+++ b/phpunit.xsd
@@ -84,15 +84,17 @@
     </xs:complexType>
     <xs:group name="argumentsGroup">
         <xs:sequence>
-            <xs:element name="array" type="arrayType" minOccurs="0" maxOccurs="unbounded"/>
-            <xs:element name="integer" type="xs:integer" minOccurs="0" maxOccurs="unbounded"/>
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
Merged manually, thanks.

PR #3263, opened 2018-08-24
Make testdox print more useful dataSet messages

MarkVaughn:
reused `getDataSetAsString` method to make this

Issue:

image

Solution:
image

### DIFF
--- a/src/Util/TestDox/NamePrettifier.php
+++ b/src/Util/TestDox/NamePrettifier.php
@@ -84,7 +84,7 @@ final class NamePrettifier
         }
 
         if ($test->usesDataProvider() && !$annotationWithPlaceholders) {
-            $result .= ' data set "' . $test->dataDescription() . '"';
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
Please present a useful use case (one that does not use emoji).

epdenouden:
@MarkVaughn I can see how automatic display of a 'summary dataset' for `@dataprovider` without descriptions are useful, this change would make for very long output when run with more complex dataproviders. For example: https://github.com/sebastianbergmann/phpunit/blob/287dd08b6b789dcc307ec71e60131c6b1affc9c4/tests/Runner/TestSuiteSorterTest.php#L167-L301

MarkVaughn:
@epdenouden at least concede the default `""` is not very helpful :)
Could probably be the index number or first value instead...

sebastianbergmann:
I think the key should be used.

PR #3265, opened 2018-08-30
Don't escape slashes in prettified JSON

willemstuursma:
* Slashes are not needed in JSON
* In particular JSON objects with lots of URL are easier on the eye

### DIFF
--- a/src/Util/Json.php
+++ b/src/Util/Json.php
@@ -28,7 +28,7 @@ final class Json
             );
         }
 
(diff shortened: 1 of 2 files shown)


keradus:
Can you please add test for this behaviour change, so it won't be accidentally changed again in the future, because your behaviour will be guarded with utest ?

willemstuursma:
@keradus yeah np brb

sebastianbergmann:
Merged manually, thanks.

PR #3277, opened 2018-09-07
Enhancement: Extract method

localheinz:
This PR

* [x] extracts a method


localheinz:

PR #3282, opened 2018-09-08
Fix: Remove unused local variables

localheinz:
This PR

* [x] removes unused local variables


tests/unit/Framework/MockObject/MockObjectTest.php
```diff
@@ -435,7 +435,6 @@ public function testMockClassStoreOverrulable(): void
 
     public function testGetMockWithFixedClassNameCanProduceTheSameMockTwice(): void
     {
-        $mock = $this->getMockBuilder(stdClass::class)->setMockClassName('FixedName')->getMock();
```
localheinz: This line has probably been duplicated on error.

PR #3283, opened 2018-09-08
[#3234] Change r_print to var_export for diff generation of array subsets

### LINKED ISSUE #3234: assertArraySubset with $strict=true doesn't display differences properly
| Q                   | A
| PHPUnit version     | 7.2.7
| PHP version         | 7.2.5

The following test will fail but no differences will be displayed.

```php
<?php

class FooTest extends TestCase
{
    public function testFoo()
    {
        $this->assertArraySubset(['foo' => 1], ['foo' => '1'], true);
    }
}
```

```console
% ./vendor/bin/phpunit FooTest.php
PHPUnit 7.2.7 by Sebastian Bergmann and contributors.

Time: 28 ms, Memory: 4.00MB

There was 1 failure:

1) FooTest::testFoo
Failed asserting that an array has the subset Array &0 (
    'foo' => 1
).
--- Expected
+++ Actual

/Users/matsui/src.nobk/phpunit-test/FooTest.php:8

FAILURES!
```

1blankz7:
Resolves #3234


sebastianbergmann:
Merged manually into `6.5` and merged to `7.3` and `master` from there.

PR #3285, opened 2018-09-08
Fix: Reduce scope of assertion

localheinz:
This PR

* [x] reduces the scope of an assertion when testing `Configuration`

### DIFF
--- a/tests/unit/Util/ConfigurationTest.php
+++ b/tests/unit/Util/ConfigurationTest.php
@@ -451,8 +451,8 @@ class ConfigurationTest extends TestCase
         );
 
         $this->assertEquals(
-            $expectedConfiguration->getTestSuiteConfiguration(),
-            $actualConfiguration->getTestSuiteConfiguration()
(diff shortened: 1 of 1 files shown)


tests/unit/Util/ConfigurationTest.php
```diff
@@ -451,8 +451,8 @@ protected function assertConfigurationEquals(Configuration $expectedConfiguratio
         );
 
         $this->assertEquals(
-            $expectedConfiguration->getTestSuiteConfiguration(),
-            $actualConfiguration->getTestSuiteConfiguration()
+            $expectedConfiguration->getTestSuiteConfiguration()->tests(),
+            $actualConfiguration->getTestSuiteConfiguration()->tests()
```
localheinz: When we compare the entire `TestSuite` object, the test becomes fragile as the set of declared classes can have changed between tests.

See https://github.com/sebastianbergmann/phpunit/blob/6.5.0/src/Framework/TestSuite.php#L131.

localheinz:

PR #3286, opened 2018-09-08
Fix: Add void return type declarations

localheinz:
This PR

* [x] adds `void` return type declarations to test methods

💁‍♂️ I ran

```
$ php-cs-fixer fix --rules=void_return
$ git add tests/unit
```


localheinz:

PR #3287, opened 2018-09-08
Refactor generator

MichelHartmann:
Moved code generation of mocked methods to a separate class.
This should make it easier to refactor this code.


src/Framework/MockObject/MockMethod.php
```diff
@@ -0,0 +1,367 @@
+<?php
+declare(strict_types=1);
```
sebastianbergmann: This should be `<?php declare(strict_types=1);`.

src/Framework/MockObject/MockMethod.php
```diff
@@ -0,0 +1,367 @@
...
+ * file that was distributed with this source code.
+ */
+namespace PHPUnit\Framework\MockObject;
+
+use ReflectionClass;
+use ReflectionException;
+use ReflectionMethod;
+use Text_Template;
+
+class MockMethod
```
sebastianbergmann: New classes should be `final` (unless they should be extended).

PR #3292, opened 2018-09-08
Enhancement: Add array return type declarations to data providers

localheinz:
This PR

* [x] adds `array` return type declarations to data providers


localheinz:

Issue #3302, opened 2018-09-13
Expected size is wrong in assertCount error message


JeroenDeDauw:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 7.3.5
| PHP version         | 7.2.6
| Installation Method | Composer

I have this and the test passes:

```php
$this->assertCount( 6, $someGenerator );
```

I changed it from 6 to 5 to verify the test is being run

```php
$this->assertCount( 5, $someGenerator );
```

Got error

> Failed asserting that actual size 0 matches expected size 5.

While I expected to get

> Failed asserting that actual size 6 matches expected size 5.

sebastianbergmann:
I cannot reproduce this:

```php
<?php declare(strict_types=1);
class Test extends PHPUnit\Framework\TestCase
{
    public function testOne(): void
    {
        $a = [1, 2, 3];
        
        $this->assertCount(4, $a);
    }
}
```
```
$ phpunit Test
PHPUnit 7.3.5 by Sebastian Bergmann and contributors.

F                                                                   1 / 1 (100%)

Time: 30 ms, Memory: 4.00MB

There was 1 failure:

1) Test::testOne
Failed asserting that actual size 3 matches expected size 4.

/home/sb/Test.php:8

FAILURES!
Tests: 1, Assertions: 1, Failures: 1.
```

JeroenDeDauw:
My input is a `Generator`, not an `array`. I just realized that `Generator` is not countable, so the "0" in the failure message makes some form of sense. Still, either assertCount should work fully with `Generator` or not at all.

sebastianbergmann:
Thank you for your report.

Please provide a **minimal**, **self-contained**, reproducing test case that shows the problem you are reporting.

Without such a minimal, self-contained, reproducing test case I will not be able to investigate this issue.

kubawerlos:
```php
    public function testGenerator()
    {
        $generator = static function () {
            yield 1;
            yield 2;
            yield 3;
        };

        static::assertCount(3, $generator());
        static::assertCount(4, $generator());
    }
```

JeroenDeDauw:
The issue does not depend on calling assertCount twice.

>        static::assertCount(4, $generator());

That assert alone does the trick. You will get the error saying 0 does not match 4.

JeroenDeDauw:
Disallowing `Generator` instances would be one way to make the behavior of `assertCount` consistent, but has the downsides that

* it is a breaking change
[...]

PR #3304, opened 2018-09-13
Fix for assertCount expected size for Generator

kubawerlos:
https://github.com/sebastianbergmann/phpunit/issues/3302

### DIFF
--- a/src/Framework/Constraint/Count.php
+++ b/src/Framework/Constraint/Count.php
@@ -22,6 +22,11 @@ class Count extends Constraint
      */
     private $expectedCount;
 
+    /**
+     * @var null|int
+     */
+    private $countOfGenerator;
+
     public function __construct(int $expected)
     {
         parent::__construct();
@@ -93,11 +98,13 @@ class Count extends Constraint
      */
     protected function getCountOfGenerator(Generator $generator): int
     {
-        for ($count = 0; $generator->valid(); $generator->next()) {
-            ++$count;
+        if ($this->countOfGenerator === null) {
+            for ($this->countOfGenerator = 0; $generator->valid(); $generator->next()) {
+                ++$this->countOfGenerator;
+            }
         }
 
-        return $count;
+        return $this->countOfGenerator;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


sebastianbergmann:
This does not solve the underlying problem that `assertCount()` modifies the generator.

kubawerlos:
I don't know if it is possible in PHP.

> the only sensible thing to do here would be to disallow `assertCount()` on generators.

Should we follow this path? If so, target it to `master` as BC break or to `7.3` as bugfix?

JeroenDeDauw (approved):

JeroenDeDauw:
@sebastianbergmann I don't see why that would be a problem. Going through a generator always does this in PHP, so it is expected behavior. Disallow it in assertCount would just force people to create such assertions themselves, which seems like loss-loss to me.

epdenouden:
From a developer perspective: whatever the underlying language feature does internally, I would expect an `assertCount` to work as the name suggests and not give me "size 0" errors.

guilliamxavier (changes requested):
Please consider:

```php
$gen2 = function () { yield 1; yield 2; };
$gen3 = function () { yield 1; yield 2; yield 3; };

$count2 = new Count(2);
Assert::assertTrue($count2->evaluate($gen2(), '', true));
Assert::assertFalse($count2->evaluate($gen3(), '', true)); // this should still work

$count3 = new Count(3);
Assert::assertFalse($count3->evaluate($gen2(), '', true));
Assert::assertTrue($count3->evaluate($gen3(), '', true)); // this should still work
```

Maybe need to reset `$this->countOfGenerator = null;` at the beginning of `function matches`?

Also please consider https://github.com/sebastianbergmann/phpunit/issues/3302#issuecomment-421365917 (i.e. there are non-`Countable`, `Traversable` things that are _not_ `instanceof Generator` but suffer the same kind of problem)...

JeroenDeDauw:
I can confirm that this PR is breaking that behavior and does not clean up the already existing confusion between Generators and Iterables that cannot be rewinded. I'll submit a modified solution in a bit

sebastianbergmann:
Superseded by #3316.

PR #3309, opened 2018-09-20
upgrade tests to new exception syntax

TomasVotruba:


TomasVotruba:
Fix in progress...

tests/unit/Framework/Constraint/ExceptionMessageRegExpTest.php
```diff
@@ -13,43 +13,37 @@
...
     {
+        $this->expectException('Exception');
+        $this->expectExceptionMessageRegExp('#Screaming preg_match#');
         \ini_set('xdebug.scream', '1');
 
         throw new \Exception('Screaming preg_match');
     }
 
-    /**
-     * @expectedException \Exception variadic
```
TomasVotruba: Why was there "variadic", a typo?

TomasVotruba:
There should be some I guess. It's not clear from class/method naming to me.
Maybe have "ExpectedExceptionAnnotationTest" with explicit comment?

tests/unit/Framework/Constraint/ExceptionMessageRegExpTest.php
keradus: no no no no no

this annotation supports one and two parameters, one of them is exception class, second, optional, is exception message.

this docblock is equivalent to
```
/**
 * @expectedException \Exception 
 * @expectedExceptionMessage variadic
 * @expectedExceptionMessageRegExp /^A variadic \w+ message/
 */
```

TomasVotruba: I see, thanks for sharing. Is there somewhere documented?

So correct replacement would be this?

```php
$this->expectException('Exception');
$this->expectExceptionMessage('variadic');
```

keradus: no, you shall still respect `expectedExceptionMessageRegExp`
`expectedExceptionMessage` is not causing `=== "variadic"`, but `/.*variadic.*/` under the hood

probably it is, but i was not using documentation for so long, i won't point it

keradus:
without having explicit tests, I would suggest to not drop those annotations for now

tests/unit/Framework/Constraint/ExceptionMessageRegExpTest.php
TomasVotruba: In the comment above you mix various rules then. So it's like this?

```php
$this->expectException('Exception');
$this->expectExceptionMessageRegExp('#variadic#');
```

TomasVotruba: Any docs?

keradus: sorry, I don't understand part of "mix various rule"

` * @expectedExceptionMessage variadic` means exceptionMessag "contains `variadic`", not "equals exactly 'variadic'"


> Any docs?

probably it is, but i was not using documentation for so long, i won't point it

TomasVotruba: You've wrote: `* @expectedExceptionMessage variadic`

@sebastianbergmann Could you confirm this?

keradus: @sebastianbergmann , I believe that I can elaborate on my answer, so maybe you won't need to spend time on it. Of course, if you want to step in, do it ;)

keradus: I'm sad you don't want to believe me @TomasVotruba nor check it on your own. I'm using this great project for so long and that intensive, I got chance to understood it and know few aspects of it that can be considered tricky, like the one you faced here ;)

To sum up:
1. one can apply that diff, both versions, before and after patch, mean the same
```diff
 /**
- * @expectedException \Exception variadic
+ * @expectedException \Exception
+ * @expectedExceptionMessage variadic
  * @expectedExceptionMessageRegExp /^A variadic \w+ message/
  */
```

2. both, message and messageRegExp expectations are respected and executed
3. message `foo` is in fact `.*foo.*` regex (but you can't provide multiple regexes annotations!) (disclaimer - yes, I would prefer personally if that would not happen, yet it does)
4. to replace annotations to code, we would have:
```php
$this->expectException('Exception');
$this->expectExceptionMessage('variadic');
$this->expectExceptionMessageRegExp('/^A variadic \w+ message/');
```

keradus: As I said, I don't rely on documentation, yet that shall be better - I took 5 minutes and grab some links for you, @TomasVotruba . I hope those will help you to understood how it works, if my elaboration is not enough.


`expectedException` annotation regex is defined here: https://github.com/sebastianbergmann/phpunit/blob/952160e0c4406a66814b51f19161b1098158f08e/src/Util/Test.php#L65
As you can see, it can defined `class`, `message` and `code`, eg `message` is handled here:
https://github.com/sebastianbergmann/phpunit/blob/952160e0c4406a66814b51f19161b1098158f08e/src/Util/Test.php#L365

Ten, util is used in `TestCase` to set expectations:
https://github.com/sebastianbergmann/phpunit/blob/64d377cc488cd3707f34f46266c8d22e21f1da06/src/Framework/TestCase.php#L1606

And asserted here https://github.com/sebastianbergmann/phpunit/blob/64d377cc488cd3707f34f46266c8d22e21f1da06/src/Framework/TestCase.php#L1169 by creating `ExceptionMessage` constraint.

So finally, we can see how message we provided in annotation is comparad to message from exception that happened:
https://github.com/sebastianbergmann/phpunit/blob/2e9ab306fa38423ecd977bb78623bc863083d547/src/Framework/Constraint/ExceptionMessage.php#L47
As we see, it's not `===`

TomasVotruba: This feature is WTF to me. I was confused by your answer and wanted to know for sure. 

Thanks for the code, that's what I needed to be able to implement it correctly.

TomasVotruba:
@sebastianbergmann What do you think about explicit test case names?

TomasVotruba:
Closing as unclear what to do to close it and not such a big deal atm.

@keradus Thanks for examples again.

PR #3311, opened 2018-09-23
Added docblock parameters.

mcneely:
IDE was showing an error because annotation was missing.

### DIFF
--- a/src/Framework/Assert.php
+++ b/src/Framework/Assert.php
@@ -414,6 +414,7 @@ abstract class Assert
      * Asserts the number of elements of an array, Countable or Traversable
      * that is stored in an attribute.
      *
+     * @param integer       $expectedCount 
      * @param object|string $haystackClassOrObject
      *
      * @throws ExpectationFailedException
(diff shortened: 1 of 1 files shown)


sebastianbergmann:

This is PHP 7 code. No need to duplicate information in a comment when there is actual code that has it.

mcneely:
Yeah, except phpstorm is still shows a syntax error: "Expected \PHPUnit\Framework\int, got int". Adding the comments in satisfies the issue. I didn't do this to waste my time. I was fixing an actual issue.

keradus:
Hi there @mcneely !
PHPStorm is not the only IDE in use, and actually, PHPStorm may or may not complain about that - depends how new version you are using and what is the configuration, including plugins.

Code shall be written to be understood by compiler, and comments/docs can be added to help developers. Here, those extra annotations provide no value, as all value it could provide is already there, on language syntax level.
In fact, most modern projects, not only PHPUnit, eg Symfony, are not providing duplicated documentation.

PR #3316, opened 2018-09-29
Fix assertCount and the Count class

### LINKED ISSUE #3302: Expected size is wrong in assertCount error message

JeroenDeDauw:
This is ready for review. Ping @sebastianbergmann 

Fixes #3302, builds upon and replaces https://github.com/sebastianbergmann/phpunit/pull/3304

Behavior changes with master:
* Error messages for non-rewindable Iterators are now shown correctly
* The `Count` class can now evaluate the same non-rewindable Iterator more than once
* Non-rewindable Iterators that are not Generators will no longer be rewound by `Count`, which for some Iterators causes an error

I took care to not change any other behavior. Though beware that the protected method `getCountOfGenerator` has been refactored away.

Tests added for the changed behavior and some clean up of existing tests, including adding assertions for existing behavior that was not tested yet.

### DIFF
--- a/src/Framework/Constraint/Count.php
+++ b/src/Framework/Constraint/Count.php
@@ -89,15 +89,20 @@ class Count extends Constraint
             $traversable = $traversable->getIterator();
         }
 
-        if ($traversable instanceof Generator) {
+        if ($traversable instanceof Iterator) {
+            if ($this->iteratorIsRewindable($traversable)) {
+                return $this->getCountOfRewindableIterator($traversable);
+            }
+
             return $this->getCountOfNonRewindableIterator($traversable);
         }
 
-        if ($traversable instanceof Iterator) {
-            return $this->getCountOfRewindableIterator($traversable);
-        }
-
         return \iterator_count($traversable);
+    }
+
+    private function iteratorIsRewindable(Iterator $iterator): bool
+    {
+        return !($iterator instanceof Generator) && !($iterator instanceof \NoRewindIterator);
     }
 
     /**
--- a/src/Framework/Constraint/Count.php
+++ b/src/Framework/Constraint/Count.php
@@ -62,22 +62,25 @@ class Count extends Constraint
         }
 
         if ($other instanceof Traversable) {
-            while ($other instanceof IteratorAggregate) {
-                $other = $other->getIterator();
-            }
-
-            $iterator = $other;
-
-            if ($iterator instanceof Generator) {
-                return $this->getCountOfGenerator($iterator);
-            }
-
-            if (!$iterator instanceof Iterator) {
-                return \iterator_count($iterator);
-            }
-
-            return $this->getCountOfRewindableIterator($iterator);
+            return $this->getCountOfTraversable($other);
         }
+    }
+
+    private function getCountOfTraversable(Traversable $traversable): int
+    {
+        while ($traversable instanceof IteratorAggregate) {
+            $traversable = $traversable->getIterator();
+        }
+
+        if ($traversable instanceof Generator) {
+            return $this->getCountOfGenerator($traversable);
+        }
+
+        if ($traversable instanceof Iterator) {
+            return $this->getCountOfRewindableIterator($traversable);
+        }
+
+        return \iterator_count($traversable);
     }
 
     /**
--- a/src/Framework/Constraint/Count.php
+++ b/src/Framework/Constraint/Count.php
@@ -31,7 +31,7 @@ class Count extends Constraint
     {
         parent::__construct();
 
-        $this->expectedCount = $expected;
+        $this->expectedCount     = $expected;
         $this->traversableCounts = new \SplObjectStorage();
     }
 
@@ -66,23 +66,6 @@ class Count extends Constraint
         }
     }
 
-    private function getCountOfTraversable(Traversable $traversable): int
-    {
-        while ($traversable instanceof IteratorAggregate) {
-            $traversable = $traversable->getIterator();
-        }
-
-        if ($traversable instanceof Generator) {
-            return $this->getCountOfGenerator($traversable);
-        }
-
-        if ($traversable instanceof Iterator) {
-            return $this->getCountOfRewindableIterator($traversable);
-        }
-
-        return \iterator_count($traversable);
-    }
-
     /**
      * Returns the total number of iterations from a generator.
      * This will fully exhaust the generator.
@@ -98,6 +81,40 @@ class Count extends Constraint
         }
 
         return $this->traversableCounts[$generator];
+    }
+
+    /**
+     * Returns the description of the failure.
+     *
+     * The beginning of failure messages is "Failed asserting that" in most
+     * cases. This method should return the second part of that sentence.
+     *
+     * @param mixed $other evaluated value or object
+     */
+    protected function failureDescription($other): string
+    {
+        return \sprintf(
+            'actual size %d matches expected size %d',
+            $this->getCountOf($other),
+            $this->expectedCount
+        );
+    }
+
+    private function getCountOfTraversable(Traversable $traversable): int
+    {
+        while ($traversable instanceof IteratorAggregate) {
+            $traversable = $traversable->getIterator();
+        }
+
+        if ($traversable instanceof Generator) {
+            return $this->getCountOfGenerator($traversable);
+        }
+
+        if ($traversable instanceof Iterator) {
+            return $this->getCountOfRewindableIterator($traversable);
+        }
+
+        return \iterator_count($traversable);
     }
 
     private function getCountOfRewindableIterator(Iterator $iterator): int
@@ -116,22 +133,5 @@ class Count extends Constraint
         }
 
         return $count;
-    }
-
-    /**
-     * Returns the description of the failure.
-     *
-     * The beginning of failure messages is "Failed asserting that" in most
-     * cases. This method should return the second part of that sentence.
-     *
-     * @param mixed $other evaluated value or object
-     */
-    protected function failureDescription($other): string
-    {
-        return \sprintf(
-            'actual size %d matches expected size %d',
-            $this->getCountOf($other),
-            $this->expectedCount
-        );
     }
 }
--- a/src/Framework/Constraint/Count.php
+++ b/src/Framework/Constraint/Count.php
@@ -22,7 +22,7 @@ class Count extends Constraint
     private $expectedCount;
 
     /**
-     * @var \SplObjectStorage
+     * @var \SplObjectStorage|null
      */
     private $traversableCounts;
 
@@ -31,7 +31,6 @@ class Count extends Constraint
         parent::__construct();
 
         $this->expectedCount     = $expected;
-        $this->traversableCounts = new \SplObjectStorage();
     }
 
     public function toString(): string
--- a/src/Framework/Constraint/Count.php
+++ b/src/Framework/Constraint/Count.php
@@ -66,40 +66,6 @@ class Count extends Constraint
         }
     }
 
-    /**
-     * Returns the total number of iterations from a generator.
-     * This will fully exhaust the generator.
-     */
-    protected function getCountOfGenerator(Generator $generator): int
-    {
-        if (!$this->traversableCounts->contains($generator)) {
-            for ($countOfGenerator = 0; $generator->valid(); $generator->next()) {
-                ++$countOfGenerator;
-            }
-
-            $this->traversableCounts->attach($generator, $countOfGenerator);
-        }
-
-        return $this->traversableCounts[$generator];
-    }
-
-    /**
-     * Returns the description of the failure.
-     *
-     * The beginning of failure messages is "Failed asserting that" in most
-     * cases. This method should return the second part of that sentence.
-     *
-     * @param mixed $other evaluated value or object
-     */
-    protected function failureDescription($other): string
-    {
-        return \sprintf(
-            'actual size %d matches expected size %d',
-            $this->getCountOf($other),
-            $this->expectedCount
-        );
-    }
-
     private function getCountOfTraversable(Traversable $traversable): int
     {
         while ($traversable instanceof IteratorAggregate) {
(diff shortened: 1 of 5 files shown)


JeroenDeDauw:
Oh my. The CS checker insists on putting the methods in a suboptimal order :(

src/Framework/Constraint/Count.php
```diff
@@ -56,64 +62,86 @@ protected function getCountOf($other): ?int
...
+    }
 
-            if (!$iterator instanceof Iterator) {
-                return \iterator_count($iterator);
-            }
+    private function getCountOfTraversable(Traversable $traversable): int
+    {
+        if ($this->iteratorCounts === null) {
+            $this->iteratorCounts = new \SplObjectStorage();
+        }
```
JeroenDeDauw: I do not understand how without this extra initialization the field can end up being null. Yet that is what apparently happens: https://travis-ci.org/sebastianbergmann/phpunit/jobs/434909673

How is this possible? It is set in the constructor. SameSize is calling `parent::__construct`. Nothing besides the constructor writes to the field and the field is private. Huh???

JeroenDeDauw:
@sebastianbergmann @kubawerlos @guilliamxavier This is ready for review :)

guilliamxavier (changes requested):
Thanks for the work! I have spotted a few issues (but I am not qualified to evaluate the whole picture)

src/Framework/Constraint/Count.php
```diff
@@ -22,11 +21,17 @@ class Count extends Constraint
...
+     */
+    private $traversableCounts;
+
     public function __construct(int $expected)
     {
         parent::__construct();
 
-        $this->expectedCount = $expected;
+        $this->expectedCount     = $expected;
+        $this->traversableCounts = new \SplObjectStorage();
```
guilliamxavier: Question: since `$this->traversableCounts` is initialized as needed in `function getCountOfTraversable`, and is never used for the most common case of `array`|`Countable`, would it be preferable to _not_ initialize it here?

src/Framework/Constraint/Count.php
```diff
@@ -116,4 +81,82 @@ protected function failureDescription($other): string
...
+            return $this->countIteratorWithoutChangingPositionIfPossible($traversable);
+        }
+
+        return $this->getCountOfTraversableThatIsNotIteratorOrIteratorAggregate($traversable);
+    }
+
+    private function countIteratorWithoutChangingPositionIfPossible(Iterator $iterator): int
+    {
+        $key   = $iterator->key();
+        $count = $this->getCountOfIterator($iterator);
```
guilliamxavier: Caution: original code uses custom iteration (without rewind) only for `Generator`, and `iterator_count` (which can rewind internally) for the rest, so I suspect this would change behavior (in edge cases). For example:

```php
$iterator = new \TestIterator([1, 2, 3]);
$iterator->next(); // this is the tricky bit

// which line below to uncomment, with original code? and with this PR?
//$n = 3;
//$n = 2;

Assert::assertCount($n, $iterator);
```

src/Framework/Constraint/Count.php
```diff
@@ -116,4 +81,82 @@ protected function failureDescription($other): string
...
+
+        $this->traversableCounts->attach($iterator, $count);
+
+        if ($this->iteratorIsRewindable($iterator)) {
+            $this->traversableCounts->detach($iterator);
```
guilliamxavier: Maybe it is possible to `attach` only if needed, so that you don't need to `detach`

src/Framework/Constraint/Count.php
```diff
@@ -116,4 +81,82 @@ protected function failureDescription($other): string
...
+            $this->rewindIterator($iterator, $key);
```
guilliamxavier: This originally was inside a condition `if ($key !== null) {`

src/Framework/Constraint/Count.php
```diff
@@ -116,4 +81,82 @@ protected function failureDescription($other): string
...
+
+    private function iteratorIsRewindable(Iterator $iterator): bool
+    {
+        try {
+            $iterator->rewind();
+        } catch (\Exception $e) {
+            return false;
+        }
+
+        return !($iterator instanceof \NoRewindIterator);
```
guilliamxavier: Thoughts:
- Reorder `getCountOfIterator` before this function
- There is no need to try rewind if it is a `NoRewindIterator`
- If you cannot test whether an iterator is rewindable without actually trying to rewind it, maybe should you merge this function with `rewindIterator` below, so that the current position is unchanged (restored)

src/Framework/Constraint/Count.php
```diff
@@ -116,4 +81,82 @@ protected function failureDescription($other): string
...
+    }
+
+    /**
+     * Returns the total number of iterations from a iterator.
+     * This will fully exhaust the iterator.
+     */
+    private function getCountOfIterator(Iterator $iterator): int
+    {
+        for ($countOfGenerator = 0; $iterator->valid(); $iterator->next()) {
```
guilliamxavier: `$countOfIterator`, or simpler (better?) just keep the original name `$count`

src/Framework/Constraint/Count.php
```diff
@@ -116,4 +81,82 @@ protected function failureDescription($other): string
...
+        while ($iterator->valid() && $key !== $iterator->key()) {
+            $iterator->next();
+        }
+    }
+
+    private function getCountOfTraversableThatIsNotIteratorOrIteratorAggregate(Traversable $traversable): int
+    {
+        $count = \iterator_count($traversable);
+
+        $this->traversableCounts->attach($traversable, $count);
```
guilliamxavier: Maybe it is possible to factorize this `attach` with the other (i.e. directly in `function getCountOfTraversable`)

tests/_files/GeneratorTest.php
```diff
@@ -0,0 +1,28 @@
+<?php
```
guilliamxavier: Is this file still needed? (if yes, then what about a more descriptive name?)

tests/unit/Framework/Constraint/CountTest.php
```diff
@@ -142,4 +142,98 @@ public function testCountTraversable(): void
...
+
+            return;
+        }
+
+        $this->fail('An exception should have been thrown');
+    }
+
+    private function newPdoStatementWithThreeValues(): \Traversable
+    {
+        if (!\class_exists('\PDO')) {
```
guilliamxavier:
```php
        if (!\extension_loaded('pdo') || !\in_array('sqlite', \PDO::getAvailableDrivers())) {
```
according to https://github.com/sebastianbergmann/phpunit/blob/be6564ea49b96b9776c53f627a49683ec2687ce5/tests/end-to-end/regression/GitHub/1351.phpt

tests/unit/Framework/Constraint/CountTest.php
```diff
@@ -142,4 +142,98 @@ public function testCountTraversable(): void
...
+            $this->markTestSkipped('derp');
```
guilliamxavier: What about `'PDO_SQLITE is required'` (still from https://github.com/sebastianbergmann/phpunit/blob/be6564ea49b96b9776c53f627a49683ec2687ce5/tests/end-to-end/regression/GitHub/1351.phpt)?

src/Framework/Constraint/Count.php
JeroenDeDauw: This is tested by `testOnlyCountRemainingGeneratorElements`. I refactored that test to ascertain the original behavior and then made sure it stayed the same.

src/Framework/Constraint/Count.php
JeroenDeDauw: As far as I can tell this is not possible, hence the approach taken.

src/Framework/Constraint/Count.php
JeroenDeDauw: Many changes where made compared to the original. If you put in this condition now the tests will fail.

JeroenDeDauw: (If you want to see how it got removed, there are plenty of small commits to look at)

src/Framework/Constraint/Count.php
JeroenDeDauw: > Reorder getCountOfIterator before this function

Done. That is some hard looking for things that can be improved there! In some way this is odd given this code is much higher quality than a lot of what I looked at while making these changes.

> If you cannot test whether an iterator is rewindable without actually trying to rewind it, maybe should you merge this function with rewindIterator below...

That was what the code was doing at some point. But then you have both the check, the rewind and a `traversableCounts->detach()` in there. Esp the detach is badly placed then.

src/Framework/Constraint/Count.php
JeroenDeDauw: I did not change this code compared to the original

src/Framework/Constraint/Count.php
JeroenDeDauw: I think that would make other things less nice and not be a good trade-off

tests/_files/GeneratorTest.php
JeroenDeDauw: Probably good to have one integration test for the originally reported issue. Agree the name was not descriptive and changed it

tests/unit/Framework/Constraint/CountTest.php
JeroenDeDauw: omg :D I forgot to put a real message there :D

src/Framework/Constraint/Count.php
```diff
@@ -22,11 +21,16 @@ class Count extends Constraint
...
+     * @var null|\SplObjectStorage
+     */
+    private $traversableCounts;
+
     public function __construct(int $expected)
     {
         parent::__construct();
 
-        $this->expectedCount = $expected;
+        $this->expectedCount     = $expected;
```
guilliamxavier: leftover from intermediate commits, can be reverted back

tests/end-to-end/generator-failing-assert-count.phpt
```diff
@@ -0,0 +1,26 @@
...
+F                                                                   1 / 1 (100%)
+
+Time: %s, Memory: %s
+
+There was 1 failure:
+
+1) Test\FailingGeneratorCountTest::testGenerator
+Failed asserting that actual size 3 matches expected size 4.
+
+%FailingGeneratorCountTest.php:%i
```
guilliamxavier:
```
%sFailingGeneratorCountTest.php:%i
```

(missing an `s` after the first `%`)
(that causes the Travis build to fail)

src/Framework/Constraint/Count.php
guilliamxavier: I'm talking about *non*-`Generator` iterators:

```php
        $it = new \TestIterator([1, 2, 3]);
        $it->next();
        $this->assertTrue((new Count(3))->evaluate($it, '', true));
```

that was passing on master but is failing with this PR

src/Framework/Constraint/Count.php
guilliamxavier: What about something like this?

```php
    private function getCountOfTraversable(Traversable $traversable): int
    {
        ...

        if ($this->traversableCounts->contains($traversable)) {
            return $this->traversableCounts[$traversable];
        }

        if ($traversable instanceof Iterator) {
            $count = $this->countIteratorWithoutChangingPositionIfPossible($traversable);
        } else {
            $count = \iterator_count($traversable);
        }

        $this->traversableCounts->attach($traversable, $count);

        return $count;
    }

    /**
     * Returns the total number of iterations from a iterator.
     * This will fully exhaust the iterator if it is not rewindable.
     */
    private function countIteratorWithoutChangingPositionIfPossible(Iterator $iterator): int
    {
        $key = $iterator->key();

        // Note: iterator_count calls rewind internally.
        try {
            $count = \iterator_count($iterator);
        } catch (\Exception $e) {
            // Fall back to manual iteration.
            for ($count = 0; $iterator->valid(); $iterator->next()) {
                ++$count;
            }

            return $count;
        }

        // Manually rewind $iterator to previous key, since iterator_count
        // moves pointer.
        if ($key !== null) {
            try {
                $iterator->rewind();
            } catch (\Exception $e) {
                return $count;
            }

            while ($iterator->valid() && $key !== $iterator->key()) {
                $iterator->next();
            }
        }

        return $count;
    }
```

It's simpler (I find) and keeps closer to the original code without needing to check for specific implementations like `NoRewindIterator` or even `Generator`. Maybe it can still be simplified or otherwise improved?

src/Framework/Constraint/Count.php
JeroenDeDauw: Done

src/Framework/Constraint/Count.php
JeroenDeDauw: Oh right. The Iterator case was also not tested >_>

JeroenDeDauw: Fixed now. Which allows getting rid of the ->detach()

guilliamxavier (approved):
This indeed seems an overall improvement (especially the tests) 👍

JeroenDeDauw:
Thanks for the detailed review!

src/Framework/Constraint/Count.php
```diff
@@ -56,64 +60,106 @@ protected function getCountOf($other): ?int
...
+            $this->moveIteratorToPosition($iterator, $key);
+        }
+
+        return $count;
     }
 
     /**
-     * Returns the total number of iterations from a generator.
-     * This will fully exhaust the generator.
+     * Returns the total number of iterations from a iterator.
```
guilliamxavier: Last comment (got lost when approving, sorry): this should now be "from a**n** iterator".
And maybe you can also improve "the total number of iterations" to the more accurate "the *remaining* number of iterations" or even "the *number of remaining iterations*"?

epdenouden:
Will check this later today, too, because kittyyyyy

sebastianbergmann:
This branch has conflicts that must be resolved. If the problem you wanted to address with this pull request still exists then please send a new pull request.

JeroenDeDauw:
Yeah well not very inclined to put effort into that again after what happened here.

Issue #3315, opened 2018-09-27
AssertContains : $ignorecase doesn't work


Tominounet:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 6.5.13
| PHP version         | 7.2
| Installation Method | PHAR

I'm using and learning how to test my code with PHPUnit. I had some test to make, so I can understand some asserts. Then I tried this one : 

![image](https://user-images.githubusercontent.com/26654281/46130896-b2e45400-c23a-11e8-802f-11a05e7728c5.png)

So I expected this test to be ok, but...

![image](https://user-images.githubusercontent.com/26654281/46130947-cc859b80-c23a-11e8-8504-f521044cde08.png)

It seems the $ignorecase parameter doesn't work, true or false doesn't change anything. Is it normal ?

sebastianbergmann:
Thank you for your report.

Please provide a **minimal**, **self-contained**, reproducing test case that shows the problem you are reporting. Do not use screenshots of code.

Without such a minimal, self-contained, reproducing test case I will not be able to investigate this issue.

mkasberg:
I'm not sure this is actually a bug. I think the code is behaving more or less as expected, and the behavior is documented in the [PHPUnit documentation](https://phpunit.readthedocs.io/en/7.3/assertions.html#assertcontains) (for both 7 and [6.5](https://phpunit.de/manual/6.5/en/appendixes.assertions.html#appendixes.assertions.assertContains)).

`assertContains()` has 2 documented forms:

* `assertContains(string $needle, string $haystack[, string $message = '', boolean $ignoreCase = false])` which accepts a string `$haystack` and a 4th parameter `$ignoreCase`
* `assertContains(mixed $needle, Iterator|array $haystack[, string $message = ''])`, which accepts an array `$haystack` but **does not** take a 4th parameter for case sensitivity.

In your example, because your second parameter is an array, PHPUnit assumes you are using the "array" form of the method, and ignores the `$ignoreCase` parameter.

Perhaps there is some small room for improvement here. PHPUnit could throw an IllegalArgumentException (or similar) in `Framework/Assert.php` if `$ignoreCase=true` was passed and the second argument is not a string. I think it would be relatively easy for me to propose some changes along these lines in a pull request and we can see what everyone thinks.

sebastianbergmann:
No feedback, closing.

epdenouden:
[...]

PR #3317, opened 2018-10-01
Throw exception when ignoreCase on array

mkasberg:
The documentation for assertContains() has 2 forms:

    assertContains(mixed $needle, Iterator|array $haystack[, string $message = ''])

and

    assertContains(string $needle, string $haystack[, string $message = '', boolean $ignoreCase = false])

Some users may become confused and try to pass `$ignoreCase=true` with
the first form. In this case, PHPUnit was silently ignoring the
parameter. Rather than silently ifnoring the parameter, we should throw
an exception to alert the user that this is an invalid parameter
combination.

There is a _small_ concern here for a backward-compatibility break. The troublesome case is where a user was passing `$ignoreCase=true` and using an array argument, but their test was passing (because the case matches). However, it still seems reasonable to warn the user that a parameter is being ignored, and the fix for the user is simple - just pass false or don't pass anything.

I've added a unit test, but for clarity, here's what the output looks like:

**Old Output**
```
$ phpunit IgnoreCaseTest.php 
PHPUnit 7.3.2 by Sebastian Bergmann and contributors.

F.                                                                  2 / 2 (100%)

Time: 62 ms, Memory: 8.00MB

There was 1 failure:

1) IgnoreCaseTest::testArrayHasValueCaseInsensitive
Ignore case with array fails.
Failed asserting that an array contains 'ABCD'.

/home/mkasberg/Desktop/phpunit-test/IgnoreCaseTest.php:9

FAILURES!
Tests: 2, Assertions: 2, Failures: 1.
```

**New Output**
```
$ ~/code/phpunit/phpunit IgnoreCaseTest.php 
PHPUnit 7.4-g435ec0594 by Sebastian Bergmann and contributors.

E.                                                                  2 / 2 (100%)

Time: 43 ms, Memory: 4.00MB

There was 1 error:

1) IgnoreCaseTest::testArrayHasValueCaseInsensitive
PHPUnit\Framework\Exception: ignoreCase is not allowed on Assert::assertContains when argument 2 is not a string.

/home/mkasberg/Desktop/phpunit-test/IgnoreCaseTest.php:9

ERRORS!
Tests: 2, Assertions: 1, Errors: 1.

```

**Sample Test Case**

```
<?php

class IgnoreCaseTest extends TestCase {

    public function testArrayHasValueCaseInsensitive() {
        $ignoreCase = true;
        $this->assertContains("ABCD", ["abcd"], 'Ignore case with array fails.', $ignoreCase);
    }

    public function testStringContainsCaseInsensitive() {
        $ignoreCase = true;
        $this->assertContains("ABCD", "abcd", 'Ignore case with string passes.', $ignoreCase);
    }
}
```

### DIFF
--- a/src/Framework/Assert.php
+++ b/src/Framework/Assert.php
@@ -174,6 +174,9 @@ abstract class Assert
     {
         if (\is_array($haystack) ||
             (\is_object($haystack) && $haystack instanceof Traversable)) {
+            if ($ignoreCase) {
+                throw new Exception('ignoreCase is not allowed on Assert::assertContains when argument 2 is not a string.');
+            }
             $constraint = new TraversableContains(
                 $needle,
                 $checkForObjectIdentity,
--- a/tests/unit/Framework/AssertTest.php
+++ b/tests/unit/Framework/AssertTest.php
@@ -76,6 +76,13 @@ class AssertTest extends TestCase
         $this->assertContains('foo', [true], '', false, true, true);
     }
 
+    public function testAssertContainsThrowsExceptionWhenArrayAndIgnoreCase(): void
+    {
+        $this->expectException(Exception::class);
+        $this->expectExceptionMessage('ignoreCase is not allowed on Assert::assertContains when argument 2 is not a string.');
+        $this->assertContains('ABC', ['abc'], '', true);
+    }
+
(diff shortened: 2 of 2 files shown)


mkasberg:
This is related to #3315 .

PR #3326, opened 2018-10-06
Make ExceptionMessage constraint strict, expectedException can handle only expected class, not message nor code

### LINKED ISSUE #3325: Make ExceptionMessage constraint strict
| Q                   | A
| PHPUnit version     | ?
| PHP version         | not related
| Installation Method | not related

Heavily descried in:
https://github.com/sebastianbergmann/phpunit/pull/3309#discussion_r219665621

`Constraint\ExceptionMessage`, used under the hood by `expectExceptionMessage` and `@expectedExceptionMessage` is not strict about expected message.

Constraint does not check if expected message is same as actual one, but only if it's somewhere present in actual one.
Making those two calls equivalent:
```
$this->expectExceptionMessage('variadic');
$this->expectExceptionMessageRegExp('/^.*variadic.*$/');
```

Which is not obvious for developer and actually most of them are not aware about this.

I would like to change behaviour of Constraint to use `===` comparison instead of `strpos`.

For me, personally, it's a bug, yet I can understood that requested change may be considered as bigger one and treated as BC breaker.

If this request will be approved, I'm all in implementing this change myself - either to lowest supported branch, if this would be treated as a bugfix, or later for 8.0 branch (which is not yet existing).

Thanks !

keradus:
Closes #3325

Commit logs are meaningful.

As  described in ticket, this change may be consider a little bigger than it appears on first look.
Eg, changing `strpos` to `===` in constraint makes (by purpose) partial message not working anymore, and `@expectedException \ClassOfException Message Code` annotation, well, make no sense anymore - as `Message` can be exactly one, single word - and without partial exception, it's pointless...

### DIFF
--- a/src/Framework/Constraint/ExceptionMessage.php
+++ b/src/Framework/Constraint/ExceptionMessage.php
@@ -59,7 +59,7 @@ class ExceptionMessage extends Constraint
         }
 
         return \sprintf(
-            "exception message '%s' contains '%s'",
+            "exception message '%s' equals '%s'",
             $other->getMessage(),
             $this->expectedMessage
         );
@@ -74,6 +74,6 @@ class ExceptionMessage extends Constraint
             return 'exception message is empty';
         }
 
-        return 'exception message contains ';
+        return 'exception message equals ';
     }
 }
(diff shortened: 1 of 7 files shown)


sebastianbergmann:
Can you please send a pull request for `7.4`? Thanks!

TomasVotruba:
Thank you :+1:

keradus:
sure, here it is #3343

guilliamxavier:
For the record: this was merged but then **reverted** in branch `6.5` (because a BC break)

https://github.com/sebastianbergmann/phpunit/issues/3325#issuecomment-428440589

PR #3327, opened 2018-10-06
CS: fix casing of callback pseudotype

keradus:


src/Framework/Assert.php
```diff
@@ -2337,7 +2337,7 @@ public static function isTrue()
     /**
      * @param callable $callback
      *
-     * @return Callback
+     * @return callback
```
localheinz: But it returns an instance of `Callback`?

Maybe it's a bit confusing because it could be interpreted as [`callback` pseudo-type](http://php.net/manual/en/language.pseudo-types.php#language.types.callback), what do you think, @keradus?

sebastianbergmann:

PR #3328, opened 2018-10-06
Cleaned doc-blocks

FrontEndCoffee:
I cleaned up the doc-blocks a bit.

Since there is still no PSR standard for this, I used the standard used by Squizlabs.

### DIFF
--- a/src/Framework/Assert.php
+++ b/src/Framework/Assert.php
@@ -68,14 +68,14 @@ use Traversable;
 abstract class Assert
 {
     /**
-     * @var int
+     * @var integer
      */
     private static $count = 0;
 
     /**
      * Asserts that an array has a specified key.
      *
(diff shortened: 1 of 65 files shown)


sebastianbergmann:

PHP used `int` and `bool` instead of `integer` and `boolean`. And these docblock will go away entirely sooner rather than later anyway.

kubawerlos:
@FrontEndCoffee it's from [Symfony Coding Standards](https://symfony.com/doc/current/contributing/code/standards.html):

> For type-hinting in PHPDocs and casting, use `bool` (instead of `boolean` or `Boolean`), `int` (instead of `integer`), `float` (instead of `double` or `real`);

and is configured [here](https://github.com/sebastianbergmann/phpunit/blob/7.4.0/.php_cs.dist#L153).

PR #3329, opened 2018-10-08
Load autoload.php from parent folder

ruudk:
I tried running phpunit and got this:
```
$ vendor/bin/phpunit
You need to set up the project dependencies using Composer:

    composer install

You can learn all about Composer on https://getcomposer.org/.
```

Turns out that the search for autoload.php is not correct when `phpunit` is inside `vendor/bin`.

The `autoload.php` file is in `vendor/autoload.php`. That's why it should search for `../autoload.php` first (when `phpunit` is inside `vendor/bin`)

### DIFF
--- a/phpunit
+++ b/phpunit
@@ -27,7 +27,7 @@ if (!ini_get('date.timezone')) {
     ini_set('date.timezone', 'UTC');
 }
 
-foreach (array(__DIR__ . '/../../autoload.php', __DIR__ . '/../vendor/autoload.php', __DIR__ . '/vendor/autoload.php') as $file) {
+foreach (array(__DIR__ . '/../autoload.php', __DIR__ . '/../../autoload.php', __DIR__ . '/../vendor/autoload.php', __DIR__ . '/vendor/autoload.php') as $file) {
     if (file_exists($file)) {
         define('PHPUNIT_COMPOSER_INSTALL', $file);
 


sebastianbergmann:
I do not understand the problem you are trying to solve.

ruudk:
My issue is that `phpunit` cannot find the autoload.php file on my system:
```
user@53b728fbe17d:/var/www/app$ pwd
/var/www/app
user@53b728fbe17d:/var/www/app$ ls -la vendor/autoload.php
-rw-r--r-- 1 user user 178 Oct  8 10:01 vendor/autoload.php
user@53b728fbe17d:/var/www/app$ phpunit
You need to set up the project dependencies using Composer:

    composer install

You can learn all about Composer on https://getcomposer.org/.
```

When I open the `vendor/bin/phpunit` file and add some logging:
```diff
 foreach (array(__DIR__ . '/../../autoload.php', __DIR__ . '/../vendor/autoload.php', __DIR__ . '/vendor/autoload.php') as $file) {
+    echo "Checking if file exists: " . $file . "\n";
     if (file_exists($file)) {
         define('PHPUNIT_COMPOSER_INSTALL', $file);
 
         break;
     }
 }
```
and run it again:
```
user@53b728fbe17d:/var/www/app$ phpunit
Checking if file exists: /var/www/app/vendor/bin/../../autoload.php
Checking if file exists: /var/www/app/vendor/bin/../vendor/autoload.php
Checking if file exists: /var/www/app/vendor/bin/vendor/autoload.php
You need to set up the project dependencies using Composer:

    composer install

You can learn all about Composer on https://getcomposer.org/.
```

As you can see, phpunit cannot find the `autoload.php` file.

PR #3331, opened 2018-10-09
Enhancement: Configure visibility_required fixer

localheinz:
This PR

* [x] configures the `visibility_required` fixer to also require visibility for `const`
* [x] runs `php-cs-fixer`

Follows https://github.com/sebastianbergmann/php-file-iterator/pull/56.


localheinz:

Issue #3180, opened 2018-06-22
maxDepth param of assertions is not used ?


keradus:
What is this param? I can't find it being used anyhow
it's passed ultimately to `Constraint\IsEqual::maxDepth`, but it is not used as well.

i believe it's a leftover and shall be cleaned up

keradus:
ssshhh, go away @stale

PR #3335, opened 2018-10-10
Fix: Remove unused field $maxDepth from IsEqual constraint

localheinz:
This PR

* [x] removes an unused field `$maxDepth` from the `IsEqual` constraint

Related #3180.

💁‍♂️ `$maxDepth` still needs to be removed from a lot of other places, which partially constitute BC issues. Happy to work on it, if there are no objections!

### DIFF
--- a/src/Framework/Constraint/IsEqual.php
+++ b/src/Framework/Constraint/IsEqual.php
@@ -35,11 +35,6 @@ class IsEqual extends Constraint
     private $delta;
 
     /**
-     * @var int
-     */
-    private $maxDepth;
-
-    /**
      * @var bool
      */
     private $canonicalize;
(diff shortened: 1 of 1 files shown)


localheinz:
@sebastianbergmann 

I’ll have a look after breakfast!

localheinz:
@sebastianbergmann 

I might be wrong, but I think it is d38ea8fbc8792391771efd8e03e8c39b9266cfbb.

To find it, I ran

```
$ git log -S maxDepth --source --all
```

Does that make sense?

localheinz:
@sebastianbergmann 

Do you want to hold off with this until work has started on PHPUnit 8, then? Just to make sure I understand correctly.

sebastianbergmann:
Yes. This will be done in `master` after PHPUnit 7.5 has been released.

OskarStark:
> Yes. This will be done in master after PHPUnit 7.5 has been released.

But it will be deprecated first, right?

PR #3348, opened 2018-10-14
Don't ignore coverage of methods without assertions

lcobucci:
PHPUnit was not adding the coverage of tests with `@doesNotPerformAssertions` or `$this->expectNotToPerformAssertions();` to the report, leading to wrong coverage results.

### DIFF
--- a/src/Framework/TestResult.php
+++ b/src/Framework/TestResult.php
@@ -754,6 +754,7 @@ class TestResult implements Countable
         }
 
         if ($this->beStrictAboutTestsThatDoNotTestAnything &&
+            $test->doesNotPerformAssertions() === false &&
             $test->getNumAssertions() == 0) {
             $risky = true;
         }
--- a/tests/_files/CoverageDoesNotPerformAssertionsTest.php
+++ b/tests/_files/CoverageDoesNotPerformAssertionsTest.php
@@ -1,0 +1,25 @@
+<?php
+/*
+ * This file is part of PHPUnit.
+ *
+ * (c) Sebastian Bergmann <sebastian@phpunit.de>
+ *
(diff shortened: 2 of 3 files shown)


sebastianbergmann:
I was able to agree to adding `@doesNotPerformAssertions` and `$this->expectNotToPerformAssertions();`. I am not able to agree to marking code as covered that is not covered by an assertion.

lcobucci:
@sebastianbergmann so do you expect people to rely on `$this->addToAssertionCount()` to test the happy path of methods that raise exceptions? Like:

```php
public function assert(): void
{
    if (/** condition */) {
        throw new Exception();
    }
} 
```

I respect your decision but I find it hard to agree that the fact that setting the test not expect an assertion is not an assertion, since it actually sets the intention of the test. I would even argue that it is similar to setting the test to expect an exception.

lcobucci:
It's important to mention that I don't really care about having 100% of coverage but having the coverage affects my mutation tests and msi and covered msi are relevant to me.

lcobucci:
@sebastianbergmann the issue #3348 and all related PRs/issues shows how people are using `@doesNotPerformAssertions`... it's kind of sad to not even have a discussion on this.

sebastianbergmann:
@lcobucci Sounds like a documentation issue to me.

lcobucci:
@sebastianbergmann alright, I can send a PR to the doc repo to solve that issue.

Could you please confirm that the way expected ways to assert that no exception has been thrown are: using `$this->addToAssertionCount()` or adding a dummy assertion after the method that is expected to not raise the exception?

Even though I might not agree with either it's important to make things clear on the docs :+1:

PR #3359, opened 2018-10-19
[WIP] Test with system memory allocator

sergeyklay:
`USE_ZEND_ALLOC=0` in the environment will stop the memory manager from functioning, all allocations fall back on the default system allocators which can be useful for debugging leaks, and should only be used for this purpose.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,8 +20,9 @@ matrix:
 
 env:
   matrix:
-    - DEPENDENCIES="high"
-    - DEPENDENCIES="low"
+    - DEPENDENCIES="high" USE_ZEND_ALLOC=1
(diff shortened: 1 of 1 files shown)


sergeyklay:
I am just testing one of my hypotheses. Please ignore this PR.

sebastianbergmann:
Can you elaborate on why this should be added to PHPUnit's build? Thanks!

sergeyklay:
@sebastianbergmann As I explained above, I simply test one of my hypotheses. Please disregard this PR. But note: there is a memory SEGFAULT for PHP 7.3. I’ve been working on this problem for the second week and only last night I realized that the problem is related to PHPUnit + PHP 7.3 + USE_ZEND_ALLOC=0

sebastianbergmann:
Okay.

PR #3377, opened 2018-10-31
Fix warning about missing covers annotations

### LINKED ISSUE #3376: Test without @covers annotation is only marked as risky when forceCoversAnnotation=true and beStrictAboutCoversAnnotation=true is set
In our project the `forceCoversAnnotation` option is enabled.
This option causes tests without covers annotations to cover nothing.
So it would be really helpful to get a warning about these tests without covers annotations.
In `CodeCoverage` there is a option `checkForMissingCoversAnnotation` which is enabled if `strictCoverage` is enabled, unfortunately this option also requires the uses annotations and we don't want to use them as they are to strict.

So as a solution I've two suggestions (I provide a PR for both):

> Introduce a new option to only check missing covers annotations #3374 

or

> Automatically check for missing covers annotations if `forceCoversAnnotation` is enabled #3375

For me the second option is my favorite because is see no benefit of having tests without covering anything.

What do you think about this feature? Which solution makes more sense in your opinion.

nicolashohm:
Fix #3376

### DIFF
--- a/src/TextUI/TestRunner.php
+++ b/src/TextUI/TestRunner.php
@@ -491,6 +491,10 @@ class TestRunner extends BaseTestRunner
                 $codeCoverage->setForceCoversAnnotation(
                     $arguments['forceCoversAnnotation']
                 );
+
+                if ($arguments['forceCoversAnnotation']) {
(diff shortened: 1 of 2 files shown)


sebastianbergmann:

Issue #3380, opened 2018-11-02
"Buffered" TestDox printer(s)


sebastianbergmann:
Thanks to #3092, #3147, and #3284, the order in which tests are run is no longer (always) based on "order in which DirectoryIterator finds *Test.php files" and "order of test methods within test class". This renders the TestDox result printer(s)'(s) output useless.

The TestDox result printer(s) should be changed to buffer test results in order to ensure the right order and then emit all output after the last test was run.

sebastianbergmann:
@rpkamp @epdenouden What do you think?

rpkamp:
It makes sense, although I think it would make sense if this were opt-in behaviour for the printer that could be enabled through some flag (like `$expectNonLinearOutput`) so that the default behaviour is as it is now (which is very handy I think) but can be switched to buffer when the normal flow doesn't make much sense.

sebastianbergmann:
Sorry for not making this clear: yes, the buffering should be optional. The test runner can detect whether the option is required or not and enable it automatically when it is.

epdenouden:
@sebastianbergmann I'm all for it when it is optional. I regularly use the visibility of the changed execution order of the TestDox printer to debug the sorters, for example in #3246 :-)

epdenouden:
@sebastianbergmann Question: being as busy as you are, shall I pick up implementing this? I have some familiarity with this part of the codebase and I use the feature in my daily work. Nevermind if you're already working on it.

sebastianbergmann:
@epdenouden That would be great -- and much appreciated. I cannot get to this before next week at the earliest.

rpkamp:
@epdenouden Seeing how the TestListener will be deprecated it would make more sense to implement this as an extension I think. But that would be easier once the CliTestDoxPrinter is also an extension, as you could then just decorate it, making it more generic in case other extensions need it as well.

That does need some more thought on how to get the printers from TestListeners to extensions though.

sebastianbergmann:
Not sure that printers should be extensions, but they should definitely use the `TestHook` interfaces instead of the `TestListener` interface. #3390 is about finding a good way to do that.

epdenouden:
Some food for thought then @rpkamp and @sebastianbergmann. I will first stomp the ordering bug #3246 and then code a proposal for this.

Unless were are doing double work when `CliTestDoxPrinter` is also getting an overhaul? Will read up on all discussions before I start coding.

epdenouden:
[...]

PR #3381, opened 2018-11-04
Introduce TestRunAwareTestListener

rpkamp:
Adds startTestRun and endTestRun hooks to TestListener so implementing
classes don't have to check whether a startTestSuite call is the first
one, plus it allows implementers to do something after the very last
test, which is usefull for implementers that want to buffer data and
only output once everything is done.

### DIFF
--- a/src/Framework/TestResult.php
+++ b/src/Framework/TestResult.php
@@ -364,15 +364,33 @@ class TestResult implements Countable
         $this->time += $time;
     }
 
+    public function startTestRun(TestSuite $suite): void
+    {
+        $this->topTestSuite = $suite;
+
+        foreach ($this->listeners as $listener) {
+            if ($listener instanceof TestRunAwareTestListener) {
+                $listener->startTestRun($this->topTestSuite);
+            }
+        }
+    }
+
+    public function endTestRun(): void
+    {
+        $this->flushListeners();
+
+        foreach ($this->listeners as $listener) {
+            if ($listener instanceof TestRunAwareTestListener) {
+                $listener->endTestRun($this->topTestSuite, $this);
+            }
+        }
+    }
+
     /**
      * Informs the result that a test suite will be started.
      */
     public function startTestSuite(TestSuite $suite): void
     {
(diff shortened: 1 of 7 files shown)


sebastianbergmann:
`BeforeFirstTestHook` and `AfterLastTestHook` already exist. Or am I missing something?

sebastianbergmann:
Maybe I should have documented the fact that the legacy `TestListener` interface that contains way too many methods will eventually go away.

Units of code that want to observe test execution should use the new `TestHook` interfaces. This is done by writing a class that implements one or more `TestHook` interfaces and registering it as a test runner extension.

rpkamp:
Okay, so basically I've made the lagacy code forward compatible? :smile: 

Okay, if the `TestListener` needs to go -which I agree would be a good thing, as it doesn't really adhere to the Interface Seggregation Principle- then what happens to the printers?

The printers are currently all implemented as `TestListener`s, but should they become extensions instead?

The whole point of this PR was to get a signal in the CliTestDoxPrinter when the last test had run so we could output any buffered content, required for https://github.com/sebastianbergmann/phpunit/issues/3380; my idea was to create a BufferedTestListener as a decorator for other `TestListener`s that would order and output all data at the end when all test are done. This decorator could then be used general purposes for all kinds of printers that need sorted output.

Also, I don't like using `printResults` as a signal to output the buffered content, as `printResult` is not part of any interface, so it relies on duck typing.

The interfaces exist, now it's a matter of figuring out how to best use them to solve this, I think.

epdenouden:
@rpkamp Perhaps the caching listener (or listening cache? :) can be of some inspiration too: https://github.com/sebastianbergmann/phpunit/blob/1349e75e079159ec19f42f75c4c79a69624be513/src/Runner/ResultCacheExtension.php#L12

rpkamp:
@epdenouden yes, that looks good.

However, since that's an extension and the printer is a TestListener I'd need an adapter. I'd like a decorator instead, but for that to work the printer would need to be an extension too :slightly_smiling_face:

Issue #3391, opened 2018-11-05
Add a Hook that fires after each test, regardless of result


rpkamp:
I currently have a TestListener that resets a service after each has run. This must be done regardless of the result of the test.

With the TestListener this is easy, implement `TestListener::endTest()` and done. However, with the new extension system no such hook exists and it seems I'd need to implement `AfterIncompleteTestHook`, `AfterRiskyTestHook`, `AfterSkippedTestHook`, `AfterTestErrorHook`, `AfterSuccessfulTestHook`, `AfterTestFailureHook` and `AfterTestWarningHook`. All with the same function body. That seems a bit overkill.

Is there another way to do this with an extension currently? If not, would you mind a PR that adds such a Hook?

sebastianbergmann:
I would not mind a pull request that adds an `AfterTestHook`.

maples3:
For anyone reading this in the future, as of PhpUnit 11 this is now done by writing an extension (implementing `PHPUnit\Runner\Extension\Extension`) to register a class that implements `PHPUnit\Event\Test\FinishedSubscriber`.

PR #3392, opened 2018-11-05
Add AfterTestHook for extensions

rpkamp:
Fixes https://github.com/sebastianbergmann/phpunit/issues/3391

So extensions have hook for some common functionality that should
happen after every test, regardless of the result.

### DIFF
--- a/src/Runner/Hook/AfterTestHook.php
+++ b/src/Runner/Hook/AfterTestHook.php
@@ -1,0 +1,21 @@
+<?php declare(strict_types=1);
+/*
+ * This file is part of PHPUnit.
+ *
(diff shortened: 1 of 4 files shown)


sebastianbergmann:
Thanks!

localheinz:
@rpkamp @sebastianbergmann 

Unless I am mistaken, I believe the `AfterTestHook` should have extended the `TestHook` interface.

Otherwise, if you implement an extension that only implements the `AfterTestHook` interface will never run:

https://github.com/sebastianbergmann/phpunit/blob/7.5.0/src/TextUI/TestRunner.php#L230-L234

PR #3398, opened 2018-11-10
Fix bug when replacing placeholders in testdox using an associative array

marcospassos:
At Croct we usually assign keys to data provider arguments to improve code readability:
```php
private function dataProvider() : array
{
    return [
        '1 + 2 = 3' => [
            'augend' => 1,
            'addend' => 2,
            'result' => 3
        ]
    ];
}
```
It works fine in all cases except when the `@testdox` has placeholders. In this case you will get an error:
```
Undefined offset: 0
```
There is also a bug when omitting optional parameters in data providers:
```php
private function dataProvider() : array
{
    return [
        [1]
    ];
}

/**
 * @testdox Message regarding $optional
 */
public function testFoo(int $value, bool $optional = true) : void
{
}
```
Warning:
```
Undefined offset: 1
```
This PR includes the fixes and tests for both scenarios.

### DIFF
--- a/src/Util/TestDox/NamePrettifier.php
+++ b/src/Util/TestDox/NamePrettifier.php
@@ -158,11 +158,15 @@ final class NamePrettifier
     {
         $reflector          = new \ReflectionMethod(\get_class($test), $test->getName(false));
         $providedData       = [];
-        $providedDataValues = $test->getProvidedData();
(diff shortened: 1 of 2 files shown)


epdenouden:
@marcospassos Looks very useful :)

sebastianbergmann:
Merged into `7.4`, thanks.

PR #3400, opened 2018-11-13
Updated teamcity logger

### LINKED ISSUE #3364: PhpStorm/Teamcity + PhpUnit 7.4.1 - Error in setUpBeforeClass
| Q                   | A
| PHPUnit version     | >=7.4.0
| PHP version         | 7.1.8

Sample:
```json
{
    "require": {
        "php": "^7.1"
    },

    "require-dev": {
        "phpunit/phpunit": "^7.4"
    }
}
```

```php
<?php
declare(strict_types=1);

require_once __DIR__ . '/../vendor/autoload.php';

class JustATest extends TestCase
{
    public static function setUpBeforeClass(): void
    {
        throw new \RuntimeException('Something\'s not quite right!');
    }

    public function testSomething(): void
    {
        $this->fail('This cannot work!');
    }
}

```

When I run this test in phpstorm i will just get "Empty test suite." with the success icon without any failed test. Full output:
```
PHPUnit 7.4.3 by Sebastian Bergmann and contributors.

Empty test suite.


Time: 64 ms, Memory: 2.00MB


ERRORS!
Tests: 1, Assertions: 0, Errors: 1.

Process finished with exit code 2
```

There seems to be an critical error in the teamcity logger (src/Util/Log/TeamCity.php). You added some "if (!$test instanceof TestCase) {return;}" in the last version and this also seem to trigger this error (or rather don't trigger an error). When I change the addError method to the following everything works like expected:
```php
public function addError(Test $test, \Throwable $t, float $time): void
    {
        if (!$test instanceof TestCase && !$test instanceof TestSuite) {
            return;
        }
```

I'm not sure whether this is everything to fully fix this bug. Propably the other added if-statements also have to be changed like this.

MaximilianKresse:
Fixed #3364
Added comparison for failed asserts in setUpBeforeClass
Changed printIgnoredTest to protected

### DIFF
--- a/src/Util/Log/TeamCity.php
+++ b/src/Util/Log/TeamCity.php
@@ -21,7 +21,6 @@ use PHPUnit\Framework\Warning;
 use PHPUnit\TextUI\ResultPrinter;
 use PHPUnit\Util\Filter;
 use ReflectionClass;
-use SAML2\Exception\Throwable;
 use SebastianBergmann\Comparator\ComparisonFailure;
 
 /**
--- a/src/Util/Log/TeamCity.php
+++ b/src/Util/Log/TeamCity.php
@@ -81,49 +81,6 @@ class TeamCity extends ResultPrinter
     }
 
     /**
-     * Print a failed test
-     */
-    protected function printFailedTest(Test $test, \Throwable $e, float $time): void
-    {
-        if (!$test instanceof TestCase && !$test instanceof TestSuite) {
-            return;
-        }
-
-        $parameters = [
-            'name'     => $test->getName(),
-            'message'  => self::getMessage($e),
-            'details'  => self::getDetails($e),
-            'duration' => self::toMilliseconds($time),
-        ];
-
-        if ($e instanceof ExpectationFailedException) {
-            $comparisonFailure = $e->getComparisonFailure();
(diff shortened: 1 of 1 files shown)


sebastianbergmann:

See https://github.com/sebastianbergmann/phpunit/issues/3364#issuecomment-438197355 for details.

Bob4ever:
What??? The comment just confirms that it's not enough what is actually in the code.

Bob4ever:
@sebastianbergmann Could you please explain the problem with the pull request in detail? Because I just can't see any problem especially not with the linked comment.

epdenouden:
Hello @Bob4ever! Also from me, thanks for spending your time on improving PHPUnit. @sebastianbergmann has linked to a comment of mine that mentions some of the issues with using the Listener/Printer interfaces for creating log formats.

I have not looked at your pull request yet, but will do so this weekend. A quick glance showed that you moved quite a bit of logic around, which is currently giving me headaches working on https://github.com/sebastianbergmann/phpunit/issues/3380

I'd be happy to work with you on modernizing the TeamCity logger. It's used extensively by IntelliJ/PHPStorm, for example. Have a good weekend!

Issue #3246, opened 2018-08-07
Execution order "by-defect" does not support parameterized tests with @dataProvider


reinholdfuereder:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 7.3.1
| PHP version         | 7.2.8
| Installation Method | PHAR

In my test set (executed via XML configuration) there is (coincidentally) (1) one incomplete, (2) one skipped and (3) one failing test. I just started playing around with the new cool feature of running tests first that failed in a previous run, but it does not behave as expected:
```
acmedev@dev56w1:~$ rm .phpunit.result.cache
rm: remove regular file ‘.phpunit.result.cache’? y
acmedev@dev56w1:~$ php /mnt/data/Kunden/acme/components/phpunit/7.3.1/phpunit.phar --configuration /home/data/acmedev/unittests/AllPlainPHPUnitTests.xml --order-by=defects --cache-result
Execution according to PHPUnit XML configuration file '/home/data/acmedev/unittests/AllPlainPHPUnitTests.xml'...
PHPUnit 7.3.1 by Sebastian Bergmann and contributors.

.................I.............................................  63 / 524 ( 12%)
............................................................... 126 / 524 ( 24%)
............................................................... 189 / 524 ( 36%)
............................................................... 252 / 524 ( 48%)
............................................................... 315 / 524 ( 60%)
............................................................... 378 / 524 ( 72%)
..........................................................F.... 441 / 524 ( 84%)
............................................................... 504 / 524 ( 96%)
.S..................                                            524 / 524 (100%)

Time: 6.41 seconds, Memory: 94.25MB

There was 1 failure:

1) Acme_Phone_FormatNumberTest::formatNumberForMobileTelLink with data set #8 ('  123-456-789', '123456789')
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'123456789'
+'      123456789'

/home/data/acmedev/unittests/unit/app/admin/classes/phone/formatnumberTest.php:40

FAILURES!
Tests: 524, Assertions: 1031, Failures: 1, Skipped: 1, Incomplete: 1.
acmedev@dev56w1:~$ php /mnt/data/Kunden/acme/components/phpunit/7.3.1/phpunit.phar --configuration /home/data/acmedev/unittests/AllPlainPHPUnitTests.xml --order-by=defects --cache-result
[...]

Issue #3396, opened 2018-11-09
Improve internals of caching and execution ordering logic


epdenouden:
### Background
_As a_ software developer
_I want_ the test execution to be predictable
_so that_ my tests feel like a stable foundation

While working on #3246 it turned out to be another edge case for the `TestSuiteSorter`. The root cause for bugs and seemingly unpredictable execution re-orders is the way the `TestSuite`s and `TestCase`s are handled behind the scenes.

Like all testing frameworks, PHPunit provides us with ways to structure our collection of tests. Unsurprisingly these structures place limits on how much we can change the existing execution order:
- automated handling of environments the tests run in: setup, teardown, environment backups
- a suite of tests are combined as methods in a class
- individual tests made dependent on others using `@depends`
- individual tests run with multiple sets of data using `@dataprovider`
- PHPT for end-to-end tests
- using the XML configuration file `TestSuite`s can be nested
- subsets of the collection can be run by specifying filters in the configuration or by command line

PHPUnit's internals are not a 1:1 mapping on the PHP code the end user is working on. PHPT testcases are quite different from the more common class+methods based tests. Tests fed by `@dataprovider` are handled by a layer of abstraction that is completely invisible from the outside.

### Solution
Previous and current edge cases (like #3246) can be fixed cleanly and efficiently by improving the unique identification of each combination of (test, data).

### Impact
- this will solve #3246
- provides a suite of additional tests for the execution reordering
- performance and memory footprint remain the same
- different cache file format

### To do
- [x] investigate all variants of test names sent to the caching listener
- [x] investigate all variants of test names and suite structure sent to `TestSuiteSorter`
- [x] implement new scheme for uniquely identifying each test+data scenario
- [x] improve result cache backend
  - [x] more efficient storage of test names with large `@dataprovider` data
  - [x] ~keep the last few (10?) test results of each unique test~
  - [x] ~automatically prune tests from cache that no longer exist~
- [x] add regression test for correctly sorting `@dataprovider` tests with failing rows

PR #3401, opened 2018-11-13
All-day breakfast at the dataprovider buffet

epdenouden:
Thanks to @reinholdfuereder some remaining issues with the reordering of `@dataprovider`-fed tests have been resolved. As described in more detail #3396 and its predecessor ticket #3246, the main problem with execution reordering has been understanding the hierarchy of the run queue.

### The fix
Lack of a consistent naming scheme for each individual combination of (test, data) reared its ugly head again. This has now been solved cleanly and I've updated all the tests to match the new scheme, so all previous functionality remains under guard. I have added a regression end-to-end test to keep an eye on the dataproviders.

### Impact
- failing dataprovider-fed tests are now sorted correctly within their `TestSuite`, in addition to the (already existing) functionality to sort failing dataprovider rows to the front of their parent test
- previous execution order sorters are unaffected
- the cache uses a different format to ID tests: `NameSpace\TestClass::testMethod`, `NameSpace\TestClass::testMethod with data set $data` or a PHPT filename
- the `TestSuiteSorter` now uses normalized FQCN+dataname everywhere
- the `ResultCacheListener` keeps the FQCN+dataname of tests but strips the dump of the data

### Additional notes
I have opted to *not* add refinements to the `TestResultCache` at this time, to keep the changes atomic. If I learned anything from #3246 is that debugging the sorters is still very time consuming. Such optimizations add complexity and would only be noticeable by maintainers of large projects and even then it's more of a nice-to-have.

As always let me know if you have any remarks or wishes.


sebastianbergmann:
Merged into `7.4`, thanks.

epdenouden:
@sebastianbergmann Thanks! This PR fixes #3246 and #3396, they can be closed, too

PR #3402, opened 2018-11-13
Fix reverse order test failures

epdenouden:
Listening to the tests in reverse revealed evil hidden messages 🤘


sebastianbergmann:
Merged into `7.4`, thanks.

PR #3408, opened 2018-11-15
Fix TestSuite sorting glitch

epdenouden:
Repaired two bugs that crept in while working on the dataprovider sorting fixes. Didn't cause any failing tests, but nested `TestSuites` were not seen by the sorter. Also found the correct regression test for #3396 in a git stash. 🤐


epdenouden:
The coverage will be up again with the buffered TestDox.

sebastianbergmann:
Merged into `7.4`, thanks.

PR #3414, opened 2018-11-21
willThrowException() only accepts Exception, not Throwable

Stadly:


gmponos:
I need this too.. any progress?

epdenouden:
Change looks useful. @stadly can you merge in the latest master, so all tests run? Thx

Stadly:
Done

epdenouden:
That's quick, thank you!

epdenouden:
@sebastianbergmann Change looks useful to me, would be nice to merge this, unless you know of something else breaking. The logging and reporting side uses `Throwable` already, too.

gmponos:
@epdenouden I am not sure if with the current change other places must change as well...

I mean what about `expectedException` function.. this must also allow `Throwables`.. 

Haven't checked the code though... so it might already be fine...

gmponos:
> The logging and reporting side uses Throwable already, too

@epdenouden ah.. when you were saying reporting you were referring to things like `expectedException`?

epdenouden:
@gmponos That is a good point. I have not done an in-depth code review of the surrounding code. 🤐 Internally I think we can upgrade it all to `Throwable` anyway. Only exception would be strict checks for `Exception`, I think.

With the reporting I only meant the method signatures in the printers and listeners. Lots of stuff going to be refactored there and `Throwable` is the new fashionable thing. :)

gmponos:
On the other hand you could accept current PR and see the other places later.. your call..

epdenouden:
Yes. The built-in collection of tests should keep us relatively safe 😂

sebastianbergmann:
Merged manually, thanks.

Issue #3390, opened 2018-11-05
Make TestHook interfaces as easy to use as TestListener


sebastianbergmann:
This issue is where the discussion that was started in #3381 should continue.

Related to #3388 and #3389.

rpkamp:
A few thoughts:

1. The `TestHook` doesn't provide any functionality similar to TestListener::startTestSuite and TestListener::endTestSuite at the moment, but both `TeamCity` and `JUnit` use those and I don't see a way round it. So we probably need a `BeforeTestSuiteHook` and `AfterTestSuiteHook`?

2. In terms of BC it's probably easiest to implement all classes that currently use TestListener again using the hooks instead, so PHPUnit itself can use the new extensions, while userland can continue to extend the TestListeners, at least until PHPUnit 9 where the TestListener and everything that implements it will be removed. That does mean there is no more coverage of the old classes in the e2e tests. Not sure if that's a bad thing though, since PHPUnit itself isn't using them anyway?

3. The `printerClass` setting could remain intact, but we could check in the TestRunner if it's a TestListener or Extension, and register it accordingly. When it's a TestListener we could additionally print a deprecation message (in 8.0).

rpkamp:
I just tried a quick proof of concept to make the TeamCity logger an extension and as it stands this is not possible.

First of all, the printResult method (which would become the executeAfterLastTest method) needs a TestResult, but doesn't get one. Adding this to the interface would be a BC break. 

Then the addError (which would become the executeAfterFailedTest I believe) needs access to the actual Throwable to get a stack trace, but only receives a message string.

I stopped trying after that, but I'm sure there would be more issues in the same vain.

epdenouden:
@rpkamp Thanks for your detailed feedback! I kept notes of what I ran into when refactoring the TestDox printer for #3380 and my list of required changes for the `TestHook` closely mirrors yours. 

I will write up a more detailed response and proposal. The bottomline is the same as Remon says: the current printers/listeners get a lot more detail about the test run. The equivalent of `startTest()` and `endTest()` are just:

```
interface BeforeTestHook extends TestHook
{
    public function executeBeforeTest(string $test): void;
}

interface AfterSuccessfulTestHook extends TestHook
{
    public function executeAfterSuccessfulTest(string $test, float $time): void;
}
```

Reimplementing the current builtin output loggers requires (much) more information to be available.

localheinz:
[...]

PR #3417, opened 2018-11-26
Buffered TestDox printer

epdenouden:
This addition provides a working solution to #3380. For Nikolaus you wanted:

> The TestDox result printer(s) should be changed to buffer test results in order to ensure the right order and then emit all output after the last test was run.

Refactoring the TextDox printer turned out be a bit of work. To clean up the internal `TestResult` magic I ended up rewriting large parts. Still some minor cleaning up to do, but it's good enough for daily use. 

### Changes
- `TestSuiteSorter` now calculates and stores the original execution order before a run
- naming scheme logic for test reordering consolidated in `TestSuiteSorter::getTestSorterUID`
- main runner passes this original order into a `CliTestDoxPrinter` extension when required
- prints unbuffered when original execution order is not provided
- when buffering the printer will flush test results in the original order as soon as it can
- new PHPT end-to-end test keeps an eye on functionality
- unit test fixes to match test naming scheme

### How to test
Edit `phpunit.xml` and switch to `verbose=false`. Then run the test suite in both buffered and unbuffered TestDox mode:
```
./phpunit --testdox | tee /tmp/new.unbuffered
./phpunit --testdox --order-by=defects,depends | tee /tmp/new.buffered
diff /tmp/new.unbuffered /tmp/new.buffered
```

### Notes
- output diff: when buffering the names of some tests have a bit of extra detail
- could clean up unbuffered message formatting and writing; rather do that after the test Hooks are done

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -2,6 +2,7 @@
 <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xsi:noNamespaceSchemaLocation="phpunit.xsd"
          bootstrap="tests/bootstrap.php"
+         cacheResult="true"
          verbose="true">
     <testsuites>
         <testsuite name="unit">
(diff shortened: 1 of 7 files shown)


epdenouden:
@sebastianbergmann Working on the TestDox has given me good insight in what is required for improving the TestHook interface as mentioned in #3390

PR #3418, opened 2018-11-26
Fix conflicting placeholder replacement and argument exporting inconsistencies in @testdox

marcospassos:
This PR fixes a bug when two variables share the same prefix resulting in a badly formatted testdox:

```php
/**
 * @testdox Message regarding $foo and $fooBar
 * @dataProvider fooProvider
 */
public function testFoo(string $foo, string $fooBar) : void
{
}
````
Outputs:
```
Message regarding fooValue and fooValueBar
```
It also fixes inconsistencies on how arguments are rendered. Currently, a string `foo` rendered unquoted, while a numeric string `124` is enclosed in quotes. The same occurs for float values. The fix addresses these issues by using the exporter logic to handle boolean, integer and float values only. The reasoning is that if one wants to use quotes, it is preferable to put it in the `@testdox`.

### DIFF
--- a/src/Util/TestDox/NamePrettifier.php
+++ b/src/Util/TestDox/NamePrettifier.php
@@ -63,20 +63,19 @@ final class NamePrettifier
         $annotations                = $test->getAnnotations();
         $annotationWithPlaceholders = false;
 
+        $callback = static function (string $variable) : string {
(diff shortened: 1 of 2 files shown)


marcospassos:
@sebastianbergmann looks like the appveyor build is broken.

sebastianbergmann:
Merged manually, thanks.

Issue #3133, opened 2018-05-14
Enable dependency resolution by default


sebastianbergmann:
* https://github.com/sebastianbergmann/phpunit/pull/3092
* 8fa2aff0793e896be20f3e85585c827bc5e60530

epdenouden:
@sebastianbergmann implementing this atm :)

epdenouden:
AppVeyor needs a little tweak for the newer PHPUnit:
```
This version of PHPUnit is supported on PHP 7.2 and PHP 7.3.
You are using PHP 7.1.11 (c:\php\7.1.11\php.exe).
```
Not sure how to fix this correctly. Guess it's set in `appveyor.yml` in the repo?
```
  matrix:
    - PHP_VERSION: '7.1.11'
      XDEBUG_VERSION: '2.5.5-7.1'
      DEPENDENCIES: '--prefer-lowest'
    - PHP_VERSION: '7.1.11'
      XDEBUG_VERSION: '2.5.5-7.1'
      DEPENDENCIES: ''
```

epdenouden:
Fixed, updated AppVeyor to use PHP Windows binary for 7.2.12 and matching Xdebug.

PR #3421, opened 2018-11-27
Enable dependency resolution by default + AppVeyor

epdenouden:
Implements #3133 

- updated PHP and Xdebug version for AppVeyor Windows CI build
- switch dependency resolver on by default in configuration handlers
- remove `--resolve-dependencies` and `--ignore-dependencies` from help messages, but keep them functional. It would be user unfriendly to have legacy tests fail when upgrading to 8.x.
- added `--order-by=no-depends` flag to give user full on/off control over dependency resolver
- add `no-depends` as an option to XML-configuration
- update tests to use `--order-by` flags, keep minimal coverage for legacy options

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -10,11 +10,11 @@ environment:
   COMPOSER_ROOT_VERSION: '7.0-dev'
 
   matrix:
-    - PHP_VERSION: '7.1.11'
-      XDEBUG_VERSION: '2.5.5-7.1'
+    - PHP_VERSION: '7.2.12'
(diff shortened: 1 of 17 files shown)


epdenouden:
@keradus I now remember why I kept `depends` in addition to the new `no-depends`: the reason is indeed to make it explicit the feature is now turned on by default. I put it in `phpunit.xml` although it strictly speaking isn't needed. People that do not read the changelog might notice 'new' config option. :)

Like the legacy CLI flags I have removed `--order-by=depends` from the messages and help, but kept the functionality to make the transition to version 8 as painless as possible.

PR #3429, opened 2018-11-30
fix inefficient loop in getHookMethods and save up to 4% in cachegrind files

dynasource:
this commit is the result of finding a significant bottleneck in cachegrind files and performance gain up to 4% (measured in a project)
    
It fixes an inefficient loop in getHookMethods in which unnecessary logic can be executed when method docblock are not defined. It also minimizes calls to the same methods.

Also, the code is verbose enough to understand without the need for extra private methods.

### DIFF
--- a/src/Util/Test.php
+++ b/src/Util/Test.php
@@ -715,26 +715,30 @@ final class Test
                         continue;
                     }
 
-                    if (self::isBeforeClassMethod($method)) {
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
Can you please send this against `7.5`? Thanks!

sebastianbergmann:
Merged manually, thanks.

PR #3434, opened 2018-12-03
TeamCity logger skips over PhptTestCase events

### LINKED ISSUE #3432: TeamCity logger doesn't show PHPT tests
| Q                   | A
| PHPUnit version     | 7.5-devel
| PHP version         | PHP 7.2.11 with Xdebug 2.6.0

After some recent work on the TeamCity logger I noticed that it doesn't emit PHPT events anymore. 

This bug was introduced in https://github.com/sebastianbergmann/phpunit/commit/be6564ea49b96b9776c53f627a49683ec2687ce5#diff-46516cb8ec446cf630723cbba025f14eR60 to help with static analysis warnings. PHP TestCases are their own class and were skipped by the logger. Because of interfering issues this was not noticed.

epdenouden:
Fixes #3432 

This is the correct set of commits. Juggling one issue too many it seems ;-)

### DIFF
--- a/src/Util/Log/TeamCity.php
+++ b/src/Util/Log/TeamCity.php
@@ -57,10 +57,6 @@ class TeamCity extends ResultPrinter
      */
     public function addError(Test $test, \Throwable $t, float $time): void
     {
-        if (!$test instanceof TestCase) {
-            return;
-        }
-
         $this->printEvent(
             'testFailed',
             [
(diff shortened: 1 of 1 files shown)


epdenouden:
@sebastianbergmann hello there, this is the fastest `git merge --yolo` I have seen in quite some time

Naktibalda:
JUnit class has the same issue.

epdenouden:
@Naktibalda Thank you for pointing that out! I will check and fix where needed.

PR #3443, opened 2018-12-08
fix(templates): define the named constant "PHPUNIT_COMPOSER_INSTALL" before requiring the autoload file

fancyweb:
The goal : be consistent with what is done in the phpunit executable itself (constant definition and then requiring the composer autoload file).

My use case : In a particular setup, I need to check if phpunit is running in an autoloaded file. However, it is impossible when the tests are ran in separate processes because this particular constant is defined after requiring the composer autoload file.

### DIFF
--- a/src/Util/PHP/Template/PhptTestCase.tpl.dist
+++ b/src/Util/PHP/Template/PhptTestCase.tpl.dist
@@ -10,8 +10,8 @@ ob_start();
 $GLOBALS['__PHPUNIT_ISOLATION_BLACKLIST'][] = '{job}';
 
 if ($composerAutoload) {
-    require_once $composerAutoload;
     define('PHPUNIT_COMPOSER_INSTALL', $composerAutoload);
+    require_once $composerAutoload;
 } else if ($phar) {
     require $phar;
 }
--- a/src/Util/PHP/Template/TestCaseClass.tpl.dist
+++ b/src/Util/PHP/Template/TestCaseClass.tpl.dist
@@ -18,8 +18,8 @@ $phar             = {phar};
 ob_start();
 
 if ($composerAutoload) {
-    require_once $composerAutoload;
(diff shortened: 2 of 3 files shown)


fancyweb:
@sebastianbergmann Also, wouldn't it be better to add a dedicated constant "PHPUNIT_RUNNING" before any tests is launched ?

sebastianbergmann:
I do not understand the problem this is supposed to solve ("I need to check if phpunit is running in an autoloaded file").

fancyweb:
Sorry, I should have explained my use case better.

In a project, I have a file that is always required via the Composer "autoload" => "files" mechanism : 
```json
"autoload": {
    "files": [
        "src/script.php"
    ]
}
```

In this file, I wanna know if PHPUnit is running or no. The only constant that is defined by the PHPUnit executable before requiring the Composer autoload file is `PHPUNIT_COMPOSER_INSTALL` (https://github.com/sebastianbergmann/phpunit/blob/22ba4e0e2a75accab7e88a0c4622fccb358ee0a2/phpunit#L32). So I have to rely on it.

It is working fine as long as the tests are ran in the same process. However, when a test is ran in a different process, then the template system is used. And currently, this constant is defined after requiring the Composer autoload file.

This is not consistent.

Also, what I suggested in my second comment is to provide a dedicated constant `PHPUNIT_IS_RUNNING` for those kind of use cases (like PHPStan does for example https://github.com/phpstan/phpstan/blob/d13673474ca7f3cdd9c6442663eaab0e419c751d/bin/phpstan#L10)

sebastianbergmann:
Your code should not care about whether or not PHPUnit is running.

fancyweb:
I agree but in some very specific use cases (class substitution for me), it can be mandatory to know this to be able to use PHPUnit.

Also, the definition of this constant is still inconsistent when the tests are ran in the same process or in separate processes.

PR #3444, opened 2018-12-10
Fix ignored test with duplicate key in dataprovider

jderusse:
With generators dataProvider may have duplicate keys.

```php
    public static function dataProvider()
    {
        yield 'foo' => ['foo'];
        yield 'foo' => ['bar'];
    }
```

This PR throws an exception when the dataprovider contains duplicate keys.

### DIFF
--- a/tests/unit/Util/TestTest.php
+++ b/tests/unit/Util/TestTest.php
@@ -12,6 +12,7 @@ namespace PHPUnit\Util;
 use PharIo\Version\VersionConstraint;
 use PHPUnit\Framework\CodeCoverageException;
 use PHPUnit\Framework\Exception;
+use PHPUnit\Framework\InvalidDataProviderException;
 use PHPUnit\Framework\TestCase;
 use PHPUnit\Framework\Warning;
 
@@ -560,6 +561,14 @@ class TestTest extends TestCase
             ['H'],
             ['I'],
         ], $dataSets);
+    }
+
+    public function testWithDuplicateKeyDataProviders(): void
+    {
+        $this->expectException(InvalidDataProviderException::class);
+        $this->expectExceptionMessage('The key "foo" as already been defined in the dataprovider "dataProvider".');
+
+        Test::getProvidedData(\DuplicateKeyDataProviderTest::class, 'test');
     }
 
     public function testTestWithEmptyAnnotation(): void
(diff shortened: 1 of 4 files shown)


tests/unit/Util/TestTest.php
```diff
@@ -562,6 +562,16 @@ public function testWithVariousIterableDataProviders(): void
         ], $dataSets);
     }
 
+    public function testWithVDuplicateKeyDataProviders(): void
```
fancyweb: extra V here

fancyweb:
IMHO, it should throw an exception, as this is certainly a mistake and it is currently not working.

epdenouden:
@jderusse if you want to fix the build, cherry-pick https://github.com/sebastianbergmann/phpunit/pull/3445/commits/d607b3ea47f8854f6cea0266be01ad8bb3cc3fe6 :)

tests/unit/Util/TestTest.php
jderusse: fixed

jderusse:
reverted to exception instead of fallback

fancyweb:
Some people in [this PHPStan issue](https://github.com/phpstan/phpstan/issues/1408) think having the same key twice in a Generator is perfectly valid and that therefore, it's a bug in PHPUnit if it's not working properly.

Majkl578:
It's indeed a PHPUnit bug that it silently drops provided entries with duplicate names.
More that it also happens if you produce same name from multiple data providers for one test, this needs to be tested and covered too.

Majkl578:
Also I think expeption is not the solution, at least when duplicate keys come from different data providers. It should be fixed (i.e. by constraining origin data provider), not prevented.

jderusse:
The key in the dataprovider is used to help the developer to find quickly which datset is responsible of a test failure.

If you have 2 dataproviders with the same Key (or the same key repeated twice in a dataprovider) How would that key help you to find the responsible of the failure? 

Each time I had one, it was a typo or a copy/paste from an other rule. Do you have a valid usecase?

I'm agree with the phpstan thread, that duplicate key are valid in PHP. But in the case of phpunit dataprovider, it don't make sense and should be reported to the developer.

Majkl578:
@jderusse You can show both data provider name and key.

jderusse:
@Majkl578  Concatening the dataprovider with the key in order to, both, give extra information and allowing duplicate keys accross dataproviders is not the purpose of that PR (warn the developer about silently ignored tests), and could be addressed in a dedicated PR as it won't fixes duplicate keys in the same dataprovider.

Majkl578:
I only answered your own question.

sebastianbergmann:
Merged manually, thanks.

Issue #2647, opened 2017-04-11
Can @depends reference a test from outside the class?


Kwadz:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 6.1.0
| PHP version         | 7.0.15
| Installation Method | Composer 

According to [this article](https://web.archive.org/web/20150217042431/https://sebastian-bergmann.de/archives/826-Test-Dependencies-in-PHPUnit-3.4.html), it seems that from PHPUnit 3.4 `@depends class::method` references the test implemented in `class::method()`.

However, when I try to execute the following tests with `phpunit --configuration phpunit.xml tests/Functional/Example` I get `This test depends on "FirstTest::testProducerFirst" to pass.`.

```php
namespace Functional\Example;

use PHPUnit\Framework\TestCase;

class FirstTest extends TestCase
{
    public function testProducerFirst()
    {
        $this->assertTrue(true);
        return 'first';
    }

    public function testProducerSecond()
    {
        $this->assertTrue(true);
        return 'second';
    }

    /**
     * @depends testProducerFirst
     * @depends testProducerSecond
     */
    public function testConsumer()
    {
        $this->assertEquals(
            ['first', 'second'],
            func_get_args()
        );
    }
}
```

```php
namespace Functional\Example;

use PHPUnit\Framework\TestCase;

class SecondTest  extends TestCase
{
    public function testProducerFirst()
    {
        $this->assertTrue(true);
        return 'first';
    }

    public function testProducerSecond()
    {
        $this->assertTrue(true);
        return 'second';
    }

    /**
     * @depends FirstTest::testProducerFirst
     * @depends testProducerSecond
     */
    public function testConsumer()
    {
        $this->assertEquals(
            ['first', 'second'],
            func_get_args()
        );
    }


}
```

matejvelikonja:
any progress on this? I would like to know also.

El-Sam:
I need this too, any progress ?

GinoPane:
Yes, it can. Just make sure that test dependant test runs before your test.

igknighton:
@GinoPane How exactly can you do that though?

GinoPane:
@igknighton Exactly like it described in the example above. The file with "FirstTest" just need to be processed before "SecondTest" and they both must be in the same test suite. This can be achieved by XML configuration for your tests.

sebastianbergmann:
This is not a discussion or support forum, sorry.

[...]

PR #3439, opened 2018-12-07
TestDox colorization

epdenouden:
While working on the [buffered TestDox printer](https://github.com/sebastianbergmann/phpunit/pull/3417) I found the output format really useful for  work on the terminal. Adding some color highlighting and removing clutter gives it a completely different look. It's much easier to spot when something needs attention.

### Changes
Basic handling of ANSI colors gets its own helper class until we migrate over to a better CLI/console library. Its own test suite gives a quick demo of most features:

image

- mark changes in `TestSuite` by underlining the class name or class `@testdox`-annotation
- when verbose the test duration mirrors the status color
- reduced typographic clutter when using color
- `@dataprovider` parameter values used in `@testdox` are highlighted
- `@testdox` has access to the name of the `@dataprovider`-row via `$_dataName`
- show whitespace using dimmed interpuncts 
- improved visualization for support for `@dataprovider` without `@testdox`

image

- mirror test status colors in exception/error messages
- use dimmed colors to make PHPT path and filenames easier to read
- use dimmed colors to draw attention to changes in PHPT working directory
- highlighted stacktraces: changes in path+file, line numbers

image

The `CLITestDoxPrinter` has been cleaned up further. It has given me fresh ideas for the `TestListener` rewrite and cleaning up the pile of slow end-to-end tests. I'll be creating issues with proposed practical solutions for this. This has been a nice warming up for working on a new implementation for the test result event system.

Any remarks or ideas are very welcome as always.

Issue #3427, opened 2018-11-29
How can a Listener conditionally skip a test?


blockjon:
I'm very excited about the ability to implement PHPUnit\Framework\TestListener in order to add a custom behavior.

In my case, I want to execute some custom logic to determine which of the tests in my suites should be skipped.

Specifically, I want to download a list of tests that should be skipped and then enforce that skip list at runtime.

How should I do this? I would prefer to not have to change all of my test classes to extend from some new class or to put annotations all over.

Pytest has a similar option where you can hook into the moment just before a test executes and set the test to be skipped just before it would otherwise actually run the test. Can I do that? It would make my life great!

blockjon:
@epdenouden Thank you for your reply! Yes I would like to either skip tests or not play them at all... either option will work.

I'm trying to create PHP support for https://www.flaptastic.com/ which has a checkbox that allows you to instantly disable a test in CI. Here's what I've got so far: https://packagist.org/packages/blockjon/flaptastic

To "disable" a test in CI means to either skip it, or to not execute it in the first place.

![image](https://user-images.githubusercontent.com/1198745/49242026-68a85c00-f3be-11e8-98f4-d1658c168e80.png)

The goal here is to be able to skip certain tests at *runtime*. This is made to be possible because the test runner (PHPUnit in this case) downloads a list of file names and test function names that are to be skipped ... so I need some smart "PHPUnit way" to programmatically skip whichever tests are in that dynamic list which changes over time from within the PHPUnit test run in the simplest way as possible.

Here's what I've got so far:
https://packagist.org/packages/blockjon/flaptastic

sebastianbergmann:
This will likely be possible once #3387 has been implemented.

blockjon:
I think I have found what might be the simplest way to do it given the current PHPUnit architecture - to have them add a trait to their own abstract TestCase file. I assume most projects implement some kind of common TestCase that it could be added to.

https://github.com/blockjon/phpunit-flaptastic/blob/master/README.md#integrating-with-your-codebase

PR #3449, opened 2018-12-11
@Depends file will be run if not ran yet + Full namespace bug when using @depends [FEATURE&BUG-FIX]

NikoGrano:
#2647 

As @sebastianbergmann said:
> That being said, yes, @depends can be used to denote a dependency on a test that is declared in another test case class.
> Keep in mind, though, that PHPUnit cannot ensure that a depended-upon test is executed before the test(s) that depend(s) on it.

**This PR will now ensure depended-upon test is executed before the test(s) that depend(s) on it.**

Some stuff to note:
- If you're depended-upon test which is failed, your test will be marked as skipped with warning: `This test depends on "%s" to pass.`
- If you're depended-upon test which is skipped, your test will be marked as skipped with warning: `This test depends on "%s" to pass.`

Fix:
In case of you're writing `@depends \Some\Project\Tests\Unit\Domain\ProductTest::test__toString` this will fail with message `This test depends on "%s" to pass.`. Why? Due the `\` at beginning of the string... I fixed this with following code:
```php
                if (\strpos($dependency, '\\') === 0)
                {
                    $dependency = \substr($dependency, 1);
                }
```

PS. I had issues creating tests for these changes. Can somebody show example or create tests?
PPS. Would it be possible to get this feature into 7.5 or 7.6 etc..?

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -56,6 +56,10 @@ use Throwable;
 abstract class TestCase extends Assert implements Test, SelfDescribing
 {
     private const LOCALE_CATEGORIES = [\LC_ALL, \LC_COLLATE, \LC_CTYPE, \LC_MONETARY, \LC_NUMERIC, \LC_TIME];
+    private const DEPENDENCY_PREFIX = 'Dependency_';
+
+    /** @var TestResult \PHPUnit\Framework\TestResult */
+    private static $DEPENDENCY_TASK_RESULTS;
 
     /**
      * @var bool
@@ -381,6 +385,11 @@ abstract class TestCase extends Assert implements Test, SelfDescribing
     {
         if ($name !== null) {
             $this->setName($name);
+        }
+
+        if (!isset(self::$DEPENDENCY_TASK_RESULTS))
+        {
+            self::$DEPENDENCY_TASK_RESULTS = new TestResult;
         }
 
         $this->data     = $data;
@@ -1308,11 +1317,11 @@ abstract class TestCase extends Assert implements Test, SelfDescribing
     protected function createMock($originalClassName): MockObject
     {
         return $this->getMockBuilder($originalClassName)
-                    ->disableOriginalConstructor()
-                    ->disableOriginalClone()
-                    ->disableArgumentCloning()
-                    ->disallowMockingUnknownTypes()
-                    ->getMock();
+            ->disableOriginalConstructor()
+            ->disableOriginalClone()
+            ->disableArgumentCloning()
+            ->disallowMockingUnknownTypes()
+            ->getMock();
     }
 
     /**
@@ -1346,12 +1355,12 @@ abstract class TestCase extends Assert implements Test, SelfDescribing
     protected function createPartialMock($originalClassName, array $methods): MockObject
     {
         return $this->getMockBuilder($originalClassName)
-                    ->disableOriginalConstructor()
-                    ->disableOriginalClone()
-                    ->disableArgumentCloning()
-                    ->disallowMockingUnknownTypes()
-                    ->setMethods(empty($methods) ? null : $methods)
-                    ->getMock();
+            ->disableOriginalConstructor()
+            ->disableOriginalClone()
+            ->disableArgumentCloning()
+            ->disallowMockingUnknownTypes()
+            ->setMethods(empty($methods) ? null : $methods)
+            ->getMock();
     }
 
     /**
@@ -1363,9 +1372,9 @@ abstract class TestCase extends Assert implements Test, SelfDescribing
     protected function createTestProxy(string $originalClassName, array $constructorArguments = []): MockObject
     {
         return $this->getMockBuilder($originalClassName)
-                    ->setConstructorArgs($constructorArguments)
-                    ->enableProxyingToOriginalMethods()
-                    ->getMock();
+            ->setConstructorArgs($constructorArguments)
+            ->enableProxyingToOriginalMethods()
+            ->getMock();
     }
 
     /**
@@ -1456,12 +1465,12 @@ abstract class TestCase extends Assert implements Test, SelfDescribing
 
         if (!\class_exists($originalClassName)) {
             eval(
-                $this->getMockObjectGenerator()->generateClassFromWsdl(
-                    $wsdlFile,
-                    $originalClassName,
-                    $methods,
-                    $options
-                )
+            $this->getMockObjectGenerator()->generateClassFromWsdl(
+                $wsdlFile,
+                $originalClassName,
+                $methods,
+                $options
+            )
             );
         }
 
@@ -1682,51 +1691,83 @@ abstract class TestCase extends Assert implements Test, SelfDescribing
         }
     }
 
+    private function processFailed(array $failed): array
+    {
+        /** @var TestFailure $failure */
+        foreach ($failed as $failure) {
+            $pos = \strpos($failure->getTestName(), ' with data set');
+
+            if ($pos !== false) {
+                $failedKeys[] = \substr($failure->getTestName(), 0, $pos);
+            } else {
+                $failedKeys[] = $failure->getTestName();
+            }
+        }
+
+        return \array_flip(\array_unique($failedKeys ?? []));
+    }
+
+    private function processPassed(array $passedKeys): array
+    {
+        foreach (\array_keys($passedKeys) as $i => $iValue) {
+            $pos = \strpos($iValue, ' with data set');
+
+            if ($pos !== false) {
+                $passedKeys[$i] = \substr($iValue, 0, $pos);
+            }
+        }
+
+        return \array_flip(\array_unique($passedKeys));
+    }
+
+    private function processSkipped(array $skipped): array
+    {
+        /** @var TestFailure $skip */
+        foreach ($skipped as $skip) {
+            $pos = \strpos($skip->getTestName(), ' with data set');
+
+            if ($pos !== false) {
+                $skippedKeys[] = \substr($skip->getTestName(), 0, $pos);
+            } else {
+                $skippedKeys[] = $skip->getTestName();
+            }
+        }
+
+        return \array_flip(\array_unique($skippedKeys ?? []));
+    }
+
+    private function getSkippedFailedPassedKeys(&$skipped, &$failed, &$passedKeys, &$passed): void
+    {
+        $skipped    = \array_merge($this->processSkipped($this->result->skipped()), $this->processSkipped(self::$DEPENDENCY_TASK_RESULTS->skipped()));
+        $failed     = \array_merge($this->processFailed($this->result->failures()), $this->processFailed(self::$DEPENDENCY_TASK_RESULTS->failures()));
+        $passed     = \array_merge($this->result->passed(), self::$DEPENDENCY_TASK_RESULTS->passed());
+        $passedKeys = $this->processPassed($passed);
+    }
+
+    private function setSkippedDependsOn(string $dependsOn): void
+    {
+        $this->status = BaseTestRunner::STATUS_SKIPPED;
+
+        $this->result->startTest($this);
+
+        $this->result->addError(
+            $this,
+            new SkippedTestError(
+                \sprintf(
+                    'This test depends on "%s" to pass.',
+                    $dependsOn
+                )
+            ),
+            0
+        );
+        $this->result->endTest($this, 0);
+    }
+
     private function handleDependencies(): bool
     {
         if (!empty($this->dependencies) && !$this->inIsolation) {
-            $className      = \get_class($this);
-            $passed         = $this->result->passed();
-            $failed         = $this->result->failures();
-            $skipped        = $this->result->skipped();
-            $passedKeys     = \array_keys($passed);
-            $numKeys        = \count($passedKeys);
-
-            for ($i = 0; $i < $numKeys; $i++) {
-                $pos = \strpos($passedKeys[$i], ' with data set');
-
-                if ($pos !== false) {
-                    $passedKeys[$i] = \substr($passedKeys[$i], 0, $pos);
-                }
-            }
-
-            $failedKeys = [];
-
-            foreach ($failed as $failure) {
-                $pos = \strpos($failure->getTestName(), ' with data set');
-
-                if ($pos !== false) {
-                    $failedKeys[] = \substr($failure->getTestName(), 0, $pos);
-                } else {
-                    $failedKeys[] = $failure->getTestName();
-                }
-            }
-
-            $skippedKeys = [];
-
-            foreach ($skipped as $skip) {
-                $pos = \strpos($skip->getTestName(), ' with data set');
-
-                if ($pos !== false) {
-                    $skippedKeys[] = \substr($skip->getTestName(), 0, $pos);
-                } else {
-                    $skippedKeys[] = $skip->getTestName();
-                }
-            }
-
-            $passedKeys  = \array_flip(\array_unique($passedKeys));
-            $failedKeys  = \array_flip(\array_unique($failedKeys));
-            $skippedKeys = \array_flip(\array_unique($skippedKeys));
+            $className = \get_class($this);
+            $this->getSkippedFailedPassedKeys($skippedKeys, $failedKeys, $passedKeys, $passed);
 
             foreach ($this->dependencies as $dependency) {
                 $deepClone    = false;
@@ -1764,29 +1805,23 @@ abstract class TestCase extends Assert implements Test, SelfDescribing
                         && $className !== \explode('::', $dependency, 2)[0]
                     ) {
                         $dependencyClass = \explode('::', $dependency, 2)[0];
-                        (new TestSuite($dependencyClass, 'Dependency_' . \ucfirst($dependencyClass)))
-                            ->run($this->result);
+                        $dependencyKey = self::DEPENDENCY_PREFIX . \ucfirst($dependencyClass);
+                        $oldErrorsCount = self::$DEPENDENCY_TASK_RESULTS->errorCount();
+                        (new TestSuite($dependencyClass, $dependencyKey))->run(self::$DEPENDENCY_TASK_RESULTS);
+                        if ($oldErrorsCount < self::$DEPENDENCY_TASK_RESULTS->errorCount())
+                        {
+                            $this->setSkippedDependsOn($dependency);
+
+                            return false;
+                        }
                     } else {
-                        $this->status = BaseTestRunner::STATUS_SKIPPED;
-
-                        $this->result->startTest($this);
-
-                        $this->result->addError(
-                            $this,
-                            new SkippedTestError(
-                                \sprintf(
-                                    'This test depends on "%s" to pass.',
-                                    $dependency
-                                )
-                            ),
-                            0
-                        );
-                        $this->result->endTest($this, 0);
+                        $this->setSkippedDependsOn($dependency);
 
                         return false;
                     }
                 }
 
+                $this->getSkippedFailedPassedKeys($skippedKeys, $failedKeys, $passedKeys, $passed);
                 $dependencyName = false;
 
                 if (isset($passed[$dependency])) {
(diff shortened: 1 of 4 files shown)


epdenouden:
@Niko9911 Oh nice, more test execution order features :-) I will have a look for sure

Niko9911:
There is also issue when returning value from depended-upon test. It will always return `null` when using data sets. I'm currently working on this issue and will commit the fix into this PR.

Meanwhile, please approve ;)

Niko9911:
I think it would be best if this could be merged into 7.5?

src/Framework/TestCase.php
```diff
@@ -1698,7 +1700,33 @@ private function handleDependencies(): bool
                 }
             }
 
-            $passedKeys = \array_flip(\array_unique($passedKeys));
+            $failedKeys = [];
+
+            foreach ($failed as $failure) {
+                $pos = \strpos($failure->getTestName(), ' with data set');
```
epdenouden: The substring ` with [data set] ` is dependent on how test names are send to the loggers. This could change in the future, for example with #3439 

You can retrieve the `TestCase` via `$failure->failedTest()` and explore if it has a data set using `$test->dataName()`

src/Framework/TestCase.php
```diff
@@ -1724,31 +1752,54 @@ private function handleDependencies(): bool
...
-                    );
+                    if (
+                        !isset($skippedKeys[$dependency])
+                        && !isset($failedKeys[$dependency])
+                        && \strpos($dependency, '::') > 0
+                        && $className !== \explode('::', $dependency, 2)[0]
+                    ) {
+                        $dependencyClass = \explode('::', $dependency, 2)[0];
+                        (new TestSuite($dependencyClass, 'Dependency_' . \ucfirst($dependencyClass)))
+                            ->run($this->result);
```
epdenouden: 😨 This is scary bit of code. It fires off a lot of logic that will have unforeseen consequences, I'm just not sure which at the moment.

If I am reading this correctly, this code does the following:
1. if the `$dependency` has not been passed yet, do
2. check if `$dependency` hasn't been failed or skipped
3. get the name of the class in which the dependency resides by splitting the FQCN
4. yolo `(new TestSuite($theDependencyWeNeed))->run($currentResult)`

I get what you are trying to achieve: auto-run dependencies without any hassle on the user side. If this logic were to be accepted, it needs more safeguards en refactoring:
- what are the side effects of instantiating a `TestSuite` here in this scope?
- what happens if this `TestSuite` has already been instantiated, but: not run? or not run completely? failed during either `setUpBeforeClass` or `setUp`?
- even if this works, `handleDependencies()` shouldn't just `->run()` the missing dependency and merge it into the current result. Leave that responsibility to the Runner and find a cleaner way to pass around the execution plan.

Tests, tests and more tests. Preferably a few end-to-end ones :-)

src/Framework/TestCase.php
epdenouden: It is used like this elsewhere indeed! Which keeps giving me headaches as I am trying to refactor it out. 😆 For the test execution reordering and TestDox original-ordering, I cooked up this one:

https://github.com/sebastianbergmann/phpunit/blob/b8f38d3f22b839ac84c80fc09641de2d8973179f/src/Runner/TestSuiteSorter.php#L77-L94

The ` with data set ` substring is a user interface element that should be independent of the inner workings of the test runner. Here's an upcoming change that might interfere with your code: https://github.com/sebastianbergmann/phpunit/pull/3439/commits/06d8c4b9bff990f18c691e9d70d4240422a91dc8

src/Framework/TestCase.php
Niko9911: > This is scary bit of code. It fires off a lot of logic
I know... Kinda the way to deal with it, when I cannot access the list of tests to execute. Another solution would be scanning all files which depends on which one and reorder them. However, I tried it without success.

> If I am reading this correctly, this code does the following:
Yup.

> what are the side effects of instantiating a TestSuite here in this scope?
image

> what happens if this TestSuite has already been instantiated, but: not run? or not run completely? failed during either setUpBeforeClass or setUp
Good question. I will check what happens.

> Tests, tests and more tests. Preferably a few end-to-end ones :-)
Not sure how to implement those at all...

> even if this works, handleDependencies() shouldn't just ->run() the missing dependency and merge it into the current result. Leave that responsibility to the Runner and find a cleaner way to pass around the execution plan.
More specific implementation how to do it?
I could just run it like this and remove the result from wrong place like shown in the image?

src/Framework/TestCase.php
Niko9911: Okey, I'm still not sure how you're gonna replace that "with data set". You would need also refactor these
https://github.com/sebastianbergmann/phpunit/blob/b8f38d3f22b839ac84c80fc09641de2d8973179f/src/Framework/TestCase.php#L1685-L1780

https://github.com/sebastianbergmann/phpunit/blob/b8f38d3f22b839ac84c80fc09641de2d8973179f/src/Framework/TestCase.php#L1097-L1116

I'm still not sure how these should be refactored. 

Also I'm not sure how I can access dependencies of test inside of TestSuiteSorter.

epdenouden:
@Niko9911 I will reply more tomorrow, logging off for the day.

Niko9911:
@epdenouden NP. Thanks for helping. I will try resolve some dangerous bugs meanwhile when using stuff chained....

Niko9911:
Okey, I have done everything I could. I tested them with my project. Tests could be nice feature to add here, but I have no Idea how to create them....

Plz Review :smiley:

Niko9911:
Seems like 1021 test makes sure this will not fail... Ah crap, I'll fix it...

Niko9911:
Tests are now passing. :smirk:

epdenouden:
@Niko9911 Quick glance: what test has been added that tests the functionality you added? Did you add a test that failed first and now works? Changes in the core runner requires strong testing, especially to guard against edge cases.

epdenouden:
@Niko9911 Keep in mind that PHPUnit is a tool used by many different projects and environments. If it works in your project that's... a good start. :-)

By the way, which project is it and it is publicly available somewhere? I am still very curious what the actual use case is. Depending on "[...] a test that is declared in another test case class" (as @sebastianbergmann  says in#2647) makes me curious what other mechanism for sharing dependencies could solve your problem.

src/Framework/TestCase.php
epdenouden: There's enough to refactor for everyone! Don't be shy, just grab a few classes and have at it. 🛠 

You can get the name of a `TestCase` with:
- `$testCase->getName($withDataSet)` which gives you the name with(out) a prettified data set
- `$testCase->dataName()` returns the name of the data set you see in `getName()`
- `$testCase->getDependencies()` returns a list of dependencies, if any

Please explore the surrounding code a bit more and write tests _before_ you make such large changes.

Niko9911:
@epdenouden Sorry, I had to blur even the project names in previous images. It's private and under strict NDA. Would like share code, but as I think you understand, it's out of my hands if I can share it or not.

The idea is to check if x tests was success fully working, so test y can be tested. For example, if Domain tests fail, commandHandler or Infra tests depending on domain test will be marked skipped, until domain tests will pass.

src/Framework/TestCase.php
Niko9911: FYI, this issue showing in the image has been fixed..

src/Framework/TestCase.php
Niko9911: I understand and I would like to fix things. However I want remind about few facts.

I'm using heavily my own time to fix & learn this stuff. It's not as simple as it looks for guy who have no experience with PHPUnit core. I will look into this after I have created some tests.

Would need this feature ASAP into PHPUnit...

Niko9911:
@sebastianbergmann Could it be possible to implement this into 7.5.x?

Niko9911:
I kinda found another bug :zzz: if depending upon other class and it is not executed yet, it works, but if it's in same class as your test function, it will fail as the function is not executed yet. Reordering functions in class works. 

I'm gonna fix this also, but it's more dangerous... (High possibility for infinite loop)

epdenouden:
> It's private and under strict NDA. Would like share code, but as I think you understand, it's out of my hands if I can share it or not.

Then make sure you have a stand-alone and self-contained test for this scenario in your proposed code.

> The idea is to check if x tests was success fully working, so test y can be tested. For example, if Domain tests fail, commandHandler or Infra tests depending on domain test will be marked skipped, until domain tests will pass.

That makes sense. You are looking for what is basically the `SKIPIF` section of an end-to-end test. This is not an uncommon request and there are multiple ways of doing that to effectively create a custom `@requires` annotation, by way of logic in `setUp()`. A recent issue here that touches on this was #3427. Also, search for 'skip' in the issues list here, there are many "how to skip test using annotations" questions.

The current solution you are proposing doesn't just "skip if other tests haven't run successfully yet", it actively changes the state of the test runner while this is running. 😨 If you want a solution based on test reordering you will always end up with hacky solutions like this, unless you want to refactor more internals.

Niko9911:
> I kinda found another bug :zzz: if depending upon other class and it is not executed yet, it works, but if it's in same class as your test function, it will fail as the function is not executed yet. Reordering functions in class works.

This won't fix due the basic logic behind the PHPUnit. I will mark tests as skipped with following msg: `Reordering same class dependency function is not implemented.  Please reorder "%s" before "%s".`

> If you want a solution based on test reordering you will always end up with hacky solutions like this, unless you want to refactor more internals.

True, Thats why I have to propose hacky solution, but it is working. Also refactoring internals would mean basically refactoring ½ test framework, which would mean it's better to start from scratch and decide what copy from "old" framework to new.

> multiple ways of doing that to effectively create a custom @requires annotation, by way of logic in setUp().

But would make much more sense to implement this in level on TestFramework.

> actively changes the state of the test runner while this is running.

What you mean? Could you elaborate?

If this is not possible to implement (merge) into the PHPUnit, could it be possible to make TestCase expendable in this manner?

Niko9911:
@epdenouden 

> I kinda found another bug :zzz: if depending upon other class and it is not executed yet, it works, but if it's in same class as your test function, it will fail as the function is not executed yet. Reordering functions in class works.

This won't fix due the basic logic behind the PHPUnit. I will mark tests as skipped with following msg: `Reordering same class dependency function is not implemented.  Please reorder "%s" before "%s".`

> If you want a solution based on test reordering you will always end up with hacky solutions like this, unless you want to refactor more internals.

True, Thats why I have to propose hacky solution, but it is working. Also refactoring internals would mean basically refactoring ½ test framework, which would mean it's better to start from scratch and decide what copy from "old" framework to new.

> multiple ways of doing that to effectively create a custom @requires annotation, by way of logic in setUp().

But would make much more sense to implement this in level on TestFramework.

> actively changes the state of the test runner while this is running.

What you mean? Could you elaborate?

If this is not possible to implement (merge) into the PHPUnit, could it be possible to make TestCase expendable in this manner?

Niko9911:
Travis.... ComeON!

epdenouden:
Will reply more later, but for now: I don't agree with just shoving in a hacky solution without proper tests. It can be done properly. You'd have to read up on the functionality and limitations of test reordering. 

I disagree fundamentally that it makes sense to implement a quickfix in a generalized testing framework, instead of in an extended TestCase, setUp() or even a Packagist package that patches PHPUnit's files.

## The test runner core is like a watch or a clock. I do not understand, at all, how you can just propose creating and running a TestSuite there.

Niko9911:
> but for now: I don't agree with just shoving in a hacky solution without proper tests

I'm working on tests... I just cannot use those I used in our project. Hold on... Tests are coming...

> how you can just propose creating and running a TestSuite there in the middle of a dependency check.

To not refactor everything. If you agree with this feature, you could implement it better as you can see, my knowledge isn't enough to any deeper...

> I disagree fundamentally that it makes sense to implement a quickfix in a generalized testing framework, instead of in an extended TestCase, setUp() or even a Packagist package that patches PHPUnit's files.

How does one implement this into `setUp()`?
Patches also mean if anything changes in PHPUnit, the patch will fail.
How I see it, even if this is generalized testing framework and as long it doesn't affect (cause bugs) and is properly tested and providing new features, why not to implement it into the framework?

@sebastianbergmann I'm highly waiting for you opinion about this, if this is worth to spent time?

src/Framework/TestCase.php
epdenouden: I get it, no worries! Like many other tools and frameworks PHPUnit is maintained by volunteer contributors. We are paid in emoji and sleepless nights.

Niko9911: @epdenouden Tests are now Implemented :tada:

Niko9911: @epdenouden Btw, cannot really remove this `$pos = \strpos($failure->getTestName(), ' with data set');`

Because if taking look into `src/Framework/TestFailure.php` there will be

```php
/**
     * Constructs a TestFailure with the given test and exception.
     *
     * @param Throwable $t
     */
    public function __construct(Test $failedTest, $t)
    {
        if ($failedTest instanceof SelfDescribing) {
            $this->testName = $failedTest->toString();
        } else {
            $this->testName = \get_class($failedTest);
        }

        if (!$failedTest instanceof TestCase || !$failedTest->isInIsolation()) {
            $this->failedTest = $failedTest;
        }

        $this->thrownException = $t;
    }
```

Same with other stuff. Meaning this needs to be refactored and refactoring this means refactoring also skipped, passed, etc... Would change little bit too much logic to be in this PR.

Niko9911:
From my perspective this PR is ready to be merged.

@epdenouden 
@sebastianbergmann

src/Framework/TestCase.php
```diff
@@ -1724,31 +1793,68 @@ private function handleDependencies(): bool
...
+                    if (
+                        !isset($skippedKeys[$dependency])
+                        && !isset($failedKeys[$dependency])
+                        && \strpos($dependency, '::') > 0
+                    ) {
+                        if ($className !== \explode('::', $dependency, 2)[0]) {
+                            $dependencyClass = \explode('::', $dependency, 2)[0];
+                            $dependencyKey   = self::DEPENDENCY_PREFIX . \ucfirst($dependencyClass);
+                            $oldErrorsCount  = self::$DEPENDENCY_TASK_RESULTS->errorCount();
+                            (new TestSuite($dependencyClass, $dependencyKey))->run(self::$DEPENDENCY_TASK_RESULTS);
```
epdenouden: Before you put any more work into this, wait for a reply for @sebastianbergmann 

The design of this feature still worries me. Creating a new `TestSuite` for the missing dependenc(y|ies) and `run()` it _while running_ the main loop, is the least optimal way of solving this. If you want to run `@depends` before the dependant test, make sure the tests are in the optimal order before you start running.

- what protection is there against instantiating (and running) the same `TestSuite` twice?
- ...or more than twice? are results from the `@depends` test merged back into the 'main' `$testResult`
- how does all this work out when sent to the various loggers? Some of them, like JUNit, log the `TestSuite` hierarchy as the `start*/add*/end*` events are fired

src/Framework/TestCase.php
```diff
@@ -57,6 +57,11 @@
 {
     private const LOCALE_CATEGORIES = [\LC_ALL, \LC_COLLATE, \LC_CTYPE, \LC_MONETARY, \LC_NUMERIC, \LC_TIME];
 
+    private const DEPENDENCY_PREFIX = 'Dependency_';
+
+    /** @var TestResult \PHPUnit\Framework\TestResult */
+    private static $DEPENDENCY_TASK_RESULTS;
```
epdenouden: This data structure is only needed for the edge case of automatically running dependencies located in another class. Does this need to be a class-wide variable?
Also: why `static`, it belongs to an actual `TestCase` instance?

epdenouden:
@Niko9911 I can have a more detailed look this weekend. The current design of creating+running a new `TestSuite` for the missing dependency is a hack and should be handled before the run starts. It would take time to brainstorm and review possible consequences of the current approach. The collection of tests that comes with PHPUnit didn't have to guard against this in the past and it isn't surprising the main test suites went green so easily.

My vote would be for this change being an external script, like Symfony's PHPUnit bridge. Or implement it like the `TestSuiteSorter`: external to the main Runner.

(Context: after two rounds of clean up of the TestDox logger, I am still finding edge cases and yesterday a bug that throws a PHP error.)

src/Framework/TestCase.php
Niko9911: > what protection is there against instantiating (and running) the same TestSuite twice


It will be ran twice, if it is not in correct order. 

> ...or more than twice? are results from the @depends test merged back into the 'main'  $testResult

Not possible to run it over twice. The result will be put into dependency-only testResult in case it's required more than one time before it is ran in correct order.

> how does all this work out when sent to the various loggers? Some of them, like JUNit, log the TestSuite hierarchy as the start*/add*/end* events are fired

I'll test this. Should not affect that.

src/Framework/TestCase.php
Niko9911: It's static for optimization reasons. 

Meaning, if multiple functions require this same dependency, the results will be used from this class property instead of running dependency again and again, maybe even 100 times, before it will run the test.

src/Framework/TestCase.php
```diff
@@ -383,6 +388,10 @@ public function __construct($name = null, array $data = [], $dataName = '')
             $this->setName($name);
         }
 
+        if (!isset(self::$DEPENDENCY_TASK_RESULTS)) {
+            self::$DEPENDENCY_TASK_RESULTS = new TestResult;
```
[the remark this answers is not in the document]
Niko9911: It's required to initialize only once. Because it's static, it's required to initialize only once. This is required due optimization reasons I mentioned above.

tests/end-to-end/test-order-reversed-without-dependency-resolution.phpt
```diff
@@ -1,13 +1,14 @@
...
+phpunit --reverse-order --ignore-dependencies ../_files/MultiDependencyTest.php
 --FILE--
 <?php
 $_SERVER['argv'][1] = '--no-configuration';
 $_SERVER['argv'][2] = '--debug';
 $_SERVER['argv'][3] = '--verbose';
-$_SERVER['argv'][4] = '--order-by=no-depends,reverse';
-$_SERVER['argv'][5] = 'MultiDependencyTest';
-$_SERVER['argv'][6] = __DIR__ . '/../_files/MultiDependencyTest.php';
+$_SERVER['argv'][4] = '--reverse-order';
```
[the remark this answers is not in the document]
Niko9911: Thanks!

Will fix this.

Niko9911:
> PHPUnit didn't have to guard against this in the past and it isn't surprising the main test suites went green so easily.

There is new tests for this logic. As I use this now heavily with edge cases 7 days/week currently, I can quickly identify and fix these issues. I've been implementing as much tests as I can.

@sebastianbergmann Would really require your opinion about this! Is this possible to be merged into PHPUnit & What is your opinion?

Niko9911:
@sebastianbergmann @epdenouden Any news?

epdenouden:
Not from me, no

epdenouden:
@Niko9911 Not sure how busy @sebastianbergmann is at the moment, it's Xmas season and all that.

As for the cross-class `@depends` support: I have taken a look at the current mechanism of `TestSuiteSorter::resolveDependencies` and what changes would be required. It is possible, e.g. by hacking in a quick lookup list. The PHPUnit runner has no clue about the tree structure of the suites and tests. Without it anything will be a patched solution at best.

I will be adding full support for 'interdependent units of work' in the near future. However, a clean and maintainable solution is required there, as that refactoring relates to solving `@runClassInSeparateProcess` and related isolation infrastructure.

epdenouden:
@Niko9911 TL;DR: I agree inter-class `@depends` is a good idea, but I fear your proposed implementation will cause _more_ edge cases in the future.

Mind you, this isn't my personal project. I just enjoy working on it and being useful around the yard. Anyway: not sure what to tell you, as you have mentioned before you do not mind a hacky temporary solution. If that is what _you_ need right now and it solves your problems, my advice would be to make your change a private patch for PHPUnit and publish it on Packagist.

I mentioned the `TestSuiteSorter` because it turned out to be rather complex to fiddle with the test running order. There are just so many scenarios to keep in mind. Currently in the middle of refactoring the collection of end-to-end tests and all these are to cover the use cases we know of so far: https://github.com/epdenouden/phpunit/tree/improved-testdox/tests/end-to-end/execution-order

Your proposed change will run missing dependencies on the spot, while the existing executio reordering does so before the run. Also it doesn't force-run missing dependencies. I am not sure yours does even.

Anyway. I have a few other loose ends to tie up, I'll wait and see what @sebastianbergmann thinks.

sebastianbergmann:
Replaced by #3519.

epdenouden:
This functionality has been inplemented as a part of #3936

Issue #2726, opened 2017-07-04
TestSuite names conflict with names of built-in classes


jbaron-gingco:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 6.2.3
| PHP version         | 7.1.5
| OS | XUbuntu 17.04 (PHP is not the packaged version, but self-compiled)
| Installation Method | Composer

If a testsuite name in phpunit.xml is the name of a built-in class, an exception is thrown.

Minimal example:

**composer.json**

    {
      "require": {
        "php": "^7.0",
        "roave/security-advisories": "dev-master"
      },
      "require-dev": {
        "php": "^7.0",
        "phpunit/phpunit": "^6.2.2"
      }
    }

**phpunit.xml**

    <?xml version="1.0" ?>
    <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
             xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/5.2/phpunit.xsd">
        <testsuites>
            <testsuite name="DateTime"/>
        </testsuites>
    </phpunit>

Same issue occurs with test suite names like "Generator", "Closure" etc., but not with something like "get_class", "strlen" etc.

**Expected Result**

No tests executed!

**Actual Result**

(The backtrace was added in line 153 of `TestSuite.php` by myself)

>#0  PHPUnit\Framework\TestSuite->__construct() called at [/home/jost/projects/gingco_appserver/branches/dev_restructuring/InstallerBuilder/tmp/test/vendor/phpunit/phpunit/src/Util/Configuration.php:916]
#1  PHPUnit\Util\Configuration->getTestSuite() called at [/home/jost/projects/gingco_appserver/branches/dev_restructuring/InstallerBuilder/tmp/test/vendor/phpunit/phpunit/src/Util/Configuration.php:875]
#2  PHPUnit\Util\Configuration->getTestSuiteConfiguration() called at [/home/jost/projects/gingco_appserver/branches/dev_restructuring/InstallerBuilder/tmp/test/vendor/phpunit/phpunit/src/TextUI/Command.php:784]
#3  PHPUnit\TextUI\Command->handleArguments() called at [/home/jost/projects/gingco_appserver/branches/dev_restructuring/InstallerBuilder/tmp/test/vendor/phpunit/phpunit/src/TextUI/Command.php:152]
[...]

PR #3454, opened 2018-12-22
Enable TestSuite to use autoloader, clean up bootstrap

epdenouden:
While doing some housekeeping I found this note:

https://github.com/sebastianbergmann/phpunit/blob/c23d78776ad415d5506e0679723cb461d71f488f/tests/bootstrap.php#L19-L20

The root cause is that a new `TestSuite` will simply not look for classnames using the autoloader, causing tests trying `new TestSuite(\SomeClass:class)` to fail. A bit confusing, as it automagically works when the class had been autoloaded beforehand.

**Changes**
- change `TestSuite` to use the autoloader for its `\class_exists()`-check
- ability to create a `TestSuite` with the same name as an existing non-`Test` class
- clean up `tests/bootstrap.php`
- fix two corner cases in tests that popped up because of the autoloader change


epdenouden:
@sebastianbergmann Both issues were indeed related to how the `TestSuite` constructor looks and then instantiates the test classes.

Allowing the constructor to just create an empty `TestSuite` when the provided class(name) isn't a `Test` fixes #2726 without any negative side-effects. Once the `TestSuite` has been created with only a name and no contents, the `getName()` isn't used for potentially risky lookups anymore.

I have tried it via `phpunit.xml` by sticking a single test in a `<testsuite>` with the name of a builtin class and that works fine, see below. The `TestSuiteTest` has been updated to reflect the change.

```
./phpunit --list-suites

PHPUnit 7.5.1-3-g3bc689926 by Sebastian Bergmann and contributors.

Available test suite(s):
 - unit
 - end-to-end
 - DateTime
```

```
./phpunit --testsuite DateTime

PHPUnit 7.5.1-3-g3bc689926 by Sebastian Bergmann and contributors.

Runtime:       PHP 7.2.11 with Xdebug 2.6.0
Configuration: /Users/ewout/proj/phpunit-new-order/phpunit.xml

Time: 123 ms, Memory: 4.00MB

There was 1 risky test:

1) OneTestCase::testCase
This test did not perform any assertions

/Users/ewout/proj/phpunit-new-order/tests/_files/OneTestCase.php:18
[...]
```

I originally started working on this for 7.5.x and with the `TestSuite("anyNameNow")` it might be worth a 7.x maintainance update.

sebastianbergmann:
By "it might be worth a maintainance update" you mean this should go into `7.5` instead of `master`?

epdenouden:
Sorry, I could have articulated that better. I don't know how much work it is for you packaging and releasing 7.5.2. It would be nice to have the fix in version 7 as the release schedule says it will be around a while. Clicking the rebase buttons is the easy part. :-D

epdenouden:
@sebastianbergmann If you have time, can you also merge this into `master` already? I'm working on some more housekeeping in `tests/` and I'd handy to have the newer bootstrap :)

sebastianbergmann:
Done. Sorry for forgetting this.

epdenouden:
No worries, thank you for the quick response! In the past I've had some merge/rebase uhmmm confusion and I wanted to prevent any self-conflicts 😬

PR #3456, opened 2018-12-28
feat(xdebug-filter-generator): improve directories paths

fancyweb:
`xdebug_set_filter` uses absolute paths for its comparisons with the whitelisted ones.

That means that `/www/./myproject` is different than `/www/myproject`.

So right now if my phpunit config is like this :

```xml
<filter>
    <whitelist>
        <directory suffix=".php">./src/</directory>
    </whitelist>
</filter>
```

Then the generated whitelisted path is not working => the code coverage report is empty since this path does not match anything. It can be a source of mistakes.

Moreover, adding a trailing slash at the end of directories paths is an improvement to avoid to match every directories or file paths that only starts with the whitelisted paths.
For example with this structure :

```
- /foo
- /foobar
```

With the whitelisted path `/foo`, both directories will be matched. This is not the case if the whitelisted path is `/foo/`. It is a gain of speed.


sebastianbergmann:
CC @sebastianheuer

fancyweb:
Sure, with this phpunit configuration file : 
```xml
<phpunit backupGlobals="false"
         backupStaticAttributes="false"
         bootstrap="vendor/autoload.php"
         colors="true"
         processIsolation="false"
         stopOnFailure="false">

    <filter>
        <whitelist>
            <directory suffix=".php">./src</directory>
        </whitelist>
    </filter>

</phpunit>
```

Before my patch : 
```php
if (!\function_exists('xdebug_set_filter')) {
    return;
}

\xdebug_set_filter(
    \XDEBUG_FILTER_CODE_COVERAGE,
    \XDEBUG_PATH_WHITELIST,
    [
        '/Users/thomas.calvet/Fancyweb/debug/./src'
    ]
);
```

After my patch : 
```php
if (!\function_exists('xdebug_set_filter')) {
    return;
}

\xdebug_set_filter(
    \XDEBUG_FILTER_CODE_COVERAGE,
    \XDEBUG_PATH_WHITELIST,
    [
        '/Users/thomas.calvet/Fancyweb/debug/src/'
    ]
);
```

sebastianbergmann:
Thanks! Can you please make this against `7.5` (instead of `master`) as it is a bug fix?

PR #3476, opened 2019-01-14
CI: Use Ubuntu 16.04

carusogabriel:
Ubuntu 16.04 is already available for usage: https://docs.travis-ci.com/user/reference/xenial/#using-xenial.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: php
-
+dist: xenial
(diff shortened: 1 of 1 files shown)


sebastianbergmann:

Both Trusty and Xenial are too old for my liking, so it does not matter really which one we use. I would rather use the default and keep our `.travis.yml` minimal.

Issue #3419, opened 2018-11-27
AppVeyor


sebastianbergmann:
https://github.com/sebastianbergmann/phpunit/pull/3418#issuecomment-441767118 is not the first time that the AppVeyor configuration ran out of date (see #2880).

@helhum @IchHabRecht Does AppVeyor really need constant attention to keep working? If so, I would rather remove it (for which I would feel bad since you invested so much effort to make it initially work) unless somebody commits to keeping an eye on it.

epdenouden:
I've added a fix for this in https://github.com/sebastianbergmann/phpunit/pull/3421 to get the build working again.

sebastianbergmann:
Can you please send the AppVeyor changes in a separate pull request? Thanks! We'll also need this in the other active branches.

epdenouden:
Of course :) I already figured and opened #3423 for that reason

weltling:
@sebastianbergmann sure, i'll help where I can. @IchHabRecht, @epdenouden do i understand correctly, that you need to always fetch some latest releases? Could you please short articulate teh proposed solution? The integration between windows.php.net and php.net could be of course improved a bit, but in many cases time and security reasons are what stands in the way. Perhaps it can also be fixed an easier way, so please tell what exactly the issue is.

Thanks.

IchHabRecht:
hi @weltling,

Thanks for getting back to me. I created a gist with the content of my email. Please see https://gist.github.com/IchHabRecht/21902a244c5cd69de4e84275b276c73a

If you want I can post this anywhere else. Just let me know.

weltling:
@IchHabRecht i see, so you want the `latest` link. As it works till now, the latest versions are all to be found in https://windows.php.net/downloads/releases/. Once a new version is uploaded, the previous one is moved to https://windows.php.net/downloads/releases/archives/. I was just talking to @sAnexeh who takes care of the windows.php.net infrastructure and next week we're gonig to work on providing some kind of redirection to the latest zips.

Btw. another solution for you could be also to stick to some version and update it from time to time. I was solving it that way like here https://github.com/nikic/php-ast/blob/master/.appveyor.yml#L141. The version changes are always reflected in https://windows.php.net/downloads/releases/sha256sum.txt and a couple of other places. So yet another solution therefore were to parse the checksum file. It would be not much different from what is provided here http://php.net/releases/?json, also no `latest` links as far as i can tell.

[...]

PR #3478, opened 2019-01-14
Automatically use latest PHP minor version for AppVeyor

epdenouden:
Implement the suggestion by @weltling in https://github.com/sebastianbergmann/phpunit/issues/3419#issuecomment-453862347

No `*-latest.tzg` link for Xdebug yet, but this is one less thing to worry about. @IchHabRecht if you have ideas to improve this further :-)


epdenouden:
@sebastianbergmann I've implemented the `*-latest` package for PHP 7.2. Once Xdebug becomes 7.3.* compatible I'll update the CI.

sebastianbergmann:
Should this not go into older branches, too?

epdenouden:
You are right of course! I will make a clean pull that works in 7.x, too.

epdenouden:
Ah, I see the 7.5.x branch uses PHP 7.1.x still. It seems we don't need the _latest_ PHP+Xdebug for the v7 family or.. perhaps I am missing the point?

epdenouden:
@sebastianbergmann Don't think this will work for the 7.x branch as the `*-latest` redirect doesn't work for https://windows.php.net/downloads/releases/archives/php-7.1-Win32-VC14-x86-latest.zip

PR #3479, opened 2019-01-14
Additional PHPT code hint and color refinements

epdenouden:
While working on self-test collection and listener/printer refactoring I enjoy playing with little output enhancements. This is the latest (and last, for now) set of little changes that have made my development cycle a bit more smooth and enjoyable.

In one screenshot:

image

- add PHPT source file to the stack trace of assertions caused by PHPT EXPECT*_EXTERNAL sections
- when colored use additional UTF eye candy to give the left margin more visual texture. I found the 'railway map' to be really helpful when scrolling through my terminal windows.
  - clear start and end of the information block
  - exception message section (if available)
  - assertion value-diff section (if available)
  - exception stack trace (if available)
- ANSI color codes are a bit more efficient

image

### DIFF
--- a/src/Runner/PhptTestCase.php
+++ b/src/Runner/PhptTestCase.php
@@ -187,9 +187,14 @@ class PhptTestCase implements Test, SelfDescribing
                     }
 
                     $hint  = $this->getLocationHintFromDiff($diff, $sections);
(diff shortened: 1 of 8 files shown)


epdenouden:
Hello line endings, my old friend

Issue #3482, opened 2019-01-15
Dumping Xdebug filter script requires resolving relative paths when phpunit.xml is not in root directory


localheinz:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 7.5.1
| PHP version         | 7.1.24
| Installation Method | Composer

### `test/Unit/phpunit.xml`

```xml
<phpunit
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
    xsi:noNamespaceSchemaLocation="../../vendor/phpunit/phpunit/phpunit.xsd"
    bootstrap="../../vendor/autoload.php"
>
    <testsuites>
        <testsuite name="Unit Tests">
            <directory>.</directory>
        </testsuite>
    </testsuites>
    <filter>
        <whitelist addUncoveredFilesFromWhitelist="true" processUncoveredFilesFromWhitelist="true">
            <directory suffix=".php">../../src</directory>
        </whitelist>
    </filter>
</phpunit>
```

### `Makefile`

```Make
.PHONY: coverage

coverage: vendor
	vendor/bin/phpunit --configuration=test/Unit/phpunit.xml --dump-xdebug-filter .build/phpunit/xdebug-filter.php
	vendor/bin/phpunit --configuration=test/Unit/phpunit.xml --coverage-text --prepend .build/phpunit/xdebug-filter.php

vendor: composer.json composer.lock
	composer validate
	composer install
```

When I run

```
$ make coverage
```

I end up with

```
vendor/bin/phpunit --configuration=test/Unit/phpunit.xml --dump-xdebug-filter .build/phpunit/xdebug-filter.php
PHPUnit 7.5.1 by Sebastian Bergmann and contributors.

Runtime:       PHP 7.1.25 with Xdebug 2.6.1
Random seed:   1547538628
Configuration: /Users/am/Sites/localheinz/json-normalizer/test/Unit/phpunit.xml

Wrote Xdebug filter script to .build/phpunit/xdebug-filter.php
vendor/bin/phpunit --configuration=test/Unit/phpunit.xml --coverage-text --prepend .build/phpunit/xdebug-filter.php
PHPUnit 7.5.1 by Sebastian Bergmann and contributors.

Runtime:       PHP 7.1.25 with Xdebug 2.6.1
Random seed:   1547538629
Configuration: /Users/am/Sites/localheinz/json-normalizer/test/Unit/phpunit.xml

[...]

PR #3483, opened 2019-01-15
Fix: Apply realpath() to items in Xdebug filter list

localheinz:
This PR

* [x] applies `realpath()` to items in Xdebug filter list

Fixes https://github.com/sebastianbergmann/phpunit/issues/3482.


localheinz:
Appears to be already fixed?

PR #3487, opened 2019-01-18
[donotmerge] Prevent tearDownAfterClass exceptions from killing PHPUnit

epdenouden:
New version with a fixed rebase: https://github.com/sebastianbergmann/phpunit/pull/3490


epdenouden:
Code coverage will be brought back by the refreshed tests I'm making for the basic end-to-end suite.

epdenouden:
@sebastianbergmann I just discovered there is already a test for this exact case, which is run and green for `TestCast` but not `TestSuite`: https://github.com/sebastianbergmann/phpunit/blob/master/tests/_files/ExceptionInTearDownTest.php tested in:

https://github.com/sebastianbergmann/phpunit/blob/e8850c1dce9250e7b430a5d666feb7f71a5d658b/tests/unit/Framework/TestCaseTest.php#L155-L165

I'll fix this nicely and update this pull request. 🚜

tests/unit/Framework/TestSuiteTest.php
```diff
@@ -227,4 +227,24 @@ public function testCreateTestForConstructorlessTestClass(): void
...
+     */
+    public function testTearDownAfterClassInTestSuite(): void
+    {
+        $suite = new TestSuite(\ExceptionInTearDownAfterClassTest::class);
+        $suite->run($this->result);
+
+        $this->assertSame(3, $this->result->count());
+        $this->assertCount(1, $this->result->failures());
+
+        /** @var TestFailure $failure */
```
sebastianbergmann: Please use `\assert($failure instanceof TestFailure);` instead.

tests/_files/ExceptionInTearDownAfterClassTest.php
```diff
@@ -0,0 +1,62 @@
...
+ * (c) Sebastian Bergmann <sebastian@phpunit.de>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+use PHPUnit\Framework\TestCase;
+
+class ExceptionInTearDownAfterClassTest extends TestCase
+{
+    public $setUp                = false;
```
sebastianbergmann: Is there a specific reason why these attributes are `public`?

tests/unit/Framework/TestSuiteTest.php
sebastianbergmann: As of b880c49bfec871a43702bca8c42b36ccc246f989 neither should be needed anymore.

tests/_files/ExceptionInTearDownAfterClassTest.php
epdenouden: No anymore no, thanks for catching this. The test started off as a version of `tests/_files/ExceptionInTearDownTest.php`. I have cleaned up test to the minimum required.

tests/unit/Framework/TestSuiteTest.php
epdenouden: Will that commit also come to 7.5? I have added the `\assert()` for now.

epdenouden:
I'm going to need to figure out a better way of juggling multiple working versions (7 vs 8) and branches of stuff I'm working on. Don't even let me command a nuclear sub. ONE REBASE ONLY.

PR #3488, opened 2019-01-19
Try support diff 3

dereuromark:
Many other internal tools now require v3, and as such phpunit 6 should, if possible, also allow this version.


sebastianbergmann:
A change like this will not happen in the last days of support for PHPUnit 6.

PR #3497, opened 2019-01-22
Add output buffering spinner to colored testdox printer

epdenouden:
The [TestDox printer will buffer test result output](https://github.com/sebastianbergmann/phpunit/pull/3417) when the execution has been changed by `--order-by`. When this happens the prettified printer will let the user know by showing a little spinner:

image

Quick way to see it at work, run:  ```phpunit --testdox --order-by=random --testsuite end-to-end```

To prevent visual noise the spinner will only appear when out of sequence results are detected by the colorized printer and hidden when it has caught up. No further configuration or changes needed.

That's it. :)

### DIFF
--- a/src/Util/TestDox/CliTestDoxPrinter.php
+++ b/src/Util/TestDox/CliTestDoxPrinter.php
@@ -46,6 +46,13 @@ class CliTestDoxPrinter extends TestDoxPrinter
         'last'    => '┴',
     ];
 
+    private const SPINNER_ICONS = [
+        " \e[36m◐\e[0m running tests",
+        " \e[36m◓\e[0m running tests",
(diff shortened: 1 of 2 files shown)


hopeseekr:
I really dislike this spinner and `testdox` is my main mode.

epdenouden:
> I really dislike this spinner and `testdox` is my main mode.

Hi there @hopeseekr! You are not the only one, that's [why an option to switch](https://github.com/sebastianbergmann/phpunit/pull/3528) it off was added.

Try this:
```
./phpunit --testdox --no-interaction --order-by=reverse
```
and you will see nothing but the version header until output can be flushed in the correct original order.

PR #3508, opened 2019-02-03
Fix FileLoader.php fatal error when loading a non existing file

mkveksas:


sebastianbergmann:
Pull requests for bug fixes should be made to the oldest branch that is supported and has the problem you are trying to fix.

That being said, can you share an example that triggers the problem you are trying to fix? Thanks!

mkveksas:
No worries, hope to find and fix more bugs in the future!

PR #3527, opened 2019-02-13
Run a single test by filename and line number

davedevelopment:
I would find this very useful, I currently use vimscript and regexp to look back to try and find the nearest method name, which I then have to mangle somewhat in to a regular expression to use with `--filter`. I'm sure other editors and IDEs might be able to benefit from this as well.


sebastianbergmann:

davedevelopment:
@sebastianbergmann fair enough, if you get two minutes to let me know why, I'd appreciate it, cheers.

Issue #3513, opened 2019-02-05
Testdox spinner cannot be disabled (and can be too verbose)


BafS:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 8.0.1
| PHP version         | 7.2
| Installation Method | Composer

Since PHPUnit 8 I have pages of "running tests ◓" (introduced by https://github.com/sebastianbergmann/phpunit/pull/3497) with the testdox option on my CI (it does support ANSI codes but it cannot clear all output). Could we have an option to disable the spinner (eg. `--no-progress`) or to detect if the output support it ? I want to keep the colors.

The current workaround is to extends CliTestDoxPrinter.php.

![image](https://user-images.githubusercontent.com/588205/52277819-e2783b00-2955-11e9-95d1-ccaae35d040d.png)

Thanks in advance

epdenouden:
@BafS Of course! The spinner should never get in your way. I will gladly make it more optional, I will start working on a config option and get a pull request tomorrow.

Question: what type of terminal/`$TERM` does this happen in and what CLI-options or `phpunit.xml` flags do you use? I'm guessing the terminal supports color but not all rendering or interactive features.

Thank you for taking the time to file an issue about this, by the way. Good to have it fixed.

BafS:
@epdenouden Thank you for your message ! The "terminal" the default gitlab web terminal.

```sh
[root@runner-5dda6155-project-56-concurrent-0 /]# echo $TERM;
xterm
```

The "problem" is just that it outputs the stream directly without clearing anything. It's not an interactive terminal but more an STDOUT stream.

I simply use `--testdox --no-cover` and the xml options are

```ini
 backupGlobals="false"
 backupStaticAttributes="false"
 verbose="true"
 colors="true"
 convertErrorsToExceptions="true"
 convertNoticesToExceptions="true"
 convertWarningsToExceptions="true"
 processIsolation="false"
 stopOnFailure="false"
```

Thanks for your help

PR #3528, opened 2019-02-13
Implement option to disable Testdox progress animation

epdenouden:
Implements #3513 allowing the user to disable the Testdox progress animation.

#### Changes ####
- add `--no-interaction` command line option
- add `noInteraction` attribute for XML-configuration files
- update configuration unit tests and CLI option tests

Unfortunately I did not manage to get a screenshot.

### DIFF
--- a/phpunit.xsd
+++ b/phpunit.xsd
@@ -281,7 +281,7 @@
         <xs:attribute name="extensionsDirectory" type="xs:string"/>
         <xs:attribute name="executionOrder" type="executionOrderType" default="default"/>
         <xs:attribute name="resolveDependencies" type="xs:boolean" default="false"/>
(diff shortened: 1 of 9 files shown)


sebastianbergmann:
Thanks!

I am not sure I like the `--no-progress` and `noProgress` names. They can be misunderstood as "no progress output at all".

epdenouden:
Agreed, the dot-printer also shows progress without animation. It seems --no-progress is relatively common as a GNU-style --arg

I'll change it to '--no-spinner' as I don't think there'll be any other new animations

sebastianbergmann:
Sounds reasonable. Thanks!

epdenouden:
While changing the name I just noticed this that composer has a `--no-interaction` flag which seems perfect and is already used by another large PHP project. Even better I think :)

PR #3534, opened 2019-02-16
Fixed wrong message in ConstraintTestCase

rdohms:
The SelfDescribing test was reporting lack of implementation of the
\Countable class not the SelfDescribing class. While not a problem, it
would be at least misleading to contributors seeing the error.

Since this is a TestCase I could not find any tests to update.


localheinz (approved):
👍

sebastianbergmann:
Merged manually, thanks.

PR #3540, opened 2019-02-22
Make resolveDependencies=true the default for phpunit.xml

### LINKED ISSUE #3539: Update resolveDependencies default value in phpunit.xsd
| Q                   | A
| PHPUnit version     | ^8.0
| PHP version         | *
| Installation Method | *

Since resolving dependencies is now the default this should be true now too.

https://github.com/sebastianbergmann/phpunit/blob/b62028f7edcbb6503f4d28382c126df8b5d89325/phpunit.xsd#L283

epdenouden:
Fixes #3539


sebastianbergmann:
This needs to go into `8.0`.

epdenouden:
I'd better make a new branch, changing the base doesn't just stick the one commit on top unfortunately, brb

Issue #3453, opened 2018-12-19
À tout de suite: fast self-test suite


epdenouden:
As a _developer of PHPUnit_
I want _a fast integration test of all basic use cases_
so that _experimenting with the new logging is easier and more fun_

### Inspiration ###
Related to the [redesign of the result event and logging system](#3390) and the updates for [test execution order logging](#3380) and [TestDox](#3439).

Working on the test execution ordering I wished I had a quick ~test suite~ collection of tests that shows me the general state of PHPUnit with a focus on the developer experience. The current test collection is too unwieldy to run on my old laptop continuously.

### What functionality to test? ###
- are all test result status and messages presented to the user; example: #3436
- do exceptions and errors bubble to the surface; example: #3364
- running order of test collection within current architecture; #3147 and all its related issues like #3093, #3156
- buffered logging of async test events to decouple execution order and result printer; example: #3380 
- output format of the loggers included with the basic package

### Implementation ###
- create a small test suite with the basics as a part of the main test collection, but which can also be reused for individual end-to-end tests
- reuse existing tests where possible
- review existing end-to-end tests: remove duplicates, obsolete regression tests
- make it easier to test loggers via unit tests

#### Basics ####
- [x] all test result types with standard messages
- [x] all test result types with verbose messages

#### Fixtures ####
- [x] TestCase class with a failing `setUpBeforeClass()`
- [x] TestCase class with a failing `tearDownAfterClass()`
- [x] TestCase class with a `setUp()` that fails always
- [ ] TestCase class with a `tearDown()` that fails

#### Dataproviders ####
- [ ] Test with `@dataprovider` and a provider that returns a simple array
- [ ] Test with `@dataprovider` and a provider that returns named rows
- [ ] Test with an incompatible `@dataprovider`
- [ ] Test with a `@dataprovider` that returns no data
- [ ] Test with a Generator `@dataprovider`
- [ ] Test with a Generator `@dataprovider` with a duplicate key
- [ ] Test with a malformed `@dataprovider` row
- [ ] skipped dataprovider; see `DataProviderSkippedTest`

#### Test execution order ####
- [ ] Test that `@depends` on a previous test (original: no output)
- [ ] Test that `@depends` on a 'later' test (original: skipped, when no-depends)
- [ ] Test that `@depends` on a Test in another class
- [ ] unbuffered logging: reordering affects both run order and output order
- [ ] buffered logging: reordering affects only run order and result, but not output order
[...]

PR #3545, opened 2019-02-27
Housekeeping: improve tearDownAfterClass handling

epdenouden:
While working on refactoring the [basic test collection](#3453) I bumped into some rough edges working with fixtures. Exceptions during `TestSuite` tear down are now logged, reusing the last test to keep output readable:

image


epdenouden:
It's line endings, will fix later today 🔬

sebastianbergmann:
Odd that the tests fail on PHP < 7.4.

epdenouden:
> Odd that the tests fail on PHP < 7.4.

It is a strange failure and it does have to do with the mixed use of `\n` and `PHP_EOL` around exception message handling. I noticed it first when colorizing the testdox printer. A risky test with custom message would also have an extra empty line that I never figured out back then. I will find it :)

sebastianbergmann:
Yay!, it's green.

epdenouden:
> Yay!, it's green.

Oh wow. I'm so proud! haha

Anyway: the line endings used all over the code turn out to be a mix of `\n` and `PHP_EOL`. The only way to get the output stable is to manually filter every line in a few places to make sure nothing slips through.

Issue #3370, opened 2018-10-24
Remove assertInternalType() and assertNotInternalType()


sebastianbergmann:
See https://github.com/sebastianbergmann/phpunit/issues/3368 for migration path.

PR #3548, opened 2019-03-02
InvokedAtIndex matcher counting fix

Firtzberg:
The InvokedAtIndex matcher seems to count all method invocations, rather than the specified method only.
@sebastianbergmann Please read the implemented test case to make sure my idea of how this should work is valid.
After I apply my fix the test passes and the other tests remain unchanged.
Basically I just moved the method name check before the invocationMatcher check.

### DIFF
--- a/src/Framework/MockObject/Matcher.php
+++ b/src/Framework/MockObject/Matcher.php
@@ -211,10 +211,6 @@ final class Matcher implements MatcherInvocation
             throw new RuntimeException('No method matcher is set');
         }
 
-        if (!$this->invocationMatcher->matches($invocation)) {
-            return false;
-        }
-
         try {
             if (!$this->methodNameMatcher->matches($invocation)) {
                 return false;
@@ -231,7 +227,7 @@ final class Matcher implements MatcherInvocation
             );
         }
 
-        return true;
+        return $this->invocationMatcher->matches($invocation);
     }
 
     /**
--- a/tests/README.md
+++ b/tests/README.md
@@ -29,4 +29,4 @@ Note: this section will change often while `tests/` is being refactored.
   - `basic/`: fast tests covering all basics
     - `configuration.basic.xml`: configuration file tailored for the `basic` suite
   - `end-to-end/`: run PHPUnit as a seperate process and observe its behaviour via console messages and the filesystem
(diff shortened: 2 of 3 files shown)


Firtzberg:
Please notice that this is a breaking change, which is why I made a PR on the master and not on the older supported version. If this gets approved, please include details in the change log for the next version.

sebastianbergmann:
It is a known issue (maybe we need to document it better) that the `InvokedAtIndex` matcher only has one counter that gets incremented for each call to a stubbed method during a test.

You are right when you say that this is a breaking change and that it has to wait for the next major version. The next major version will be PHPUnit 9. The development for PHPUnit 9 will start in December 2019 after the release of PHPUnit 8.5. This cannot be merged before then.

That being said, and while I agree that the current behaviour is wrong (or at least confusing), I would rather not have a breaking change like this. It is completely different from a breaking change such as #3370: after `assertInternalType()` is removed there will be an error when a test uses it. The change you propose does not change API but semantics. The tests will continue to work without an error but there will be failures and we will be unable to detect in PHPUnit's code whether this failure is because of this change or not.

If we want to make this change then we would need to
* introduce a new method as an alternative to `at()` that counts the "right way" in PHPUnit 8.x
* [soft-deprecate](https://github.com/sebastianbergmann/phpunit/wiki/Development-and-Release-Process#stage-1-soft-deprecation) `at()` in PHPUnit 8.x and recommend using the new method instead
* [deprecate](https://github.com/sebastianbergmann/phpunit/wiki/Development-and-Release-Process#stage-2-deprecation) `at()` in PHPUnit 9
* [remove](https://github.com/sebastianbergmann/phpunit/wiki/Development-and-Release-Process#stage-3-removal) `at()` in PHPUnit 10

Firtzberg:
I completely agree.
Please take over from here.

sebastianbergmann:
This branch has conflicts that must be resolved, sorry for dropping the ball on this.

Can you please send a new pull request that implements a new method as an alternative to `at()` as discussed in https://github.com/sebastianbergmann/phpunit/pull/3548#issuecomment-468996144? I will take care of the rest then. Thanks!

Firtzberg:
Can you suggest a name?

sebastianbergmann:
`atIndex()` maybe?

Issue #3519, opened 2019-02-07
Using @depends to depend on classes


sebastianbergmann:
As of today, the `@depends` annotation cannot be used to express that a test (or all tests of a test case class) depend on all the tests of another test case class to pass. This fact is not clear in the documentation (https://github.com/sebastianbergmann/phpunit-documentation-english/issues/121 exists to remedy this). It is also not clear from the error output you get if you try to use `@depend TestClassName`. Consider the two test case classes shown below:

```php
<?php declare(strict_types=1);
use PHPUnit\Framework\TestCase;

/**
 * @depends XyzTest
 */
final class AbcTest extends TestCase
{
    public function testOne(): void
    {
        $this->assertTrue(true);
    }
}
```
```php
<?php declare(strict_types=1);
use PHPUnit\Framework\TestCase;

final class XyzTest extends TestCase
{
    public function testOne(): void
    {
        $this->assertTrue(true);
    }
}
```
Executing the tests shown above results in the confusing output shown below:
```
$ ./phpunit
PHPUnit 8.0.1 by Sebastian Bergmann and contributors.

Runtime:       PHP 7.3.2 with Xdebug 2.7.0rc1
Configuration: /home/sb/example/phpunit.xml

S.                                                                  2 / 2 (100%)

Time: 35 ms, Memory: 4.00MB

There was 1 skipped test:

1) AbcTest::testOne
This test depends on "AbcTest::XyzTest" to pass.

OK, but incomplete, skipped, or risky tests!
Tests: 2, Assertions: 1, Skipped: 1.
```
Once #3517 is implemented this error output would be less confusing. But it would still not be clear that using `@depends` to depend on all tests of a test case class to pass is not possible.

[In the past I said that](https://github.com/sebastianbergmann/phpunit/issues/2782#issuecomment-336390765) "*[i]t does not make sense to me for a test to depend on all tests of a test case class*". In a recent discussion, @spriebsch changed my mind. Especially in the context of integration tests, or larger tests in general, this makes a lot of sense.

What does not make sense, in my opinion, is to pass the results of all tests of a test case class to the test(s) that depend(s) on that test case class.

sebastianbergmann:
@epdenouden Would be great to get your input on this.

spriebsch:
[...]

PR #3549, opened 2019-03-05
Implement using @depends for depending on class

epdenouden:
Exploring #3519. 

### Use case ###
_As a_ developer of a larger collection of tests
_I want to_ express dependencies on whole `TestCase` classes
_so that_ I don't waste time on maintaining lists of `@depends`

### Example ###
https://github.com/sebastianbergmann/phpunit/blob/131c59e237f1b4cd15c3513ab4a5a919537ac3f6/tests/_files/DependencyOnClassTest.php#L18-L25

### Changes ###
- [x] keep track of passing test suites in `TestResult`, seperate from passing tests
- [x] handle `@depends SomeTest::class` annotation skip/warn in `TestCase` dependency handler
- [ ] support `@depends SomeTest::class` annotation in execution reordering

ping @spriebsch

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -1737,83 +1737,98 @@ abstract class TestCase extends Assert implements Test, SelfDescribing
      */
     private function handleDependencies(): bool
     {
-        if (!empty($this->dependencies) && !$this->inIsolation) {
-            $className  = \get_class($this);
-            $passed     = $this->result->passed();
-            $passedKeys = \array_keys($passed);
+        if (empty($this->dependencies) || $this->inIsolation) {
+            return true;
+        }
 
-            foreach ($passedKeys as $key => $value) {
-                $pos = \strpos($value, ' with data set');
+        $className  = \get_class($this);
+        $passed     = $this->result->passed();
+        $passedKeys = \array_keys($passed);
 
-                if ($pos !== false) {
-                    $passedKeys[$key] = \substr($value, 0, $pos);
-                }
+
+        foreach ($passedKeys as $key => $value) {
+            $pos = \strpos($value, ' with data set');
+
+            if ($pos !== false) {
+                $passedKeys[$key] = \substr($value, 0, $pos);
             }
+        }
 
-            $passedKeys = \array_flip(\array_unique($passedKeys));
+        $passedKeys = \array_flip(\array_unique($passedKeys));
 
-            foreach ($this->dependencies as $dependency) {
-                $deepClone    = false;
-                $shallowClone = false;
+        foreach ($this->dependencies as $dependency) {
+            if (\substr($dependency, -7) === '::class') {
+                $dependencyClassName = \substr($dependency, 0, -7);
 
-                if (\strpos($dependency, 'clone ') === 0) {
-                    $deepClone  = true;
-                    $dependency = \substr($dependency, \strlen('clone '));
-                } elseif (\strpos($dependency, '!clone ') === 0) {
-                    $deepClone  = false;
-                    $dependency = \substr($dependency, \strlen('!clone '));
-                }
-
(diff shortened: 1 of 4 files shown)


sebastianbergmann:
@epdenouden Are you sure this should go into `7.5`? I would consider this a new feature and not a bugfix, to be honest.

spriebsch:
I think I like what I see (since you pinged me). @sebastianbergmann, what would be the problem with adding a new feature in a minor version (as long as it does not break BC)?

sebastianbergmann:
I have no problem with accepting this into PHPUnit 8.1.

epdenouden:
> @epdenouden Are you sure this should go into `7.5`? I would consider this a new feature and not a bugfix, to be honest.

@sebastianbergmann I started exploring the code in `master` and got it working rather quickly. Then realized you might have wanted it in 7, too, so I created a new branch for this. Personally I do not mind either way.

My assumptions when creating the `7.5`-based branch:
- safe: version 7.5 has all infrastructure needed
- safe: no BC breaks or changing test results, as the use-case simply wasn't available before
- it is aimed at larger test collections which are more likely to run not-the-latest version
- support for this was already in my earliest sketches for the reordering features

@spriebsch Ah, I thought it was you who had brought this up in #3519! Sorry if I poked you for something unrelated, hope you like your present anyway 🎁

sebastianbergmann:
It is new functionality. PHPUnit 7.5 does not get new functionality anymore.

epdenouden:
> It is new functionality. PHPUnit 7.5 does not get new functionality anymore.

Ok thanks for the clarification. I will target 8.x and implement an improved dependency handler that understands dependencies between suites. Will be done after the weekend.

epdenouden:
> It is new functionality. PHPUnit 7.5 does not get new functionality anymore.

Okidoki, I'll target v8  🏗 

This will take a bit more time to implement properly. I've had a good look at the `TestSuiteSorter` logic and found out again why I didn't manage to get `@depends` working for all of the following scenarios:
1. **already supported**: determine optimal running order for tests in a `TestCase <converted to> TestSuite`
2. **already supported**: determine optimal running order for `@dataprovider` rows as individual tests in a `TestCase <converted to> DataproviderTestSuite`
3. **already supported**: understand result of a data-driven test as a result in context of a suite
4. determine optimal running order for test classes when there are interdependencies

The current logic was designed to be a simple as I didn't understand the technical details of the various scenarios last time around. Now that I do I will upgrade the logic to support inter-class dependencies as a next step.

epdenouden:
Note to self: check interaction between `--filter` options and test execution reordering. The sorter logic should only trigger on `@depends` and cached results related to tests being run. My suspicion is it currently takes all defined tests into account.

PatricNox:
This already exists, with capital D 

@Depends Class::Method

epdenouden:
hi @PatricNox! What is going to be build is allowing for `@depends SomeClass::class` which will force a check on all tests run for that class.

> This already exists, with capital D
> @Depends Class::Method

I cannot find this in the PHPUnit codebase; is it an optional package that provides this?

epdenouden:
Superseded by much improved #3936

PatricNox:
@epdenouden Not sure if origins from its own package, but my belief it was an variant made in phpunit, maybe I was wrong?

https://stackoverflow.com/a/54626398/9476349 (feb 2019)
https://github.com/sebastianbergmann/phpunit/pull/3936 (nov 2019 - merged functionality)

epdenouden:
@PatricNox I am not sure I understand your question? Yes, there have been experiments for this functionality previously.

PatricNox:
> @PatricNox I am not sure I understand your question? Yes, there have been experiments for this functionality previously.

Not a question, but an answer to _your_ question.

> I cannot find this in the PHPUnit codebase; is it an optional package that provides this

PR #3555, opened 2019-03-11
Use TestHook Interface consistently

keulinho:
Remove Hook interface and use TestHook for it
When using the AfterTestHook you had to explicitly implement also the TestHookInterface because of the 
`assert` check in TestRunner.php:1085

### DIFF
--- a/src/TextUI/TestRunner.php
+++ b/src/TextUI/TestRunner.php
@@ -807,7 +807,7 @@ final class TestRunner extends BaseTestRunner
         return $this->loader;
     }
 
-    public function addExtension(TestHook $extension): void
+    public function addExtension(Hook $extension): void
     {
         $this->extensions[] = $extension;
     }
@@ -1083,7 +1083,7 @@ final class TestRunner extends BaseTestRunner
                     );
                 }
 
-                \assert($extensionObject instanceof TestHook);
+                \assert($extensionObject instanceof Hook);
 
                 $this->addExtension($extensionObject);
             }


sebastianbergmann:
Can you elaborate on which problem this is supposed to solve and whether the proposed chance has BC implications? Thanks!

keulinho:
In TestRunner.php:1085 is a check that the extension is an instance of `TestHook`, so it is currently not possible to add Hooks as extensions that just implement the `Hook` Interface.
The problem occurs when you want to add e.g. a `AfterTestHook` that previously just extended the `Hook` interface.
Basically:
```
class MyHook implements AfterTestHook
```
does not work currently, because `MyHook` only implements the `Hook` Interface.
A workaround is following:
```
class MyHook implements TestHook, AfterTestHook
```

But as it is currently not possible to add hust a hook as a extension IMHO there is no need for that extra interface and it leads to consistency problems like this.

Practically there should not be a BC because `Hook` and `TestHook` are both just empty interfaces and as stated above it is not possible to write custom hooks that just implement the `Hook` interface and add it as a extension.
Theoretically it is a BC because we deleted the `Hook` interface, a non BC way would be to deprecate the `Hook` interface and let it extend the `TestHook` interface rather than vice versa.

sebastianbergmann:
With "there is no need for that extra interface" you miss the whole point of having `Hook` and `TestHook`. This is probably my fault as I, obviously, did not document it (well enough).

An object that wants to hook into PHPUnit's test runner needs to implement the `Hook` interface.

The `TestHook` interface is used to "group" hooks such as `BeforeTestHook` and `AfterTestHook` that are related to a single test that is being executed. The case can probably be made for `TestHook` being superfluous but I quite like the distinction of `Hook` and the more specific `TestHook` supertypes.

`BeforeFirstTestHook` and `AfterLastTestHook` are examples of `Hook` that are not a `TestHook`, meaning they are not related to a single test being executed.

keulinho:
Ok but what is then the point of that assert Statement `\assert($extensionObject instanceof TestHook);`
in TestRunner.php:1085?
Doesn't it have to be `\assert($extensionObject instanceof Hook);` instead? And the 
`public function addExtension(TestHook $extension): void`
of the TestRunner has to be 
`public function addExtension(Hook $extension): void`
Because now the TestRunner knows that there are Hooks and TestHooks and right now just TestHooks are allowed to be extensions, but it has to be Hooks right?
> An object that wants to hook into PHPUnit's test runner needs to implement the Hook interface.
Because currently this is not true as it currently needs to implement the `TestHook` interface

keulinho:
@sebastianbergmann I've updated the PR after your clarification:
`AfterTestHook` extends now `TestHook`
Implementations of `Hook` can now be added as extension to the TestRunner

PR #3557, opened 2019-03-12
Add getter support for private var dependencyInput

slaughter550:


sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it. PHPUnit 7.5 is closed for new functionality.

I am not convinced that this addition is useful. If you make a new pull request against `master` then please explain why you need this.

PR #3558, opened 2019-03-12
Add getter support for dependencyInput

slaughter550:
This is needed to access `dependencyInput` as it became private in 7.x and above. 

This variable is used because we have a custom implementation of `runTest` as we auto generate a dynamic testsuite for our framework. 

Same reason as this PR a while ago: https://github.com/sebastianbergmann/phpunit/pull/2230


sebastianbergmann:
Please use commit messages better than `Update TestCase.php` in the future. Thanks!

slaughter550:
@sebastianbergmann no problem, remote right now and I didn't catch it. Thanks so much!

PR #3561, opened 2019-03-13
Flush all remaining results from buffered Testdox printer

### LINKED ISSUE #3560: testdox does not work with filters
| Q                   | A
| PHPUnit version     | master
| PHP version         | 7.2.14
| Installation Method | git

reproduction:
```bash
git glone git@github.com:sebastianbergmann/phpunit.git
cd phpunit
composer install
./phpunit --testdox ## works
./phpunit --testdox --filter CliTestDoxPrinter
```
when adding a filter, a spinner is shown, testdox format is not printed.

epdenouden:
When running with `--filter` the buffering Testdox printer would omit results it couldn't put in the correct order. Fixed by flushing any remaining results in the buffer at the end of the run.

The target for the change is `master` as it relies on a previous improvement to the output flusher in #3545 

Restoring the old behaviour shows the results disappearing:

image

Fixes #3560

### DIFF
--- a/src/Util/TestDox/TestDoxPrinter.php
+++ b/src/Util/TestDox/TestDoxPrinter.php
@@ -252,7 +252,7 @@ class TestDoxPrinter extends ResultPrinter
             $this->writeTestResult($prevResult, $this->testResults[$this->testFlushIndex++]);
         } else {
             do {
-                $flushed = $forceFlush;
+                $flushed = false;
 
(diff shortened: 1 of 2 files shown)


sebastianbergmann:
Sorry, but this should go into `8.0`.

epdenouden:
> Sorry, but this should go into `8.0`.

That'd be best yes, will check if it then causes a small merge conflict when adding it to `master`.

epdenouden:
Superceded by #3562 which can be merged with `8.0` without conflicts

PR #3570, opened 2019-03-22
Add missing global assertion functions

mihaeu:
I use the `Functions.php` in most of my projects for brevity (and to avoid discussions regarding `Assert::`, `self::`, `$this->`), but noticed that `assertEqualsWithDelta` is missing. A quick grep confirmed that it's not the only missing function. 

Since I'm not sure if you want to add the other functions I just wrote a lazy test which ensures that functions are not forgotten in the future.

Open questions:
 - not sure where you'd like the test to be, but I got confused by all the stuff in the `end-to-end` dir
 - test that documentation is up to date?
 - remove documentation and add `@see` to `Assert::` methods
 - run all tests from `AssertTest.php` against the mapped functions just to be safe?

### DIFF
--- a/build/generate_global_assert_wrappers.php
+++ b/build/generate_global_assert_wrappers.php
@@ -80,12 +80,14 @@ foreach ($class->getMethods() as $method) {
         continue;
     }
 
-    $buffer .= \sprintf(
-        "%s\n%s\n{\n    Assert::%s(...\\func_get_args());\n}\n\n",
-        \str_replace('     *', ' *', $method->getDocComment()),
-        \str_replace('public static ', '', \trim($lines[$method->getStartLine() - 1])),
-        $method->getName()
+    $docComment = \str_replace(
+        ['*/', '     *'],
+        ["*\n * @see Assert::" . $method->getName() . "\n */", ' *'],
+        $method->getDocComment()
     );
+    $signature = \str_replace('public static ', '', \trim($lines[$method->getStartLine() - 1]));
+    $body      = "{\n    Assert::" . $method->getName() . "(...\\func_get_args());\n}";
+    $buffer .= "$docComment\n$signature\n$body\n\n";
 }
 
 $buffer .= $constraintMethods;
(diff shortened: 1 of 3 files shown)


sebastianbergmann:
Thank you, @mihaeu. I hope you did not write these yourself but rather used https://github.com/sebastianbergmann/phpunit/blob/master/build/generate_global_assert_wrappers.php to generate them.

As these are just wrappers they are currently untested.

Adding `@see` annotations to the global function wrappers to reference the wrapped methods makes sense. This should be implemented in https://github.com/sebastianbergmann/phpunit/blob/master/build/generate_global_assert_wrappers.php, of course.

mihaeu:
> Thank you, @mihaeu. I hope you did not write these yourself but rather used https://github.com/sebastianbergmann/phpunit/blob/master/build/generate_global_assert_wrappers.php to generate them.

Nope, not by hand, but not using the script either :smile: 

```bash
./phpunit -c phpunit.xml tests/unit/Framework/Assert/FunctionsTest.php | ag 'Mapping for' | sed -r "s/Mapping for//;s/is missing.*//;s/Assert:://;s/\s*//g" | xargs -I{} grep 'public static function {}' src/Framework/Assert.php | sed -r "s/function ([^\(]+)(.*): void/function \1\2: void\n{\n    Assert::\1(...\\\func_get_args());\n}\n/g;s/\s+public static//"
```
I'll update the generated functions using your script.

mihaeu:
Feel free to request any change. Am in Thailand on the beach with plenty of spare time for yak shaving :wink: (got to this issue by finally upgrading dePHPend to PHP 7.3 and PHPUnit 8 btw :stuck_out_tongue:)

epdenouden:
> Am in Thailand on the beach with plenty of spare time for yak shaving

CLOSE. THE. LAP. TOP.

mihaeu:
> > Am in Thailand on the beach with plenty of spare time for yak shaving
> 
> CLOSE. THE. LAP. TOP.

Much appreciated @epdenouden :kissing_heart: I am however on a six-month sabbatical on which I spend most of my time diving, doing yoga and cycling across the USA. I do however also want to use this time to give something back to projects which helped me a lot :wink: so a little laptop time actually is ok and good for my mental health I figure :stuck_out_tongue:

Issue #3578, opened 2019-03-26
Scoped PHAR prophecy issue


TimoBakx:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 8.0.6
| PHP version         | 7.3.3
| Installation Method | PHAR scoped

Using Prophecy `TypeToken` causes some problems when using the scoped PHAR. Running the same test with the regular PHAR of the same version does not cause any exception or failure.

**Source classes:**
```
namespace App;

class ServiceToMock
{
    public function performAction(string $string): void
    {
    }
}
```

```
namespace App;

class MyService
{
    /**
     * @var ServiceToMock
     */
    private $serviceToMock;

    public function __construct(ServiceToMock $serviceToMock)
    {
        $this->serviceToMock = $serviceToMock;
    }

    public function doSomething(): void
    {
        $this->serviceToMock->performAction('Some string');
    }
}
```

**Test case:**
```
namespace App\Tests;

use App\MyService;
use App\ServiceToMock;
use PHPUnit\Framework\TestCase;
use Prophecy\Argument\Token\TypeToken;

final class MyServiceTest extends TestCase
{
    public function testDoSomething(): void
    {
        $mock = $this->prophesize(ServiceToMock::class);

        $mock->performAction(new TypeToken('string'))
            ->shouldBeCalledOnce();

        $subject = new MyService($mock->reveal());

        $subject->doSomething();
    }
}
```

**Output scoped:**
```
$ php ~/Downloads/phpunit-scoped-8.0.6.phar 
PHPUnit 8.0.6 by Sebastian Bergmann and contributors.

Testing Project Test Suite
E                                                                   1 / 1 (100%)

Time: 86 ms, Memory: 12.00 MB

There was 1 error:

1) App\Tests\MyClassTest::testGetSomeString
Error: Class 'Prophecy\Argument\Token\TypeToken' not found

/Users/timobakx/workspace/opensource/tests/tests/MyClassTest.php:17

ERRORS!
Tests: 1, Assertions: 0, Errors: 1.
```

**Output unscoped:**
```
$ php ~/Downloads/phpunit-8.0.6.phar  -v
PHPUnit 8.0.6 by Sebastian Bergmann and contributors.

Runtime:       PHP 7.3.3 with Xdebug 2.7.0RC2
Configuration: /Users/timobakx/workspace/opensource/tests/phpunit.xml.dist

Testing Project Test Suite
.                                                                   1 / 1 (100%)

[...]

PR #3580, opened 2019-03-27
Add 'Prophecy' namespace to the scoping whitelist

sebastianfeldmann:
In oder to setup or check test doubles with Prophecy it has to be
part of the public API of PHPUnit, even it not provided by PHPUnit
itself.

Issues:
- #3578 - Scoped PHAR Prophecy issue


sebastianbergmann:
Merged manually into `8.0`, thanks.

PR #3600, opened 2019-04-08
Fix class name in doc block

lcobucci:
When putting the annotation back we forgot to use the aliased name. This fixes it, making static analysers and IDEs work as expected.

More info: 31df2e805fe0cf184bdfcb528a696dc1daae40c3


lcobucci:
@sebastianbergmann thanks!

PR #3605, opened 2019-04-12
Add support for test suite chunking directly in phpunit.

jwage:
At my job we had 1 hour of functional tests. We built https://github.com/jwage/phpchunkit on top of phpunit to get the chunking functionality and with this we were able to achieve test builds in ~3 minutes. I thought it might be useful to have this built in to phpunit. If you use travis, circle ci, etc. it becomes trivial to configure multiple jobs, one for each chunk.

Before I continue work on this I wanted to know if something like this would be accepted? I am new to the internals of phpunit so I am not 100% sure if I even implemented this functionality the "correct" way.

Example, run your tests in two separate chunks across two different servers/processes

**server1**

```console
$ phpunit --chunk=1 --num-chunks=2 --group=functional
```

**server2**

```console
$ phpunit --chunk=2 --num-chunks=2 --group=functional
```

Thoughts?

### DIFF
--- a/src/Runner/Filter/ChunkIterator.php
+++ b/src/Runner/Filter/ChunkIterator.php
@@ -1,0 +1,78 @@
+<?php declare(strict_types=1);
+/*
+ * This file is part of PHPUnit.
+ *
+ * (c) Sebastian Bergmann <sebastian@phpunit.de>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+namespace PHPUnit\Runner\Filter;
+
+use InvalidArgumentException;
+use PHPUnit\Framework\TestSuite;
+use PHPUnit\Framework\WarningTestCase;
+use PHPUnit\Util\RegularExpression;
+use RecursiveFilterIterator;
+use RecursiveIterator;
+
+/**
+ * @internal This class is not covered by the backward compatibility promise for PHPUnit
+ */
+final class ChunkIterator extends RecursiveFilterIterator
+{
+    /**
+     * @var int
+     */
+    private $chunk;
+
+    /**
+     * @var int
+     */
+    private $numChunks;
+
+    /**
+     * @var Test[][]
+     */
+    private $chunks;
+
+    /**
+     * @throws \Exception
+     */
+    public function __construct(RecursiveIterator $iterator, array $args)
+    {
+        parent::__construct($iterator);
+
+        $this->chunk = $args['chunk'];
+        $this->numChunks = $args['numChunks'];
+
+        if ($this->chunk > $this->numChunks) {
+            throw new InvalidArgumentException(sprintf('You only configured %s chunks but passed chunk %s',
+                $this->numChunks,
+                $this->chunk
+            ));
+        }
+
+        $tests = iterator_to_array($iterator);
+        $numTests = count($tests);
+        $numTestsPerChunk = (int) round($numTests / $this->numChunks);
+
+        $this->chunks = array_chunk(iterator_to_array($iterator), $numTestsPerChunk);
+    }
+
+    public function accept(): bool
+    {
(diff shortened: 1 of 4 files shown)


src/Runner/Filter/ChunkIterator.php
```diff
@@ -0,0 +1,75 @@
...
+     * @throws \Exception
+     */
+    public function __construct(RecursiveIterator $iterator, array $args)
+    {
+        parent::__construct($iterator);
+
+        $this->chunk = $args['chunk'];
+        $this->numChunks = $args['numChunks'];
+        $this->numTests = count($iterator);
+        $this->numTestsPerChunk = floor($this->numTests / $this->numChunks);
```
jwage: Maybe we can use `array_chunk` here to simplify the logic. Let php chunk it and then in `accept()` we can check if the current test is in the chunk being ran.

sebastianbergmann:
My initial thought is that I think that I prefer the approach outlined in #3387. That way, PHPUnit is used to create a list of tests and can be given a list of tests to execute. In between those two steps, that list of tests can be chunked, filtered, sorted, etc. outside of PHPUnit.

jwage:
Interesting. Thanks for the link. 

I did something similar here to achieve my parallelization https://github.com/jwage/phpchunkit

I just built my own runner that finds the tests, chunks them and generates a temporary phpunit xml file to run with.

While what you describe makes sense, it leaves the responsibility up to the end user to implement something custom. Which is not necessarily bad, it is just a higher barrier. With some kind of built in chunking plus travis, circleci, etc. it is just a matter of configuration to achieve parallel test running. How can we make it that easy?

jwage:
Shall we close this then?

sebastianbergmann:
I need to think about this some more. But it's the weekend, you know ;-) Maybe @epdenouden has some input?

theofidry:
Random user opinion: I quite miss parallelism in PHP and I think it's hurting a lot the PHP ecosystem especially for testing frameworks. There is quite a few attempts in the wild but they are all unreliable because:

- Requires to re-implement the test runner (e.g. for paratest for PHPUnit) which is bound to be incomplete or getting out of sync one way or another
- Requires a dubious PHP extension (I have high hopes on `krakjoe/parallel` but it's not widely used yet and more importantly not in the core and likely requires one to recompile his/her PHP binary since requires the ZTS distribution)

Anyway just to say from my PoV both approach suggested here and in #3387 are perfectly valid... I quite like #3387 since it opens the door to more advanced stuff. I however dislike that it's pushing the parallelism integration down to third-party instead of being handled by PHPUnit as this will most likely result in a small adoption rate.

The solution suggested here however is elegant and pragmatic: it won't cover all cases but will get the job done for most people. So _maybe_ a solution would be to:

- Implement #3387
- Leverage #3387 to implement the chunk feature suggested here within PHPUnit

This way you provide a minimalist yet good way to parallelise PHPUnit tests in the core and for the few for whom it's not enough, they have an extension point to work on instead of trying to push more things in PHPUnit guarding it from becoming overly complex and bloated for this extra feature.

jwage:
In my opinion, even if PHP has good parallelism support, really big test suites will still need to be handled differently because you need to be able to run the chunks across a BIG infrastructure, not just one server. For example, with this PR it would be easily possible to have a process with docker that spins up all the servers need to run all the chunks in parallel across many servers, and tear it all down when done. This PR enables people to generate the chunks and run them directly with PHPUnit, then they can decide to either parallelize those chunks in one server and multiple PHP threads, or across many servers however they are able to set it up.

I am starting a new project right now where we are modernizing a legacy application and we have unit, functional and selenium tests. The selenium tests will quickly be very slow when ran serially and I have the infrastructure to be able to run those Selenium tests in parallel, I just need the mechanism to do it. This is why I think PHPUnit should have this mechanism built in.

sebastianbergmann:

PR #3614, opened 2019-04-26
PHPUnit\Framework\Constraint\Attribute should be deprecated (and ignored from code coverage)

donbuehl:


sebastianbergmann:
Merged manually, thanks!

PR #3632, opened 2019-04-26
Refactor and test TestBuilder

nicolashohm:


src/Framework/IncompleteTestCase.php
```diff
@@ -39,11 +39,6 @@ final class IncompleteTestCase extends TestCase
      */
     private $message = '';
 
-    /**
-     * @var bool
-     */
-    private $useOutputBuffering = false;
```
nickel715: phpstorm said this private attribute is unused

src/Framework/TestBuilder.php
```diff
@@ -56,151 +56,157 @@ public function build(\ReflectionClass $theClass, string $name): Test
...
+                    $methodName,
+                    $className,
+                    $data,
+                    $runTestInSeparateProcess,
+                    $preserveGlobalState,
+                    $runClassInSeparateProcess,
+                    $backupSettings
                 );
-
-                if (empty($data)) {
```
nickel715: I removed this if as `TestUtil::getProvidedData` never returns empty array this code is unreachable. To make sure it breaks if someone removes the empty array check I implemented the test `TestBuilderTest.php::testCreateWithEmptyData`

PR #3633, opened 2019-04-26
Added some more test cases

didslm:
Added some more test cases in the test case test class in order to increase the coverage percentage

### DIFF
--- a/tests/unit/Framework/TestCaseTest.php
+++ b/tests/unit/Framework/TestCaseTest.php
@@ -9,8 +9,11 @@
  */
 namespace PHPUnit\Framework;
 
+use PHPUnit\Framework\MockObject\MockBuilder;
 use PHPUnit\Framework\MockObject\MockObject;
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. 

Here are some general remarks:

* Please run `./tools/php-cs-fixer fix` (before your commits) to ensure that the code conforms to PHPUnit's coding style. See [here](https://travis-ci.org/sebastianbergmann/phpunit/jobs/525045272) for feedback from continuous integration.

* Please do not test private implementation details using means such as `setAccessible(true)`

* Please use `expectExceptions()` for testing that an exception is raised

I have decided not to merge this pull request as I do not think that they provide value beyond just generating code coverage. The tests you wrote are still valuable, though, as they show that the code you wanted to test with them probably needs refactoring to make it easier to test.

PR #3636, opened 2019-04-27
Fix: Remove unused field

localheinz:
This PR

* [x] removes an apparently unused field


localheinz:

PR #3644, opened 2019-04-27
Fix: Remove unused field

localheinz:
This PR

* [x] removes an unused field

Follows #3636.


localheinz:

PR #3645, opened 2019-04-27
Handled the empty @depends on tests

didslm:
Added the condition to check if the method in dependency is null and they just specified '@depends' without anything than the test should just skip it and mention that there is no method specified

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -1891,7 +1891,7 @@ abstract class TestCase extends Assert implements Test, SelfDescribing
         $this->result->addError(
             $this,
             new SkippedTestError(
-                \sprintf('This test has dependency declared but no method specified.')
+                \sprintf('This method is invalid at depends annotation.')
             ),
             0
         );
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -1880,7 +1880,7 @@ abstract class TestCase extends Assert implements Test, SelfDescribing
         return true;
     }
 
-    private function markSkippedForNotSpecifyingDependency()
+    private function markSkippedForNotSpecifyingDependency(): void
     {
         $this->status = BaseTestRunner::STATUS_SKIPPED;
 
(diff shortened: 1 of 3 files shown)


src/Framework/TestCase.php
```diff
@@ -1876,6 +1882,23 @@ private function handleDependencies(): bool
...
+    private function markSkippedForNotSpecifyingDependency(): void
+    {
+        $this->status = BaseTestRunner::STATUS_SKIPPED;
+
+        $this->result->startTest($this);
+
+        $this->result->addError(
+            $this,
+            new SkippedTestError(
+                \sprintf('This method is invalid at depends annotation.')
```
localheinz: Perhaps this is better?

```suggestion
                \sprintf('This method has an invalid @depends annotation.')
```

What do you think?

localheinz (approved):
Apart from the wording, this looks good to me!

👍

src/Framework/TestCase.php
Diarselimi: I thought about this but then I thought maybe the @depends keyword can change in the future than we have to change the message too!

tests/end-to-end/execution-order/dependencies-isolation.phpt
```diff
@@ -48,5 +48,8 @@ This test depends on "DependencyFailureTest::testTwo" to pass.
 3) DependencyFailureTest::testFour
 This test depends on "DependencyFailureTest::testOne" to pass.
 
+4) DependencyFailureTest::testHandlesDependsAnnotationWithNoMethodSpecified
+This method is invalid at depends annotation.
```
epdenouden: Update this text and it's green :-)

epdenouden:
Didn't the existing `@depends` validator complain about an invalid (a.k.a. missing) parameter already? I'm going to poke around there for the #3519, so now I'm curious...

Diarselimi:
@epdenouden It did complain that it cannot find the method and fail the test ex: if no method specified
``` ClassName:: ```

epdenouden (approved):

PR #3647, opened 2019-04-27
Enhancement: Add tests for SkippedTestCase

localheinz:
This PR

* [x] adds tests for `SkippedTestCase`

PR #3646, opened 2019-04-27
Enhancement: Add tests for IncompleteTestCase

localheinz:
This PR

* [x] adds tests for `IncompleteTestCase`

PR #3648, opened 2019-04-27
Fix: Remove unnecessary default value assignment

localheinz:
This PR

* [x] removes unnecessary default value assignments

Follows #3646 and #3647.


localheinz:

PR #3652, opened 2019-04-27
Streamline the types in Runner/BaseTestRunner.php

oliverklee:


oliverklee:
@sebastianbergmann I've just noticed that this PR only has changes to the doc blocks, but not to the code. I propose backporting this to 7.5 and 8.1 to get better static analysis there as well.

sebastianbergmann:
Makes sense to me. Can you send a pull request against `7.5`? Thanks!

oliverklee:
> Makes sense to me. Can you send a pull request against `7.5`?

Done: #3684

PR #3655, opened 2019-04-27
Make return type of TestCase::getName non-nullable

oliverklee:
Many places in the code rely on the return type not being null.
So now this is ensured to actually never be null.


localheinz (approved):
👍

PR #3657, opened 2019-04-27
Streamline types in Runner/DefaultTestResultCache.php

oliverklee:


src/Runner/DefaultTestResultCache.php
```diff
@@ -178,17 +180,24 @@ public function serialize(): string
         ]);
     }
 
+    /**
+     * @param string $serialized
```
sebastianbergmann: Is there a reason to annotate the type in a docblock instead of denoting it in syntax?

src/Runner/DefaultTestResultCache.php
```diff
@@ -178,17 +180,24 @@ public function serialize(): string
...
+     */
     public function unserialize($serialized): void
     {
         $data = \unserialize($serialized);
 
         if (isset($data['times'])) {
+            /** @var string $testName */
```
sebastianbergmann: I would prefer `\assert(\is_string($testName);`.

src/Runner/DefaultTestResultCache.php
oliverklee: Yes, because the declaration needs to be the same as the declaration of PHP's own `unserialize` (which does not have a type declaration). If I add a type declaration, PhpStorm complains, and running the tests is not possible anymore at all.

src/Runner/DefaultTestResultCache.php
oliverklee: Done. I'll repush in a second.

src/Runner/DefaultTestResultCache.php
```diff
@@ -24,7 +24,7 @@ final class DefaultTestResultCache implements \Serializable, TestResultCache
     /**
      * Provide extra protection against incomplete or corrupt caches
      *
-     * @var array<string, string>
+     * @var int[]
```
oliverklee: Note: the original annotation saying that this array contains strings was wrong.

sebastianbergmann:
After merging other pull requests, this pull request now has conflicts that must be resolved. Can you have a look, please? Thanks!

oliverklee:
I have resolved the conflict and repushed.

src/Runner/DefaultTestResultCache.php
```diff
@@ -50,7 +50,7 @@ final class DefaultTestResultCache implements \Serializable, TestResultCache
      * $this->defects[$testName] = BaseTestRunner::TEST_SKIPPED;
      * </code>
      *
-     * @var array array<string, int>
+     * @var int[]
      */
     private $defects = [];
```
muglug: I think this annotation is correct, though

src/Runner/DefaultTestResultCache.php
```diff
@@ -62,11 +62,11 @@ final class DefaultTestResultCache implements \Serializable, TestResultCache
      * $this->times[$testName] = 1.234;
      * </code>
      *
-     * @var array<string, float>
+     * @var float[]
```
muglug: This annotation is also correct (and both save you having to use `@var` docblocks above `foreach` loops)

src/Runner/DefaultTestResultCache.php
```diff
@@ -183,19 +185,26 @@ public function serialize(): string
...
 
+    /**
+     * @param string $serialized
+     */
     public function unserialize($serialized): void
     {
         $data = \unserialize($serialized);
 
         if (isset($data['times'])) {
+            /** @var float $testTime */
```
muglug: You're using both `@var` and `assert` for asserting basically the same thing. Nothing wrong with either IMO, but you may want to pick one and use it consistently.

oliverklee:
When I researched this a few years back, PhpStorm supported notations like `int[]`, but not notations like `array<string, int>`, and this also was not part of the PHPDocumentor format. This might have changed. I'll research what of these tools/standards now allow the `array<x, y>` format and then report back:

- PhpStorm
- Psalm
- PHPStan
- Phan
- PSR-5
- PHPDocumentor

oliverklee:
So, if we also want to support PhpStorm's static analysis, we'll need to keep using the `int[]` notation (and not be able to define the type of the keys. Do you have any information on whether this has been added to PhpStorm since 2017 (the date of the support page linked)?

oliverklee:
That being said, as far as Psalm-related type coverage is concerned (which this PR mainly is about), the `array<>` notation is fine. I'll change it back in the places where I've only changed the notation.

epdenouden:
> So, if we also want to support PhpStorm's static analysis, we'll need to keep using the `int[]` notation (and not be able to define the type of the keys. I've just tested this using the mouseover types:

That's a bit of a bummer. I remember talking with @keradus of PHP-CS-Fixer about this and thought that Jetbrain's stuff supported the detailed type definitions.

src/Runner/DefaultTestResultCache.php
epdenouden: Nice catch! The early prototypes did use `string` descriptions

oliverklee:
These are the two relevant PhpStorm tickets I could find:

https://youtrack.jetbrains.com/issue/WI-43843 (most specific)
https://youtrack.jetbrains.com/issue/WI-6558 (more general)

src/Runner/DefaultTestResultCache.php
oliverklee: Done.

src/Runner/DefaultTestResultCache.php
oliverklee: Done.

src/Runner/DefaultTestResultCache.php
oliverklee: Done.

oliverklee:
Repushed.

oliverklee:
php-cs-fixer again … I've repushed.

oliverklee:
The appveyor build failure has nothing to do with the changes in this branch - it's a problem with their caches.

src/Runner/DefaultTestResultCache.php
sebastianbergmann: I see, or rather I did not see that `\Serializable` is implemented here.

PR #3661, opened 2019-04-27
Fix: Test name will not be null

localheinz:
This PR

* [x] adds tests for `TestCase` related to blank or empty names
* [x] fixes a condition and adjusts an exception message

Follows #3655.


localheinz:

Issue #3590, opened 2019-04-05
7.5.8 withConsecutive fails with mutable parameters


root-talis:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 7.5.8
| PHP version         | 7.1.27
| Installation Method | Composer

May be related to #3266.

withConsecutive seems to store invocation parameters as references, which means that when a single mutable object is expected to be passed to a mock two times and it is modified in between the invocation, the tests fail.

```
<?php

use PHPUnit\Framework\MockObject\MockObject;
use PHPUnit\Framework\TestCase;

/** @test */
class MyTest extends TestCase {
    /** @test */
    public function fooTest()
    {
        /** @var \DateTime|MockObject $b */
        $b = $this->createMock(\DateTime::class);

        $foo = function(\DateTime $arg) use ($b) {
            $b->diff($arg);
            $arg->setDate(1970, 4, 5);
            $b->diff($arg);
        };

        $validationValues = [];

        $b->method('diff')
            ->withConsecutive(
                [$this->callback(function(\DateTime $it) use (&$validationValues) {
                    echo 'Validating first call with Y = '.$it->format('Y').PHP_EOL;
                    $validationValues[0] = $it->format('Y');
                    return true;
                })],
                [$this->callback(function(\DateTime $it) use (&$validationValues) {
                    echo 'Validating second call with Y = '.$it->format('Y').PHP_EOL;
                    $validationValues[1] = $it->format('Y');
                    return true;
                })]
            );

        $foo(new \DateTime('2019-04-05'));

        // actually this equals to [1970, 1970] at this point, which is an error
        $this->assertEquals([0 => 2019, 1 => 1970], $validationValues);
    }
}
```

`vendor/phpunit/phpunit/phpunit src/MyTest.php`:

```
PHPUnit 7.5.8 by Sebastian Bergmann and contributors.

F                                                                   1 / 1 (100%)Validating first call with Y = 2019
Validating first call with Y = 2019
Validating second call with Y = 1970
Validating first call with Y = 1970
Validating second call with Y = 1970


Time: 32 ms, Memory: 4.00 MB

There was 1 failure:

1) MyTest::fooTest
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
 Array (
-    0 => 2019
-    1 => 1970
+    0 => '1970'
+    1 => '1970'
 )

/home/talis/temp/src/MyTest.php:39

FAILURES!
[...]

PR #3663, opened 2019-04-27
Adapted the verification methods the ConsecutiveParameters class and the withConsecutive exp…

vasilake-v:
Adapted the verification methods the ConsecutiveParameters class
The `withConsecutive` expectation constraints

Fix for https://github.com/sebastianbergmann/phpunit/issues/3590

### DIFF
--- a/src/Framework/MockObject/Matcher/ConsecutiveParameters.php
+++ b/src/Framework/MockObject/Matcher/ConsecutiveParameters.php
@@ -37,7 +37,12 @@ class ConsecutiveParameters extends StatelessInvocation
     private $invocations = [];
 
     /**
-     * @throws \PHPUnit\Framework\Exception
+     * @var array
+     */
+    private $parameterVerificationResults = [];
+
+    /**
+     * @param array $parameterGroups
      */
     public function __construct(array $parameterGroups)
     {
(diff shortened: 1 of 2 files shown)


slava-v:
I think this class needs complete refactor in terms of making it a Collection of `Parameter` objects by extending the `ArrayObject` or by implementing the `ArrayAccess`. 
As soon as it's behavior is almost similar to the `Parameter` classes and the only difference is that it it should check the sequence of `match` and invokations. But this is for later when I'll have the time (or somebody else for that matter)

src/Framework/MockObject/Matcher/ConsecutiveParameters.php
```diff
@@ -58,41 +63,85 @@ public function toString(): string
...
+
+    /**
+     * Verify a single invocation
+     *
+     * @param BaseInvocation|null $invocation
+     * @param int $callIndex
+     *
+     * @return bool
+     */
+    private function verifyInvocation(?BaseInvocation $invocation, $callIndex): bool
```
sebastianbergmann: Please use `int $callIndex` here.

src/Framework/MockObject/Matcher/ConsecutiveParameters.php
```diff
@@ -58,41 +63,85 @@ public function toString(): string
...
         }
 
+        return true;
+    }
+
+    /**
+     * Verify a single invocation
+     *
+     * @param BaseInvocation|null $invocation
+     * @param int $callIndex
```
sebastianbergmann: The `@param` annotations can be removed.

tests/unit/Framework/MockObject/Matcher/ConsecutiveParametersTest.php
```diff
@@ -65,4 +65,38 @@ public function testIntegrationExpectingException(): void
 
         $mock->foo('invalid');
     }
+
+    /**
+     * Test for the issue https://github.com/sebastianbergmann/phpunit/issues/3590
+     */
+    public function testMutableObjectsChangeSuccess()
```
sebastianbergmann: This method should be `:void`.

tests/unit/Framework/MockObject/Matcher/ConsecutiveParametersTest.php
```diff
@@ -65,4 +65,38 @@ public function testIntegrationExpectingException(): void
...
+    {
+        /** @var \DateTime|MockObject $b */
+        $b = $this->getMockBuilder(\DateTime::class)
+            ->setMethods(['diff'])
+            ->getMock();
+
```
sebastianbergmann: Please align the `->` symbols.

tests/unit/Framework/MockObject/Matcher/ConsecutiveParametersTest.php
```diff
@@ -65,4 +65,38 @@ public function testIntegrationExpectingException(): void
...
+                    $validationValues[0] = $it->format('Y');
+                    return $it->format('Y') === '2019';
+                })],
+                [$this->callback(function (\DateTime $it) use (&$validationValues)  {
+                    $validationValues[1] = $it->format('Y');
+                    return $it->format('Y') === '1970';
+                })]
+            );
+
+        $arg = \DateTime::createFromFormat('Y-m-d', '2019-01-01');
```
sebastianbergmann: If possible, please use `\DateTimeImmutable` instead of `\DateTime`. In this case this means using two separate objects (instead of using `setDate()`).

src/Framework/MockObject/Matcher/ConsecutiveParameters.php
```diff
@@ -58,41 +63,85 @@ public function toString(): string
...
      *
-     * @param int $callIndex
+     * to avoid duplication, it was exported into a separate method
      *
-     * @throws ExpectationFailedException
+     * @param $callIndex
+     * @return bool
      */
-    private function verifyInvocation(BaseInvocation $invocation, $callIndex): void
+    private function shouldInvocationBeVerified($callIndex): bool
```
sebastianbergmann: `$callIndex` is `int`, right? If so, please add the appropriate type declaration.

src/Framework/MockObject/Matcher/ConsecutiveParameters.php
```diff
@@ -58,41 +63,85 @@ public function toString(): string
...
+     * As soon as
+     *  1. this should be verified to avoid duplicate calls when a callback is specified
+     *  2. check is performed for all _and_ individual callbacks
      *
-     * @param int $callIndex
+     * to avoid duplication, it was exported into a separate method
      *
-     * @throws ExpectationFailedException
+     * @param $callIndex
+     * @return bool
```
sebastianbergmann: Please do not duplicate the information from type declarations in docblock annotations.

tests/unit/Framework/MockObject/Matcher/ConsecutiveParametersTest.php
slava-v: I am not sure it makes sense to do that here, as soon as this unit-test replicates the scenario described within the issue #3590 description, where the expected call does change the object state, therefore must be performed on the same object.

src/Framework/MockObject/Matcher/ConsecutiveParameters.php
```diff
@@ -37,8 +37,10 @@ class ConsecutiveParameters extends StatelessInvocation
     private $invocations = [];
 
     /**
-     * @throws \PHPUnit\Framework\Exception
```
oliverklee: Is the `@throws` annotation not applicable for the constructor anymore?

src/Framework/MockObject/Matcher/ConsecutiveParameters.php
```diff
@@ -37,8 +37,10 @@ class ConsecutiveParameters extends StatelessInvocation
...
+     * @var array
```
oliverklee: I think this we can make this more specific as `ExpectationFailedException[]`.

src/Framework/MockObject/Matcher/ConsecutiveParameters.php
slava-v: no, it's not applicable.
The constructor doesn't trigger anything that could throw that exception, therefore it was removed.

src/Framework/MockObject/Matcher/ConsecutiveParameters.php
slava-v: ... same as above. The `__construct` doesn't throw any exception. It prefills it's object properties only

src/Framework/MockObject/Matcher/ConsecutiveParameters.php
oliverklee: If I understand you correct, removing the annotation is not related to the other changes in this PR. I propose we move it to a separate cleanup PR then.

PR #3667, opened 2019-04-28
Fix: Mark field and methods related to deprecated TestListener as deprecated

localheinz:
This PR

* [x] marks fields and methods on `TestResult` that are related to the deprecated `TestListener` as deprecated (and ignores methods in regard to code coverage)

Related to #3388.


localheinz:

PR #3668, opened 2019-04-28
Fix: Move class into directory matching namespace

localheinz:
This PR

* [x] moves a class into the directory matching its namespace

💁‍♂ The class is named `PHPUnit\Runner\NullTestResultCache`, but it currently resides in a file with the path `/src/Util/NullTestResultCache.php`.


sebastianbergmann:
Cherry-picked into `8.1` and merged to `master` from there. Thanks!

localheinz:

PR #3669, opened 2019-04-29
Fix: Move classes into directory matching namespace

localheinz:
This PR

* [x] moves classes into directory matching the namespace

💁‍♂ Not sure, since we are not using PSR-0 or PSR-4 autoloading, it actually doesn't matter much in which directories classy constructs reside. Still, I believe it's good practice to map namespaces to directories. What do you think?


sebastianbergmann:

This is one of the cases where I like to diverge from the default namespace->directory relationship. These interfaces belong in the `PHPUnit\Runner` namespace but I would like to have them grouped together in a directory.

PR #3679, opened 2019-05-03
Explicity return floats for DefaultTestResultCache::getTime

oliverklee:
This method has a float return type, and the fallback return value
should also be an explicit float, not an int.

### DIFF
--- a/tests/unit/Runner/DefaultTestResultCacheTest.php
+++ b/tests/unit/Runner/DefaultTestResultCacheTest.php
@@ -1,0 +1,34 @@
+<?php declare(strict_types=1);
+/*
+ * This file is part of PHPUnit.
+ *
+ * (c) Sebastian Bergmann <sebastian@phpunit.de>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+namespace PHPUnit\Runner;
+
+use PHPUnit\Framework\TestCase;
+
+/**
+ * @covers \PHPUnit\Runner\DefaultTestResultCache
+ * @small
+ */
+final class DefaultTestResultCacheTest extends TestCase
+{
+    /**
+     * @var DefaultTestResultCache
+     */
+    private $subject;
+
+    protected function setUp(): void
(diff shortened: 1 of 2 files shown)


tests/unit/Runner/DefaultTestResultCacheTest.php
```diff
@@ -0,0 +1,34 @@
...
+     * @var DefaultTestResultCache
+     */
+    private $subject;
+
+    protected function setUp(): void
+    {
+        $this->subject = new DefaultTestResultCache();
+    }
+
+    public function testGetTimeForInexistentTestNameReturnsZero(): void
```
localheinz: The word _inexistent_ is _non-existent_, what do you think about using _non-existent_ instead?

Also, the test method name reflects the current behavior, perhaps it makes sense to add `AsFloat` as suffix?

tests/unit/Runner/DefaultTestResultCacheTest.php
```diff
@@ -0,0 +1,34 @@
...
+final class DefaultTestResultCacheTest extends TestCase
+{
+    /**
+     * @var DefaultTestResultCache
+     */
+    private $subject;
+
+    protected function setUp(): void
+    {
+        $this->subject = new DefaultTestResultCache();
```
localheinz: I believe it's a matter of personal taste, but personally I would prefer not to set up the _system under test_ in the `setUp()` method, what do you think?

localheinz (approved):
👍

tests/unit/Runner/DefaultTestResultCacheTest.php
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+namespace PHPUnit\Runner;
+
+use PHPUnit\Framework\TestCase;
+
+/**
+ * @covers \PHPUnit\Runner\DefaultTestResultCache
```
localheinz: 👍

PR #3682, opened 2019-05-05
Allow using duration for the --order-by option as well as for the executionOrder attribute in phpunit.xml

### LINKED ISSUE #3681: Implement --order-by=duration CLI option
| Q                   | A
| PHPUnit version     | 8.0.6
| PHP version         | 7.3.2
| Installation Method | Composer / PHAR

`duration` execution order is configurable only via config file, but can't be configured via CLI
introduced in #3293

```
ker@dus:~ λ vendor/bin/phpunit --order-by=duration
PHPUnit 8.0.6 by Sebastian Bergmann and contributors.

unrecognized --order-by option: duration
```
```
ker@dus:~ λ vendor/bin/phpunit --help | grep order-by
  --order-by=<order>          Run tests in order: default|reverse|random|defects|no-depends
```

localheinz:
This PR

* [x] allows specifying `duration` as value for the `order-by` option via CLI as well as for the `executionOrder` attribute in `phpunit.xml`

Fixes #3681.

Also see https://github.com/sebastianbergmann/phpunit-documentation-english/pull/143.

### DIFF
--- a/src/TextUI/Help.php
+++ b/src/TextUI/Help.php
@@ -100,7 +100,7 @@ final class Help
             ['arg'    => '--printer <printer>', 'desc' => 'TestListener implementation to use'],
             ['spacer' => ''],
 
-            ['arg'  => '--order-by=<order>', 'desc' => 'Run tests in order: default|reverse|random|defects|no-depends|duration'],
+            ['arg'  => '--order-by=<order>', 'desc' => 'Run tests in order: default|defects|duration|no-depends|random|reverse'],
             ['arg'  => '--random-order-seed=<N>', 'desc' => 'Use a specific random seed <N> for random order'],
             ['arg'  => '--cache-result', 'desc' => 'Write test results to cache file'],
             ['arg'  => '--do-not-cache-result', 'desc' => 'Do not write test results to cache file'],
--- a/src/TextUI/Help.php
+++ b/src/TextUI/Help.php
@@ -100,7 +100,7 @@ final class Help
             ['arg'    => '--printer <printer>', 'desc' => 'TestListener implementation to use'],
             ['spacer' => ''],
 
-            ['arg'  => '--order-by=<order>', 'desc' => 'Run tests in order: default|reverse|random|defects|no-depends'],
+            ['arg'  => '--order-by=<order>', 'desc' => 'Run tests in order: default|reverse|random|defects|no-depends|duration'],
             ['arg'  => '--random-order-seed=<N>', 'desc' => 'Use a specific random seed <N> for random order'],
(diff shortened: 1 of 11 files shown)


localheinz:
@sebastianbergmann 

Should I target `7.5` instead of `master`?

sebastianbergmann:
IMO, this is not a bugfix but new functionality. Can you please make this pull request against the `master` branch instead? Thanks!

src/TextUI/Help.php
```diff
@@ -100,7 +100,7 @@ final class Help
             ['arg'    => '--printer <printer>', 'desc' => 'TestListener implementation to use'],
             ['spacer' => ''],
 
-            ['arg'  => '--order-by=<order>', 'desc' => 'Run tests in order: default|reverse|random|defects|no-depends'],
+            ['arg'  => '--order-by=<order>', 'desc' => 'Run tests in order: default|reverse|random|defects|no-depends|duration'],
```
localheinz: Do you have any preferences in regard to the order 

* in which the options are listed in the help output
* in which the options are handled in the `switch`-statement

?

Personally, I would probably keep them sorted like this

* `default` first
* everything else alphabetically

If you like this, I will adjust!

localheinz:
@sebastianbergmann 

Adjusted and pointing against `master`!

src/TextUI/Help.php
sebastianbergmann: I like :)

epdenouden: Nice addition :)

localheinz: Thanks, adjusted!

localheinz:
@sebastianbergmann @epdenouden 

Just noticed that we are apparently not handling via `phpunit.xml`, would you like me to adjust this in this PR as well?

sebastianbergmann:
Are you saying that currently neither CLI option nor configuration directive exists to control duration sorting? If so, yes please.

localheinz:
@sebastianbergmann 

Already done! 😉

localheinz:
@sebastianbergmann 

Do I need to update documentation somewhere?

sebastianbergmann:
Thanks!

localheinz:

localheinz:
@epdenouden @sebastianbergmann 

I have been wondering, perhaps the `reverse` value should be a modifier for the existing strategies?

That is, one could sort tests by

* `default` (whatever that is, the order in which tests are found on the specific file system)
* `duration`
* `random`

and then `reverse` could be a modifier to order in the opposite direction (doesn't make a lot of sense with `random`, but anyway).

What do you think?


* `duration`
* `

PR #3685, opened 2019-05-09
Cleanup: Improve a test name

oliverklee:
Make it obvious that the test is about the return type.


oliverklee:
This change implements one of the suggestions from code review.

PR #3686, opened 2019-05-09
Bugfix: Explicity return floats for DefaultTestResultCache::getTime

oliverklee:
This method has a float return type, and the fallback return value
should also be an explicit float, not an int.

This is the 8.1 backport of #3679.


localheinz (approved):
👍

PR #3687, opened 2019-05-09
Add onlyMethods + addMethods and soft deprecate setMethods

DFoxinator:
The purpose of this PR is to provide important testing functionality when using setMethods() on a mock, to which there is currently no workaround to accomplish otherwise (as far as I can see, please correct me if I'm wrong). Right now, any time setMethods is used, the test using it is completely unprotected against any of the methods sent to setMethods not actually existing in the class. In practice, for the large majority of use-cases where setMethods is appropriate, this means that the tests using it are likely not functioning as the test writer expected because the methods in setMethods are not asserted in any way to actually exist, so you can easily just delete methods while refactoring code and have properly written tests still pass with no warning. In practice, this behavior is made even more confusing because so far every developer I've talked to mixes up the strictness of createMock (which tells you when you try to mock a method that didn't exist) with also offering that functionality when you use setMethods, which unfortunately is not the case.

This change provides a simple backwards-compatible way to address this issue, and provide the functionality that most casual PHPUnit users already think they are getting. Anyone who wants the stricter functionality can just search and replace setMethods with the stricter version that asserts the methods actually exist. Also, just to add, the name I added for this new method is kind of just one I came up with quickly - so feedback on that and anything else in this is definitely appreciated.

### DIFF
--- a/src/Framework/MockObject/MockBuilder.php
+++ b/src/Framework/MockObject/MockBuilder.php
@@ -87,6 +87,11 @@ final class MockBuilder
     private $generator;
 
     /**
+     * @var bool
+     */
+    private $alreadyUsedMockMethodConfiguration = false;
+
+    /**
      * @param string|string[] $type
      *
      * @psalm-param class-string<MockedType>|string|string[] $type
@@ -200,6 +205,12 @@ final class MockBuilder
      */
     public function onlyMethods(array $methods): self
     {
+        if ($this->alreadyUsedMockMethodConfiguration) {
+            throw new RuntimeException('Can\'t use onlyMethods after already configuring mock methods on the mock.');
+        }
+
+        $this->alreadyUsedMockMethodConfiguration = true;
+
         $reflection = new \ReflectionClass($this->type);
 
         foreach ($methods as $method) {
@@ -228,6 +239,12 @@ final class MockBuilder
      */
     public function addMethods(array $methods): self
     {
+        if ($this->alreadyUsedMockMethodConfiguration) {
+            throw new RuntimeException('Can\'t use addMethods after already configuring mock methods on the mock.');
+        }
+
+        $this->alreadyUsedMockMethodConfiguration = true;
+
         $reflection = new \ReflectionClass($this->type);
 
         foreach ($methods as $method) {
--- a/src/Framework/MockObject/MockBuilder.php
+++ b/src/Framework/MockObject/MockBuilder.php
@@ -193,21 +193,22 @@ final class MockBuilder
      * Specifies the subset of methods to mock, requiring each to exist in the class
      *
      * @param string[] $methods
+     *
+     * @throws RuntimeException
      */
     public function setRealMethods(array $methods): self
     {
-        if ($methods) {
-            $reflection = new \ReflectionClass($this->type);
+        $reflection = new \ReflectionClass($this->type);
 
-            foreach ($methods as $method) {
-                if (!$reflection->hasMethod($method)) {
-                    throw new RuntimeException(
-                        \sprintf(
-                            'Trying to set mock method "%s" which cannot be configured because it does not exist',
-                            $method
-                        )
-                    );
-                }
+        foreach ($methods as $method) {
+            if (!$reflection->hasMethod($method)) {
+                throw new RuntimeException(
+                    \sprintf(
+                        'Trying to set mock method "%s", but it does not exist in class "%s"',
+                        $method,
+                        $this->type
+                    )
+                );
             }
         }
 
--- a/src/Framework/MockObject/MockBuilder.php
+++ b/src/Framework/MockObject/MockBuilder.php
@@ -181,6 +181,8 @@ final class MockBuilder
 
     /**
      * Specifies the subset of methods to mock. Default is to mock none of them.
+     *
+     * @deprecated https://github.com/sebastianbergmann/phpunit/pull/3687
      */
     public function setMethods(array $methods = null): self
     {
@@ -196,7 +198,7 @@ final class MockBuilder
      *
      * @throws RuntimeException
      */
-    public function setRealMethods(array $methods): self
+    public function onlyMethods(array $methods): self
     {
         $reflection = new \ReflectionClass($this->type);
 
@@ -204,7 +206,35 @@ final class MockBuilder
             if (!$reflection->hasMethod($method)) {
                 throw new RuntimeException(
                     \sprintf(
-                        'Trying to set mock method "%s", but it does not exist in class "%s"',
+                        'Trying to set mock method "%s" with onlyMethods, but it does not exist in class "%s". Use addMethods() for methods that don\'t exist in the class.',
+                        $method,
+                        $this->type
+                    )
+                );
+            }
+        }
+
+        $this->methods = $methods;
+
+        return $this;
+    }
+
+    /**
+     * Specifies methods that don't exist in the class which you want to mock
+     *
+     * @param string[] $methods
+     *
+     * @throws RuntimeException
+     */
+    public function addMethods(array $methods): self
+    {
+        $reflection = new \ReflectionClass($this->type);
+
+        foreach ($methods as $method) {
+            if ($reflection->hasMethod($method)) {
+                throw new RuntimeException(
+                    \sprintf(
+                        'Trying to set mock method "%s" with addMethod, but it exists in class "%s". Use onlyMethods() for methods that exist in the class.',
                         $method,
                         $this->type
                     )
--- a/src/Framework/MockObject/MockBuilder.php
+++ b/src/Framework/MockObject/MockBuilder.php
@@ -193,6 +193,8 @@ final class MockBuilder
     {
         $this->methods = $methods;
 
+        $this->alreadyUsedMockMethodConfiguration = true;
+
         return $this;
     }
 
@@ -206,7 +208,12 @@ final class MockBuilder
     public function onlyMethods(array $methods): self
     {
         if ($this->alreadyUsedMockMethodConfiguration) {
-            throw new RuntimeException('Can\'t use onlyMethods after already configuring mock methods on the mock.');
+            throw new RuntimeException(
+                \sprintf(
+                    'Can\'t use onlyMethods on "%s" mock because mocked methods were already configured.',
+                    $this->type
+                )
+            );
         }
 
         $this->alreadyUsedMockMethodConfiguration = true;
@@ -240,7 +247,12 @@ final class MockBuilder
     public function addMethods(array $methods): self
     {
         if ($this->alreadyUsedMockMethodConfiguration) {
-            throw new RuntimeException('Can\'t use addMethods after already configuring mock methods on the mock.');
+            throw new RuntimeException(
+                \sprintf(
+                    'Can\'t use addMethods on "%s" mock because mocked methods were already configured.',
+                    $this->type
+                )
+            );
         }
 
         $this->alreadyUsedMockMethodConfiguration = true;
--- a/src/Framework/MockObject/MockBuilder.php
+++ b/src/Framework/MockObject/MockBuilder.php
@@ -190,6 +190,31 @@ final class MockBuilder
     }
 
     /**
+     * Specifies the subset of methods to mock, requiring each to exist in the class
+     */
+    public function setRealMethods(array $methods = null): self
+    {
+        if ($methods) {
+            $reflection = new \ReflectionClass($this->type);
+
+            foreach ($methods as $method) {
+                if (!$reflection->hasMethod($method)) {
+                    throw new RuntimeException(
+                        \sprintf(
+                            'Trying to set mock method "%s" which cannot be configured because it does not exist',
+                            $method
+                        )
+                    );
+                }
+            }
+        }
+
+        $this->methods = $methods;
+
+        return $this;
+    }
+
+    /**
      * Specifies the subset of methods to not mock. Default is to mock all of them.
      */
     public function setMethodsExcept(array $methods = []): self
--- a/src/Framework/MockObject/MockBuilder.php
+++ b/src/Framework/MockObject/MockBuilder.php
@@ -191,8 +191,10 @@ final class MockBuilder
 
     /**
      * Specifies the subset of methods to mock, requiring each to exist in the class
+     *
+     * @param string[] $methods
      */
-    public function setRealMethods(array $methods = null): self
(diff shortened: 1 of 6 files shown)


src/Framework/MockObject/MockBuilder.php
```diff
@@ -178,6 +178,31 @@ public function setMethods(array $methods = null): self
         return $this;
     }
 
+    /**
+     * Specifies the subset of methods to mock, requiring each to exist in the class
+     */
+    public function setRealMethods(array $methods = null): self
+    {
+        if ($methods) {
+            $reflection = new \ReflectionClass($this->type);
```
psalm-shepherd: Argument 1 of ReflectionClass::__construct expects class-string<T>|object, parent type string provided
```
            $reflection = new \ReflectionClass($this->type);
                                               ^^^^^^^^^^^
```

psalm-shepherd:
Psalm found some errors

src/Framework/MockObject/MockBuilder.php
psalm-shepherd: Argument 1 of ReflectionClass::__construct expects class-string<T>|object, parent type string provided
```
            $reflection = new \ReflectionClass($this->type);
                                               ^^^^^^^^^^^
```

psalm-shepherd:
Psalm found some errors

src/Framework/MockObject/MockBuilder.php
DFoxinator: That doesn't seem to be accurate I don't think?

muglug: There are a few ways to fix this.

1. change `if ($methods) {` to `if ($methods && class_exists($this->type)) {`
2. add `@psalm-var class-string` to the property type definition of `MockBuilder::$type`
3. add the error to the baseline with `vendor/bin/psalm --set-baseline=.psalm/baseline.xml`

muglug: There's also a potential bug that @sebastianbergmann may have more insight into - the constructor allows a `string|array<string>`, but the property only allows `string`. I don't see how an `array` would work here.

DFoxinator: @muglug thanks a bunch for the help with that - I went with the 3rd approach for now, will see if there's any other feedback for that and please let me know if you see anything else.

psalm-shepherd:
Psalm didn’t find any errors!

src/Framework/MockObject/MockBuilder.php
```diff
@@ -178,6 +178,31 @@ public function setMethods(array $methods = null): self
         return $this;
     }
 
+    /**
+     * Specifies the subset of methods to mock, requiring each to exist in the class
+     */
+    public function setRealMethods(array $methods = null): self
```
localheinz: What do you think about removing the `null` default value?

oliverklee: Can we add an `@param string[] $methods` annotation to help static analysis?

DFoxinator: I think either way could make sense. The only reason I had defaulted to null is to keep it 100% backwards compatible with setMethods. When testing this is was a little beneficial in a large codebase I work on to just do a big search and replace and see the tests light up and not have to worry about compatibility, but let me know what you think, either way works.

DFoxinator: Thanks, added!

localheinz: I understand that the method signature is consistent with that of `setMethods()`, but my questions here are: 

- What does `null` mean in this context? 
- Why would anyone want to invoke this method with`null`, or without any argument at all?

DFoxinator: I think you have good points, I appreciate that feedback. It does seem to not really be useful/make sense, so might as well remove. Updated and removed!

src/Framework/MockObject/MockBuilder.php
```diff
@@ -178,6 +178,33 @@ public function setMethods(array $methods = null): self
...
     }
 
+    /**
+     * Specifies the subset of methods to mock, requiring each to exist in the class
+     *
+     * @param string[] $methods
+     */
+    public function setRealMethods(array $methods): self
+    {
+        if ($methods) {
```
oliverklee: This `if` then is not needed anymore, right?

src/Framework/MockObject/MockBuilder.php
```diff
@@ -178,6 +178,33 @@ public function setMethods(array $methods = null): self
...
+            $reflection = new \ReflectionClass($this->type);
+
+            foreach ($methods as $method) {
+                if (!$reflection->hasMethod($method)) {
+                    throw new RuntimeException(
```
oliverklee: We need an `@throws` annotation for this method.

src/Framework/MockObject/MockBuilder.php
```diff
@@ -178,6 +178,33 @@ public function setMethods(array $methods = null): self
...
+                        \sprintf(
+                            'Trying to set mock method "%s" which cannot be configured because it does not exist',
```
oliverklee: Can we also have the class name in the exception message? This would make debugging problems easier.

DFoxinator: Good idea. I added it into the message and also changed up the message a little.

src/Framework/MockObject/MockBuilder.php
```diff
@@ -178,6 +178,34 @@ public function setMethods(array $methods = null): self
...
+     *
+     * @throws RuntimeException
+     */
+    public function setRealMethods(array $methods): self
+    {
+        $reflection = new \ReflectionClass($this->type);
+
+        foreach ($methods as $method) {
+            if (!$reflection->hasMethod($method)) {
+                throw new RuntimeException(
```
oliverklee: Do we need a leading slash before `RuntimeExeception` here (and in the `@throws` annotation)?

oliverklee: And we should have unit tests for the error case.

DFoxinator: I don't think the leading slash is needed in this case? It seems to be resolved correctly.

There is a test for the exception, but do you mean for the message itself?

oliverklee: I don't get why `RuntimeException` gets resolved correctly, though - it's not `use`d, and this class is within the `PHPUnit\Framework\MockObject` namespace. Do you have an idea?

oliverklee: Oh, right, now I see the test.

DFoxinator: They are both in PHPUnit\Framework\MockObject namespace though right? So i think that would be why.

oliverklee: Ah, you're right. There indeed is a `PHPUnit\Framework\MockObject\RuntimeException` class. How confusing.

oliverklee: In other words: This is not what I would have expected from the code, and the (existing) code does not follow the "minimize the number of surprises" principle here.

sebastianbergmann:
> any time setMethods is used, the test using it is completely unprotected against any of the methods sent to setMethods not actually existing in the class

The whole point of `setMethods()` is to configure methods that do not (yet) exist. Can you please provide a use case for the `setRealMethods()` method you are proposing here?

DFoxinator:
The use case is when you need to mock some methods on a class but not all of them. As far as I know, you can't use createMock for that. I've always used createMock for all dependencies (where you don't need any functionality from the actual class), but for testing actual service classes where those dependencies are passed in, createMock doesn't seem like an option.

So for any use case where you want to test a class but also mock out some methods (in my case, that's generally for mocking public methods on the service and also protected/trait methods that make external calls), from my understanding, you'd use MockBuilder and setMethods for that right now, and I've always seen that in practice in the codebases I've worked on. But the problem I've encountered here is using setMethods makes the test ineffective because there's no check to see if the methods being mocked actual exist. So in a recent case. I saw prod code that was broken because a test was properly mocking data using setMethods and a mock for the method, but the method itself had been removed from the code but was still being called from that service that was being tested.

sebastianbergmann:
You are correct: `createMock()` does not support partial mocking. But `createPartialMock()` does. However, since that is just a wrapper for the Mock Builder API and because it uses `setMethods()` internally there is no check for method existence.

DFoxinator:
Yeah, that makes sense. It seems for this use case right now there really is no workaround, and in the codebase I work on, in the hundreds of places MockBuilder setMethods/createPartialMock is used, the behavior that was intended was the test would fail if any of the methods that were attempted to be mocked were removed from the code. The only use case we could think of where you wouldn't want that check tied to setMethods is for magic methods maybe, so hence the idea for this new setRealMethods for the majority of cases.

sebastianbergmann:
I am not happy with the name `setRealMethods()`. For that matter, I am also not happy with `setMethods()`.

How about:

* Introduce `addMethods()` for adding additional methods that do not (yet) exist in the doubled interface or class
* Introduce `onlyMethods()` for selecting which set of methods in the doubled interface or class should be doubled
* Figure out whether `addMethods()` and `onlyMethods()` should be combinable (I don't think they should be)
* Deprecate `setMethods()` in PHPUnit 8.x (soft)
* Trigger a warning when names for methods that do not exist are passed to `createPartialMock()` in PHPUnit 8.x
* Let `createPartialMock()` use `onlyMethods()` in PHPUnit 9
* Deprecate `setMethods()` in PHPUnit 9 (hard)
* Remove `setMethods()` in PHPUnit 10

DFoxinator:
@sebastianbergmann makes sense, sounds like a solid plan to me. I will try to update this PR later today based on that plan. Thanks!

DFoxinator:
@localheinz @oliverklee I updated this PR with the changes outlined by @sebastianbergmann above, please review when you have a chance and let me know if you have any feedback. Thanks.

DFoxinator:
@sebastianbergmann is it possible for this to be considered for 8.2.0?

sebastianbergmann:
Possible and likely. I was not able to work Open Source this week and may not have time to work on Open Source next week. But the week leading up to the PHPUnit 8.2 release is currently blocked for Open Source work.

DFoxinator:
Sounds good, thanks for the update. If there's anything more I can do to help with this (ex. documentation or whatever else is needed) feel free to let me know during that week leading up to 8.2.

DFoxinator:
There were some conflicts/CI failures from latest merges so I updated this to get it to pass/be mergeable.

sebastianbergmann:
Thanks!

DFoxinator:
@sebastianbergmann is there anything else that's needed for this for it to get into 8.3?

DFoxinator:
@sebastianbergmann another conflict was introduced so I updated the PR to resolve. Please let me know if there are any blockers for this getting merged. I'm happy to help with anything that needs to be done for this. I'd really like to be able to use this functionality in the PHPUnit tests I work with soon.

sebastianbergmann:
Thank you for keeping this up-to-date. This PR almost made it to the very top of my TODO stack. Sorry for keeping you waiting.

DFoxinator:
No problem, and sounds good!

DFoxinator:
@sebastianbergmann I was taking a look at the available PHPUnit documentation, is there a way I'd be able to contribute documentation/usage example for these too? Mainly for https://phpunit.readthedocs.io/en/8.2/test-doubles.html#mock-objects I guess, but for the 8.3 manual maybe

sebastianbergmann:
Merged manually, thanks.

DFoxinator:
Thanks @sebastianbergmann!

GrahamCampbell:
Can this be backported to PHPUnit 7.5. Laravel 6 LTS will need to support all PHPUnit versions 7 through 10.

GrahamCampbell:
@sebastianbergmann Can this be backported to PHPUnit 7.5? Laravel 6 LTS will need to support all PHPUnit versions 7 through 10.

sebastianbergmann:
PHPUnit 7 has reached its end of life and will get no more releases.

yyaremenko:
the PHPUnit documentation still has setMehods() and has no addMethods()

https://phpunit.readthedocs.io/en/9.1/test-doubles.html#mock-objects

DFoxinator:
I'm happy to submit a docs PR if that's possible - do you know if those docs are updatable via a repo? cc @sebastianbergmann

DFoxinator:
I guess https://github.com/sebastianbergmann/phpunit-documentation-english/ would end up there?

sebastianbergmann:
Yes, https://github.com/sebastianbergmann/phpunit-documentation-english/ is the correct repository for the English documentation of PHPUnit.

acewebsolutions:
All sounds good except case scenario:
What if I am trying to mock a non-existent class. 

In this case setMethods would have allowed me to do so, whereas addMethods and onlyMethods will first check the methods against the class I have mocked, which will result in an Class "NonExistentClass" does not exist error.

Maybe a check can be added if $this->allowMockingUnknownTypes = true then the ReflectionClass check is bypassed

stronk7:
Just linking from here to existing documentation change PR, because it seems that current version docs still are missing this stuff: https://github.com/sebastianbergmann/phpunit-documentation-english/pull/205

Ciao :-)

rulatir:
I need to mock an interface and supply mock implementation of a few methods via `->willReturnCallback()`, while leaving all _other_ methods to be automatically stubbed.

With current set of non-deprecated, non-removed method-mocking APIs, this appears **completely impossible**. Specifically, `onlyMethods(['method1','method2`])->getMock()` results in a class that implements the interface but fails to implement the other methods and causes an error.

How do I partially mock an interface AND have everything else auto-stubbed?

Levivb:
Is there an alternative to `setMethodsExcept`? 

It's deprecated and links to this issue, but it seems there's no alternative?

dylan-marty-boldcommerce:
> Is there an alternative to `setMethodsExcept`?
> 
> It's deprecated and links to this issue, but it seems there's no alternative?

Alternatives were tucked into some of the thoughts on a previous comment here: https://github.com/sebastianbergmann/phpunit/pull/3687#issuecomment-492537584

> - [...] addMethods() for adding additional methods that do not (yet) exist in the doubled interface or class
> - [...] onlyMethods() for selecting which set of methods in the doubled interface or class should be doubled

I would imagine for most use cases, you're looking for `onlyMethods` instead.

PR #3689, opened 2019-05-10
Fix magic method __sleep signature

sshilko:
https://www.php.net/manual/en/language.oop5.magic.php#object.sleep

Visibility for magic method __sleep must be public. Found: private
__sleep must return array


sebastianbergmann:

PR #3694, opened 2019-05-16
Do not ignore constructor arguments based on type being mocked

morozov:
Currently, it is impossible to mock the message or code on a `Throwable`. On the one hand, `getMessage()` and `getCode()` are `final` in `Exception`, on the other, the mock generator ignores constructor arguments if the type being mocked is an interface.

The existing behavior doesn't seem correct because on the one hand, PHP currently allows passing more arguments than the method expects (including class constructors). On the other, if it stops doing so at some point later, it should be a test error, not the PHPUnit's call to ignore the arguments.

### DIFF
--- a/src/Framework/MockObject/Generator.php
+++ b/src/Framework/MockObject/Generator.php
@@ -593,9 +593,7 @@ class Generator
     {
         $this->evalClass($code, $className);
 
-        if ($callOriginalConstructor &&
-            \is_string($type) &&
-            !\interface_exists($type, $callAutoload)) {
+        if ($callOriginalConstructor) {
             if (\count($arguments) === 0) {
                 $object = new $className;
             } else {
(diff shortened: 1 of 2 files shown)


sebastianbergmann:
Just curious: in what scenario does stubbing/mocking `Throwable` or `Exception` make sense? I have never though about this before the recent patches, much less needed it.

morozov:
In [doctrine/dbal](https://github.com/doctrine/dbal), we have some logic of handling/generalizing driver exceptions based on their properties. We do not mock `Throwable` or `Exception` directly but we do mock `interface DriverException implements Throwable`.

sebastianbergmann:
Can you point me to a test that does this? Thanks!

morozov:
See [`Doctrine\Tests\DBAL\Driver\AbstractDriverTest::testConvertsException()`](https://github.com/doctrine/dbal/blob/4d0d2affceb800088f24904dd5aed316db97b145/tests/Doctrine/Tests/DBAL/Driver/AbstractDriverTest.php#L260) for example.

PR #3699, opened 2019-05-21
Fix #3697 Respect @coversNothing at coverage driver start

### LINKED ISSUE #3697: @codeCoverageIgnore on a docblock should prevent driver start/stop
| Q                         | A
| --------------------------| ---------------
| php-code-coverage version | 6.1.4
| PHP version               | 7.2.18
| Driver                    | PHPDBG
| Xdebug version (if used)  | x.y.z
| Installation Method       | Composer
| Usage Method              | PHPUnit
| PHPUnit version (if used) | 7.5.9

At the moment if you specify `@codeCoverageIgnore` on an individual test function or testcase class then that entire test or group of tests is not included in coverage generation. However, this only filters the data which is collected, rather than preventing the collection in the first place.

Whilst the collection is still required in some situations, specifically where a single line is ignored (`// @codeCoverageIgnore`), or a range of lines via the `@codeCoverageIgnoreStart` and `@codeCoverageIgnoreEnd` tags, where the entire class or an entire function is ignored then I feel it would be prudent to not start the driver at all.

This becomes a problem when testing large and complex datasets as the fact that the driver is logging at all causes high memory consumption leading to out of memory. To give a more concrete example, we have one dataset related to machine learning where generation of the data consumes 76GB of memory via opcache alone. At the moment our only solutions are to either add a `@group` tag to the relevant test and ignore it from phpunit when we wish to generate coverage at all, or to add a hack before and after the test to stop/restart the opcache logging if it is enabled.

I'd like to propose that  php-code-coverage either:
* does not start/stop the driver if the `@codeCoverageIgnore` tag is applied to an entire testcase class, or test function within that testcase class; or
* a new tag is added to support the same functionality.

I'm aware that xdebug now supports some filters, but unfortunately phpdbg does not do so yet.

If you feel that this would be a beneficial change I'm more than happy to help in its development.

andrewnicols:
As mentioned in #3697, the existing `@coversNothing` annotation works as expected for classes, but for methods the coverage data is still generated and it is later filtered out.

This change adds support for `@coversNothing` in the test method annotation, and attempts to address inverted assertions in the unit testing of the existing functionality.

### DIFF
--- a/src/Util/Test.php
+++ b/src/Util/Test.php
@@ -156,6 +156,12 @@ final class Test
     {
         $annotations = $test->getAnnotations();
 
+        // If there is no @covers annotation but a @coversNothing annotation on
+        // the test method then code coverage data does not need to be collected
+        if (isset($annotations['method']['coversNothing'])) {
+            return false;
+        }
+
         // If there is at least one @covers annotation then
         // code coverage data needs to be collected
         if (isset($annotations['method']['covers'])) {
--- a/tests/_files/CoverageClassNothingTest.php
+++ b/tests/_files/CoverageClassNothingTest.php
@@ -1,0 +1,22 @@
+<?php declare(strict_types=1);
+/*
+ * This file is part of PHPUnit.
+ *
+ * (c) Sebastian Bergmann <sebastian@phpunit.de>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+use PHPUnit\Framework\TestCase;
+
+/**
+ * @coversNothing
+ */
+class CoverageClassNothingTest extends TestCase
+{
+    public function testSomething(): void
+    {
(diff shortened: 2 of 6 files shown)


andrewnicols:
Note: I see that in `shouldCoversAnnotationBeUsed`, the `@coversNothing` annotation has a higher precedence than `@covers` on the same method. There's nothing in the documentation to say which shoudl have higher precedent and unit tests pass with it either way.

I've currently modified it for the 'new' behaviour, but I'll swap it tomorrow (late here in Australia) unless I hear otherwise.

andrewnicols:
Is this eligible for backport to 7.5? If so, do I need to prepare a patch and raise a separate pull request?

Thanks

sebastianbergmann:
If you send a pull request for `7.5` then I will consider it, yes. I will, however, not invest time in backporting it myself. Sorry.

PR #3701, opened 2019-05-22
forceCoversAnnotations not respected in Util\Test::requiresCodeCoverageDataCollection()

andrewnicols:
| Q                   | A
| PHPUnit version     | master
| PHP version         | 7.3.5
| Installation Method | Git

If you have specified the `forceCoversAnnotation` setting in your configuration, I would expect this to influence the result of `requiresCodeCoverageDataCollection()` such that the Coverage Driver is only started if any coverage is required.

It appears that the configuration value is only used at the end of the run when generating the coverage report, and is ignored when starting the driver.

Related to #3697/#3699.

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -87,6 +87,11 @@ abstract class TestCase extends Assert implements SelfDescribing, Test
     /**
      * @var bool
      */
+    protected $forceCoversAnnotation = true;
+
+    /**
+     * @var bool
+     */
     private $runClassInSeparateProcess;
 
     /**
@@ -1044,6 +1049,16 @@ abstract class TestCase extends Assert implements SelfDescribing, Test
     public function isInIsolation(): bool
     {
         return $this->inIsolation;
+    }
+
+    public function setForceCoversAnnotation(bool $forceCoversAnnotation): void
+    {
+        $this->forceCoversAnnotation = $forceCoversAnnotation;
+    }
+
+    public function getForceCoversAnnotation(): bool
+    {
+        return $this->forceCoversAnnotation;
     }
 
     public function getResult()
--- a/src/Framework/TestSuite.php
+++ b/src/Framework/TestSuite.php
@@ -41,6 +41,11 @@ class TestSuite implements \IteratorAggregate, SelfDescribing, Test
     protected $runTestInSeparateProcess = false;
 
     /**
+     * @var bool
+     */
+    protected $forceCoversAnnotation = false;
+
+    /**
      * The name of the test suite.
      *
      * @var string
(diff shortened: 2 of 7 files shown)


andrewnicols:
I feel that this change makes sense from a coverage performance perspective. When `forceCoversAnnotation` is set, coverage reports should require the `@covers` annotation, and therefore it only makes sense to start the coverage driver in this situation.

I'm not sure whether I have pulled the `forceCoversAnnotation` configuration correctly into the TestCase, but this did seem to be the most correct way of doing so from my understanding of the surrounding code.

Similar to the change in 3697/3699, I have modified the existing `testCanSkipCoverage` test. That test was creating a new instance of the testcase without specifying the name of the test, and therefore was missing the method annotations. As a result, it was only testing whether the `@covers` and `@coversNothing` was applied to the class, and not checking overrides on individual test methods.

I've updated those tests to check the `testSomething` test in each of those examples, and corrected the outcome.
Each of the existing tests did explicitly specify a `@covers` annotation on either the class, or the test method, so I have also added a new test which does not specify any annotations to test this properly.

Note: This does not affect the `getLinesToBeCovered` function or associated tests because that is only called after coverage has been generated, and is only passed the test class and name, and only inspects the annotations. It is not aware of the TestRunner configuration at all.

I suspect that it will be easier to merge this change before 3697/3699 and have that issue build upon the changes here.

sebastianbergmann:

andrewnicols:
Hi @sebastianbergmann ,

Thanks for your time to review this. It'd be good to understand your rationale for future reference, and to see if there is something else we can do instead. The `@coversNothing` change in #3697/#3699 should really help with our issues, but this would probably have been a more future-proof fix.

Thanks again for your time and dedication to this project.

sebastianbergmann:
The `@forceCoversAnnotations` should not have anything to with whether or not code coverage data is collected.

andrewnicols:
Thanks for the quick explanation.

The documentation for `forceCoversAnnotation` states:
```
Code Coverage will only be recorded for tests that use the @covers annotation documented in @covers.
```

Therefore I'd argue that it the presence of the configuration setting is entirely relevant to whether code coverage data should be collected.

With the setting enabled, coverage will not be generated regardless of whether or not it is collected, so what sense does it make to collect it?

I see in php-code-coverage that it is possible to ignore the `forceCodeCoverage` configuration when calling `stop()` on the coverage tool, but we don't ever pass this argument from phpunit when calling stop. If this were to change in future, then it would make sense to not ignore the coverage generation at the point that such a change were made.

Thanks again for your time.

andrewnicols:
Hi @sebastianbergmann,

Don't mean to bug you but I realised you may not have notifications turned on and therefore may not have seen my previous comment as I forgot to cc you.

andrewnicols:
Hi @sebastianbergmann ,

As I say, it would be extremely useful to have rationale as to this decision. I feel that I have justified why it makes sense to make this change, citing project documentation which states that this is meant to be the case, and providing rationale as to why it would make sense.

It is pointless to generate coverage data only to discard it completely unused.

PR #3708, opened 2019-05-26
Built-in assertion and mock type definitions

Ocramius:
As discussed briefly with @sebastianbergmann, adding a few type declarations to the codebase.

The final aim (not sure if fully to be reached in this PR) is to:

 * [x] identify redundant assertions - `self::assertTrue(true);` should likely become some sort of type check error ~~out of current patch scope~~
 * [x] identify impossible type errors, such as `if (! $a instanceof B) { self::fail(); } typeMismatch($a /* requires B */);`
 * [x] allow type-checking `MockObject` instances wherever an explicit class name is given
 * [x] make the build green

Notes:

 * [x] Prophecy is not yet covered - patches need to be applied upstream in order to make a `Prophecy` instance templated
 * [x] I don't yet have a strategy for testing that the added type checks do detect (or not) type errors: thinking of using `.phpt` files

Upstream issues currently causing false positives here:

 * [x] https://github.com/vimeo/psalm/issues/1692
 * [x] https://github.com/vimeo/psalm/issues/1691 (won't fix - let's drop trait types from the patch)
 * [x] https://github.com/vimeo/psalm/issues/1743 ~~avoided via different test design~~
 * [x] https://github.com/vimeo/psalm/issues/1744


composer.json
```diff
@@ -49,7 +49,8 @@
         "sebastian/version": "^2.0.1"
     },
     "require-dev": {
-        "ext-PDO": "*"
+        "ext-PDO": "*",
+        "vimeo/psalm": "dev-master@dev"
```
Ocramius: Not sure what strategy is preferred here in order to verify type checks as part of the CI pipeline. @sebastianbergmann how do you prefer the dev dependency to be used?

src/Framework/Assert.php
```diff
@@ -819,6 +819,8 @@ public static function assertAttributeNotEquals($expected, string $actualAttribu
      *
      * @throws ExpectationFailedException
      * @throws \SebastianBergmann\RecursionContext\InvalidArgumentException
+     *
+     * @psalm-assert TEmptyMixed $actual
```
Ocramius: All of this file will need at least 2 tests for each touched assertion, to be picked from the following:

 * [ ] positive test (no issues)
 * [ ] negative test (impossible type)
 * [ ] redundant test (assertion always passes)

src/Framework/MockObject/MockObject.php
```diff
@@ -28,4 +28,8 @@ public function __phpunit_hasMatchers(): bool;
     public function __phpunit_setReturnValueGeneration(bool $returnValueGeneration): void;
 
     public function expects(Invocation $matcher): BuilderInvocationMocker;
+
+    // @TODO get rid of this - required by psalm-only right now - bug in parsing @method?
+    // @TODO this addition is a BC break and needs to be reverted before merge.
+    public function method($constraint) : BuilderInvocationMocker;
```
Ocramius: Need https://github.com/vimeo/psalm/issues/1692 to get rid of this

tests/static-analysis/TestUsingMocks.php
```diff
@@ -0,0 +1,113 @@
+<?php declare(strict_types=1);
```
Ocramius: Will need to add a CI step to verify this. Also unsure if we should split it into multiple type-check `.phpt` scenarios

Ocramius:
@weirdan can you give this a skim, since you worked on a lot of this stuff?

src/Framework/Assert.php
muglug: I think this should be `@psalm-assert empty $actual`

composer.json
muglug: PHPUnit currently uses a Psalm binary in the `tools` directory, installed via phive and then committed. When I've fixed stuff in Psalm and it's in a release it'll presumably get ported over.

src/Framework/Assert.php
Ocramius: I'll flesh these out one by one with tests as I get to work on them. I used the various `T*` from the plugin documentation, so that may be incorrect.

muglug: Oh that's _definitely_ incorrect. Where's that?

Ocramius: https://github.com/vimeo/psalm/blob/3.3.0/docs/plugins_type_system.md

I guess that's post-parsing types

muglug: Yeah, it's complicated.

`@psalm-assert` takes anything that the `Reconciler` class can understand.

It's not documented anywhere, but it takes any unadorned type (e.g. `int`, `Foo`, `scalar`), and the special case `empty`.

It also takes negations of the same e.g. `!int`, `!Foo`, `!numeric`, `!empty` etc.

composer.json
sebastianbergmann: Psalm is a tool. I strongly believe that tools must be handled separately from runtime dependencies. Therefore I do not manage tool dependencies with Composer but with Phive. `tools/psalm` is what I use locally and on Travis.

weirdan:
>  I don't yet have a strategy for testing that the added type checks do detect (or not) type errors: thinking of using .phpt files

I've used Codeception gherkin tests (with custom modules) in plugins, however using it here would introduce potential dependency problems (Codeception depends on PHPUnit). PHPT runner extracts the code from `.phpt` file and runs it with `php` - you would have to find a way to pass this code to Psalm instead.

Another point to consider is what Psalm's target version is going to be. Psalm is moving fast, and the code used to be passing with flying colors yesterday not necessarily will pass today. Error messages also change sometimes. Unfortunately composer does not allow specifying version constraints on suggested packages, so there's no good way to declare peer dependency constraints (as npm calls it).

Slamdunk:
> The final aim (not sure if fully to be reached in this PR) is to:
> 
> * identify redundant assertions - `self::assertTrue(true);` should likely become some sort of type check error
> * identify impossible type errors, such as `if (! $a instanceof B) { self::fail(); } typeMismatch($a /* requires B */);`
> * allow type-checking `MockObject` instances wherever an explicit class name is given

I'm sorry if I ask dumb questions, but what are the benefits of coupling this goals with a specific tool? Until now I've successfully used @phpstan [phpunit extension](https://github.com/phpstan/phpstan-phpunit), which requires zero modification of this library and, AFAIK, provides the same functionalities.

Is this intended to be used with [Roave/you-are-using-it-wrong](https://github.com/Roave/you-are-using-it-wrong)? If yes, again, what would be the benefits? `you-are-using-it-wrong` only do the checks during installation, while psalm and phpstan do the works every run.

@vimeo Psalm is a great tool, but I'm worried when I see tool-specific edits to libraries, look such a huge burden from a maintainer perspective, even if they are just docblocks :worried:

Ocramius:
@Slamdunk no dependencies to be added here: this would just make tyoe checks work OOTB (without any plugins) downstream for those using psalm. Phpstan is working on supporting these annotations too, but isn't quite there yet. I strongly believe that no plugin should be needed for what is to be provided by clear declarations in the upstream sources.

Similar discussion to https://github.com/doctrine/collections/pull/177

composer.json
Ocramius: Works for me 👍

src/Framework/Assert.php
```diff
@@ -1597,6 +1617,10 @@ public static function assertAttributeSame($expected, string $actualAttributeNam
      *
      * @throws ExpectationFailedException
      * @throws \SebastianBergmann\RecursionContext\InvalidArgumentException
+     *
+     * @psalm-template ExpectedType
+     * @psalm-param ExpectedType $expected
+     * @psalm-assert !=ExpectedType $actual
```
bdsl:
```suggestion
```
I think this might be wrong - NotSame doesn't mean that the types have to be different.

src/Framework/TestCase.php
```diff
@@ -1645,6 +1688,10 @@ protected function getObjectForTrait($traitName, array $arguments = [], $traitCl
      * @throws \Prophecy\Exception\Doubler\ClassNotFoundException
      * @throws \Prophecy\Exception\Doubler\DoubleException
      * @throws \Prophecy\Exception\Doubler\InterfaceNotFoundException
+     *
+     * @psalm-template RealInstanceType of object
+     * @psalm-param class-string<RealInstanceType>|null $classOrInterface
+     * @psalm-return ObjectProphecy<RealInstanceType>
```
bdsl: To be useful I suppose this needs to have the @template tag annotation added to https://github.com/phpspec/prophecy/blob/7e272180527c34a97680de85eb5aba0847a664e0/src/Prophecy/Prophecy/ObjectProphecy.php#L30 (or a stub for it)

src/Framework/Assert.php
muglug: It doesn’t actually seem to have an effect. This should be a `RedundantCondition`: https://psalm.dev/8be275b3ec

muglug: Oh it only complains when both are known string literals.

src/Framework/MockObject/MockObject.php
muglug: Done

src/Framework/Assert.php
muglug: The above code is now behaves as expected.

@bdsl the `!=ExpectedType` there means that Psalm expects the exact values to be different. In the case that it can determine the types are always different, it will emit a `RedundantCondition` or `RedundantConditionGivenDocblockType`, depending on the input.

ondrejmirtes:
I can get behind generics, but I don't understand why PHPUnit's code needs to be littered with `@psalm-assert` annotations. I prefer using external extension for this... Adding annotations to PHPUnit so that Psalm can work out of the box seems to me like a dependency direction with an arrow on the wrong side...

Ocramius:
> I don't understand why PHPUnit's code needs to be littered with `@psalm-assert` annotations. 

Sources dictate types, not external runtime extensions. I've already explained it at https://github.com/sebastianbergmann/phpunit/pull/3708#issuecomment-496105405

Besides that, no explicit dependency, since prefixed annotations are ignored by tooling ;-)

ondrejmirtes:
Also, PHPStan plans to support generics with exactly this syntax (`@template` etc. with optional vendor prefix) because it's pretty easy, but this syntax is too obscure and proprietary:

* `@psalm-assert TEmptyMixed $actual`
* `@psalm-assert !=ExpectedType`

If I didn't know what `assertNotSame` actually does, I wouldn't be able to figure out what the annotation does. (PHPStan's phpstan-phpunit extension already understands these method calls.)

ondrejmirtes:
I get it it's not a technical dependency, but if only Psalm can (and will ever benefit) from these annotations, should they really be maintained as part of PHPUnit source code?

Ocramius:
I think so, yes. It is a win-win from a technical/engineering perspective:

 * works out of the box
 * no runtime overhead
 * no added dependencies
 * ignored by tooling that is unaware of these
 * immediate technical benefits for downstream projects

muglug:
@ondrejmirtes if you build equivalent support for `@phpstan-assert`, I promise you Psalm will support that too.

I absolutely believe they should be part of the codebase, though. TypeScript has the concept of [type predicates](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards) which essentially do the same thing.

ondrejmirtes:
Yeah, I can also do that (it would be doable even with the current public API - TypeSpecifyingExtension - anyone could build an extension that reads and interprets `@psalm-assert`), but the current syntax is really weird - I guess that names like `TEmptyMixed` are internal Psalm class names, or what is it? Is there a list of all possible asserts-after-method-call?

muglug:
`TEmptyMixed` is _not_ accepted docblock syntax anywhere. `@psalm-assert empty $foo` is.

Ocramius:
I've mentioned that in https://github.com/sebastianbergmann/phpunit/pull/3708#discussion_r287616283 - will get rid of most of these internal specs as I go forward and write tests ;-)

ondrejmirtes:
OK :)

src/Framework/Assert.php
Ocramius: After inspecting this further, I only attacked the problem from a "correct type" point of view.

src/Framework/Assert.php
Ocramius: > @bdsl the `!=ExpectedType` there means that Psalm expects the exact values to be different. 

I got rid of this hint: as much as it may be about exact equality, I will keep myself on types-only for now. Please send a PR (against my branch) if you have a way to specifically test this.

Ocramius:
@muglug you might be interested in the current failures. Gonna try isolating them into single issues:

```
INFO: PropertyNotSetInConstructor - tests/static-analysis/TestUsingMocks.php:26:13 - Property PHPUnit\Framework\TestCase::$backupGlobals is not defined in constructor of PHPUnit\StaticAnalysis\TestUsingMocks or in any methods called in the constructor
final class TestUsingMocks extends TestCase

ERROR: InvalidReturnType - tests/static-analysis/happy-path/assert-empty.php:7:13 - The declared return type 'false' for PHPUnit\Framework\StaticAnalysis\HappyPath\consume is incorrect, got 'bool'
/** @return false */

ERROR: DocblockTypeContradiction - tests/static-analysis/happy-path/assert-empty.php:12:12 - string(a non-empty string) does not contain string()|string(0)
    return $value === 'a non-empty string';

ERROR: InvalidReturnStatement - tests/static-analysis/happy-path/assert-empty.php:12:12 - The type 'bool' does not match the declared return type 'false' for PHPUnit\Framework\StaticAnalysis\HappyPath\consume
    return $value === 'a non-empty string';

ERROR: InvalidReturnType - tests/static-analysis/happy-path/assert-is-not-scalar.php:8:28 - The declared return type 'object' for PHPUnit\Framework\StaticAnalysis\HappyPath\AssertIsNotScalar\consume is incorrect, got 'scalar|object'
function consume($value) : object

ERROR: InvalidReturnStatement - tests/static-analysis/happy-path/assert-is-not-scalar.php:12:12 - The type 'scalar|object' does not match the declared return type 'object' for PHPUnit\Framework\StaticAnalysis\HappyPath\AssertIsNotScalar\consume
    return $value;

ERROR: MismatchingDocblockReturnType - tests/static-analysis/happy-path/assert-is-scalar.php:10:18 - Docblock has incorrect return type 'scalar', should be 'string'
 * @psalm-return scalar

ERROR: MixedInferredReturnType - tests/static-analysis/happy-path/assert-is-scalar.php:10:18 - Could not verify return type 'scalar' for PHPUnit\Framework\StaticAnalysis\HappyPath\AssertIsScalar\consume
 * @psalm-return scalar

ERROR: InvalidReturnType - tests/static-analysis/happy-path/assert-not-empty.php:7:13 - The declared return type 'false' for PHPUnit\Framework\StaticAnalysis\HappyPath\AssertNotEmpty\consume is incorrect, got 'bool'
/** @return false */

ERROR: InvalidReturnStatement - tests/static-analysis/happy-path/assert-not-empty.php:12:12 - The type 'bool' does not match the declared return type 'false' for PHPUnit\Framework\StaticAnalysis\HappyPath\AssertNotEmpty\consume
    return $value === '';

------------------------------
9 errors found
------------------------------
1 other issues found.
You can hide them with --show-info=false
------------------------------

Checks took 0.26 seconds and used 95.430MB of memory
Psalm was able to infer types for 100% of the codebase

```

Ocramius:
The `scalar` issue was incorrect in my own definitions.

Here's an updated report:

```

INFO: PropertyNotSetInConstructor - tests/static-analysis/TestUsingMocks.php:26:13 - Property PHPUnit\Framework\TestCase::$backupGlobals is not defined in constructor of PHPUnit\StaticAnalysis\TestUsingMocks or in any methods called in the constructor
final class TestUsingMocks extends TestCase

ERROR: InvalidReturnType - tests/static-analysis/happy-path/assert-empty.php:7:13 - The declared return type 'false' for PHPUnit\Framework\StaticAnalysis\HappyPath\consume is incorrect, got 'bool'
/** @return false */

ERROR: DocblockTypeContradiction - tests/static-analysis/happy-path/assert-empty.php:12:12 - string(a non-empty string) does not contain string()|string(0)
    return $value === 'a non-empty string';

ERROR: InvalidReturnStatement - tests/static-analysis/happy-path/assert-empty.php:12:12 - The type 'bool' does not match the declared return type 'false' for PHPUnit\Framework\StaticAnalysis\HappyPath\consume
    return $value === 'a non-empty string';

ERROR: InvalidReturnType - tests/static-analysis/happy-path/assert-not-empty.php:7:13 - The declared return type 'false' for PHPUnit\Framework\StaticAnalysis\HappyPath\AssertNotEmpty\consume is incorrect, got 'bool'
/** @return false */

ERROR: InvalidReturnStatement - tests/static-analysis/happy-path/assert-not-empty.php:12:12 - The type 'bool' does not match the declared return type 'false' for PHPUnit\Framework\StaticAnalysis\HappyPath\AssertNotEmpty\consume
    return $value === '';

------------------------------
5 errors found
------------------------------
1 other issues found.
You can hide them with --show-info=false
------------------------------

Checks took 0.40 seconds and used 95.418MB of memory
Psalm was able to infer types for 100% of the codebase
```

Ocramius:
Fixed all \o/ 

```

INFO: PropertyNotSetInConstructor - tests/static-analysis/TestUsingMocks.php:26:13 - Property PHPUnit\Framework\TestCase::$backupGlobals is not defined in constructor of PHPUnit\StaticAnalysis\TestUsingMocks or in any methods called in the constructor
final class TestUsingMocks extends TestCase

------------------------------
No errors found!
------------------------------
1 other issues found.
You can hide them with --show-info=false
------------------------------

Checks took 0.36 seconds and used 95.410MB of memory
Psalm was able to infer types for 100% of the codebase
```

Ocramius:
@sebastianbergmann how do you feel about fixing `PHPUnit\Framework\TestCase#$backupGlobals` not having a default value? Should I change its documented type to `bool|null`, assign it `false` by default, or leave it alone?

src/Framework/TestCase.php
```diff
@@ -1411,6 +1430,10 @@ protected function createPartialMock($originalClassName, array $methods): MockOb
      * Returns a test proxy for the specified class.
      *
      * @throws Exception
+     *
+     * @psalm-template RealInstanceType of object
+     * @psalm-param class-string<RealInstanceType>|string[] $originalClassName
```
psalm-shepherd: Parameter $originalClassName has wrong type 'array<array-key, string>|class-string<RealInstanceType>', should be 'string'
```
     * @psalm-param class-string<RealInstanceType>|string[] $originalClassName
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

psalm-shepherd:
Psalm also found errors in other files

```
src/Framework/Assert.php:1701:13 - Argument 1 of PHPUnit\Framework\Assert::assertinstanceof expects class-string<ExpectedType>, parent type string provided
src/Framework/Assert.php:1750:13 - Argument 1 of PHPUnit\Framework\Assert::assertnotinstanceof expects class-string<ExpectedType>, parent type string provided
src/Framework/TestCase.php:1522:17 - Argument 1 of PHPUnit\Framework\MockObject\Generator::getmockforabstractclass expects string, possibly different type array<array-key, string>|class-string<RealInstanceType> provided
```

src/Framework/TestCase.php
muglug: Remove `|string[]`

psalm-shepherd:
Psalm didn’t find any errors!

src/Framework/TestCase.php
```diff
@@ -597,6 +597,9 @@ public function getExpectedExceptionMessageRegExp(): ?string
         return $this->expectedExceptionMessageRegExp;
     }
 
+    /**
+     * @psalm-param class-string<|Throwable> $exception
```
weirdan: Looks broken. Did you mean `class-string<Throwable>` ?

Ocramius: Indeed: this is supposed to be a `\`

Ocramius:
Green!

@sebastianbergmann up to you now - need also opinion on https://github.com/sebastianbergmann/phpunit/pull/3708#issuecomment-499251673

tests/static-analysis/config.xml
```diff
@@ -0,0 +1,51 @@
...
+>
+    <projectFiles>
+        <directory name="tests/static-analysis" />
+        <ignoreFiles>
+            <directory name="vendor" />
+        </ignoreFiles>
+    </projectFiles>
+
+    <issueHandlers>
+        <LessSpecificReturnType errorLevel="error" />
```
muglug: you can remove all of these - `error` is the default level

muglug (approved):
Looks awesome!

sebastianbergmann:
I would leave `PHPUnit\Framework\TestCase#$backupGlobals` alone for now.

sebastianbergmann:
Merged, thanks!

Ocramius:
\o/

PR #3711, opened 2019-05-29
Unique Log Files

Goahnary:
This allows the user to make a unique log file for every run of phpunit by prepending the log file name with a timestamp and a random string (to assure tests run within a second of each other are unique).

This would be especially useful when running tests in parallel as the current way will overwrite the log file so that only the last test run will be logged.

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -15,10 +15,6 @@
         </testsuite>
     </testsuites>
 
-    <logging>
-        <log type="testdox-text" target="./tests/logs/testdox.log" timestamps="true"/>
-    </logging>
-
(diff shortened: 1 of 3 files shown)


sebastianbergmann:
This branch has conflicts that must be resolved. I would consider a new pull request against `master` that implements this. Please introduce a configuration attribuve in `phpunit.xml` as well as a CLI option. Do not perform any work based on those settings in the code that is responsible for parsing the XML configuration file or the CLI options. The CLI test runner's code should look at the configuration option and then {pre|suf]fix the logfile names.

PR #3717, opened 2019-06-07
Removed `array ...$arguments` declaration: arguments are `mixed` by design

Ocramius:
Without this, the interpreted signature is `with(array ...$arguments)`, which makes `$arguments` an `array[]`.


sebastianbergmann:
Merged manually, thanks.

PR #3723, opened 2019-06-14
Bugfix: Fix unescaped dash in character group in regex

mvorisek:


sebastianbergmann:
Can you please elaborate what you intend to fix with this change? Thanks!

mvorisek:
PHP 7.3 uses newer POSIX regex engine and regexes like `~[\d-_]~` became invalid. The fixed regex may became invalid someday too, so I fixed it. `-` as character is safe to use only as regular range operator or when it is placed the most left or the most right in the character group.

PR #3731, opened 2019-06-24
SegFault in set upping test in \PHPUnit\Framework\TestSuite:574

JacksonGH:
When I try to test my project I have SegFault. I found the reason for this behavior, it was unset array iteracted by foreach.


sebastianbergmann:

PR #3738, opened 2019-06-25
Allow using associative arrays in data providers

bkrukowski:
Let's consider the following test:

```php
    /**
     * @dataProvider additionProvider
     *
     * @param int $a
     * @param int $b
     * @param int $expected
     */
    public function testAddition(int $a, int $b, int $expected): void
    {
        $this->assertSame($expected, $a + $b);
    }

    public function additionProvider(): array
    {
        return [
            'first' => ['a' => 2, 'b' => 2, 'expected' => 4],
            'second' => ['expected' => 4, 'a' => 2, 'b' => 2],
        ];
    }
```

PHPUnit will return:

```
1) MyTest::testAddition with data set "second" (4, 2, 2)
Failed asserting that 6 is identical to 2.
```

For first data set test is green. For second data set test is red. Looks weird, right? Would be nice to have official support for associative arrays in data providers. What do you think?

### DIFF
--- a/src/Util/Test.php
+++ b/src/Util/Test.php
@@ -528,6 +528,7 @@ final class Test
                     throw new Exception(
                         \sprintf(
                             'Data set %s is invalid: %s.',
+                            \is_int($key) ? '#' . $key : '"' . $key . '"',
                             $exception->getMessage()
                         ),
                         0,
--- a/src/Util/Test.php
+++ b/src/Util/Test.php
@@ -14,6 +14,7 @@ use PHPUnit\Framework\Assert;
 use PHPUnit\Framework\CodeCoverageException;
 use PHPUnit\Framework\InvalidCoversTargetException;
 use PHPUnit\Framework\InvalidDataProviderException;
+use PHPUnit\Framework\InvalidDataSetException;
 use PHPUnit\Framework\SelfDescribing;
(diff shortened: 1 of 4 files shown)


src/Util/Test.php
```diff
@@ -520,6 +521,20 @@ public static function getProvidedData(string $className, string $methodName): ?
                         )
                     );
                 }
+
+                try {
+                    $data[$key] = DataSetTransformer::transform($reflector, $value);
```
psalm-shepherd: Cannot access array value on non-array variable $data[$key] of type iterable<mixed, mixed>
```
                    $data[$key] = DataSetTransformer::transform($reflector, $value);
                    ^^^^^^^^^^^
```

psalm-shepherd:
Psalm found some errors

psalm-shepherd:
Psalm didn’t find any errors!

sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it as it would break backward compatibility.

bkrukowski:
@sebastianbergmann can we consider having this feature in next major version?

sebastianbergmann:
No, I do not think that the drawbacks (every existing test that uses a data provider needs to be reviewed and potentially updated) outweigh what little benefit would be gained. Sorry.

PR #3747, opened 2019-07-05
Use strlen to validate non-empty string for StringStartsWith

MasterOdin:
Fixes regression caused by 91936e4f7f1bbaca710da2c07be0e8bab32a5f3e which would cause tests to fail if the string to test for was just `"0"` due to how PHP handles truthy-ness in `empty`.


sebastianbergmann:
Thanks!

PR #3781, opened 2019-08-02
onlyMethods/addMethods - blank arrays fix

DFoxinator:


DFoxinator:
@sebastianbergmann I was implementing these into a large test suite manage, and I came across this issue/behavior that I think is broken. It doesn't seem like expected behavior that onlyMethods([]) would result in all methods in the class being mocked. I think the correct behavior is no methods should be mocked because you're saying mock onlyMethods with a blank list. This functionality becomes more critical for when createPartialMock uses onlyMethods instead of setMethods. Because as-is, the behavior of mocking no methods on the class would break. 

Please let me know what you think.

PR #3787, opened 2019-08-07
Cast exception message to string

jderusse:
Since PHPUnit 8, the `PHPUnit\Framework\Constraint\ExceptionMessage` contains a `strict_types=1` annotation. 
This triggers an `TypeError` when exception's message is not a string, but a class with a `__toString` method like symfony/symfony does in `AutowiringFailedException`.

This PR cast the exception message to string


jderusse:
close. Wrong target branche

PR #3791, opened 2019-08-16
Use correct article ("a" or "an") in InvalidArgumentHelper

Tarasovych:
Development
- [x] code complete
- [x] fix PHPUnit self-tests locally

Create pull request
- [x] add title
- [x] add description of change
- [x] request review

This PR might change InvalidArgumentHelper exception message, especially article before `$type`.
If `$type` starts with vowel, "an" will be used, "a" otherwise.

### DIFF
--- a/src/Util/InvalidArgumentHelper.php
+++ b/src/Util/InvalidArgumentHelper.php
@@ -27,7 +27,7 @@ final class InvalidArgumentHelper
                 $value !== null ? ' (' . \gettype($value) . '#' . $value . ')' : ' (No Value) ',
                 $stack[1]['class'],
                 $stack[1]['function'],
-                \in_array($type[0], ['a', 'e', 'i', 'o', 'u']) ? 'an' : 'a',
+                \in_array(\lcfirst($type)[0], ['a', 'e', 'i', 'o', 'u']) ? 'an' : 'a',
                 $type
             )
         );
--- a/src/Util/InvalidArgumentHelper.php
+++ b/src/Util/InvalidArgumentHelper.php
@@ -27,7 +27,7 @@ final class InvalidArgumentHelper
                 $value !== null ? ' (' . \gettype($value) . '#' . $value . ')' : ' (No Value) ',
                 $stack[1]['class'],
                 $stack[1]['function'],
-                \in_array($type[0], ['a', 'e', 'i', 'o', 'u']) ? 'a' : 'an',
+                \in_array($type[0], ['a', 'e', 'i', 'o', 'u']) ? 'an' : 'a',
                 $type
             )
         );
--- a/src/Util/InvalidArgumentHelper.php
+++ b/src/Util/InvalidArgumentHelper.php
@@ -22,11 +22,12 @@ final class InvalidArgumentHelper
 
         return new Exception(
             \sprintf(
-                'Argument #%d%sof %s::%s() must be a %s',
+                'Argument #%d%sof %s::%s() must be %s %s',
                 $argument,
                 $value !== null ? ' (' . \gettype($value) . '#' . $value . ')' : ' (No Value) ',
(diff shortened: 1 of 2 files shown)


epdenouden:
Hi @Tarasovych!

This is indeed a nice enhancement to start working on PHPUnit! All you need to do now is add tests, especially since you are working on a... test framework. 😬 

Tip: run the tests locally before doing commits and especially before you make a pull request. This change was small enough to quickly review and comment on. Like PHPUnit many projects rely on volunteer work from developers and might simply reject proposed changes if they do not come with a good description and tests.

Do not hestiate to ask for help if you feel stuck.

Tarasovych:
@epdenouden Hi! I think, this PR is ready for review now.

epdenouden:
Excellent!! I will have a good look this tomorrow, thanks for your quick work. :)

tests/unit/Framework/AssertTest.php
```diff
@@ -1524,6 +1524,19 @@ public function testAssertCountThrowsExceptionIfElementIsNotCountable(): void
         $this->fail();
     }
 
+    public function testInvalidArgumentExceptionUseCorrectArticleInErrorMessage(): void
```
epdenouden: minor niggle: Uses instead of Use, third person singular

tests/unit/Framework/AssertTest.php
```diff
@@ -1524,6 +1524,19 @@ public function testAssertCountThrowsExceptionIfElementIsNotCountable(): void
         $this->fail();
     }
```
epdenouden: Are you already familiar with the @testdox annotation? It allows you to format the output. Normally the TestDox output formatter will break up CamelCase like in "InvalidArgumentException".

Example in this case:
/** @testdox InvalidArgumentException uses correct article for variable type */

src/Util/InvalidArgumentHelper.php
```diff
@@ -22,11 +22,12 @@ public static function factory(int $argument, string $type, $value = null): Exce
 
         return new Exception(
             \sprintf(
-                'Argument #%d%sof %s::%s() must be a %s',
+                'Argument #%d%sof %s::%s() must be %s %s',
                 $argument,
                 $value !== null ? ' (' . \gettype($value) . '#' . $value . ')' : ' (No Value) ',
                 $stack[1]['class'],
                 $stack[1]['function'],
+                \in_array($type[0], ['a', 'e', 'i', 'o', 'u']) ? 'an' : 'a',
```
epdenouden: Can you check if `$type` is already guaranteed lowercase here? Just to be sure and prevent possible edge cases like `must be a Iterator` or `must be a Object`.

epdenouden (approved):
Nice work, just a few very minor niggles :)

epdenouden (changes requested):
So many buttons

tests/unit/Framework/AssertTest.php
Tarasovych: > Uses instead of Use

Is it a requirement for all test methods? If yes, it'd be nice if `php-cs-fixer` will inform about issues like this. Or maybe a note should be added to the CONTRIBUTING.md?

epdenouden: Thanks for bringing that up, I don't think the contribution guidelines are that detailed. For me this was also a nice-to-have, it doesn't affect the functionality.

At the next Code Sprint meetup early September I want to brainstorm ideas to make contributing easier.

epdenouden (approved):

epdenouden:
@Tarasovych Thanks for the quick turnaround and the attention to detail. Feels good working like this. 👍 

@sebastianbergmann Pull looks fine to me. Let me know if you have any review-comments regarding my review process.

PR #3797, opened 2019-08-27
Add support for multiple whitelisted directories in CLI args

rask:
Previously inserting multiple `--whitelist` args would result in the
last one being the only effective one.

New implementation aggregates them and adds them to the code coverage
filtering one by one, allowing developers to use multiple directories
in whitelisting without relying on the XML configuration file.

This introduces no BC breaks, as the old behavior works in the same
way as always from the end user's perspective, internally an array
with a single item is used instead.

The way it can be used if this is merged:

    phpunit --coverage-text --whitelist=mydir1 --whitelist=mydir2/file123.php --whitelist=mydir3/file1.php ./testsuite

### DIFF
--- a/src/TextUI/Command.php
+++ b/src/TextUI/Command.php
@@ -724,7 +724,11 @@ class Command
                     break;
 
                 case '--whitelist':
-                    $this->arguments['whitelist'] = $option[1];
+                    if (!isset($this->arguments['whitelist'])) {
+                        $this->arguments['whitelist'] = [];
+                    }
+
+                    $this->arguments['whitelist'][] = $option[1];
 
                     break;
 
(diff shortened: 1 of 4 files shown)


sebastianbergmann:
Out of curiosity: why can you not rely on the configuration file?

rask:
A monorepo with various projects relying on collective code, meaning developers often run only a subset of tests to save time, processing resources, and nerves. Coverage collection whitelisting cannot be committed into source control in out setup, and having to juggle the whitelist in the XML file during testing and development is tedious at times (and people accidentally commit their whitelist configuration to upstream). Having a CLI interface to define multiple whitelisted files/directories would simplify workloads and test runs a bit.

sebastianbergmann:
Thank you for the explanation. I would probably address this with multiple XML configuration files but will, of course, consider this pull request.

PR #3800, opened 2019-08-28
Document replacement methods

greg0ire:
This PR is incomplete, because I could not find a suitable replacement for `startTestSuite` and `endTestSuite`. It might be `BeforeFirstTestHook::executeBeforeFirstTest` and `AfterLastTestHook::executeAfterLastTest` but I'm not sure of that, for 2 reasons:
1. These methods do not have a `TestSuite` argument.
2. The adapter establishes no such correspondance: https://github.com/sebastianbergmann/phpunit/blob/b61d4d78a2657f5c65a296ffc86944b473211c88/src/Runner/Hook/TestListenerAdapter.php#L133-L139

Please comment on how people should migrate away from these 2 methods.

### DIFF
--- a/src/Framework/TestListener.php
+++ b/src/Framework/TestListener.php
@@ -16,31 +16,43 @@ interface TestListener
 {
     /**
      * An error occurred.
+     *
+     * @deprecated Use `AfterTestErrorHook::executeAfterTestError` instead
      */
     public function addError(Test $test, \Throwable $t, float $time): void;
 
     /**
      * A warning occurred.
+     *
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
Merged manually, thanks.

greg0ire:
Thank you, any hint about my question above?

sebastianbergmann:
The concept of a "test suite" is one that should, and will, eventually vanish. Therefore there is no equivalent in the new API.

greg0ire:
Ok that makes sense, thanks for the heads up.

PR #3805, opened 2019-09-01
Fix a few new bugs Psalm found

muglug:
Around the use of `??` and a `@throws` docblock class that doesn’t exist.


sebastianbergmann:
Merged manually, thanks.

PR #3807, opened 2019-09-03
Match method name in MockBuilder exception message

baspeeters:
Matched name of `MockBuilder->onlyMethods()` in its thrown exception's message and added coverage for the exception messages.


sebastianbergmann:
Merged manually, thanks.

PR #3810, opened 2019-09-06
#3120 implemented `TestCase#createStub()`

### LINKED ISSUE #3120: createStub(), createConfiguredStub(), createPartialStub()
PHPUnit's built-in test doubles functionality pre-dates the [de-facto standard classification for test doubles](http://xunitpatterns.com/Test%20Double.html) established by Gerard Meszaros in his book "xUnit Test Patterns".

> * A stub is used to "replace a real component on which the [system under test] depends so that the test has a control point for the indirect inputs of the [system under test]. This allows the test to force the [system under test] down paths it might not otherwise execute."
>
> * A mock is used "as an observation point that is used to verify the indirect outputs of the [system under test] as it is exercised. Typically, the [mock] also includes the functionality of a [stub] in that it must return values to the [system under test] if it hasn't already failed the tests but the emphasis is on the verification of the indirect outputs. Therefore, a [mock] is lot more than just a [stub] plus assertions; it is used a fundamentally different way."

The API of PHPUnit's built-in test doubles functionality for creating test doubles — originally `getMock()`, then `getMockBuilder()`, and now `createMock()` etc. — creates a stub. If, and only if, the `expects()` method is called on this stub to set up expectations then that stub becomes a mock.

The fact that the test double returned by `createMock()` is initially "just" a stub can be confusing, especially for developers new to PHPUnit in particular and/or unit testing and test doubles in general.

It should be explored whether the addition of `createStub()`, `createConfiguredStub()`, `createPartialStub()` methods is feasible. These methods should create test double objects that do not have the `expects()` method and therefore are stubs that cannot become mocks.

Ocramius:
Fixes #3120 

In practice, a `MockObject` is also a `Stub`, so we are just aliasing methods, but it is also true that the type-checker now does the heavy lifting for us. If people use `MockObject` methods after being told explicitly that a `Stub` is returned, that's a type error downstream.

Auto-completion now restricts methods correctly, so we're good :+1:

### DIFF
--- a/src/Framework/MockObject/Builder/InvocationMocker.php
+++ b/src/Framework/MockObject/Builder/InvocationMocker.php
@@ -29,7 +29,7 @@ use PHPUnit\Framework\MockObject\Stub\Stub;
 /**
  * @internal This class is not covered by the backward compatibility promise for PHPUnit
  */
-final class InvocationMocker implements MethodNameMatch
+final class InvocationMocker implements InvocationStubber, MethodNameMatch
 {
     /**
      * @var MatcherCollection
@@ -68,7 +68,7 @@ final class InvocationMocker implements MethodNameMatch
         return $this;
     }
 
-    public function will(Stub $stub): Identity
+    public function will(Stub $stub): self
     {
         $this->matcher->setStub($stub);
 
@@ -92,9 +92,7 @@ final class InvocationMocker implements MethodNameMatch
         return $this->will($stub);
     }
 
-    /**
-     * @param mixed $reference
-     */
+    /** {@inheritDoc} */
     public function willReturnReference(&$reference): self
     {
         $stub = new ReturnReference($reference);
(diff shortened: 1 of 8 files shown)


src/Framework/MockObject/Builder/InvocationStubber.php
```diff
@@ -0,0 +1,44 @@
...
+ */
+namespace PHPUnit\Framework\MockObject\Builder;
+
+use PHPUnit\Framework\MockObject\Stub\Stub;
+
+/** @internal This class is not covered by the backward compatibility promise for PHPUnit */
+interface InvocationStubber
+{
+    /**
+     * @TODO is "will" actually sensible in a stub context? Should a stub produce side-effects?
```
Ocramius: Requires discussion

src/Framework/MockObject/Builder/InvocationMocker.php
```diff
@@ -68,7 +68,7 @@ public function id($id): self
         return $this;
     }
 
-    public function will(Stub $stub): Identity
+    public function will(Stub $stub): self
```
Ocramius: See https://github.com/vimeo/psalm/issues/2102 about current failures. For now, will update baseline

PR #3811, opened 2019-09-06
Make .travis.yml warning-free

oliverklee:
- properly quote build stage names
- have a proper environment-value pair for the used tool in the stages


oliverklee:
This is the 7.5 version of the change. I'll create a separate version for 8.3/master in a minute.

Issue #3750, opened 2019-07-07
Whitelist for PHAR Scoping


sebastianbergmann:
[Currently](https://github.com/sebastianbergmann/phpunit/blob/ce8db891d630bf835d3d402374e01d60ee0db392/build/scoper.inc.php), we exclude
* `PHPUnit\*`,
* `SebastianBergmann\CodeCoverage\*`,
* `PharIo\*`,
* `PHP_Token*`,
* `Prophecy\*`,

from the [prefixing of all code bundled in PHAR distribution with random/unique namespace](https://github.com/sebastianbergmann/phpunit/issues/2015).

I would like to reduce the amount of code that does not get prefixed.

When it comes to `PHPUnit\*`, only classes, interfaces, and traits that are not annotated with `@internal` should be excluded. I will investigate whether this works by manually creating a list of these code units. In the long run, it would be great if php-scoper supported this natively.

I am not sure why `SebastianBergmann\CodeCoverage\*`, `PharIo\*`, and `PHP_Token*` are excluded. Maybe @sebastianfeldmann can provide some insight here? Excluding `PharIo\*`, for instance, means that @theseer and @sebastianheuer cannot use the scoped `phpunit.phar` for testing their `PharIo` code as intended.

With regard to `Prophecy\*` I am sure that some classes, interfaces, or traits of Prophecy need to be excluded (so that code complete and/our type annotations in test code continue to work).

sebastianbergmann:
#### `PharIo\*`
```patch
diff --git a/build/scoper.inc.php b/build/scoper.inc.php
index 9dd770d2a..827dbbece 100644
--- a/build/scoper.inc.php
+++ b/build/scoper.inc.php
@@ -12,7 +12,6 @@
     'whitelist' => [
         'PHPUnit\*',
         'SebastianBergmann\CodeCoverage\*',
-        'PharIo\*',
         'PHP_Token*',
         'Prophecy\*',
     ],
```
```
$ ant phar
.
.
.
    [phpab] phar archive '/usr/local/src/phpunit/build/phpunit-8.3-g20665d017.phar' generated.
```
```
$ cd /tmp
$ git clone git@github.com:phar-io/manifest.git
$ cd manifest
```
```patch
diff --git a/composer.json b/composer.json
index 9888b5a..4a4d92b 100644
--- a/composer.json
+++ b/composer.json
@@ -38,8 +38,5 @@
     "branch-alias": {
         "dev-master": "1.0.x-dev"
[...]

PR #3814, opened 2019-09-06
Remove PHP_Token and CodeCoverage from whitelist

sebastianfeldmann:
Since CodeCoverage and PHP_Token are not part of the public API of
PHPUnit they can safely be removed from the whitelist.

Issue #3750


localheinz (approved):
👍

Issue #3159, opened 2018-06-04
assertSame and assertEquals incorrectly pass on some floats


luispabon:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 7.1.5
| PHP version         | 7.2.5
| Installation Method | Composer

These two assertions pass:

```php
        self::assertEquals(0.06999999999, 0.07);
        self::assertSame(0.06999999999, 0.07);
```

They shouldn't.

keradus:
i will take `assertSame` for discussion
docs claims it uses `===` for comparision: https://github.com/sebastianbergmann/phpunit/blob/master/src/Framework/Constraint/IsIdentical.php#L18

it is not true, for numbers, it has epsilon counted in:
https://github.com/sebastianbergmann/phpunit/blob/master/src/Framework/Constraint/IsIdentical.php#L67

The thing is, that EPSILON can't be configured.

It was introduced in https://github.com/sebastianbergmann/phpunit/pull/273/files#diff-c228e9b2f01ec4a07f0f59042e40ded6R73 , 7 years ago...

Docs doesn't match expectations, can you clarify that @sebastianbergmann , it would help clearing the issue.

epdenouden:
!stale

luispabon:
Bump?

luispabon:
Bumpey

gabma:
This might look like a commun PHP issue with floats ? So not related to PHPUnit.
https://www.leaseweb.com/labs/2013/06/the-php-floating-point-precision-is-wrong-by-default/

luispabon:
Could be, but I don't think so:

```
~ php -a
Interactive mode enabled

php > $foo = 0.06999999999;
php > $bar = 0.07;
php > var_dump($foo == $bar);
bool(false)
php > var_dump($foo === $bar);
bool(false)
```

keradus:
Its a concrete part of code responsible for this on phpunit level

sebastianbergmann:
The more I think about it the more I come to the conclusion that two floating point numbers cannot (really) be compared for identity and that using `assertSame()` on `float` values does not make sense.

I do not remember why `private const EPSILON = 0.0000000001` is defined and used in the `IsIdentical` constraint that is used by `assertSame()`. To be honest, until I looked the code of `IsIdentical` today, I was not even aware of that.

[...]

PR #3817, opened 2019-09-06
Make epsilon for float comparison customizable - #3159

SebastianKull:
it would be possible to make the epsilon from float comparison customizable.
see #3159 

grafik

### DIFF
--- a/src/Framework/Constraint/IsIdentical.php
+++ b/src/Framework/Constraint/IsIdentical.php
@@ -55,10 +55,15 @@ final class IsIdentical extends Constraint
      */
     public function evaluate($other, string $description = '', bool $returnResult = false)
     {
+        $epsilon = self::EPSILON;
+        if (defined('CUSTOM_EPSILON')) {
+            $epsilon = CUSTOM_EPSILON;
+        }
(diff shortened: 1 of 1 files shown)


src/Framework/Constraint/IsIdentical.php
```diff
@@ -55,10 +55,15 @@ public function __construct($value)
      */
     public function evaluate($other, string $description = '', bool $returnResult = false)
     {
+        $epsilon = self::EPSILON;
+        if (defined('CUSTOM_EPSILON')) {
```
Ocramius: What about defining `EPSILON` conditionally in a namespace, and always having it declared?

Basides that, while the global state is immutable, I'm not sure I like the approach. What alternatives (if any) do we have?

Ocramius (changes requested):

sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it. See [https://github.com/sebastianbergmann/phpunit/issues/3159#issuecomment-528908608](here) for details.

PR #3819, opened 2019-09-06
Fix: Rename job

localheinz:
This PR

* [x] renames a previously added job for GitHub actions


localheinz:

PR #3820, opened 2019-09-06
Enhancement: Run psalm using GitHub actions

localheinz:
This PR

* [x] runs `vimeo/psalm` (as installed with `phive` to `/tools`) using GitHub actions

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,3 +18,25 @@ jobs:
         uses: docker://php:7.3-cli
         with:
           args: ./tools/php-cs-fixer fix --diff-format=udiff --dry-run --show-progress=dots --using-cache=no --verbose
+
+  type-checker:
+    name: Type Checker
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@master
+
+      - name: Update dependencies with composer
+        uses: docker://composer:1.9.0
+        with:
+          args: composer update --no-interaction --no-ansi --no-progress --no-suggest
+
+      - name: Run vimeo/psalm on public API
+        uses: docker://php:7.3-cli
+        with:
+          args: ./tools/psalm --config=.psalm/static-analysis.xml --no-progress --show-info=false
+
+      - name: Run vimeo/psalm on internal code
+        uses: docker://php:7.3-cli
+        with:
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -18,3 +18,20 @@ jobs:
...
+
+      - name: Update dependencies with composer
+        uses: docker://php:7.3-cli
+        with:
+          args: composer update
+
+      - name: Run vimeo/psalm
+        uses: docker://php:7.3-cli
+        with:
+          args: ./tools/psalm --config=.psalm/static-analysis.xml --no-progress --show-info=false && ./tools/psalm --config=.psalm/config.xml --no-progress --shepherd --stats --show-info=false
```
localheinz: Not sure, I need to investigate if there's a better way to run multiple commands - perhaps as a separate step?

.travis.yml
```diff
@@ -50,16 +50,6 @@ notifications:
...
 jobs:
   include:
-    - stage: "Static Code Analysis"
-      php: 7.3
-      env:
-        - TOOL="psalm"
-      install:
-        - phpenv config-rm xdebug.ini
-      script:
-        - travis_retry ./tools/composer update $DEFAULT_COMPOSER_FLAGS
```
localheinz: Note to self: need to specify flags.

.github/workflows/ci.yml
```diff
@@ -18,3 +18,25 @@ jobs:
...
+
+  type-checker:
+    name: Type Checker
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@master
+
+      - name: Update dependencies with composer
+        uses: docker://composer:1.9.0
```
Slamdunk: Composer docker image can have a PHP version different from the targeted one of the following step, so there can be a mismatch of the installed dependencies, right?

Ocramius:
For now, this would suffice:

```
./tools/psalm --config=.psalm/config.xml --set-baseline=.psalm/baseline.xml
```

Please be sure to review the diff on `baseline.xml` before merging.

Ocramius (approved):

.github/workflows/ci.yml
localheinz: Yes, that's definitely a problem we need to solve by providing dedicated Docker images!

localheinz:

PR #3824, opened 2019-09-06
Enhancement: Run roave/backward-compatibility-check using GitHub actions

localheinz:
This PR

* [x] runs `roave/backward-compatibility-check` (as installed with `phive` to `/tools`) using GitHub actions

Replaces #3823.

💁‍♂ As discussed earlier, we'll follow Kent Beck here with _Make it work, make it right, make it fast_!

### DIFF
--- a/.docker/php-7.3/Dockerfile
+++ b/.docker/php-7.3/Dockerfile
@@ -1,0 +1,18 @@
+FROM php:7.3-cli
+
+RUN apt-get update \
+    && apt-get install --assume-yes \
+        apt-utils \
+        git \
+        libzip-dev \
+        zip \
+        zlib1g-dev \
+    && apt-get clean --assume-yes \
+    && docker-php-ext-install \
+        zip
+
+RUN echo 'memory_limit = 256M' >> /usr/local/etc/php/conf.d/docker-php-memory-limit.ini
+
+COPY --from=composer:1.9.0 /usr/bin/composer /usr/local/bin/composer
+
+ENV COMPOSER_ALLOW_SUPERUSER=1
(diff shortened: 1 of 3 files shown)


localheinz:
@sebastianbergmann 

Not sure what to do here - we have a few options:

- use our own `Dockerfile` (slower or requires us to push to Dockerhub for faster builds)
- use an existing `Dockerfile` (for example, https://github.com/Nyholm/roave-bc-check-docker, faster, but then we might run into issues because of using a different version from the one installed with `phive`)

.github/workflows/ci.yml
```diff
@@ -40,3 +40,20 @@ jobs:
...
           args: ./tools/psalm --config=.psalm/config.xml --no-progress --shepherd --show-info=false --stats
+
+  backward-compatibility:
+    name: Backward Compatibility
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@master
+
+      - name: Build Docker image for PHP 7.3
```
localheinz: I've added this as a separate (!) step so it's more obvious how much time is consumed for building the Docker image for PHP 7.3.

.docker/php-7.3/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+        apt-utils \
+        git \
+        libzip-dev \
+        zip \
+        zlib1g-dev \
+    && apt-get clean --assume-yes \
+    && docker-php-ext-install \
+        zip
+
+RUN echo 'memory_limit = 256M' >> /usr/local/etc/php/conf.d/docker-php-memory-limit.ini
```
sebastianbergmann: To be safe, we should use `memory_limit=-1`, I think.

localheinz: Can do!

PR #3826, opened 2019-09-07
Properly initialize TestSuiteIterator.position

oliverklee:
Now this field always is an integer, following up on the promise
given by the type annotation.


oliverklee:
I'll check whether this will change (i.e., hopefully improve) the psalm baseline for 8.3, and then report back here whether this PR can be directly forward-ported to 8.3 or needs a separate PR for that branch.

PR #3837, opened 2019-09-07
Throw meaningful exception in TestSuiteIterator.getChildren()

oliverklee:
Also refactor the `getChildren()` and `hasChildren()`
to reduce code duplication.

Also drop some `rewind()` calls from the tests now that
the `TestSuite` already is properly rewound directly
after instantiation.

### DIFF
--- a/src/Framework/TestSuiteIterator.php
+++ b/src/Framework/TestSuiteIterator.php
@@ -73,12 +73,19 @@ final class TestSuiteIterator implements RecursiveIterator
 
     /**
      * Returns the sub iterator for the current element.
+     *
+     * @throws \UnexpectedValueException if the current element is no TestSuite
      */
     public function getChildren(): self
     {
-        return new self(
-            $this->tests[$this->position]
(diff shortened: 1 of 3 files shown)


src/Framework/TestSuiteIterator.php
```diff
@@ -73,19 +73,26 @@ public function next(): void
 
     /**
      * Returns the sub iterator for the current element.
+     *
+     * @throws \UnexpectedValueException if the current element is no TestSuite
```
sebastianbergmann: Exceptions raised by PHPUnit must have the appropriate marker interface. We need something along the lines of
```php
namespace PHPUnit\Framework;

final class UnexpectedValueException extends \UnexpectedValueException implements Exception
{
}
```

oliverklee: Done and repushed. After this has been merged, I'll create a separate PR for 8.3/master as the new exception class needs to be in a different location there.

Issue #3727, opened 2019-06-20
Problems hidden by PHPUnit's error handler


sebastianbergmann:
When PHPUnit does not register its error handler
```patch
diff --git a/src/Framework/TestResult.php b/src/Framework/TestResult.php
index d8a87e52f..378c2950d 100644
--- a/src/Framework/TestResult.php
+++ b/src/Framework/TestResult.php
@@ -627,7 +627,7 @@ public function run(Test $test): void
 
         $errorHandlerSet = false;
 
-        if ($this->convertErrorsToExceptions) {
+        if (false && $this->convertErrorsToExceptions) {
             $oldErrorHandler = \set_error_handler(
                 [ErrorHandler::class, 'handleError'],
                 \E_ALL | \E_STRICT
```
then the following problems are revealed

- [x] `strpos(): Empty needle in /usr/local/src/phpunit/src/Framework/Constraint/StringStartsWith.php on line 46`
- [x] `Undefined index: stdout in /usr/local/src/phpunit/src/Runner/PhptTestCase.php on line 180`

#### Complete output
```
PHPUnit 7.5.13 by Sebastian Bergmann and contributors.

Runtime:       PHP 7.3.6
Configuration: /usr/local/src/phpunit/phpunit.xml

.............................................................   61 / 2074 (  2%)
.............................................................  122 / 2074 (  5%)
.............................................................  183 / 2074 (  8%)
.............................................................  244 / 2074 ( 11%)
.............................................................  305 / 2074 ( 14%)
.............................................................  366 / 2074 ( 17%)
.............................................................  427 / 2074 ( 20%)
.............................................................  488 / 2074 ( 23%)
.............................................................  549 / 2074 ( 26%)
.................................PHP Warning:  strpos(): Empty needle in /usr/local/src/phpunit/src/Framework/Constraint/StringStartsWith.php on line 46
F...........................  610 / 2074 ( 29%)
.............................................................  671 / 2074 ( 32%)
...........................................S.................  732 / 2074 ( 35%)
.............................................................  793 / 2074 ( 38%)
.............................................................  854 / 2074 ( 41%)
.............................................................  915 / 2074 ( 44%)
.............................................................  976 / 2074 ( 47%)
............................................................. 1037 / 2074 ( 50%)
............................................................. 1098 / 2074 ( 52%)
............................................................. 1159 / 2074 ( 55%)
............................................................. 1220 / 2074 ( 58%)
[...]

PR #3839, opened 2019-09-07
Prevent undefined stdout index error - #3727

DanielRuf:
I guess we should use the `$this->output` to prevent the undefined index error.

See #3727


sebastianbergmann:
Can you make this against `7.5`, please? Thanks!

DanielRuf:
Done.

localheinz (approved):
👍

PR #3840, opened 2019-09-07
Refactor Configuration to remove Code-Duplicates

SebastianKull:
WIP
Feedback is welcome

### DIFF
--- a/src/Util/Configuration.php
+++ b/src/Util/Configuration.php
@@ -1197,6 +1197,7 @@ final class Configuration
                 true
             );
         }
+
         return [
             'class'     => $class,
             'file'      => $file,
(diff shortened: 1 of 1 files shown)


src/Util/Configuration.php
```diff
@@ -1242,4 +1182,25 @@ private function parseGroupConfiguration(string $root): array
 
         return $groups;
     }
+
+    private function setResult($element): array
```
sebastianbergmann: The name `setResult()` does not "speak to me"; it is not clear what the method does. Can we find a better name?

The `$element` parameter should have a type declaration. `DOMElement` looks like the appropriate type, then we can also get rid of the `instanceof` on the first line of the method's body.

PR #3843, opened 2019-09-07
Fix Psalm warnings in Constraints

oliverklee:

PR #3846, opened 2019-09-07
Keep php-cs-fixer from removing or simplifying "return null;"

oliverklee:
`return null;` semantically is something different than `return;`
and should be kept.

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -171,7 +171,7 @@ return PhpCsFixer\Config::create()
             'semicolon_after_instruction' => true,
             'set_type_to_cast' => true,
             'short_scalar_cast' => true,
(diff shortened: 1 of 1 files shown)


kubawerlos:
@oliverklee @sebastianbergmann can you elaborate about this change?

> `return null;` semantically is something different than `return;`

Of course it is, and `simplified_null_return` cares about that and change it only when should be changed:

> * **simplified_null_return**
>
>  A return statement wishing to return ``void`` should not return ``null``.

Do you have an example when the fixer removed `null` when it shouldn't? Maybe there is some bug in fixer need to be handled.

oliverklee:
@kubawerlos Yes, php-cs-fixer was removing most of the `return null;` from PR #3843.

PR #3849, opened 2019-09-07
Refactor mock runtime

MichelHartmann:
Refactored runtime code for mocks.

* The old "Matchers" are separated into three responsibilities:
  * method invocation
  * parameters
  * method name
* Some interfaces that were used only by a single class are removed.
* Renamed some concepts to better reflect what they are doing


sebastianbergmann:
I think we can safely ignore the [API breaks](https://github.com/sebastianbergmann/phpunit/pull/3849/checks?check_run_id=215169922) that this refactoring introduces.

mickaelandrieu:
> I think we can safely ignore the API breaks that this refactoring introduces.

I know these classes are not covered in any way by the BC Break, but I'm wondering what we can do to replace the use of `getInvocations` method of the `AnyInvokedCount` class ?

Issue #3671, opened 2019-04-30
Add @covers annotations to all tests


oliverklee:
I'd like to tackle this so the code coverage report is more accurate.

@sebastianbergmann  Any objections? Or things I'll need to take into consideration when doing this?

sebastianbergmann:
I am afraid that this will not be as easy as it should be. You may want to start with the tests for the `PHPUnit\Util` classes. Most of them should be isolated enough for this to be not as frustrating as I expect `PHPUnit\Framework` to be, for instance.

oliverklee:
.

PR #3850, opened 2019-09-07
Add some covers annotations - #3671

edditor:
This PR does:

* add some @covers annotations #3671

### DIFF
--- a/tests/unit/Util/GetoptTest.php
+++ b/tests/unit/Util/GetoptTest.php
@@ -14,9 +14,14 @@ use PHPUnit\Framework\TestCase;
 
 /**
  * @small
+ * @covers \PHPUnit\Util\Getopt
  */
 final class GetoptTest extends TestCase
 {
+    /**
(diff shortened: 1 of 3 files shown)


tests/unit/Util/GetoptTest.php
```diff
@@ -14,9 +14,14 @@
 
 /**
  * @small
+ * @covers \PHPUnit\Util\Getopt
  */
 final class GetoptTest extends TestCase
 {
+    /**
+     * @covers \PHPUnit\Util\Getopt::getopt
```
localheinz: Not sure, perhaps the class-level annotation will suffice?

sebastianbergmann:
This branch has conflicts that must be resolved.

localheinz (approved):
👍

localheinz:
Thank you, @edditor!

tests/unit/Util/GetoptTest.php
sebastianbergmann: Yes, please only use class-level annotations.

Issue #2300, opened 2016-09-22
Exclude tests from package distribution


neomerx:
Currently if phpunit is installed as a project dependency, composer downloads it with tests and some other not necessary files. And some of the phpunit names get IDE code helpers messy

![image](https://cloud.githubusercontent.com/assets/10420662/18758392/942bcdea-8100-11e6-8f84-f302c7d61009.png)

The solution is to instruct git/composer to ignore some of the files and folders by adding to `.gitattributes` a list of files/folders

```
/tests              export-ignore
/.gitattributes     export-ignore
/.gitignore         export-ignore
/.php_cs            export-ignore
/.travis.yml        export-ignore
...
```

alcohol:
This does break anyone installing phpunit from github zipballs and attempting to install the dependencies, because of the explicit classmap reference in `autoload-dev` to the `tests` directory. Possible solutions here would be to switch to using PSR-0/PSR-4 instead of classmap (if this is something that you would want to address).

sebastianbergmann:
@alcohol If you think that this breaks something that should not be broken then please open a new ticket.

PR #3855, opened 2019-09-08
Don't export development-only files

oliverklee:
The files that are used for developing PHPUnit
(but not for developing _with_ PHPUnit) should not be
exported.

This should reduce the size of the Composer packages
that are installed with the `--prefer-dist` option
(which is the default) considerably.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,17 @@
+/.docker export-ignore
+/.github export-ignore
+/.psalm export-ignore
 /build export-ignore
+/tests export-ignore
 /tools export-ignore
+/.editorconfig export-ignore
+/.gitattributes export-ignore
+/.gitignore export-ignore
+/.php_cs.dist export-ignore
+/.phpstorm.meta.php export-ignore
+/.travis.yml export-ignore
+/build.xml export-ignore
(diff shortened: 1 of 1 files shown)


oliverklee:
To keep things simple, this 7.5 version also includes some files and directories have have been introduced in 8.3.

The files are ordered the same way they are ordered in PhpStorm.

kubawerlos:
Can we have `/.* export-ignore` for all the dot-starting-files? The convention is that these files are internal and not to be released after all, right?

sebastianbergmann:
I am reluctant to merge this as there were [problems in the past](https://github.com/sebastianbergmann/phpunit/issues/2300#issuecomment-249176276) when `tests` was ignored for export.

AFAIK, @remicollet relies on `tests` being exported. Maybe he can provide some insight?

oliverklee:
> Can we have `/.* export-ignore` for all the dot-starting-files?

I would be fine with that. @sebastianbergmann What do you think?

oliverklee:
Symfony recently had this change, too:
https://github.com/symfony/yaml/commit/79800cc42e2316bdc0330bbd2ae34fb51ee06e68

PR #3809, opened 2019-09-06
Enhancement: Run php-cs-fixer using GitHub Actions

localheinz:
This PR

* [x] runs `friendsofphp/php-cs-fixer` (as installed with `phive` to `/tools`) as GitHub action

PR #3856, opened 2019-09-09
Enhancement: Use built-in PHP 7.3 to speed up build for 7.5

localheinz:
This PR

* [x] uses the built-in binary for PHP 7.3 to run `php-cs-fixer` on `7.5.`

Follows #3809.

💁‍♂ As pointed out by @OskarStark, PHP binaries are available on `ubuntu-latest`. This should speed up the build significantly!


localheinz:

OskarStark:
Glad I could help 😃

PR #3857, opened 2019-09-09
Enhancement: Use built-in PHP 7.3 to speed up build for 8.3

localheinz:
This PR

* [x] uses the built-in binary for PHP 7.3 to run jobs on  on `8.5.`

Follows #3809, #3820, #3823.


sebastianbergmann:
Thanks! Can you have a look that I did not break anything by mergin this into `master`?

localheinz:
@sebastianbergmann 

Will do!

localheinz:
@sebastianbergmann 

Running `roave-backward-compatibility-check` [fails](https://github.com/sebastianbergmann/phpunit/runs/215960959#step:3:1), but I think that should be expected!

👍

sebastianbergmann:
No, it should not fail. And it no longer does: https://github.com/sebastianbergmann/phpunit/runs/215965976

PR #3863, opened 2019-09-11
Count Constaint - check \Countable first

jasverix:
A problem in a project where a class has been created `EmptyIteratorWithCount` meant to iterate nothing, but still have a count on it. It is used to display a count somewhere in the project where an iterator is expected, but we don't want to actually fetch any data.

Unit-tests on this project suddenly started to fail after upgrading to latest PhpUnit because it checks EmptyIterator first.


sebastianbergmann:
Merged manually, thanks.

Issue #3877, opened 2019-10-02
startTestSuite is call on every unit test


fezfez:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 8.3.5, 8.2.5, 8.1.6, 8.0.6, 7.5.16, 7.4.5, 7.3.5, 7.2.7, 7.1.5
| PHP version         | 7.3.4, 7.4RC1, 7.2.25
| Installation Method | Composer

#### Summary

i have added 
```php 
var_dump(__METHOD__.' '.$suite->getName()); 
```

on 

https://github.com/sebastianbergmann/phpunit/blob/4d2e3f801b316766fbfc3431f6fda6137b63fab0/src/Framework/TestResult.php#L381-L383

and

```php 
var_dump(__METHOD__.' '. $result->count());
```
on

https://github.com/sebastianbergmann/phpunit/blob/4d2e3f801b316766fbfc3431f6fda6137b63fab0/src/Framework/TestSuite.php#L516

#### Current behavior

```
PHPUnit 8.3.5 by Sebastian Bergmann and contributors.

Runtime:       PHP 7.4.0RC1
Configuration: /var/www/phpunit.xml.dist
Random seed:   1570011937

string(34) "PHPUnit\Framework\TestSuite::run 0"
string(45) "PHPUnit\Framework\TestResult::startTestSuite "
string(34) "PHPUnit\Framework\TestSuite::run 0"
string(49) "PHPUnit\Framework\TestResult::startTestSuite unit"
string(34) "PHPUnit\Framework\TestSuite::run 0"
string(103) "PHPUnit\Framework\TestResult::startTestSuite RoaveUnitTest\NoLeaks\PHPUnit\EmptyBaselineMemoryUsageTest"
.string(34) "PHPUnit\Framework\TestSuite::run 1"
string(115) "PHPUnit\Framework\TestResult::startTestSuite RoaveUnitTest\NoLeaks\PHPUnit\CollectTestExecutionMemoryFootprintsTest"
......string(34) "PHPUnit\Framework\TestSuite::run 7"
string(104) "PHPUnit\Framework\TestResult::startTestSuite RoaveUnitTest\NoLeaks\PHPUnit\MeasuredTestRunMemoryLeakTest"
.......string(35) "PHPUnit\Framework\TestSuite::run 14"
[...]

PR #3878, opened 2019-10-03
#3877 : Add E2E test to prove the bug

fezfez:


sebastianbergmann:
#3877 has been closed.

PR #3882, opened 2019-10-04
Fixes #3881

### LINKED ISSUE #3881: Regression with multiple test case classes declared in a single sourcecode file
While cleaning up the reproducing example from #3879, I ran into the following issue:
```php
<?php

abstract class AbstractTest extends TestCase
{
    /**
     * @dataProvider provider
     */
    public function testOne(string $parameter): void
    {
        $this->assertSame('test', $parameter);
    }

    public function provider(): array
    {
        return [
            ['test'],
        ];
    }
}

final class Test extends AbstractTest
{
}
```
With PHPUnit 8.4.0 I get
```
PHPUnit 8.4.0 by Sebastian Bergmann and contributors.

W                                                                   1 / 1 (100%)

Time: 34 ms, Memory: 6.00 MB

There was 1 warning:

1) Warning
Cannot instantiate class "AbstractTest".

WARNINGS!
Tests: 1, Assertions: 0, Warnings: 1.
```
With PHPUnit 8.3.5 I get
```
PHPUnit 8.3.5 by Sebastian Bergmann and contributors.

Time: 34 ms, Memory: 6.00 MB

OK (1 test, 1 assertion)
```

realFlowControl:
This PR fixes the regression that came with removing a lot of code in commit 82e7076. I accidentally also removed a check if a class is abstract and should not be instantiated.

### DIFF
--- a/src/Runner/StandardTestSuiteLoader.php
+++ b/src/Runner/StandardTestSuiteLoader.php
@@ -52,6 +52,10 @@ final class StandardTestSuiteLoader implements TestSuiteLoader
                 );
             }
 
+            if ($class->isAbstract()) {
+                continue;
+            }
+
             if (\substr($loadedClass, $offset) === $suiteClassName &&
                 $class->getFileName() == $filename) {
                 $suiteClassName = $loadedClass;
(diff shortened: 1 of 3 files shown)


sebastianbergmann:
Merged manually, thanks.

Issue #3879, opened 2019-10-04
Tests with data providers in parent class don't work anymore


l-x:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 8.4.0
| PHP version         | 7.3.9
| Installation Method | Composer

#### Summary

Tests classes that contain a test method utilizing a data provider defined in a parent class will produce a warning.

#### Previous Behavior

No warnings will be shown

#### Current Behavior

Warning:
```
The data provider specified for FuTest::testFu is invalid.
Cannot instantiate abstract class AbstractFuTestCase
```

#### How to reproduce

```php
abstract class AbstractFuTestCase extends \PHPUnit\Framework\TestCase
{
    public function dataProvider(): array
    {
        return [
            ['fu'],
            ['bar'],
        ];
    }

    /**
     * @dataProvider dataProvider
     */
    public function testFu(string $bar): void
    {
        $this->assertTrue(true);
    }
}

class FuTest extends AbstractFuTestCase
{

}
```
Just run the `FuTest`

pavbis:
Can also confirm the issue. Just updated to 8.4.0.

sebastianbergmann:
@Ocramius This looks like it could be related to #3836. PHPUnit tries to create an instance of `AbstractFuTestCase` instead of `FuTest` to call the data provider method on.

Ocramius:
@sebastianbergmann basically the patch takes the parent class instead of the child class when considering an annotation, right?

dpauli:
The problem also appear on simple abstraction, see: https://github.com/sebastianbergmann/phpunit/issues/3881#issuecomment-538315994

sebastianbergmann:
I am not sure whether this is related to annotation parsing. Might as well be related to #3381 (or something else). I cannot look into this right now, will do so over the weekend.

l-x:
To clarify the problem I reported: It is not necessary to put the two classes in a single file to reproduce this behaviour.

sebastianbergmann:
@Ocramius According to `git bisect`, 930906229e3a30d17224cd4a173565faf0952c53 is the first "bad commit":
```
$ cat run.sh               
#!/bin/sh
composer update
./phpunit --fail-on-warning /tmp/FuTest.php
```
```
$ git bisect start
$ git bisect good 8.3.5
$ git bisect bad 8.4.0
$ git bisect run ./run.sh
```
[...]

PR #3885, opened 2019-10-06
Let the DocBlock::ofMethod handle an ~optional~ className

wgevaert:
Fixes issue #3879


sebastianbergmann:
@Ocramius Can you have a look at this, please? I cannot really put a finger on it, but at first glance this seems more like a workaround than a solution.

src/Util/Annotation/DocBlock.php
```diff
@@ -111,7 +114,7 @@ public static function ofMethod(\ReflectionMethod $method): self
             $method->getEndLine(),
             $method->getFileName(),
             $method->getName(),
-            $method->getDeclaringClass()->getName()
+            $className ?? $method->getDeclaringClass()->getName()
```
Ocramius: If fetching the specific class is the use-case, then I suggest making the new argument mandatory, and naming `$className` so that it reflects the fact that this is a specific class in the hierarchy.

src/Util/Annotation/Registry.php
```diff
@@ -62,7 +62,7 @@ public function forClassName(string $class): DocBlock
 
     /**
      * @throws Exception
-     * @psalm-param class-string $className
```
Ocramius: Ow, not sure how this made it through 😱

src/Util/Annotation/Registry.php
```diff
@@ -80,6 +80,6 @@ public function forMethod(string $class, string $method): DocBlock
             );
         }
 
-        return $this->methodDocBlocks[$class][$method] = DocBlock::ofMethod($reflection);
+        return $this->methodDocBlocks[$class][$method] = DocBlock::ofMethod($reflection, $class);
     }
```
Ocramius: `$class` should probably be named `$classInHierarchy` or such

Ocramius (approved):
Fix seems correct to me, @sebastianbergmann: we are now recording docblocks by hierarchy level in a class hierarchy.

The only problem I see with the patch is that it introduces a change in the test suite, rather than an addition, and that makes it hard to validate it. I suggest reverting test changes, and adding targeted tests instead.

src/Util/Annotation/DocBlock.php
```diff
@@ -101,7 +101,10 @@ public static function ofClass(\ReflectionClass $class): self
         );
     }
 
-    public static function ofMethod(\ReflectionMethod $method): self
+    /**
+     * @psalm-param class-string $className
+     */
+    public static function ofMethod(\ReflectionMethod $method, string $className = null): self
```
gleb-svitelskiy: Now there are two ways to create DocBlock object
```
DocBlock::ofClass(
    new \ReflectionClass($class)
);
DocBlock::ofMethod(
    new \ReflectionMethod($class, $method)
);
```
Unfortunately, can not obtain class namespace from method reflection object, only class that declares method.
Maybe it's better to describe only one method to create a DocBlock object e.g.
```
public static function fromReflection(\ReflectionClass $class, \ReflectionMethod $method = null): self
```
or make a public constructor
```
public function __construct(\ReflectionClass $class, \ReflectionMethod $method = null)
```
@Ocramius what do you think about that?

src/Util/Annotation/DocBlock.php
gleb-svitelskiy: Absolutely, it's not optional.

src/Util/Annotation/Registry.php
wgevaert: What's with this psalm-param? I do not really understand it, I copied it to more locations, seemed logical to me

Ocramius: This is a restriction on the `string` type, stating that the string must be a class name. The `@psalm-param` is an annotation we have to use here, in lack of better type support in phpdocumentor/phpstorm/etc.

All good BTW 👍

src/Util/Annotation/DocBlock.php
Ocramius: What about separate named ctors (as separate static methods)?

Also, please keep the ctor private: that's what gave us this flexibility in first place 👍

wgevaert: There is a difference between docblocks of classes and those of methods right? I think the current structure makes that quite clear?

tests/_files/VariousIterableDataProviderTest.php
```diff
@@ -7,9 +7,9 @@
  * For the full copyright and license information, please view the LICENSE
  * file that was distributed with this source code.
  */
-class VariousIterableDataProviderTest
+class VariousIterableDataProviderTest extends AbstractVariousIterableDataProviderTest
 {
-    public static function asArrayProvider()
+    public function asArrayProvider()
```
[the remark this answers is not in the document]
wgevaert: I renamed them and wrote similar tests. I hope it is clear, would be happy to hear how it could be done better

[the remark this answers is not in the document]
Ocramius: I'd generally suggest reverting the old asset changes: they do prevent regressions (like the one that you are fixing - and BTW thanks very much for your efforts!)

src/Util/Annotation/DocBlock.php
Ocramius: Yeah, all good with that

wgevaert:
Currently having problems with the following:
Adding the `class-string` to the classNames for psalm gave problems in `Framework/DataProviderTestSuite`, in the following snippet:
```
         [$className, $methodName] = \explode('::', $this->getName());

        return TestUtil::getSize($className, $methodName);
```
So I made this
```
         [$className, $methodName] = \explode('::', $this->getName());
 
+        /** @psalm-suppress ArgumentTypeCoercion */
         return TestUtil::getSize($className, $methodName);
```
Now php-Cs-fixer wants it to look like 
```
         [$className, $methodName] = \explode('::', $this->getName());
 
-        /** @psalm-suppress ArgumentTypeCoercion */
+        /* @psalm-suppress ArgumentTypeCoercion */
         return TestUtil::getSize($className, $methodName);
```
but then psalm does not recognize the psalm-annotation. What should I do?

tests/_files/VariousIterableDataProviderTest.php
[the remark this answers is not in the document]
wgevaert: I've now kept the old test, but named it `testWithVariousIterableStaticDataProviders` instead of `testWithVariousIterableDataProviders`, and then also made tests for when this provider is non-static, of the parent, called from the parent, etc.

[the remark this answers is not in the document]
Ocramius: Thanks!

Ocramius:
If suppressing won't work, then I suggest updating the baseline file. See `psalm --help | grep baseline`.

If all is good, we should see that one error added to the file, and if lucky even some error removals in the diff 👍

wgevaert:
Hurray!

tests/_files/AbstractVariousIterableDataProviderTest.php
```diff
@@ -0,0 +1,56 @@
+<?php declare(strict_types=1);
+/*
+ * This file is part of PHPUnit.
+ *
+ * (c) Sebastian Bergmann <sebastian@phpunit.de>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+abstract class AbstractVariousIterableDataProviderTest
```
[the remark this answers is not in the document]
Ocramius: Solved

Ocramius (approved):

sebastianbergmann:
Cherry-picked into `8.4` and merged to `master` from there. Thank you!

wgevaert:
Did you cherry-pick the whole pull-request or only the first commit? I did make quite some adjustments, thanks to ocramius' feedback, that should certainly be included as well

PR #3890, opened 2019-10-09
small refactor to reduce complexity #hacktoberfest

codisart:
based on code climate analysis

### DIFF
--- a/src/Util/Filter.php
+++ b/src/Util/Filter.php
@@ -22,13 +22,6 @@ final class Filter
      */
     public static function getFilteredStacktrace(\Throwable $t): string
     {
-        $prefix = false;
-        $script = \realpath($GLOBALS['_SERVER']['SCRIPT_NAME']);
-
-        if (\defined('__PHPUNIT_PHAR_ROOT__')) {
-            $prefix = __PHPUNIT_PHAR_ROOT__;
-        }
-
         $filteredStacktrace = '';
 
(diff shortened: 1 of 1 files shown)


src/Util/Filter.php
```diff
@@ -17,18 +17,16 @@
  */
 final class Filter
 {
+    /** @var Blacklist */
+    private static $blacklist;
```
sebastianbergmann: What is your reasoning for using static attributes here?

codisart: I choose to add the attributes to avoid add many params on the private methods that I added.

sebastianbergmann: But is there a specific reason why you declared these attributes `static`?

codisart: It seems to me that we can not use non static attribut ion a static method.

https://3v4l.org/vqjCY

sebastianbergmann: Fair point, I blocked out the fact that `Filter` still uses static methods.

sebastianbergmann: I would prefer, though, not to have these attributes and that `$blacklist` and `$prefix` are passed as parameters to the private method. Thanks!

codisart: ok I will do the change. 

Will you be interested to remove the static method from `Filter` in a future refactoring ?

PR #3894, opened 2019-10-13
Added --no-interaction to the usage text

morozov:
The `--no-interaction` command-line option was introduced in #3528 but wasn't added to the usage text.


sebastianbergmann:
Merged manually, thanks.

PR #3901, opened 2019-10-15
Use finally insead of isset on error

BackEndTea:
By using the `finally` keyword we don't need to check if the error 
was thrown, and then throw it again after the checks.

This _should_ also help a bit more in case one needs to debug where
an error came from, as the original error is still raised, instead of 
being caught and thrown again.

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -1662,21 +1662,15 @@ abstract class TestCase extends Assert implements SelfDescribing, Test
         if ($this->prophet !== null) {
             try {
                 $this->prophet->checkPredictions();
-            } catch (Throwable $t) {
-                /* Intentionally left empty */
-            }
-
-            foreach ($this->prophet->getProphecies() as $objectProphecy) {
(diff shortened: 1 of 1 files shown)


localheinz (approved):
Using `finally` without `catch` is ugly, in my opinion, but it works. 

👍

BackEndTea:
I think its more clear than the current solution. 
Plus, this is what finally is intended for. Code blocks that need to be executed regardless of whether an error is thrown.

Could you explain why it is ugly? Because to me it doesn't seem ugly, so i'd like to hear your idea on it.

localheinz:
@BackEndTea 

Only came across a usage of `try`-`finally` (without the `catch`) in a closed-source project a couple of months ago, and I thought it was weird to `try` something without `catch`ing any exceptions that might be thrown. 

Here it makes perfect sense.

👍

asgrim (approved):

BackEndTea:
Thank you @localheinz, @asgrim & @sebastianbergmann

PR #3919, opened 2019-10-30
Revert #3830 (Cleanup and small refactorings)

realFlowControl:
This pull request reverts #3830 and adds a test for another regression introduced by merging #3830 and trying to cleanup the mess. In summary #3830 did backfire quite large :disappointed:

### DIFF
--- a/src/Runner/StandardTestSuiteLoader.php
+++ b/src/Runner/StandardTestSuiteLoader.php
@@ -25,7 +25,8 @@ final class StandardTestSuiteLoader implements TestSuiteLoader
      */
     public function load(string $suiteClassName, string $suiteClassFile = ''): ReflectionClass
     {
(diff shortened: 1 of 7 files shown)


flow-control:
looks like i need to write some additional tests for that ...

flow-control:
Hey @sebastianbergmann,
the only things left to test are multiple of those blocks in the `src/Runner/StandardTestSuiteLoader.php`
```php
try {
    $class = new ReflectionClass($suiteClassName);
} catch (\ReflectionException $e) {
    throw new Exception(
        $e->getMessage(),
        (int) $e->getCode(),
        $e
    );
}
```
And i don't know how to test these. Could you assist?

/Flo

sebastianbergmann:
They should be safe to ignore. Maybe I'm just too defensive when it comes to the Reflection API :) Thanks!

PR #3927, opened 2019-11-05
Enhancement: Cache dependencies installed with composer between builds

localheinz:
This PR

* [x] caches dependencies installed with `composer` between builds

💁‍♂ This is inspired by the work of @bendavies on https://github.com/localheinz/php-library-template/pull/131. 

For reference, see https://github.blog/changelog/2019-11-04-github-actions-adds-dependency-caching/.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,6 +56,14 @@ jobs:
       - name: Checkout
         uses: actions/checkout@master
 
+      - name: "Cache dependencies installed with composer"
+        uses: actions/cache@v1
+        with:
+          path: ~/.composer/cache
+          key: ${{ matrix.php-binary }}-composer-${{ matrix.dependencies }}-${{ github.sha }}
+          restore-keys: |
+            ${{ matrix.php-binary }}-composer-${{ matrix.dependencies }}-
+
       - name: Install lowest dependencies with composer
         if: matrix.dependencies == 'lowest'
         run: ${{ matrix.php-binary }} ./tools/composer update --no-ansi --no-interaction --no-progress --no-suggest --prefer-lowest


.github/workflows/ci.yml
```diff
@@ -56,6 +56,14 @@ jobs:
       - name: Checkout
         uses: actions/checkout@master
 
+      - name: "Cache dependencies installed with composer"
+        uses: actions/cache@v1.0.0
+        with:
+          path: ~/.composer/cache
+          key: ${{ matrix.php-binary }}-composer-${{ matrix.dependencies }}-${{ hashFiles('**/composer.json') }}
```
bendavies: the problem with `${{ hashFiles('**/composer.json') }}` here is it does not allow the cache to be additive when `matrix.dependencies == 'highest'` starts downloading **new** dependencies.
GitHub Actions will not overwrite the cache so some deps will never be cached.
The cache key needs to be unique to fix this.
Using this would allow the cache to be additive over time:
`${{ matrix.php-binary }}-composer-${{ matrix.dependencies }}-${{ github.sha }}`

.github/workflows/ci.yml
```diff
@@ -56,6 +56,14 @@ jobs:
       - name: Checkout
         uses: actions/checkout@master
 
+      - name: "Cache dependencies installed with composer"
+        uses: actions/cache@v1.0.0
```
bendavies:
```suggestion
        uses: actions/cache@v1
```

sebastianbergmann:
Is this ready to be merged?

localheinz:
@sebastianbergmann 

Unless @bendavies has any objections, let's ship it!

.github/workflows/ci.yml
localheinz: :+1

bendavies:
LGTM

.github/workflows/ci.yml
bendavies: They will be used, as the `restore-keys` keys will be used to find the latest matching cache hit by substring.
They won't be used for long, but that doesn't matter, more recent ones will be.

localheinz:
Thank you, @bendavies and @sebastianbergmann!

PR #3936, opened 2019-11-08
Support using @depends to depend on classes

### LINKED ISSUE #3519: Using @depends to depend on classes

epdenouden:
## Use case
As a _developer of PHPUnit test collections_
I want to _declare dependencies between tests and whole `TestCase` classes_
so that _large or interdependent collections can be more easily managed_

This implements feature request #3519 by adding support for declaring interdependency between a test and a `TestCase` class using the existing `@depends` annotation. This is often useful for managing slow integration suites: "don't even try this if `SomeTestClass` didn't pass."

[Detailed use case](https://github.com/sebastianbergmann/phpunit/issues/3519) with [additional context provided](https://github.com/sebastianbergmann/phpunit/issues/3519#issuecomment-461384340) by @spriebsch.

## Example
```php
class DependencyOnClassTest extends TestCase
{
    /**
     * Guard support for using annotations to depend on a whole successful TestSuite
     *
     * @depends DependencySuccessTest::class
     *
     * @see https://github.com/sebastianbergmann/phpunit/issues/3519
     */
    public function testThatDependsOnASuccessfulClass(): void
    {
        $this->assertTrue(true);
    }
	// please refer to the PHPUnit source for the full version of this example
}
```

Example from the [new unit test `DependencyOnClassTest`](https://github.com/epdenouden/phpunit/blob/feature/issue-3519-depends-on-class/tests/_files/DependencyOnClassTest.php#L13-L37) and [configuration file](https://github.com/epdenouden/phpunit/blob/feature/issue-3519-depends-on-class/tests/_files/configuration.depends-on-class.xml)

## Implementation
- implement improved dependency resolver, while still using the proven `TestSuiteSorter`
- implement `@depends SomeTest::class` dependency check mechanism
- track run state of individual `TestSuite` objects based on its child tests
- a `TestCase::class` is considered passed if it doesn't contain any failing tests
- test housekeeping to cover the new resolver and supporting infrastructure

### Groundwork for future improvements
- **next project**: move execution reordering to the core iterator
- refactoring the on-demand `@dataProvider` prototype to production code will be much easier with all the new plumbing

## Testing
The new functionality is guarded by multiple unit and end-to-end tests.
* [non-trivial scenario end-to-end test](https://github.com/epdenouden/phpunit/blob/feature/issue-3519-depends-on-class/tests/end-to-end/execution-order/depends-on-class.phpt)
* new [`@depends` resolver mechanism](https://github.com/epdenouden/phpunit/blob/55423b694797697c638c1492f08bed4033d6adc0/tests/unit/Framework/TestSuiteTest.php#L254-L337)
* found some rare use cases in need of attention, like `@depends`+`@size` handling

## To do
- [x] implement full end-to-end scenario with config file and multiple test suites
- [x] add `@depends`-annotation normalization as this is provided by user input and not reflection
- [x] `TestSuite` pass/fail tracking in `TestResult`
- [x] skip-test behaviour for whole `TestSuite` based on `@depends ::class`
- [x] implement tree-traversing `provides`/`requires` dependency resolver
  - [x] `TestSuite`
  - [x] `DataProviderTestSuite`
  - [x] `TestCase`
  - [x] `PhptTestCase`
- [x] replace resolver in `TestSuiteSorter` with new one
- [x] get code to LTS release quality
  - [x] improve simplistic `TestSuite` dirty flagging
  - [x] full feature code coverage
  - [x] improve project code coverage

## Notes
I realize there will be edge-cases with a new feature like this. Give it a try and let me know what you think!

Ping @barbu110


epdenouden:
@sebastianbergmann I'll fix coverage tomorrow, pls don't merge yet

src/Framework/TestCase.php
```diff
@@ -1157,6 +1167,9 @@ public function runBare(): void
     public function setName(string $name): void
     {
         $this->name = $name;
+
+        // Normalized name for dependency resolver
+        $this->providedTests = [\get_class($this) . '::' . $name];
```
mvriel, line 1172: I do not know if this is relevant but this normalized name may be duplicate between properties, methods and constants because these may have the same name :)

epdenouden: This line by itself can indeed overwrite an existing key in the `providedTests`. Fortunately this is part of the contruction logic of new test objects and the `testName` always refers to the `TestCase::`-methods. :relieved:

tests/unit/Util/TestClassTest.php
```diff
@@ -1103,22 +1103,26 @@ public function testTestWithThrowsProperExceptionIfMultiLineDatasetCannotBeParse
...
     }
 
     /**
      * @depends Foo
      * @depends ほげ
+     * @depends
+     * @depends AnotherClass::Foo
      *
-     * @todo Remove fixture from test class
+     * @xtodo Remove fixture from test class
```
sebastianbergmann, line 1125: `@xtodo`?

epdenouden: fixed

src/Framework/DataProviderTestSuite.php
```diff
@@ -26,25 +26,35 @@ final class DataProviderTestSuite extends TestSuite
      */
     public function setDependencies(array $dependencies): void
     {
-        $this->dependencies = $dependencies;
+        $this->dependencies  = $dependencies;
+        $this->requiredTests = \PHPUnit\Util\Test::trimDependencyOptions($dependencies);
```
localheinz, line 30:
```suggestion
        $this->requiredTests = TestUtil::trimDependencyOptions($dependencies);
```

src/Framework/TestCase.php
```diff
@@ -1166,23 +1179,8 @@ public function setName(string $name): void
...
-    }
-
-    /**
-     * @internal This method is not covered by the backward compatibility promise for PHPUnit
-     */
-    public function hasDependencies(): bool
-    {
-        return \count($this->dependencies) > 0;
+        $this->dependencies  = $dependencies;
+        $this->requiredTests = \PHPUnit\Util\Test::trimDependencyOptions($dependencies);
```
localheinz, line 1183:
```suggestion
        $this->requiredTests = TestUtil::trimDependencyOptions($dependencies);
```

src/Framework/TestCase.php
```diff
@@ -1407,6 +1389,28 @@ public function addWarning(string $warning): void
         $this->warnings[] = $warning;
     }
 
+    /**
+     * Returns the normalized test name as class::method
```
localheinz, line 1393:
```suggestion
     * Returns the normalized test name as class::method.
```

src/Framework/TestCase.php
```diff
@@ -1407,6 +1389,28 @@ public function addWarning(string $warning): void
...
+     *
+     * @return string[]
+     */
+    public function provides(): array
+    {
+        return $this->providedTests;
+    }
+
+    /**
+     * Returns a list of normalized dependency names, class::method
```
localheinz, line 1403:
```suggestion
     * Returns a list of normalized dependency names, class::method.
     *
```

src/Framework/TestCase.php
```diff
@@ -1407,6 +1389,28 @@ public function addWarning(string $warning): void
...
+     * This list can differ from the raw dependencies as the resolver has
+     * no need for the
```
localheinz, line 1405: Is there something missing here?

src/Framework/TestCase.php
```diff
@@ -1942,89 +1946,100 @@ private function checkRequirements(): void
 
     private function handleDependencies(): bool
     {
-        if (!empty($this->dependencies) && !$this->inIsolation) {
-            $className  = \get_class($this);
-            $passed     = $this->result->passed();
-            $passedKeys = \array_keys($passed);
+        if (empty($this->dependencies) || $this->inIsolation) {
```
localheinz, line 1949:
```suggestion
        if ([] === $this->dependencies || $this->inIsolation) {
```

src/Framework/TestCase.php
```diff
@@ -1942,89 +1946,100 @@ private function checkRequirements(): void
...
+            return true;
+        }
 
-            foreach ($passedKeys as $key => $value) {
-                $pos = \strpos($value, ' with data set');
+        $className  = \get_class($this);
```
localheinz, line 1953:
```suggestion
```

Variable is unused.

src/Framework/TestCase.php
```diff
@@ -1942,89 +1946,100 @@ private function checkRequirements(): void
...
+        }
 
-            $passedKeys = \array_flip(\array_unique($passedKeys));
+        $passedKeys = \array_flip(\array_unique($passedKeys));
 
-            foreach ($this->dependencies as $dependency) {
-                $deepClone    = false;
-                $shallowClone = false;
+        foreach ($this->dependencies as $dependency) {
+            if (empty($dependency)) {
```
localheinz, line 1969: Since we touch this, what do you think about using something that is a bit more explicit instead of `empty()`?

src/Framework/TestCase.php
```diff
@@ -104,6 +104,16 @@ abstract class TestCase extends Assert implements SelfDescribing, Test
      */
     protected $preserveGlobalState = true;
 
+    /**
+     * @var string[]
```
localheinz, line 108: I have noticed that a few properties have been added which appear to be lists of `string`s.

In some places `@var string[]` is used and in other places `@var array<string>`.

What do you think about using the same notation in all modified places?
```

File: src/Framework/TestResult.php [hunk 12]
```diff
@@ -379,6 +389,8 @@ public function addFailure(Test $test, AssertionFailedError $e, float $time): vo
      */
     public function startTestSuite(TestSuite $suite): void
     {
+        $this->currentTestSuiteIsFailed = false;
```
localheinz, line 392:
```suggestion
        $this->currentTestSuiteHasFailed = false;
```

Perhaps _has failed_ is better?

Or, for consistency with `lastTestFailed`:

```suggestion
        $this->currentTestSuiteFailed = false;
```

src/Framework/TestResult.php
```diff
@@ -578,6 +598,15 @@ public function passed(): array
         return $this->passed;
     }
 
+    /**
+     * Returns the names of the TestSuites that have passed.
+     * This enables @depends-annotations for TestClasName::class
```
localheinz, line 603:
```suggestion
     *
     * This enables @depends-annotations for TestClassName::class.
```

src/Framework/TestCase.php
epdenouden: fixed

src/Framework/TestResult.php
```diff
@@ -578,6 +598,15 @@ public function passed(): array
...
+     */
+    public function passedClasses(): array
```
localheinz, line 605:
```suggestion
    public function passedTestClasses(): array
```

What do you think?

src/Framework/TestCase.php
epdenouden: Added period and empty line after sentence

src/Framework/TestCase.php
epdenouden: yes, I forgot to include the

src/Runner/TestSuiteSorter.php
```diff
@@ -366,19 +359,11 @@ private function resolveDependencies(array $tests): array
...
-                 * @throws \SebastianBergmann\RecursionContext\InvalidArgumentException
-                 */
-                static function ($test) {
-                    return self::getTestSorterUID($test);
-                },
-                $tests
-            );
-
-            if (!$tests[$i]->hasDependencies() || empty(\array_intersect($this->getNormalizedDependencyNames($tests[$i]), $todoNames))) {
+            if (empty(\array_diff($tests[$i]->requires(), $provided))) {
```
localheinz, line 365:
```suggestion
            if ([] === \array_diff($tests[$i]->requires(), $provided)) {
```

src/Framework/TestCase.php
epdenouden: rest of the sentence

src/Util/Test.php
```diff
@@ -353,9 +353,48 @@ public static function getDependencies(string $className, string $methodName): a
             );
         }
 
+        // Normalize dependency name to className::methodName
```
localheinz, line 356: Perhaps it makes sense to extract a method here, so we can replace the comment with a suitable method name?

src/Util/Test.php
```diff
@@ -353,9 +353,48 @@ public static function getDependencies(string $className, string $methodName): a
...
+            if (!empty($testName) && \strpos($testName, '::') === false) {
+                $name = $modifier . $className . '::' . $testName;
+            }
+        }
+
         return \array_unique($dependencies);
     }
 
+    /**
+     * Remove superfluous annotation detail not needed by dependency resolver
```
localheinz, line 378:
```suggestion
     * Remove superfluous annotation detail not needed by dependency resolver.
```

src/Util/Test.php
```diff
@@ -353,9 +353,48 @@ public static function getDependencies(string $className, string $methodName): a
...
+     */
+    public static function trimDependencyOptions(array $dependencies): array
+    {
+        $requires = [];
+
+        foreach ($dependencies as $annotationValue) {
+            if (\trim($annotationValue) !== '') {
+                $annotation            = \explode(' ', $annotationValue, 2);
+                $requires[]            = \end($annotation);
+            }
```
localheinz, line 392:
```suggestion
            if (\trim($annotationValue) === '') {
                continue;
            }
            
            $annotation = \explode(' ', $annotationValue, 2);
            $requires[] = \end($annotation);
```

tests/_files/NumericGroupAnnotationTest.php
```diff
@@ -7,11 +7,17 @@
...
+/**
+ * @author Companion Cube
+ * @ticket t123456
+ */
 class NumericGroupAnnotationTest extends \PHPUnit\Framework\TestCase
 {
     /**
      * @testdox Empty test for @ticket numeric annotation values
      * @ticket  3502
+     * @author C. Lippy
```
localheinz, line 20: 😆

src/Framework/TestCase.php
epdenouden: fixed

src/Framework/TestCase.php
epdenouden: fixed; however no idea how this came to be

src/Framework/TestCase.php
epdenouden: I like it, will change

src/Framework/TestCase.php
epdenouden: That would be better, I agree. Myself I got used to `string[]` but the newer notitation is more flexible. Will bring the new properties in line with the Psalm checks

src/Framework/TestCase.php
localheinz: There's a fixer for this, we could enable it. What do you think?

src/Framework/TestCase.php
localheinz: 😄

src/Framework/TestCase.php
epdenouden: I don't mind either way, really. What does this fixer do, try to guess where sentences end or just add periods and newlines?

src/Framework/TestResult.php
epdenouden: Good catch. Either it's "is marked (as) failed" or "has failed". I think just `-Failed' is better as it is shorter and to-the-point

src/Util/Test.php
localheinz: See #4294.

sebastianbergmann:
Thank you so much, @epdenouden and @localheinz, for working on this!

src/Framework/DataProviderTestSuite.php
```diff
@@ -17,34 +17,44 @@
...
+                // @codeCoverageIgnoreStart
             }
-
             $test->setDependencies($dependencies);
         }
     }
 
-    public function getDependencies(): array
+    /**
+     * @return array<string>
```
localheinz, line 43: What do you think about omitting the DocBlock?

It will be inherited from the parent class.

src/Framework/DataProviderTestSuite.php
```diff
@@ -17,34 +17,44 @@
...
+     */
+    public function provides(): array
     {
-        return $this->dependencies;
+        return [$this->getName()];
     }
 
-    public function hasDependencies(): bool
+    /**
+     * @return array<string>
```
localheinz, line 51: What do you think about omitting the DocBlock?

It will be inherited from the parent class.

src/Framework/TestSuite.php
```diff
@@ -78,11 +78,14 @@ class TestSuite implements \IteratorAggregate, SelfDescribing, Test
     protected $foundClasses = [];
 
     /**
-     * Last count of tests in this suite.
-     *
-     * @var null|int
+     * @var null|array<string>
```
localheinz, line 81: Not sure, but perhaps this makes sense?

```suggestion
     * @var null|array<callable-string>
```

src/Framework/TestSuite.php
```diff
@@ -78,11 +78,14 @@ class TestSuite implements \IteratorAggregate, SelfDescribing, Test
...
+     */
+    protected $providedTests;
+
+    /**
+     * @var null|array<string>
```
localheinz, line 86: Not sure, but perhaps this makes sense?

```suggestion
     * @var null|array<callable-string>
```

src/Framework/TestSuite.php
```diff
@@ -497,22 +499,18 @@ public function addTestFiles(iterable $fileNames): void
...
-        if ($preferCache && $this->cachedNumTests !== null) {
-            return $this->cachedNumTests;
-        }
-
-        $numTests = 0;
+        $this->numTests = 0;
 
         foreach ($this as $test) {
-            $numTests += \count($test);
+            $this->numTests += \count($test);
```
localheinz, line 510: What do you think about making this explicit?

```suggestion
            $this->numTests += $test->count();
```

src/Runner/TestSuiteSorter.php
```diff
@@ -83,12 +83,12 @@ final class TestSuiteSorter
     private $cache;
 
     /**
-     * @var string[] A list of normalized names of tests before reordering
+     * @var array<string> A list of normalized names of tests before reordering
```
localheinz, line 86:
```suggestion
     * A list of normalized names of tests before reordering.
     *
     * @var array<string>
```

src/Runner/TestSuiteSorter.php
```diff
@@ -83,12 +83,12 @@ final class TestSuiteSorter
...
      */
     private $originalExecutionOrder = [];
 
     /**
-     * @var string[] A list of normalized names of tests affected by reordering
+     * @var array<string> A list of normalized names of tests affected by reordering
```
localheinz, line 91:
```suggestion
     * A list of normalized names of tests affected by reordering.
     * 
     * @var array<string> 
```

src/Util/Test.php
```diff
@@ -353,9 +353,48 @@ public static function getDependencies(string $className, string $methodName): a
...
+
+            if (\count($parts) == 1) {
+                $modifier = '';
+                $testName = $parts[0];
+            } else {
+                $modifier = $parts[0] . ' ';
+                $testName = $parts[1];
+            }
+
+            if (!empty($testName) && \strpos($testName, '::') === false) {
```
localheinz, line 369:
```suggestion
            if ('' !==$testName && \strpos($testName, '::') === false) {
```

src/Framework/TestResult.php
epdenouden: fixed

src/Runner/TestSuiteSorter.php
epdenouden: more explicit, fixed

src/Util/Test.php
epdenouden: I agree and this will surely happen in the near future, as I'll keep cleaning up parts of the runner and support logic while working on the Iterator and data providers.

The word 'dependency' is used in multiple contexts at the moment:
- the parsed user input directly from `@depends`
- the dependencies as expressed in the internal test collection tree
- as a bonus it causes the return value of a dependency to be inserted as parameter

src/Util/Test.php
epdenouden: fixed

src/Framework/TestResult.php
epdenouden: I like it better, it's more descriptive. Fixed.

tests/_files/NumericGroupAnnotationTest.php
epdenouden: 📎  It looks like you're writing a PHP test framework. Would you like help?

src/Util/Test.php
epdenouden: The Empire Returns Early

src/Framework/DataProviderTestSuite.php
epdenouden: Hang on, `DataProviderTestSuite::getDependencies()` doesn't exist anymore... what diff is the pull request showing?

epdenouden: Otherwise yes, `@inheritDoc` or automagic is fine with me

src/Framework/DataProviderTestSuite.php
epdenouden: same as with `DPTS::getDependencies()`

src/Framework/TestSuite.php
epdenouden: I did not know that existed. :) Much more descriptive, thx @localheinz

src/Framework/DataProviderTestSuite.php
epdenouden: figured out that Github shows an older version of the source in the inline view here, when clicking through to the full review it shows a newer version.

src/Framework/TestSuite.php
epdenouden: fixed

src/Runner/TestSuiteSorter.php
epdenouden: fixed

src/Runner/TestSuiteSorter.php
epdenouden: fixed

src/Util/Test.php
epdenouden: fixed

src/Framework/TestSuite.php
epdenouden: I will refactor the counting mechanism(s?) for the data provider changes :)

PR #3938, opened 2019-11-08
Fix: Use hash of composer.json instead of commit hash for cache key

localheinz:
This PR

* [x] uses a hash of the contents of `composer.json` instead of the commit hash for the cache key

💁‍♂ For reference, see https://github.com/actions/cache/issues/65.


localheinz:

PR #3941, opened 2019-11-13
Add Windows testing and fix failing tests

johnstevenson:
I've put this on master because it is sort of a new feature, but I'll happily PR it for 7.5 if required.

I see that you're are using Github actions, so you may well be planning to test on Windows there. In the meantime, I've included a simple appveyor configuration for basic testing (similar to what we use on Composer and XdebugHandler).

Test results:
https://ci.appveyor.com/project/johnstevenson/phpunit/builds/28837509

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,0 +1,44 @@
+build: false
+clone_depth: 5
+
+environment:
+  # Set the PHP version (from Chocolatey)
+  PHPCI_CHOCO_VERSION: 7.3.11
+  # Set appropriate Xdebug download link from https://pecl.php.net/package/xdebug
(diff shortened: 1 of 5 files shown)


epdenouden:
ping @localheinz

sebastianbergmann:
We uses AppVeyor in the past for CI on Windows but it is a pain to maintain, for instance because [concrete PHP versions](https://github.com/sebastianbergmann/phpunit/pull/3941/files#diff-11c909939117928998b102a1fff7d363R6) (`7.3.11` instead of `7.3`) need to be configured. As older versions become unavailable the build breaks and the configuration needs to be updated.

sebastianbergmann:
I would like for tests to be run on Windows, but not through a separate service. I will gladly accept a pull request that uses GitHub Actions for this.

johnstevenson:
Ok, will do

sebastianbergmann:
You may want to coordinate with @localheinz.

PR #3943, opened 2019-11-14
Fix: Reduce number of builds

localheinz:
This PR

* [x] reduces the number of builds


localheinz:

PR #1726, opened 2015-05-26
Support to pass expected exception to a callback

UlrichEckhardt:
In some cases I want to inspect more of the thrown exceptions than just the type, message and errorcode. Up to now, I have to do it manually:

```
try {
    $testee->run();
    $this->fail("expected exception not raised");
} catch (\MyException $e) {
    $this->assertEquals(MyException::CATEGORY_FATAL, $e->category());
}
```
This is both error-prone and ugly. The pull request here adds another function to register expected exceptions to complement the existing two. It allows this kind of test instead:
```
$this->setExpectedExceptionCallback(
    'MyException',
    function ($e) {
        $this->assertEquals(MyException::CATEGORY_FATAL, $e->category());
    }
);
$testee->run();
```
An additional advantage is that the wrong thrown exception is handled as usual, giving test failures instead of errors on failure.

PR #3958, opened 2019-11-25
Support to pass an expected exception to a callback for validation

UlrichEckhardt:
Heya!

I found https://github.com/sebastianbergmann/phpunit/pull/1726 in my personal backlog, and rebased in onto the current master branch. Anyhow, so much for the background of this... ;)

### Rationale
In some cases I want to inspect more of the thrown exceptions than just the type, message and errorcode. Up to now, I have to do it manually:

```
try {
    $testee->run();
    $this->fail("expected exception not raised");
} catch (\MyException $e) {
    $this->assertEquals(MyException::CATEGORY_FATAL, $e->category());
}
```

There are a few downsides to this:
 * I need to check the failure to throw explicitly.
 * I should check for wrong exceptions. I don't it above, because it's just too much code.

### Solution
The pull request here adds another function to validate expected exceptions using a callback. It allows this kind of test instead:

```
$this->expectExceptionCallback(
    function (\MyException $e) {
        $this->assertEquals(MyException::CATEGORY_FATAL, $e->category());
    }
);
$testee->run();
```
This code allows you to
 * verify that the exception type matches.
 * verify that the exception is thrown.
 * write custom validation code for the exception type.

What do you think?

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -531,10 +531,36 @@ abstract class TestCase extends Assert implements SelfDescribing, Test
     public function expectExceptionCallback(callable $callback): void
     {
         if (!$this->expectedException) {
-            // TODO: If we could find the argument type of the callback,
-            // we could both tighten this expectation to provide better
-            // diagnostics and validate the callback argument.
-            $this->expectedException = \Exception::class;
+            // default to a regular exception
+            $expectedException = \Exception::class;
+
+            do { // once
+                $fn = new \ReflectionFunction($callback);
+
+                if ($fn->getNumberOfRequiredParameters() > 1) {
+                    throw new Exception('Callback requires more than one parameter.');
+                }
+
+                $params = $fn->getParameters();
+
+                if (\count($params) == 0) {
+                    // callback doesn't have any fixed parameters
+                    break;
+                }
+
+                $type = $params[0]->getType();
+
+                if ($type === null) {
+                    // no type annotation found
+                    break;
+                }
+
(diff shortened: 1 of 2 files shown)


UlrichEckhardt:
Just a note: I tried to fix the issues raised by the autotest. Some of them were pretty easy. However, I'd like some guidance on the remaining two:
 * Firstly, why does it report backward compatibility issues in multiple cases? Is that test relevant to my PR or can it be ignored?
 * Secondly, the type checker. It reports a wrong parameter for the constructor of `ReflectionFunction`, but firstly it works and secondly it's also documented to work that way. Suggestions?

sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it as I believe the current means for expecting/verifying exceptions to be sufficient.

UlrichEckhardt:
Can you elaborate a bit? Of course, this only provides syntactic sugar, but that is IMHO pretty much in line with many other interfaces (`assert...` functions, `Constraint`-generating factory functions).

UlrichEckhardt:
Ping @sebastianbergmann?

On a related note, an even better approach might be to allow passing a `Constraint` to `expectException()`, to reuse existing infrastructure to specify a match.

UlrichEckhardt:
Ping @sebastianbergmann?

sebastianbergmann:
I have said all I have to say on this subject.

UlrichEckhardt:
Okay, you kind-of made your position clear.

Issue #3937, opened 2019-11-08
JUnit's logging of risky tests is inconsistent with other types


oprypkhantc:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 8.4.3
| PHP version         | 7.3
| Installation Method | Composer/PHAR

#### Summary

JUnit's logger error output is inconsistent between an error and a risky test. Real errors include test name in body of `<error>` XML tag, while risky tests don't.

This is a problem as tools like `paratest` rely on that output to display the information. It's almost impossible to detect the source of a risky test when using paratest (which relies on JUnit's logs). 

#### Current behavior

```
<error type="PHPUnit\Framework\RiskyTestError">Risky Test
</error>
```

#### How to reproduce

1. Create an empty test case with an empty test in it.
2. Run phpunit with `--log-junit` option.
3. Open the produced XML file.

#### Expected behavior

Risky tests should either also dump test name or "error", "failure", "warning" should not.
```
<error type="PHPUnit\Framework\RiskyTestError">Tests\Super\Test\Namespace\SomeTest::testSomething
Risky Test
</error>
```
to be consistent with
```
<error type="LogicException">Tests\Super\Test\Namespace\SomeTest::testSomething
LogicException: asdaddsa

/var/www/html/tests/Super/Test/Namespace/SomeTest.php:129
</error>
```

Fedcomp:
When using phpunit with paratest with option to stop tests with timeout (`@small @medium @big` etc) phpunit fails to provide any additional info in JUnit other than "risky test". Can this be reviewed?

Issue #3162, opened 2018-06-07
Support new standard for reporting test results in XML format


sebastianbergmann:
https://github.com/ota4j-team/opentest4j/issues/9#issuecomment-382630222

nfalco79:
The JUnit5 XSD file referenced in #2964 is not by junit. It is a test resource get from jenkins xunit plugin.

ColinLaws:
I hope this input is in the right spot or is appropriate for this thread. I have been having issues with using the test output of PHPUnit with systems that support JUnit. Proper JUnit support would be ideal (at least in the scope of my own personal implementation), as I have seen JUnit as a widely accepted test format.

keradus:
> Proper JUnit support would be ideal 

issue is that nobody knows how proper support would look alike, as there is no official data scheme for JUnit format

nfalco79:
> 
> 
> > Proper JUnit support would be ideal
> 
> issue is that nobody knows how proper support would look alike, as there is no official data scheme for JUnit format

yes and no.
JUnit java library has not define an official schema for XML report. You can choose to follow one of already supported schemas ([maven surefire](http://maven.apache.org/surefire/maven-surefire-plugin/xsd/surefire-test-report.xsd) or [junit ant task](https://github.com/windyroad/JUnit-Schema/blob/master/JUnit.xsd)).

ColinLaws:
@keradus Forgive me, I'm definitely not the expert when it comes to this. I'm just an end user who consumes PHPUnit and would like to use it in Azure DevOps (and they don't support nested test suites). I wish everyone could agree on an open testing standard, but I don't know if there'd be a way to make it fit everyone's needs given there may be language specific data.

mindplay-dk:
JUnit 5 does come with an XSD:

https://github.com/junit-team/junit5/blob/master/platform-tests/src/test/resources/jenkins-junit.xsd

This appears to match most of the third-party XSD's I've been able to find.

Nested `testsuite` elements? I'm surprised it works - does anything besides PHPUnit emit those?

(Kind of crazy how the entire OS world is using JUnit XML as the de-facto standard, while literally nobody seems to know exactly what it is they supposedly agree on - and we all just go along with it...)

epdenouden:
> This appears to match most of the third-party XSD's I've been able to find.

Lovely, that sounds just like the kind of well-described standard to suit a testig framework. 😨
[...]

PR #3960, opened 2019-11-26
Make JUnit's logging of risky tests consistent with other types

oprypkhantc:
Closes https://github.com/sebastianbergmann/phpunit/issues/3937

### DIFF
--- a/src/Util/Log/JUnit.php
+++ b/src/Util/Log/JUnit.php
@@ -179,18 +179,7 @@ class JUnit extends Printer implements TestListener
             return;
         }
 
-        $error = $this->document->createElement(
-            'error',
-            Xml::prepareString(
-                "Risky Test\n" .
-                Filter::getFilteredStacktrace($t)
-            )
-        );
-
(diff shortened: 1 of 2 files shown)


Fedcomp:
can this be reviewed?

sebastianbergmann:
TL;DR: A risky test is not an error.

The slightly longer answer: not all features of PHPUnit can be mapped in a sensible way to "JUnit XML", whatever that may be. See #3162.

autaut03:
There's no alternative for risky tests in JUnit xml other than error. 

Hence, PHPUnit reports it as an error currently. The only change this PR introduces is the addition of stack trace and test name. It has always been an error.

Fedcomp:
Thank you for the answer!
In my case i use paratest and i would like to see additional reasons test failed like phpunit does for risky tests.
Right now in paratest i don't even see the script name! so i consider this pull request a reasonable fix for now. I spent a really long time debugging race condition tests in paratest and this pr would probably save me a lot of time.

PR #3962, opened 2019-11-27
Fix: Do not run roave/backward-compatibility-check on master

localheinz:
This PR

* [x] stops running `roave/backward-compatibility-check` on the `master` branch


localheinz:

PR #3966, opened 2019-12-01
Remove template that doesn’t exist on destination object

muglug:
There was a bug in Psalm where this wasn’t caught - now it will be.


sebastianbergmann:
Merged manually, thanks.

PR #3969, opened 2019-12-05
Fix: Run roave/backward-compatibility-check with ext/intl installed

localheinz:
This PR

* [x] runs `roave/backward-compatibility-check` with `ext/intl` installed


localheinz:

PR #3970, opened 2019-12-06
Fix: Check out entire history for backward-compatibility job

localheinz:
This PR

* [x] checks out the entire history for the `backward-compatibility` job

💁‍♂ Hopefully this fixes the following error:

```
Comparing from 42afe2b8b42077b26148272bd03da4ab6e46a072 to 3ee1c1fd6fc264480c25b6fb8285edefe1702dab...

In Process.php line 255:

  The command "'git' 'checkout' '42afe2b8b42077b26148272bd03da4ab6e46a072'" failed.                                                                       

  Exit Code: 128(Invalid exit argument)                                        

  Working directory: /tmp/api-compare-42afe2b8b42077b26148272bd03da4ab6e46a072_5de9ec71d5270                                                              

  Output:                                                                      

  ================                                                             
  Error Output:                                                                
  ================                                                             

  fatal: reference is not a tree: 42afe2b8b42077b26148272bd03da4ab6e46a072     
```

For reference, see https://github.com/sebastianbergmann/phpunit/runs/336217322#step:4:22.

PR #3971, opened 2019-12-06
Fix: Fetch tags

localheinz:
This PR

* [x] fetches tags before running `roave/backward-compability-check`

Follows #3970.

💁‍♂ For reference, see https://github.com/actions/checkout/issues/85.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,6 +48,9 @@ jobs:
         with:
           fetch-depth: 0
 
+      - name: "Fetch tags"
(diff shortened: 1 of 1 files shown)


localheinz:
@sebastianbergmann 

Perhaps use #3972 instead, what do you think? `actions/checkout@v1` still fetches tags. We could switch to `actions/checkout@v2` once it has been released.

sebastianbergmann:
Makes sense to me. Read this after clicking the "Merge" button :)

localheinz:

😆

PR #3972, opened 2019-12-06
Fix: Avoid using master for GitHub actions

localheinz:
This PR

* [x] avoids using `master` as version constraint for composed GitHub actions

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@master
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -14,7 +14,7 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@master
+        uses: actions/checkout@v1
```
localheinz: Need to see if this conflicts with #3970 and #3971.

localheinz:

PR #3974, opened 2019-12-06
Fix: Avoid using master for GitHub actions

localheinz:
This PR

* [x] avoids using `master` as version constraint for composed GitHub actions

Follows #3972.


localheinz:

PR #3977, opened 2019-12-07
8.5 php notice/warning

jayjfletcher:
After upgrading to 8.5 I noticed the following notice/warning:
Notice: Undefined index: warnings in /vendor/phpunit/phpunit/src/TextUI/TestRunner.php on line 334
Warning:  Invalid argument supplied for foreach() in /vendor/phpunit/phpunit/src/TextUI/TestRunner.php on line 334

this fixes that.

### DIFF
--- a/src/TextUI/TestRunner.php
+++ b/src/TextUI/TestRunner.php
@@ -331,8 +331,7 @@ final class TestRunner extends BaseTestRunner
             }
         }
         
-        if(isset($arguments['warnings']))
-        {
+        if (isset($arguments['warnings'])) {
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request but it works around a symptom instead of solving the problem's root cause: `$arguments['warnings']` should always exist, hence checking for its existence must be superfluous.

I will investigate this further.

sebastianbergmann:
I am not able to reproduce the issue you are trying to fix here. Please open a bug ticket with a minimal, self-contained, reproducing example.

PR #3979, opened 2019-12-09
Enhancement: Add PHP 8 to build matrix

localheinz:
This PR

* [x] adds PHP 8 to the build matrix

💁‍♂ For reference, see https://github.com/shivammathur/setup-php/issues/104.

PR #3980, opened 2019-12-09
Fix: Do not use deprecated inputs

localheinz:
This PR

* [x] stops using deprecated inputs

💁‍♂ For reference, see https://github.com/sebastianbergmann/phpunit/pull/3979/checks?check_run_id=339672157#step:3:1:

Screen Shot 2019-12-09 at 09 53 38


localheinz:

PR #3982, opened 2019-12-09
Windows build automation

johnstevenson:
Update to https://github.com/sebastianbergmann/phpunit/pull/3941

@localheinz I've put the action in a separate file at this stage, just to make it simpler (rather than lots of conditions). Please confirm what sort of configuration you wish on Windows.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -93,10 +93,14 @@ jobs:
   tests:
     name: Tests
 
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
 
     strategy:
       matrix:
+        os:
+          - ubuntu-latest
+          - windows-latest
+
         php-version:
           - 7.2
           - 7.3
@@ -118,14 +122,18 @@ jobs:
           extensions: dom, json, libxml, mbstring, pdo_sqlite, soap, xml, xmlwriter
           ini-values: assert.exception=1, zend.assertions=1
 
-      - name: Determine composer cache directory
-        id: determine-composer-cache-directory
-        run: echo "::set-output name=directory::$(./tools/composer config cache-dir)"
+      - name: Determine composer cache directory on Linux
+        if: matrix.os == 'ubuntu-latest'
+        run: echo "::set-env name=COMPOSER_CACHE_DIR::$(./tools/composer config cache-dir)"
+
+      - name: Determine composer cache directory on Windows
+        if: matrix.os == 'windows-latest'
+        run: ECHO "::set-env name=COMPOSER_CACHE_DIR::~\AppData\Local\Composer"
 
       - name: Cache dependencies installed with composer
         uses: actions/cache@v1
         with:
-          path: ${{ steps.determine-composer-cache-directory.outputs.directory }}
+          path: ${{ env.COMPOSER_CACHE_DIR }}
           key: php${{ matrix.php-version }}-composer-${{ matrix.dependencies }}-${{ hashFiles('**/composer.json') }}
           restore-keys: |
             php${{ matrix.php-version }}-composer-${{ matrix.dependencies }}-
@@ -145,6 +153,6 @@ jobs:
         run: ./phpunit --coverage-clover=coverage.xml
 
       - name: Send code coverage report to Codecov.io
-        env:
-          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
-        run: bash <(curl -s https://codecov.io/bash) || true
+        uses: codecov/codecov-action@v1
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}


johnstevenson:
I've updated the action to use pcov (as per the recent update on master) but since this results in skipping 5 xdebug-required tests, I've included xdebug in the matrix.

I have no idea why coverage has decreased (since this PR has not changed the affected file: `Util/PHP/AbstractPhpProcess.php`).

@localheinz I could do with some guidance here regarding what you want the Windows tests to actually do. Thanks.

johnstevenson:
@sebastianbergmann In the interests of moving this forward, I'll outline what I have done regarding the separate workflow for Windows.

The main (ubuntu) action takes the following steps:
- Coding guidelines (PHP 7.3)
- Type Checker ( PHP7.3)
- Backward Compatibility (PHP 7.3)
- Lint XML Configuration
- Tests (PHP 7.3, 7.4, 8.0) with pcov
  - Lowest dependencies
  - Highest dependencies

Note that 6 tests which require xdebug are skipped because the extension is not loaded. Rather than repeat all this on Windows, I created a separate workflow which only runs:
- Tests (PHP 7.3, 7.4, 8.0)
  - pcov
  - xdebug (but not on 8.0)

It is probably unnecessary to run the tests with pcov and xdebug, but this was my starting point and I was hoping for some feedback. But if you are okay with ignoring the xdebug related tests, I could either:

- remove these from the separate Windows workflow
- or, discard the Windows workflow and  add Windows to the main Tests workflow

Or anything else that is suggested.

johnstevenson:
@localheinz So what needs to be done then?

.github/workflows/ci-win.yml
```diff
@@ -0,0 +1,64 @@
...
+            coverage: xdebug
+        
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
+        with: 
+          fetch-depth: 1
+
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v1
```
localheinz:
```suggestion
        uses: shivammathur/setup-php@v2
```

.github/workflows/ci-win.yml
```diff
@@ -0,0 +1,64 @@
...
+          extensions: curl, intl, mbstring, openssl, pdo_sqlite, soap          
+          ini-values: assert.exception=1, zend.assertions=1                  
+
+      - name: Restore cached composer dependencies 
+        uses: actions/cache@v1
+        with:
+          path: ~\AppData\Local\Composer
+          key: ${{ matrix.php-versions }}-composer-${{ hashFiles('**/composer.json') }}
+          restore-keys: |
+            ${{ matrix.php-versions }}-composer-
```
localheinz: What do you think about letting `composer` determine the cache directory instead?

```suggestion
      - name: Determine composer cache directory
        id: determine-composer-cache-directory
        run: echo "::set-output name=directory::$(php ./tools/composer config cache-dir)"
          
      - name: Restore cached composer dependencies 
        uses: actions/cache@v1
        with:
          path: ${{ steps.determine-composer-cache-directory.outputs.directory }}
          key: ${{ matrix.php-versions }}-composer-${{ hashFiles('**/composer.json') }}
          restore-keys: |
            ${{ matrix.php-versions }}-composer-      
```

.github/workflows/ci-win.yml
localheinz: Also see #4153.

.github/workflows/ci-win.yml
localheinz: Also see #4155.

localheinz:
@johnstevenson 

Not sure about Windows, I haven't used it in years - you probably know more than I do!

What do you think, though, about reordering commits, so we can see how they are related:

- set up GitHub Actions for Windows
- fix failures related to running tests on Windows

🤓

localheinz:
@sebastianbergmann 

Slightly related to this pull request, do you think it would make sense to rename the `CI` workflow to something like `CI-Linux`?

.github/workflows/ci.yml
```diff
@@ -145,6 +153,6 @@ jobs:
         run: ./phpunit --coverage-clover=coverage.xml
 
       - name: Send code coverage report to Codecov.io
-        env:
-          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
-        run: bash <(curl -s https://codecov.io/bash) || true
```
localheinz: Looks like this does not work in Windows, that's why I switched to using the action.

localheinz:
@johnstevenson @sebastianbergmann 

What do you think, does that make sense?

localheinz (approved):

.github/workflows/ci.yml
```diff
@@ -118,14 +122,18 @@ jobs:
...
+
+      - name: Determine composer cache directory on Windows
+        if: matrix.os == 'windows-latest'
+        run: ECHO "::set-env name=COMPOSER_CACHE_DIR::~\AppData\Local\Composer"
 
       - name: Cache dependencies installed with composer
         uses: actions/cache@v1
         with:
-          path: ${{ steps.determine-composer-cache-directory.outputs.directory }}
+          path: ${{ env.COMPOSER_CACHE_DIR }}
```
localheinz: I think that by making the path depend on the environment variable, then setting the environment variable depending on which operating system we are running on, the entire thing boils down to adding an `os` matrix, and we can skip the separate workflow file!

👍

localheinz:
Thank you, @johnstevenson and @sebastianbergmann!

.github/workflows/ci.yml
johnstevenson: You need to specify the bash shell on Windows because Powershell is the default, so:
```
        run: bash <(curl -s https://codecov.io/bash) || true
        shell: bash
```

Perhaps the action is better though, for future maintenance.

johnstevenson:
@localheinz Cool. Thanks for your work.

PR #3990, opened 2019-12-17
Add CLI option to restart without Xdebug

johnstevenson:
This adds a `--no-xdebug` CLI option to restart the main process without
loading the Xdebug extension, using the composer/xdebug-handler package.

The process is restarted using persistent settings, which ensures that
Xdebug is not loaded in any PHP sub-process.

This may are may not be something you wish to incorporate. If it is, then it is perhaps worth also considering an alternative strategy that would automatically remove Xdebug unless it is needed.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,6 +29,7 @@
         "ext-mbstring": "*",
         "ext-xml": "*",
         "ext-xmlwriter": "*",
+        "composer/xdebug-handler": "^1.4",
         "doctrine/instantiator": "^1.2.0",
         "myclabs/deep-copy": "^1.9.1",
         "phar-io/manifest": "^1.0.3",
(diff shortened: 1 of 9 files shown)


sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it as I do not think that functionality such as this has a place in PHPUnit itself. If you do not need Xdebug then simply do not load the extension.

johnstevenson:
Okay, no worries.

moosapour-arash:
if this works without any problems, i find it very useful.
i have a docker image with xdebug installed in my CI, i want to calculate test coverage with xdebug only in certain branches (e.g master) but not in others because it takes a lot more time. at the moment i have to do some hackish way(change xdebug.ini in bash) to unload the xdebug before running tests. if i could tell the phpunit to ignore it, that would be much simpler

PR #3994, opened 2019-12-22
Use better typing for internal property

muglug:
This was caught by the latest version of Psalm, and should make things a little more resilient.


sebastianbergmann:
Merged manually into `8.5`, thanks!

Issue #4012, opened 2020-01-11
Remove class name as CLI argument


flow-control:
Hello there,

with #3859 we deprecated calling PHPUnit with a class name. This deprecation was released with PHPUnit `v8.5` and we can remove calling with a class name in the upcoming PHPUnit `v9`.

This Issue is about removing calling PHPUnit with a class name.
I will create a pull request in the next few days.

/Flo

PR #4013, opened 2020-01-11
Remove class name as CLI option

realFlowControl:
This pull request removes the class name as CLI option as stated in #4012


sebastianbergmann:
Looks good so far. Now we only need to remove the actual code :) Do you know when you will have time to finish this? Would be great to have to this ready by the end of next week. Thanks!

flow-control:
End of next week sound feasible as i was planing to invest more time this weekend

flow-control:
Hey @sebastianbergmann,

not yet finalized, but I will continue tomorrow.
One thing I like to notice: In the current state I removed the magic, that looked around in the included file to find _some_ class that is not `abstract` and a subclass of `\PHPUnit\Framework\TestCase` if everything else fails. I could easily add this behavior again and was removing it because I think this should be part of this cleanup. Just ping me if you like to have that fallback added back again.

/Flo

flow-control:
Rebased on `master` branch

flow-control:
I'll have a look at the code coverage report. That decreased with rebasing on `master` branch :disappointed:

Issue #4001, opened 2020-01-06
PHP 7.4 fixtures


dpauli:
There are some little changes in PHP 7.4 with could break the normal usage of PHPUnit. Perhaps these things can be fixed:

- Failure return values of `fread()` is now `false` and not `''`, see https://github.com/sebastianbergmann/phpunit/blob/master/src/Util/PHP/DefaultPhpProcess.php#L147

sebastianbergmann:
A pull request for this would be appreciated.

PR #4019, opened 2020-01-17
#4001 Check for 'false' line for PHP 7.4.

panvid:
Fix PHP 7.4. combatibility.


sebastianbergmann:
[PHPUnit 7.5 is not supported on PHP 7.4.](https://phpunit.de/supported-versions.html)

PR #4037, opened 2020-02-10
Bug: can't rely on already loaded class

Slamdunk:
Hi, in order to be able to have a feedback for each test, but run them quickly, [Paratest](https://github.com/paratestphp/paratest) calls within a single process `PHPUnit\TextUI\Command::main(false)` [multiple times](https://github.com/paratestphp/paratest/blob/master/bin/phpunit-wrapper).

Starting for PHPUnit 9.0, if a test already loaded another test class, `PHPUnit\Runner\StandardTestSuiteLoader` fails to get a useful diff here:

https://github.com/sebastianbergmann/phpunit/blob/ec3adec334f7bbe8c61607bdacefbf5046a6df94/src/Runner/StandardTestSuiteLoader.php#L28-L36

hence throws a `ClassNotFoundException`, even though the class is perfectly already loaded.

The bug is present also in 8.5, but the workaround was to specify the class name alongside the class file, behavior suppressed in 9.0

Ping @andreasschroth

### DIFF
--- a/tests/end-to-end/regression/GitHub/4037-namespaced.phpt
+++ b/tests/end-to-end/regression/GitHub/4037-namespaced.phpt
@@ -1,0 +1,34 @@
+--TEST--
+https://github.com/sebastianbergmann/phpunit/pull/4037
+--FILE--
+<?php declare(strict_types=1);
+require __DIR__ . '/../../../bootstrap.php';
+
+$_SERVER['argv'][1] = '--configuration';
+$_SERVER['argv'][2] = __DIR__ . '/4037-namespaced/';
+$_SERVER['argv'][3] = __DIR__ . '/4037-namespaced/LastNamespacePart/Issue4037ATest.php';
+
+PHPUnit\TextUI\Command::main(false);
+
+$_SERVER['argv'][1] = '--configuration';
+$_SERVER['argv'][2] = __DIR__ . '/4037-namespaced/';
+$_SERVER['argv'][3] = __DIR__ . '/4037-namespaced/LastNamespacePart/Issue4037BTest.php';
+
+PHPUnit\TextUI\Command::main(false);
+
+@unlink(__DIR__ . '/4037-namespaced/.phpunit.result.cache');
+--EXPECTF--
+PHPUnit %s by Sebastian Bergmann and contributors.
+
+.                                                                   1 / 1 (100%)
+
+Time: %s, Memory: %s
+
+OK (1 test, 1 assertion)
+PHPUnit %s by Sebastian Bergmann and contributors.
+
+.                                                                   1 / 1 (100%)
+
+Time: %s, Memory: %s
+
+OK (1 test, 1 assertion)
--- a/tests/end-to-end/regression/GitHub/4037-namespaced/LastNamespacePart/Issue4037ATest.php
+++ b/tests/end-to-end/regression/GitHub/4037-namespaced/LastNamespacePart/Issue4037ATest.php
@@ -7,7 +7,6 @@
  * For the full copyright and license information, please view the LICENSE
  * file that was distributed with this source code.
  */
-
 namespace FirstNamespacePart\LastNamespacePart;
 
 use PHPUnit\Framework\TestCase;
--- a/tests/end-to-end/regression/GitHub/4037-namespaced/LastNamespacePart/Issue4037ATest.php
+++ b/tests/end-to-end/regression/GitHub/4037-namespaced/LastNamespacePart/Issue4037ATest.php
@@ -1,0 +1,22 @@
+<?php declare(strict_types=1);
+/*
+ * This file is part of PHPUnit.
+ *
+ * (c) Sebastian Bergmann <sebastian@phpunit.de>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace FirstNamespacePart\LastNamespacePart;
+
+use PHPUnit\Framework\TestCase;
+
+final class Issue4037ATest extends TestCase
+{
+    public function testA(): void
+    {
+        require __DIR__ . '/Issue4037BTest.php';
+        $this->assertTrue(Issue4037BTest::ok());
+    }
+}
--- a/tests/end-to-end/regression/GitHub/4037-namespaced/LastNamespacePart/Issue4037BTest.php
+++ b/tests/end-to-end/regression/GitHub/4037-namespaced/LastNamespacePart/Issue4037BTest.php
@@ -1,0 +1,26 @@
+<?php declare(strict_types=1);
+/*
+ * This file is part of PHPUnit.
+ *
+ * (c) Sebastian Bergmann <sebastian@phpunit.de>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace FirstNamespacePart\LastNamespacePart;
+
+use PHPUnit\Framework\TestCase;
+
+final class Issue4037BTest extends TestCase
+{
+    public static function ok(): bool
+    {
+        return true;
+    }
+
+    public function testB(): void
+    {
+        $this->assertTrue(true);
+    }
+}
--- a/tests/end-to-end/regression/GitHub/4037-namespaced/LastNamespacePart/Issue4037BTest.php
+++ b/tests/end-to-end/regression/GitHub/4037-namespaced/LastNamespacePart/Issue4037BTest.php
@@ -7,7 +7,6 @@
  * For the full copyright and license information, please view the LICENSE
  * file that was distributed with this source code.
  */
-
 namespace FirstNamespacePart\LastNamespacePart;
 
 use PHPUnit\Framework\TestCase;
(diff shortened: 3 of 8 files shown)


sebastianbergmann:
CC @flow-control

flow-control:
Hey there,

thanks for also providing a test case. I'll take a look.

/Flo

Slamdunk:
The issue gets tricky when considering namespaced test class :/

flow-control:
Correct, in this case https://github.com/sebastianbergmann/phpunit/blob/ec3adec334f7bbe8c61607bdacefbf5046a6df94/src/Runner/StandardTestSuiteLoader.php#L28
should have you covered, but it is not, as this checks relative to the `PHPUnit\Runner` namespace ;-)

So the first easy fix for this should be to change that check to:
```php
if (!\class_exists('\' . $suiteClassName, false)) {
```
But this will not help you with namespaced tests. As i can not commit on your branch, would you mind making this proposed change?
I will have to think about a nice solution for namespaced test class

flow-control:
@Slamdunk forgot what i have written, should have tested this out before writing ;-)

flow-control:
I have a fix for the non namespaced classes, but I don't know how to get this into your branch. I could post the diff, but there is probably a better solution.
@Slamdunk How should we continue from here?

Slamdunk:
> As i can not commit on your branch, would you mind making this proposed change?

You can always commit a PR to any public repo :)

I've pushed a namespaced test case too.

A first solution that comes into my mind is to make `StandardTestSuiteLoader` stateful: track every every `TestCase` subclass found (along with their filename) and use the saved list when needed.

flow-control:
> > As i can not commit on your branch, would you mind making this proposed change?
> 
> You can always commit a PR to any public repo :)

Haha, right. I go an open up a pull request to your branch then.

> A first solution that comes into my mind is to make `StandardTestSuiteLoader` stateful: track every every `TestCase` subclass found (along with their filename) and use the saved list when needed.

That could work, I will tinker with that. The problem I see is that having it state full makes it harder to test it.

Slamdunk:
> That could work, I will tinker with that. The problem I see is that having it state full makes it harder to test it.

Of course, but there aren't much alternatives:

1. Keep track of what already is in memory, thus making something stateful (`StandardTestSuiteLoader` or its user)
1. `file_get_contents` with an AST analyzer, stupid (regex) or clever (tokenizer) one

The second one is pretty, but heavy I think.

Slamdunk:
The third option of using a sub-process is a no-go for everyone, I think we agree on this :)

flow-control:
> A first solution that comes into my mind is to make `StandardTestSuiteLoader` stateful: track every every `TestCase` subclass found (along with their filename) and use the saved list when needed.

Your test case is a hard nut to crack ;-)
The `Issue4037BTest.php` is not loaded while `Issue4037ATest.php` is loaded, so the `StandardTestSuiteLoader` does not load it itself (and therefore does not have this class in the `$loadedClasses` array). I can not only rely on `get_declared_classes()` because there might be another class named `Issue4037BTest` in another namespace ;-)

I am playing with `token_get_all()` to find about the internals of the file (namespace and class name), sounds like the correct solution to me, what do you think?

sebastianbergmann:
> I am playing with token_get_all() to find about the internals of the file (namespace and class name), sounds like the correct solution to me, what do you think?

No. The point of the changed made in PHPUnit was to make our code less complicated. Introducing static analysis of test source files (if only using the tokenizer) is the opposite of "less complicated".

While we're on the subject of `StandardTestSuiteLoader`: I would very much like to get rid of the support for changing the way PHPUnit loads tests (basically anything related to the `TestSuiteLoader` interface and the `StandardTestSuiteLoader` implementation thereof as well as the `testSuiteLoaderClass` and `testSuiteLoaderFile` configuration settings sooner rather than later. AFAICS, there is no benefit to this and the downside is, again, hard-to-maintain code.

Slamdunk:
> I am playing with `token_get_all()` to find about the internals of the file (namespace and class name), sounds like the correct solution to me, what do you think?

That's the second solution I've listed: my experience on developing [PHP-CS-Fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer) tells me it can be slow and error-prone.

Saving a map of filename-to-classes seems light and easy to maintain, to me.

You should note that `StandardTestSuiteLoader` is **already** stateful: relying on `get_declared_classes` function, which is stateful, makes it stateful as well!

sebastianbergmann:
@Slamdunk I briefly looked at the test case you submitted in this pull request. I do not understand it. Both https://github.com/sebastianbergmann/phpunit/pull/4037/files#diff-d94cdd7abdc97673d2933ce1c7ad8b03 and https://github.com/sebastianbergmann/phpunit/pull/4037/files#diff-0d3b7b035b6000760f4dd2808c2b22dd seem to be two separate test cases implemented in a single PHPT case. Or this about the fact that `PHPUnit\TextUI\Command::main()` does not work when invoked multiple times within the same PHP process? If that is the case then my response to this ticket will be "won't fix".

@flow-control Does the bug reported here affect PHPUnit itself in any way? If not then I am inclined to say that we should not fix it.

Slamdunk:
> Or this about the fact that `PHPUnit\TextUI\Command::main()` does not work when invoked multiple times within the same PHP process?

Exactly

> If that is the case then my response to this ticket will be "won't fix".
> 
> @flow-control Does the bug reported here affect PHPUnit itself in any way? If not then I am inclined to say that we should not fix it.

I can understand your will to close this as "won't fix", but I'd like to remark that relying on diffing on `get_declared_classes` multiple times, without considering it a global state, is brittle as hell in my POV.

sebastianbergmann:
Our goal with the refactorings that we have been doing for quite a while, @flow-control's work on test loading being part of this year's cleanup for PHPUnit, are clear: reduce complexity to make PHPUnit's code easier to understand, maintain, and contribute to.

I recognize that a solution for running tests in parallel is desirable. Based on what I have learned from tickets such as this one (I have never looked at Paratest's design, architecture, or implementation in detail), I believe that Paratest's approach is flawed. I am also not convinced that parallel test execution belongs into PHPUnit's own test runner. I outlined my solution for parallel test execution in #3838.

I also believe that GitHub tickets are the wrong medium to discuss a topic as complex and important as this one. I think it would be beneficial for everyone involved, PHPUnit and Paratest developers alike, if we met and discussed this in person. I would very much appreciate if somebody from the Paratest development team would visit a [PHPUnit code sprint](https://phpunit.de/code-sprints/index.html).

> I can understand your will to close this as "won't fix", but I'd like to remark that relying on diffing on get_declared_classes multiple times, without considering it a global state, is brittle as hell in my POV.

Which is one of the reasons I want to remove it for PHPUnit 10.

flow-control:
> @flow-control Does the bug reported here affect PHPUnit itself in any way? If not then I am inclined to say that we should not fix it.

If you have both test cases in the same directory, and then in `Issue4037ATest::testA` require the file `Issue4037BTest.php` (which is already loaded from the SuiteLoader), you get an exception, but this originates from the TestClass and not from PHPUnit.

```bash
$ ./phpunit --no-configuration tests/end-to-end/regression/GitHub/4037/
PHPUnit 9.0.0-13-gdada762b7 by Sebastian Bergmann and contributors.

PHP Fatal error:  Cannot declare class Issue4037BTest, because the name is already in use in /home/bowman/Work/flow-control/phpunit/tests/end-to-end/regression/GitHub/4037/Issue4037BTest.php on line 0
PHP Stack trace:
PHP   1. {main}() /home/bowman/Work/flow-control/phpunit/phpunit:0
PHP   2. PHPUnit\TextUI\Command::main() /home/bowman/Work/flow-control/phpunit/phpunit:61
PHP   3. PHPUnit\TextUI\Command->run() /home/bowman/Work/flow-control/phpunit/src/TextUI/Command.php:160
PHP   4. PHPUnit\TextUI\TestRunner->run() /home/bowman/Work/flow-control/phpunit/src/TextUI/Command.php:200
PHP   5. PHPUnit\Framework\TestSuite->run() /home/bowman/Work/flow-control/phpunit/src/TextUI/TestRunner.php:636
PHP   6. PHPUnit\Framework\TestSuite->run() /home/bowman/Work/flow-control/phpunit/src/Framework/TestSuite.php:597
PHP   7. Issue4037ATest->run() /home/bowman/Work/flow-control/phpunit/src/Framework/TestSuite.php:597
PHP   8. PHPUnit\Framework\TestResult->run() /home/bowman/Work/flow-control/phpunit/src/Framework/TestCase.php:772
PHP   9. Issue4037ATest->runBare() /home/bowman/Work/flow-control/phpunit/src/Framework/TestResult.php:692
PHP  10. Issue4037ATest->runTest() /home/bowman/Work/flow-control/phpunit/src/Framework/TestCase.php:1028
PHP  11. Issue4037ATest->testA() /home/bowman/Work/flow-control/phpunit/src/Framework/TestCase.php:1408
[bowman@rocinante phpunit]$ vim /home/bowman/Work/flow-control/phpunit/src/Framework/TestCase.php
```

I don't see a use case for this, the Exception comes from within the Test itself, not from PHPUnit and you could fix this by using `require_once`.

I understand that this can somehow happen in https://github.com/paratestphp/paratest, it might happen in PHPUnit also, but only from a test class, not from the core.

Slamdunk:
> I outlined my solution for parallel test execution in #3838.

I'm sorry @sebastianbergmann #3838 seems unrelated: which is the right issue/pr to look at?

sebastianbergmann:
I meant #3387, sorry.

andreasschroth:
@sebastianbergmann I'm one of the maintainers of Paratest now. :-) We met a few times for consulting some years ago if you still remember. Not sure if I can make it to the PHPUnit Code Spirit, but thanks for the invitation. I'll see if I can make it, but no promises.

PR #4040, opened 2020-02-10
Deprecate TestSuiteLoader

realFlowControl:
This pull request will deprecate `TestSuiteLoader`, `StandardTestSuiteLoader` and the `testSuiteLoaderClass` and `testSuiteLoaderFile` configuration settings.

See #4039

PR #4042, opened 2020-02-11
Fix Undefined index: disableCodeCoverageIgnore

realFlowControl:
This will fix the `Notice: Undefined index: disableCodeCoverageIgnore in TestRunner.php:549` when you you have xDebug extension enabled and call without the `--disable-coverage-ignore` CLI argument

From: https://github.com/sebastianbergmann/phpunit/pull/4040#issuecomment-584516329

### DIFF
--- a/src/TextUI/TestRunner.php
+++ b/src/TextUI/TestRunner.php
@@ -546,7 +546,7 @@ final class TestRunner extends BaseTestRunner
                 );
             }
 
-            if ($arguments['disableCodeCoverageIgnore']) {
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it because it is easier for me to just revert the commit that broke this.

sebastianbergmann:
Interestingly enough, I do not get that notice neither locally (`error_reporting=-1`) nor on GitHub actions.

flow-control:
Interesting indeed, because I have this error locally.

PR #4043, opened 2020-02-11
Fix mocking with anonymous classes

villfa:
A PHP Parse error is thrown when trying to mock an anonymous class.

I encountered this error by writing a test this way:
```php
public function testRunner()
{
    $mock = $this->createMock(new class {
        public function foo() {}
    });
    $mock->expects($this->once())->method('foo');

    $runner = new Runner();
    $runner->run(function () use ($mock) {
        $mock->foo();
    });
}
```
I recognize it's a bit convoluted (I have rewritten the test without using any mock then) but I hope this fix will help someone with a valid use case.

### DIFF
--- a/src/Framework/MockObject/Generator.php
+++ b/src/Framework/MockObject/Generator.php
@@ -902,6 +902,15 @@ final class Generator
             $type = \implode('_', $type);
         }
 
+        // in case of an anonymous class we have to use an alias because of the special characters
+        if (\strpos($type, '@') !== false && \class_exists($type, false)) {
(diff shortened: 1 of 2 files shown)


sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it as doubling an anonymous class makes no sense to me.

If at all, then the code generator should be changed to raise an exception which in turns leads to PHPUnit flagging the test as broken (with a warning).

PR #4095, opened 2020-02-13
Performance improvement of StringContains

willemstuursma:
I've created a faster version of StringContains, in particular for large haystacks. 

I noticed a 4.5x speed up for asserting 1,000x that a 2MB string contains an other string. 

We don't need the multi-byte ware function `mb_strpos` as it is approximately O(n^2) where `n` is the length of the haystack. 

* We don't care about the _character_ position of the substring. 
* In most encodings, the same characters will always be represented by the same bytes. This holds for ISO-8859-1 and friends, and UTF-8. 

In UTF-16 and UTF-32, it is _theoretically_ possible that same byte sequences map to different characters, but as all PHPUnit's own output is not compatible with this encoding I don't think it will cause any issues in practice as I don't think anyone is running PHPUnit with PHP's internal encoding set to UTF-16/UTF-32.

NB. There is the issue that UTF-8 allows the same _characters_ to be represented by different combinations of unicode code points (and thus bytes), but that wasn't supported by `mb_strpos` anyway as it doesn't normalize the string internally.

### DIFF
--- a/src/Framework/Constraint/StringContains.php
+++ b/src/Framework/Constraint/StringContains.php
@@ -66,9 +66,21 @@ final class StringContains extends Constraint
         }
 
         if ($this->ignoreCase) {
+            /*
+             * We must use the multi byte safe version so we can accurately compare non latin upper characters with
+             * their lowercase equivalents.
+             */
             return \mb_stripos($other, $this->string) !== false;
         }
 
-        return \mb_strpos($other, $this->string) !== false;
+        /*
+         * Use the non multi byte safe functions to see if the string is contained in $other.
+         *
+         * This function is very fast and we don't care about the character position in the string.
+         *
+         * Additionally, we want this method to be binary safe so we can check if some binary data is in other binary
+         * data.
+         */
+        return \strpos($other, $this->string) !== false;
     }
 }


be-heiglandreas:
The only way to normalize strings in unicode would be to use `\Normalizer::normalize($string, Normalizer::FORM_C)` on both strings before passing them to the `strpos`-function. 

But TBH I would leave that to the person writing the test to be able to find issues with different normalizations (which can be a PITA especially with filenames on MacOS-systems - been there...)

And as that will only work on unicode-strings it would require a lot of guesswork before the actual `strpos` regarding the current strings encoding. 

Additionally while it might be possible to find a binary match in UTF16 or UTF32 when looking for a single character the possibility to find such a match decreases rapidly with the number of characters tried to match. This seems such an edge-case that the possible performance increase for the majority of the use-cases IMO verifies this risk.

be-heiglandreas (approved):
The only way to normalize strings in unicode would be to use `\Normalizer::normalize($string, Normalizer::FORM_C)` on both strings before passing them to the `strpos`-function. 

But TBH I would leave that to the person writing the test to be able to find issues with different normalizations (which can be a PITA especially with filenames on MacOS-systems - been there...)

And as that will only work on unicode-strings it would require a lot of guesswork before the actual `strpos` regarding the current strings encoding. 

Additionally while it might be possible to find a binary match in UTF16 or UTF32 when looking for a single character the possibility to find such a match decreases rapidly with the number of characters tried to match. This seems such an edge-case that the possible performance increase for the majority of the use-cases IMO verifies this risk.

heiglandreas (approved):

src/Framework/Constraint/StringContains.php
```diff
@@ -66,9 +66,21 @@ protected function matches($other): bool
         }
 
         if ($this->ignoreCase) {
+            /*
+             * We must use the multi byte safe version so we can accurately compare non latin upper characters with
+             * their lowercase equivalents.
+             */
             return \mb_stripos($other, $this->string) !== false;
```
heiglandreas: To speed the case-insensitive comparison up we could use `$other = mb_strtolower($other)` (as used in line 45) here to convert the searched phrase to lower case which is much faster here than using `mb_stripos` and then do the comparison binary safe in line 84.

heiglandreas (commented):
This modification would increase speed not only for binary safe comparisons but also for case-insensitive comparisons.

sebastianbergmann:
Thank you for your review, @heiglandreas.

src/Framework/Constraint/StringContains.php
sebastianbergmann: Feel free to send a pull request for this.

willemstuursma:
Thanks @sebastianbergmann.

PR #4102, opened 2020-02-17
Add missing deprecation warnings

pl-github:
When switching the tests to PHPUnit 9, a few errors occurred because some methods were removed in PHPUni 9 that did not yet give a deprecation warning in PHPUnit 8.

**What I have changed:**

- Assert::contains()
  - added deprecation warning
  - removed testAssertThatContains() (seems to be the normal behavior for deprecated code)
- TestCase:: expectExceptionMessageRegExp()
  - added deprecation warning
  - replaced calls in tests with expectExceptionMessageMatches() to supress warnigs in test runs
- TestCase::setUseErrorHandler()
  - added deprecation warning

### DIFF
--- a/src/Framework/Assert.php
+++ b/src/Framework/Assert.php
@@ -3128,6 +3128,8 @@ abstract class Assert
      */
     public static function contains($value, bool $checkForObjectIdentity = true, bool $checkForNonObjectIdentity = false): TraversableContains
     {
+        self::createWarning('contains() is deprecated and will be removed in PHPUnit 9. Use containsEqual() or containsIdentical() instead.');
+
         return new TraversableContains($value, $checkForObjectIdentity, $checkForNonObjectIdentity);
     }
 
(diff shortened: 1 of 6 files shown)


sebastianbergmann:

`assertContains()` is not deprecated, and neither is `contains()`. `expectExceptionMessageRegExp()` and `setUseErrorHandler()` have `@deprecated` annotations ("soft deprecation") which has to be enough as I do not think that we should add "hard deprecations" in a bugfix release.

pl-github:
You are right, it is only a small benefit and I didn`t know about the soft / hard deprecation strategy.

Many thanks for your continuous work on PHPUnit.

It`s true `assertContains()` isn`t deprecated, but for me `contains()` looks deprecated, too.

```php
    /**
     * @deprecated Use containsEqual() or containsIdentical() instead
     */
    public static function contains($value, bool $checkForObjectIdentity = true, bool $checkForNonObjectIdentity = false): TraversableContains
```

PR #4127, opened 2020-02-29
Tests for assertFileEqualsIgnoringCase

simivar:
#4050


sebastianbergmann:
Merged manually (because this also needs to be in `8.5` and `9.0`.

Issue #4144, opened 2020-03-24
Empty objects are converted to empty arrays in JSON comparison failure diff


uuf6429:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 8.5.2
| PHP version         | 7.4.3
| Installation Method | Composer

#### Summary

If asserting equality between 2 json strings fails, any empty objects (`{}`) are converted to empty arrays (`[]`) causing a different diff than expected.

At first, I saw this in PHPStorm and thought it's a PHPStorm issue, but apparently it's not ([link to PHPStorm report](https://youtrack.jetbrains.com/issue/WI-51891#focus=streamItem-27-4026117.0-0)).

![image](https://user-images.githubusercontent.com/230049/77436173-753a9a00-6de3-11ea-8642-e60b51acf2d4.png)

The problem seems to be caused by [Json.php:26](https://github.com/sebastianbergmann/phpunit/blob/master/src/Util/Json.php#L26) (the `true` parameter). I think all PHPUnit versions are affected.

#### Current behavior

```diff
Failed asserting that '{"key": {}}' matches JSON string "{"key": {"sub": 1}}".
--- Expected
+++ Actual
@@ @@
 {
-    "key": {
-        "sub": 1
-    }
+    "key": []
 }
```

#### Expected behavior

```diff
Failed asserting that '{"key": {}}' matches JSON string "{"key": {"sub": 1}}".
--- Expected
+++ Actual
@@ @@
 {
-    "key": {
-        "sub": 1
-    }
+    "key": {}
 }
```

#### How to reproduce

Look at the results of the test failure:
```php
    public function testPoC(): void
    {
        $this->assertJsonStringEqualsJsonString(
            '{"key": {"sub": 1}}',
            '{"key": {}}'              // <- notice that it's an empty object, not array
        );
    }
```

uuf6429:
I also just noticed another issue (also related to the display of json diffs). The values displayed in the diff are not the ones being compared (since the value is canonicalized first). This leads to larger diffs that tend to hide the real issue.

sebastianbergmann:
@localheinz Can you have a look at this? You know this functionality better than me.

localheinz:
@sebastianbergmann @uuf6429 

Apologies for the delay!

Looks like you already opened a few pull requests, @uuf6429 - I think we can go with #4145 and then merge `8.5` into `9.0`, `9.0` into `master`.

👍

PR #4146, opened 2020-03-24
Fix json diff failure issues (9.0)

uuf6429:
See #4144


localheinz:
@sebastianbergmann 

Are we going to merge #4145 instead, and then merge `8.5`?

sebastianbergmann:
Duplicate of #4145 which I will merge into `8.5` and from there to `9.0` and `master`.

PR #4147, opened 2020-03-24
Fix json diff failure issues (master)

uuf6429:
See #4144


localheinz:
@sebastianbergmann 

See https://github.com/sebastianbergmann/phpunit/pull/4146#issuecomment-605608691.

sebastianbergmann:
Duplicate of #4145 which I will merge into `8.5` and from there to `9.0` and `master`.

PR #4151, opened 2020-03-26
Exclude internal tests from composer dist packages

Tobion:
See discussion for example https://github.com/symfony/symfony/pull/33579

It was already done in phpunit in https://github.com/sebastianbergmann/phpunit/commit/d99e588e3cbab28db86f8da22e22c72d8cc38543 but reverted because it also removed the changelog and license file by mistake.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,7 +1,16 @@
-/.docker    export-ignore
-/build      export-ignore
-/tools      export-ignore
-/tools/*    binary
+/.docker export-ignore
+/.github export-ignore
+/.psalm export-ignore
+/build export-ignore
+/tests export-ignore
+/tools export-ignore
+/tools/* binary
+/.editorconfig export-ignore
+/.gitattributes export-ignore
+/.gitignore export-ignore
+/.php_cs.dist export-ignore
+/build.xml export-ignore
+/phive.xml export-ignore
+/phpunit.xml export-ignore
 
 *.php diff=php
-


sebastianbergmann:

simonschaufi:
@sebastianbergmann could you explain why you left out `/tests export-ignore`? Nobody needs the internal tests of phpunit. I also found this interesting comment: https://github.com/Codeception/Codeception/pull/5527#issue-282555074 and this one: https://www.emergeinteractive.com/insights/detail/does-irresponsible-web-development-contribute-to-global-warming/ Save the :earth_africa: and reduce global warming!

sebastianbergmann:
AFAIK, @remicollet and other downstream packagers rely on tests etc. being included in release tarballs. I would gladly add them to `export-ignore` if that were not the case.

Tobion:
They can still have the tests by using prefer-source. That's not a valid argument to not change it.

sebastianbergmann:
I think you are right. And I think that `prefer-source` may not have existed when @remicollet asked me to not exclude tests. I will go through my repositories and make the necessary changes.

remicollet:
This has been discussed ad-nauseam.

"prefer-source' is a composer option, so only for composer user.

tests are part of the project so have to be part of the public tarball, like documentation, license, or any other file used by the project.

Again THIS IS A COMPOSER ISSUE, not being able to manage file "roles"

PHP is the only language in the world to NOT have any INSTALLER (and composer is not an installer, only a dependency manager).

perl have cpan,
python have pip
nodejs have npm
...

Even the old PEAR was able to manage file "roles"

BTW, using git snapshot instead of official tarball still possible

This issue make all non-php developers laugh, so continue, in our strange configned world, always nice to have something to laugh about.

simonschaufi:
@remicollet what is the difference between npm and composer? For me they do the same.

remicollet:
"npm install foo" is able to install only needed scripts, without using any .gitattributes hack.

Issue #4150, opened 2020-03-26
"Type Checker" and "Backward Compatibility" jobs cannot install dependencies via Composer


sebastianbergmann:
Fails on GitHub Actions:
```
$ ./tools/composer update --no-interaction --no-ansi --no-progress --no-suggest
Loading composer repositories with package information
Updating dependencies (including require-dev)
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Installation request for phpunit/phpunit dev-3a2dd9fb398b8879ed56b87080da966419873023 -> satisfiable by phpunit/phpunit[dev-3a2dd9fb398b8879ed56b87080da966419873023].
    - phpspec/prophecy-phpunit 2.0.x-dev requires phpunit/phpunit ^9.1 -> satisfiable by phpunit/phpunit[9.1.x-dev].
    - Can only install one of: phpunit/phpunit[9.1.x-dev, dev-3a2dd9fb398b8879ed56b87080da966419873023].
    - Installation request for phpspec/prophecy-phpunit ^2.0 -> satisfiable by phpspec/prophecy-phpunit[2.0.x-dev].
```
Works locally:
```
$ ./tools/composer update --no-interaction --no-ansi --no-progress --no-suggest
Loading composer repositories with package information
Updating dependencies (including require-dev)
Package operations: 31 installs, 0 updates, 0 removals
  - Installing sebastian/code-unit (dev-master dde01b7): Cloning dde01b7318 from cache
  - Installing sebastian/environment (5.0.1): Loading from cache
  - Installing phpunit/php-text-template (2.0.0): Loading from cache
  - Installing phpunit/php-token-stream (4.0.0): Loading from cache
  - Installing phpunit/php-file-iterator (3.0.0): Loading from cache
  - Installing sebastian/code-unit-reverse-lookup (2.0.0): Loading from cache
  - Installing sebastian/version (3.0.0): Loading from cache
  - Installing theseer/tokenizer (1.1.3): Loading from cache
  - Installing phpunit/php-code-coverage (8.0.1): Loading from cache
  - Installing sebastian/recursion-context (4.0.0): Loading from cache
  - Installing sebastian/exporter (4.0.0): Loading from cache
  - Installing phpunit/php-timer (3.0.0): Loading from cache
  - Installing sebastian/object-reflector (2.0.0): Loading from cache
  - Installing sebastian/global-state (4.0.0): Loading from cache
[...]

PR #4153, opened 2020-03-29
Enhancement: Update shivammathur/setup-php

localheinz:
This PR

* [x] updates `shivammathur/setup-php`

Follows https://github.com/sebastianbergmann/phpunit/issues/4150#issuecomment-605590992.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,7 +49,7 @@ jobs:
           fetch-depth: 0
 
       - name: "Install PHP with extensions"
-        uses: shivammathur/setup-php@v1
+        uses: shivammathur/setup-php@v2
         with:
           php-version: 7.4
           coverage: none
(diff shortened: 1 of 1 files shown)


localheinz:
@sebastianbergmann 

Are we missing the `COMPOSER_ROOT_VERSION` environment variable here, so dependencies can be properly resolved?

sebastianbergmann:
> Are we missing the `COMPOSER_ROOT_VERSION` environment variable here, so dependencies can be properly resolved?

Don't know. All I know is that I do not need to set `COMPOSER_ROOT_VERSION` locally. `ant setup` "just works" for me.

localheinz:
@sebastianbergmann 

Same for me, but hang on - perhaps we can work around it.

sebastianbergmann:
The change from `shivammathur/setup-php@v1` to `shivammathur/setup-php@v2` should go into `8.5` and be merged from there to `9.0` and `master`.

localheinz:
@sebastianbergmann 

Adjusted!

PR #4154, opened 2020-03-29
Fix: Configure COMPOSER_ROOT_VERSION

localheinz:
This PR

* [x] configures `COMPOSER_ROOT_VERSION`

Fixes #4150.

PR #4157, opened 2020-03-29
Fix: Disable caching

localheinz:
This PR

* [x] disables caching

Related to #4150.

💁‍♂ What puzzles me: the

- `backward-compatibility`
- `type-checker`

jobs fail, but the `tests` job passes. The obvious difference is caching. The commands for installing dependencies with `composer` are identical. For consistency, we should probably disable caching to expose the problem.


localheinz:
@sebastianbergmann 

Once you confirm, I will happily change the base branch to `8.5.`

Alternatively, we can close and instead rely on #4154.

PR #4158, opened 2020-03-29
Enhancement: Update actions/checkout

localheinz:
This PR

* [x] updates `actions/checkout`

Related to https://github.com/sebastianbergmann/phpunit/issues/4150#issuecomment-605590992.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,7 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v1
+        uses: actions/checkout@v2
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -44,10 +44,13 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v1
+        uses: actions/checkout@v2
         with:
           fetch-depth: 0
 
+      - name: Fetch tags
```
localheinz: For reference, see https://github.com/actions/checkout#fetch-all-tags.

PR #4173, opened 2020-04-11
Ensure Warnings is always an array

Lewiscowles1986:
Hacking around with Laravel, I started writing tests and was greeted with a message that foreach was invalid. Clicked the traceback and saw this. Made this edit / hack and it works :shrug: .

Thanks for working on this library for so long

Running debian buster with php 7.3

sample repo https://github.com/Lewiscowles1986/larahack-2020-04-10/tree/ceb681afa87b7fa6a2f1d2418fba299e728f7c56

### DIFF
--- a/src/TextUI/TestRunner.php
+++ b/src/TextUI/TestRunner.php
@@ -121,6 +121,10 @@ final class TestRunner extends BaseTestRunner
      */
     public function run(Test $suite, array $arguments = [], bool $exit = true): TestResult
     {
+        if (!isset($arguments['warnings']) || !\is_array($arguments['warnings'])) {
+            $arguments['warnings'] = [];
+        }
+
         if (isset($arguments['configuration'])) {
             $GLOBALS['__PHPUNIT_CONFIGURATION_FILE'] = $arguments['configuration'];
         }


src/TextUI/TestRunner.php
```diff
@@ -338,8 +338,10 @@ public function run(Test $suite, array $arguments = [], bool $exit = true): Test
             }
         }
 
-        foreach ($arguments['warnings'] as $warning) {
-            $this->writeMessage('Warning', $warning);
+        if(isset($arguments['warnings'])) {
```
Lewiscowles1986, line 341:
```suggestion
        if (isset($arguments['warnings'])) {
```

Lewiscowles1986:
Apologies for the coding guidelines missing a space. All via GitHub UI, I can squash the history to a single commit if it is required (i.e. you don't support squash via Git for some reason)

Ayesh:
Looking at the vague nature of the commit messages, I certainly think squashing and force-pushing can help keep the commit history clean and useful. In git (not GitHub), you can `git rebase HEAD~3 -i` and work out changes easily.

Lewiscowles1986:
Just a tip @Ayesh if the commit messages need improving, then you'd want an `-i` flag in there to reword if using rebase.

I did not want that, but have taken your comments on commit message clarity on board.

It's not as helpful as you might think to assume people don't know how to use git. Even worse when the advice doesn't include the directive to reword the commit.

Ayesh:
I had the the `-i` flag there, and was to regarding the mention of squash in GitHub UI. I apologize if my comment above sounded negative, that certainly was not intended.

sebastianbergmann:
This should either go into `8.5` or `9.1`. That being said, though, I would rather see this solved by ensuring that `$arguments['warnings']` always exists.

Lewiscowles1986:
I can pivot it to always be there

Lewiscowles1986:
@sebastianbergmann perhaps inelegant, but I've done as you asked, keeping this in another place that this one area seeks to modify an array.

I can keep going down the stack, but I feel like as nothing else is blowing up for me, it might be better to leave the patch on a known-edge until it becomes clear it's needed elsewhere.

Lewiscowles1986:
The coverage decrease is in a pig of a place. I can remove the array check as it's internal, but that seems a worse situation to be in. I'll see if I can find a place through the coverage check utility to add a trivial non array case.

sebastianbergmann:

I would like to understand under what circumstances `$arguments['warnings']` is not set or not an array. For this, I have to ask you to please provide a **minimal**, **self-contained**, reproducing test case that shows the problem you are reporting. A test case that requires Laravel, for instance, is not minimal. Without such a minimal, self-contained, reproducing test case I will not be able to investigate this issue.

Please open a ticket with such a minimal, self-contained, reproducing and I will investigate this issue further. Below is a summary of my research into this issue so far.

#### PHPUnit 8.5

* `$arguments['warnings']` is accessed [here](https://github.com/sebastianbergmann/phpunit/blob/8.5.3/src/TextUI/TestRunner.php#L334)
* This is in the `PHPUnit\TextUI\TestRunner::doRun()` method
* The `$arguments` array is passed to `PHPUnit\TextUI\TestRunner::doRun()` [here](https://github.com/sebastianbergmann/phpunit/blob/8.5.3/src/TextUI/Command.php#L200)
* This is in the  `PHPUnit\TextUI\Command::run()` method
* `$this->arguments` is passed as the argument to `PHPUnit\TextUI\TestRunner::doRun()`
* `$this->arguments['warnings']` is initialized to `[]` [here](https://github.com/sebastianbergmann/phpunit/blob/8.5.3/src/TextUI/Command.php#L56)
* The only other access to `$this->arguments['warnings']` in `PHPUnit\TextUI\Command` is [this](https://github.com/sebastianbergmann/phpunit/blob/8.5.3/src/TextUI/Command.php#L56)
* Therefore we can assume that `$arguments['warnings']` exists [here](https://github.com/sebastianbergmann/phpunit/blob/8.5.3/src/TextUI/TestRunner.php#L120)

#### PHPUnit 9.1

* `$arguments['warnings']` is accessed [here](https://github.com/sebastianbergmann/phpunit/blob/9.1.1/src/TextUI/TestRunner.php#L341)
* This is in the `PHPUnit\TextUI\TestRunner::run()` method
* The `$arguments` array is passed to `PHPUnit\TextUI\TestRunner::run()` [here](https://github.com/sebastianbergmann/phpunit/blob/9.1.1/src/TextUI/Command.php#L103)
* This is in the  `PHPUnit\TextUI\Command::run()` method
* `$this->arguments` is passed as the argument to `PHPUnit\TextUI\TestRunner::run()`
* `$this->arguments` is set [here](https://github.com/sebastianbergmann/phpunit/blob/9.1.1/src/TextUI/Command.php#L277)
* The `ArgumentsMapper` initializes `$result['warnings']` [here](https://github.com/sebastianbergmann/phpunit/blob/9.1.1/src/TextUI/Arguments/ArgumentsMapper.php#L30)
* Therefore we can assume that `$arguments['warnings']` exists [here](https://github.com/sebastianbergmann/phpunit/blob/9.1.1/src/TextUI/TestRunner.php#L122)

sebastianbergmann:
I have separated the passing of arguments and warnings from `Command` to `TestRunner` from each other:

* [PHPUnit 8.5](https://github.com/sebastianbergmann/phpunit/commit/9041d82604effd0d0ab59ce67eafbc3462a2ba83)
* [PHPUnit 9.1](https://github.com/sebastianbergmann/phpunit/commit/a09abe3563d16eb2faac4b4b92305380cb3d0b00)

This may or may not solve your issue, but it is cleaner. These changes will go into the next release of PHPUnit 8.5 and PHPUnit 9.1, respectively.

Lewiscowles1986:
I didn't notice the messages until now.

Just to be clear. Laravel doesn't reach into your class that I know of. I'll add it to a backlog of tasks to make a less involved example.

For my own understanding: You believe this is not clear enough the cause of internal state being missing; would prefer internal state to be set, rather than to use a guard.

PR #4184, opened 2020-04-21
Document fluent methods as such

greg0ire:
This will help static analyzers understand calls to these methods
better.

For example, it makes that kind of error go away:

```
ERROR: InvalidPropertyAssignmentValue - tests/Doctrine/Tests/DBAL/StatementTest.php:43:23 - $this->conn with declared type 'Doctrine\DBAL\Connection&PHPUnit\Framework\MockObject\MockObject' cannot be assigned type 'PHPUnit\Framework\MockObject\MockObject' (see https://psalm.dev/145)
        $this->conn = $this->getMockBuilder(Connection::class)
            ->setConstructorArgs([[], $driver])
            ->getMock()
```

### DIFF
--- a/src/Framework/MockObject/Builder/InvocationMocker.php
+++ b/src/Framework/MockObject/Builder/InvocationMocker.php
@@ -53,6 +53,9 @@ final class InvocationMocker implements InvocationStubber, MethodNameMatch
         $this->configurableMethods = $configurableMethods;
     }
 
+    /**
+     * @return $this
+     */
     public function id($id): self
     {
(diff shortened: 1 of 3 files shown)


sebastianbergmann:
@muglug Does Psalm really need `@return $this` to understand this? Thanks!

muglug:
Yeah - `@return $this` (or the direct equivalent `@return static`) is more specific than `@return self`

sebastianbergmann:
Thank you, @greg0ire and @muglug!

PR #4187, opened 2020-04-23
Mark all methods that never return as such

greg0ire:
This should allow static analyzers to understand better why methods
calling these methods also never return.

### DIFF
--- a/src/Framework/Assert.php
+++ b/src/Framework/Assert.php
@@ -3460,6 +3460,8 @@ abstract class Assert
      * Mark the test as incomplete.
      *
      * @throws IncompleteTestError
+     *
+     * @psalm-return never-return
      */
     public static function markTestIncomplete(string $message = ''): void
     {
@@ -3471,6 +3473,8 @@ abstract class Assert
      *
      * @throws SkippedTestError
      * @throws SyntheticSkippedError
+     *
+     * @psalm-return never-return
      */
     public static function markTestSkipped(string $message = ''): void
     {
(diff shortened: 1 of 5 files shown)


sebastianbergmann:
Does `@psalm-return never-return` really provide a benefit here as all these methods are declared `void`?

greg0ire:
I think so. [Before the fix](https://psalm.dev/r/bedd6acaff), [after the fix](https://psalm.dev/r/195d77a406).

sebastianbergmann:
Can you send this against `8.5`, please? Thanks!

greg0ire:
Done, sorry, that's what I intended to do in the first place :sweat_smile:

muglug (approved):
Looks good!

sebastianbergmann:
@muglug Thanks for looking into this.

@greg0ire I'll merge this. Thanks!

@muglug Can you explain why `@psalm-return never-return` is necessary? Psalm should see that the code does not have a `return` statement. I feel like I am missing something here.

greg0ire:
Apparently, in the "Other" section of [this page](https://psalm.dev/docs/annotating_code/type_syntax/atomic_types/), it says that void and no-return are separate things:

> void - can be used in a return type when a function does not return a value.`

vs

> no-return is the 'return type' for a function that can never actually return, such as die(), exit(), or a function that always throws an exception. It may also be written as never-return or never-returns, and is also known as the bottom type.

And I think I understood why that is.

In my case we have

TestCase::markTestIncomplete(): void;

called by `getQuotedAlterTableChangeColumnLengthSQL() : array`

with just that and

```
{
    $this->markTestIncomplete('Not implemented yet');
}
```

a static analyzer will think that `getQuotedAlterTableChangeColumnLengthSQL` is supposed to return an `array`, because `void` returns nothing, but does not stop the execution flow, so you will get an error asking why there is no `return []` in that method.

PR #4190, opened 2020-04-25
Suggestion on regular expressions in PHP

danon:
I noticed regular expressions are used quite heavily in phpunit, so I thought I might suggest using a dedicated lib for them. If you have any questions, I'm here! :)


sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it as I do not think there is enough benefit to justify the additional dependency.

Danon:
Thanks for Talking time to consider :)

Issue #4039, opened 2020-02-10
Deprecate custom TestSuiteLoader


flow-control:
Hi there,

as we started with refactoring the `StandardTestSuiteLoader` in #3859 it is now time to go even further and deprecate the possibility to have a custom `TestSuiteLoader` at all. This would mean to deprecate the Interface, the `StandardTestSuiteLoader` and the `testSuiteLoaderClass` and `testSuiteLoaderFile` configuration settings.

I would be happy to provide a PR for this.

/Flo

This originates from https://github.com/sebastianbergmann/phpunit/pull/4037#issuecomment-584133301

sebastianbergmann:
Yes, please.

sebastianbergmann:
@flow-control This can be closed, right?

flow-control:
@sebastianbergmann yes

PR #4192, opened 2020-04-26
StandardTestSuiteLoader get_declared_classes check

symfonyaml:
| Branch? | master |
|---------|---|
| Version | 9 only |
| Bug fix | yes |
| New feature | no |
| Deprecations | no |

in `StandardTestSuiteLoader::load()`, if class exists, we still check for a `$loadedClasses`, but this variable `$loadedClasses` does not exists in this context.

```php
/* StandardTestSuiteLoader.php */
...
        // if class exists
        if (!\class_exists($suiteClassName, false)) {
           ...
        }
        // var $loadedClasses is not set, so this statement is always true.
        if (empty($loadedClasses)) {
           ...
        }
```

### DIFF
--- a/src/Runner/StandardTestSuiteLoader.php
+++ b/src/Runner/StandardTestSuiteLoader.php
@@ -27,9 +27,9 @@ final class StandardTestSuiteLoader implements TestSuiteLoader
     {
         $suiteClassName = \str_replace('.php', '', \basename($suiteClassFile));
 
-        if (!\class_exists($suiteClassName, false)) {
-            $loadedClasses = \get_declared_classes();
+        $loadedClasses = \get_declared_classes();
 
+        if (!\class_exists($suiteClassName, false)) {
             FileLoader::checkAndLoad($suiteClassFile);
 
             $loadedClasses = \array_values(
--- a/src/Runner/StandardTestSuiteLoader.php
+++ b/src/Runner/StandardTestSuiteLoader.php
@@ -35,10 +35,10 @@ final class StandardTestSuiteLoader implements TestSuiteLoader
             $loadedClasses = \array_values(
                 \array_diff(\get_declared_classes(), $loadedClasses)
             );
-        }
 
-        if (empty($loadedClasses)) {
-            throw ClassNotFoundException::byFilename($suiteClassName, $suiteClassFile);
+            if (empty($loadedClasses)) {
+                throw ClassNotFoundException::byFilename($suiteClassName, $suiteClassFile);
+            }
         }
 
         if (!\class_exists($suiteClassName, false)) {
--- a/tests/end-to-end/standardtestsuiteloader-issue-4192-1.phpt
+++ b/tests/end-to-end/standardtestsuiteloader-issue-4192-1.phpt
@@ -1,0 +1,21 @@
+--TEST--
+phpunit ../../_files/ConcreteTest.php
+--FILE--
+<?php declare(strict_types=1);
+require_once __DIR__.'/../../vendor/autoload.php';
+$cmd = new \PHPUnit\TextUI\Command();
+$cmd->run([
+    'phpunit',
+    (new \ReflectionClass(\ConcreteTest::class))->getFileName()
+], false);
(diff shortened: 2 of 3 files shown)


sebastianbergmann:
Please provide a **minimal**, **self-contained**, reproducing test case that shows the problem you are trying to fix.

symfonyaml:
> Please provide a **minimal**, **self-contained**, reproducing test case that shows the problem you are trying to fix.

I created a tiny bash script that hopefully can show you exactly what's happening:
The script will run 2 tests:
 - `Example-Working.php` This one works because the `App\Test` class was not never loaded in that file.
 - `Example-Not-Working.php` This one does not work because the `App\Test` class is already loaded. Then the **[diff of get_declared_classes() in the StandardTestSuiteLoader class](https://github.com/sebastianbergmann/phpunit/blob/1a4025d885e7d100931267cb8304ded8d3cca09e/src/Runner/StandardTestSuiteLoader.php#L36)** is empty, because `App\Test` class already loaded.

```bash
#!/bin/sh

echo 'install phpunit 9'
echo "{
    \"require\": {\"phpunit/phpunit\": \"^9.0\"},
    \"autoload\": {\"psr-4\": {\"App\\\\\": \"\"}}
}" > ./composer.json

composer install -q

echo 'create dummy test'
echo "<?php
namespace App;
class Test extends TestCase {
    public function testA()
    {
        \$this->assertTrue(true);
    }
}" > ./Test.php

echo 'Working example:'
echo "#!/usr/bin/env php
<?php
require_once __DIR__.'/vendor/autoload.php';
\$cmd = new \PHPUnit\TextUI\Command();
\$cmd->run([
    'vendor/bin/phpunit',
    __DIR__.'/Test.php'
], false);" > ./Example-Working.php
php ./Example-Working.php
echo 'Test was successful because class Test.php is not loaded in that file'

echo 'Not working example: (expects a Class Test could not be found error)'
echo "#!/usr/bin/env php
<?php
require_once __DIR__.'/vendor/autoload.php';
\$cmd = new \PHPUnit\TextUI\Command();
\$cmd->run([
    'vendor/bin/phpunit',
    (new \ReflectionClass(\App\Test::class))->getFileName()
], false);" > ./Example-Not-Working.php
php ./Example-Not-Working.php
```

sebastianbergmann:
@flow-control Can you have a look at this? Thanks!

flow-control:
Hey there :vulcan_salute:,

basically this method is only used in case you invoke PHPUnit from CLI with a path to a test file, not when invoked with directory or test suite from `phpunit.xml`. So the only case this solves is when you somehow load the test file in let's say your bootstrap file:

```bash
$ ./phpunit --bootstrap bootstrap.php path/to/SomeTest.php
```
with `bootstrap.php` already including the `path/to/SomeTest.php` file.

Without the patch applied PHPUnit will exit with `Class 'SomeTest' could not be found in '/path/to/SomeTest.php'`. With the PR applied PHPUnit will execute the tests in `SomeTest` Class.

Another note to add: This PR will not add any functionality that would interfere with #4039.

Looks good to me :+1: 

/Florian

flow-control:
Hey @symfonyaml :vulcan_salute:,

would you mind adding a test for the case this PR fixes and could you also have a look at the type check?

Kind regards
Florian

symfonyaml:
Hey @flow-control ,
Sure I can do that, I'll try m best :)

sebastianbergmann:
Also note that the type checker has complaints about these code chages.

symfonyaml:
@sebastianbergmann @flow-control I have created some tests.
I cant figure out the type check issue though sorry
```php
ReflectionClass::__construct expects class-string|object,
$class = new \ReflectionClass($suiteClassName);
```

flow-control:
This is a bug in Psalm: 
https://twitter.com/psalmphp/status/1255902171540979712
Got fixed in Psalm `master`: https://github.com/vimeo/psalm/commit/8ab5a0f50497be609823fb4a9f3978eb6ee8bbec

PR #4193, opened 2020-04-26
Fix magic method `__sleep` return value

carusogabriel:
Found this one during the development of https://wiki.php.net/rfc/magic-methods-signature :)


sebastianbergmann:
Cherry-picked into `8.5` and merged to `9.1` and `master` from there. Thanks!

Issue #4049, opened 2020-02-12
Tests for assertFileEqualsCanonicalizing()


sebastianbergmann:

pwsdotru:
For now assertFileEqualsCanonicalizing work incorrect. This assert just compare two string with simple equals (===) and not use param $canonical in comparator.

PR #4203, opened 2020-05-03
#4049 Add tests for assertFileEqualsCanonicalizing

pwsdotru:
By issue #4049


pwsdotru:
Test for compare canonicalizing string not worked, because comparator for files just compare strings - without canonicalizing

sebastianbergmann:
Cherry-picked into `10.5`, will merge to `main` from there. Thanks!

PR #4206, opened 2020-05-04
Update `mb_` functions in StringContainers Constraint to explicitly set "UTF-8" encoding

### LINKED ISSUE #4098: Consistently use UTF-8
Quoting a conversation with @nikic and @heiglandreas:

* [you should explicitly an utf-8 encoding on the mb_stripos, so this isn't affected by global encoding ](https://chat.stackoverflow.com/transcript/message/48600812#48600812)
* [And for obvious reasons, you should absolutely not, under any circumstances, attempt to support encodings other than utf-8 and binary inside phpunit ](https://chat.stackoverflow.com/transcript/message/48600814#48600814)
* [if you use mb_ you should be specifying encoding everywhere, at least in libraries ](https://chat.stackoverflow.com/transcript/message/48600819#48600819)

Ayesh:
Fixed #4098 .

There were only two `mb_` function calls, and they now explicitly get "UTF-8" as the encoding. `UTF-8` is used (as opposed to `utf-8` or `8bit`) to be consistent with `\PHPUnit\Util\Xml::convertToUtf8()`.

Thank you.

### DIFF
--- a/src/Framework/Constraint/StringContains.php
+++ b/src/Framework/Constraint/StringContains.php
@@ -42,7 +42,7 @@ final class StringContains extends Constraint
     public function toString(): string
     {
         if ($this->ignoreCase) {
-            $string = \mb_strtolower($this->string);
+            $string = \mb_strtolower($this->string, 'UTF-8');
         } else {
             $string = $this->string;
         }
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
Cherry-picked into `9.1` and merged to `master` from there.

Issue #4057, opened 2020-02-12
Tests for assertDirectoryIsNotReadable()


sebastianbergmann:

0silverback0:
im new to open source and PHP but I'm ready to help in any way

PR #4208, opened 2020-05-05
Not use folders test/files for create temporary folders

pwsdotru:
Use system temp folder instead "test/files" for create temporary folders for tests testAssertDirectoryIsNotReadable and testAssertDirectoryIsNotWritable

I think test cannot create folder and files inside repository.

### DIFF
--- a/tests/unit/Framework/AssertTest.php
+++ b/tests/unit/Framework/AssertTest.php
@@ -516,7 +516,7 @@ XML;
 
     public function testAssertDirectoryIsNotReadable(): void
     {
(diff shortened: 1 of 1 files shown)


Ayesh (approved):
I sent #4057 and #4058 that added these tests. I think this change makes sense, because these are temp directories and should not mutate the `_file` directory. I'm sorry this had to be corrected, but I'm happy it gets changed :)

sebastianbergmann:
Cherry-picked into `9.1` and merged to `master` from there. Thanks!

PR #4252, opened 2020-05-29
Fixed #4251, LogicalXor returns wrong result.

### LINKED ISSUE #4251: LogicalXor returns wrong result?
| Q                   | A
| PHPUnit version     | 9.2-g61f7c75a1
| PHP version         | 7.3
| Installation Method | git

#### Summary

I wonder, whether the current [logic of LogicalXor](https://github.com/sebastianbergmann/phpunit/blob/61f7c75a11eb1c6908c87235757a428de2bae6d1/src/Framework/Constraint/LogicalXor.php#L64) is correct (for me it looks wrong, but maybe I'm wrong here). If it's expected to resemble the classical ``xor`` operator, then it can't be "short circuited" as it is done in current implementation, i.e. the classical result of chain of ``xor``s' always depends on all its arguments. The current implementation assumes that if there is a subexpression of ``xor``s' chain which evaluates to ``false``, then the remaining arguments can't change the result (which is wrong, IMHO). It even looks, like the [current test for Xor logic](https://github.com/sebastianbergmann/phpunit/blob/61f7c75a11eb1c6908c87235757a428de2bae6d1/tests/unit/Framework/Constraint/LogicalXorTest.php#L45) is wrong, as it expects ``0 xor 1 xor 0 xor 1 xor 0`` to be ``true``,  while in reality it's ``false``:

``` console
ptomulik@barakus:$ vendor/bin/psysh 
Psy Shell v0.10.4 (PHP 7.3.15-3 — cli) by Justin Hileman
>>> 0 xor 1 xor 0 xor 1 xor 0
=> false
```

#### Current behavior

The proposed test (below) does not pass, with the following errors (the first case is most astonishing!):

```console
ptomulik@barakus:$ ./phpunit  tests/unit/Framework/Constraint/LogicalXorTest.php        
[...]

..F.....FF..FFFF                                                  16 / 16 (100%)

Time: 00:02.405, Memory: 18.00 MB

There were 7 failures:

1) Framework\Constraint\LogicalXorTest::testEvaluateReturnsCorrectResult with data set #0 (0)
Failed asserting that false is false
Failed asserting that true is identical to false.

[...]

2) Framework\Constraint\LogicalXorTest::testEvaluateReturnsCorrectResult with data set #6 (0, 0, 0)
Expectation failed for method name is "evaluate" when invoked 1 time(s).
Method was expected to be called 1 times, actually called 0 times.

[...]

3) Framework\Constraint\LogicalXorTest::testEvaluateReturnsCorrectResult with data set #7 (0, 0, 1)
Failed asserting that false xor false xor true is true
Failed asserting that false is identical to true.

[...]

4) Framework\Constraint\LogicalXorTest::testEvaluateReturnsCorrectResult with data set #10 (1, 0, 0)
Failed asserting that true xor false xor false is true
Failed asserting that false is identical to true.

[...]

5) Framework\Constraint\LogicalXorTest::testEvaluateReturnsCorrectResult with data set #11 (1, 0, 1)
Failed asserting that true xor false xor true is false
Failed asserting that true is identical to false.

[...]

6) Framework\Constraint\LogicalXorTest::testEvaluateReturnsCorrectResult with data set #12 (1, 1, 0)
Expectation failed for method name is "evaluate" when invoked 1 time(s).
Method was expected to be called 1 times, actually called 0 times.

[...]

7) Framework\Constraint\LogicalXorTest::testEvaluateReturnsCorrectResult with data set #13 (1, 1, 1)
Failed asserting that true xor true xor true is true
Failed asserting that false is identical to true.

[...]

FAILURES!
Tests: 16, Assertions: 43, Failures: 7.

```

#### How to reproduce

The following test may be added to reproduce the bug:

```php
final class LogicalXorTest extends TestCase                                                                             
{                                                                                                                       
    // [...]
    public function provideEvaluateReturnsCorrectResult()
    {
        return [
            [0],       [1],
            [0, 0],    [0, 1],    [1, 0],    [1, 1],
            [0, 0, 0], [0, 0, 1], [0, 1, 0], [0, 1, 1],
            [1, 0, 0], [1, 0, 1], [1, 1, 0], [1, 1, 1],
        ];
    }

    /**
     * @dataProvider provideEvaluateReturnsCorrectResult
     */
    public function testEvaluateReturnsCorrectResult(int ...$args)
    {
        $other = 'Foo';
        $constraints = array_map(function (bool $arg) use ($other) {
            $constraint = $this->getMockBuilder(Constraint::class)->getMock();

            $constraint
                ->expects($this->once())
                ->method('evaluate')
                ->with($this->identicalTo($other))
                ->willReturn($arg);
            $constraint
                ->expects($this->any())
                ->method('toString')
                ->with()
                ->willReturn($arg ? 'true' : 'false');
            return $constraint;
        }, $args);

        $initial = (bool)array_shift($args);
        $expected = array_reduce($args, function (bool $carry, bool $item) {
            return $carry xor $item;
        }, $initial);
        array_unshift($args, $initial);

        $constraint = LogicalXor::fromConstraints(...$constraints);

        $message = 'Failed asserting that '.$constraint->toString().' is '. ($expected ? 'true' : 'false');
        $this->assertSame($expected, $constraint->evaluate($other, '', true), $message);
    }
    // [...]
}
```

#### Expected behavior

The proposed test should pass without errors.

ptomulik:
This shall fix #4251

### DIFF
--- a/src/Framework/Constraint/LogicalXor.php
+++ b/src/Framework/Constraint/LogicalXor.php
@@ -63,19 +63,19 @@ final class LogicalXor extends Constraint
      */
     public function evaluate($other, string $description = '', bool $returnResult = false): ?bool
     {
-        $success    = true;
-        $lastResult = null;
+        $constraints = $this->constraints;
 
-        foreach ($this->constraints as $constraint) {
-            $result = $constraint->evaluate($other, $description, true);
-
-            if ($result === $lastResult) {
-                $success = false;
-
-                break;
-            }
-
-            $lastResult = $result;
+        if (($initial = \array_shift($constraints)) !== null) {
+            $success = \array_reduce(
+                $constraints,
+                function ($carry, Constraint $constraint) use ($other, $description) {
+                    return $carry xor $constraint->evaluate($other, $description, true);
+                },
+                $initial->evaluate($other, $description, true)
+            );
+        } else {
+            // $constraints was empty or not an array...
+            $success = false;
         }
 
         if ($returnResult) {
--- a/tests/unit/Framework/Constraint/LogicalXorTest.php
+++ b/tests/unit/Framework/Constraint/LogicalXorTest.php
@@ -18,31 +18,26 @@ use PHPUnit\Framework\TestCase;
  */
 final class LogicalXorTest extends TestCase
 {
-    public function testFromConstraintsReturnsConstraint(): void
+    public function providerFromConstraintsReturnsConstraint()
     {
-        $other = 'Foo';
-        $count = 5;
+        return [
+            [0], [0, 1], [0, 1, 0],
+        ];
+    }
 
-        $constraints = \array_map(function () use ($other) {
-            static $count = 0;
-
-            $constraint = $this->getMockBuilder(Constraint::class)->getMock();
-
-            $constraint
-                ->expects($this->once())
-                ->method('evaluate')
-                ->with($this->identicalTo($other))
-                ->willReturn($count % 2 === 1);
-
-            ++$count;
-
-            return $constraint;
-        }, \array_fill(0, $count, null));
+    /**
+     * @dataProvider providerFromConstraintsReturnsConstraint
+     */
+    public function testFromConstraintsReturnsConstraint(int ...$args): void
+    {
+        $other       = 'Foo';
+        $constraints = \array_map(function (bool $arg) use ($other) {
+            return $this->getMockBuilder(Constraint::class)->getMock();
+        }, $args);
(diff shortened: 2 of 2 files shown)


sebastianbergmann:
Please fix the errors found by the type checker as well as the coding guidelines violations.

ptomulik:
> Please fix the errors found by the type checker as well as the coding guidelines violations.

Done.

sebastianbergmann:
Superseded by #4264.

Issue #4255, opened 2020-05-29
Improvements on withConsecutive with return


jaapio:
As a developer using the build-in mocks of PHPUnit can be complicated. One of the things I do have to look up aways is the way to have `withConsecutive` mocks. On the other hand, when using the `willReturnOnConsecutiveCalls` in combination `willReturnConsecutive` is a real issue. 

Stubs do have a somewhat better method called `willReturnMap` but that one is using a less complicated check to match the arguments.

In an example where we have something like this:

```
        $this->client->expects($this->exactly($calles))
            ->method('doRequest')
            ->withConsecutive(
                [
                    /* cut checks for simplification */
                    /* cut checks for simplification */
                ],
                [
                    /* cut checks for simplification */
                    /* cut checks for simplification */
                ]
            )->willReturnOnConsecutiveCalls(
                new Reponse(),
                new Reponse(),
            );
```
The problem here is that there is a relation between `willReturnOnConsecutiveCalls` and `withConsecutive` but it is not really clear to everyone that this relation is here. Especially when the number of checks done in the `withConsecutive` arrays is growing, it gets unreadable. because the lines of argument matchers are not fitting into your screen anymore. 

Using `willReturnMap` is not an option in this case, because the subject under test has some internal complexity that needs to be tested. And the arguments of `doRequest` are objects so checking needs a callback. 

In the past, I created a manual test double that was implementing a certain interface and just stores the calls, like some kind of spy object. But in the end, it would be a nice feature to have this in the mocks so those extra classes are not required anymore.  

What I would like to propose is a combination of `willReturnMap` and `withConsecutive` and `willReturnOnConsecutiveCalls` or maybe an extension of `willReturnMap` that will allow a developer to add `Constraints` into the return map. I'm a bit in doubt if this will work properly since the new method will mix the concepts of `parameterMatchers` and `stubs` for the return types. Which I dislike in `willReturnMap` 

Another approach would be more like a builder type of implementation. So we could use the existing calls to create a mock. 

```   
        $mock->expects($this->once())->method('doRequest')->with()->willReturn();
        $mock->expects($this->once())->method('doRequest')->with()->willReturn();
```

This doesn't work currently because the matcher will throw an error.  A more suitable solution would be something like the example below. The order of calls to `with` and `willReturn` would set the expected order of calls just like `withConsecutive` and `willReturnOnConsecutiveCalls`

```
        $mock->expects($this->once())->method('doRequest')
        ->with()->willReturn()
[...]

PR #4258, opened 2020-06-01
Throw exception on non unique mocked method

jaapio:
This patch will make a mock throw an exception when multiple matchers
can be applied to a invoke. When allowing this, the results of tests are
not predictable.

In preparation on #4255 I found that a test case like the one below would never pass. Because the written test is just wrong. However, the mocking framework does allow this. But does not handle it properly. From now on an exception will be thrown. 

```
        $mock->expects($this->any())
            ->method($this->stringStartsWith('foo'))
            ->with('foo')
            ->willReturn('result');

        $mock->expects($this->any())
            ->method('foo')
            ->with('bar')
            ->willReturn('result');
```
refs #4255

### DIFF
--- a/src/Framework/MockObject/InvocationHandler.php
+++ b/src/Framework/MockObject/InvocationHandler.php
@@ -113,30 +113,11 @@ final class InvocationHandler
     public function invoke(Invocation $invocation)
     {
         $exception      = null;
-        $hasReturnValue = false;
         $returnValue    = null;
+        $match          = $this->findMatcher($invocation);
 
-        foreach ($this->matchers as $match) {
-            try {
-                if ($match->matches($invocation)) {
-                    $value = $match->invoked($invocation);
-
-                    if (!$hasReturnValue) {
-                        $returnValue    = $value;
-                        $hasReturnValue = true;
-                    }
-                }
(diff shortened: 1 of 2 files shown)


sebastianbergmann:
Cherry-picked into `9.2` and merged from there. Thanks!

GrahamCampbell:
This change broke `laravel/framework`'s test suite.

GrahamCampbell:
```
1) Illuminate\Tests\Database\DatabaseConnectionTest::testBeginTransactionMethodRetriesOnFailure
Non unique mocked method invocation: PDO::beginTransaction

/home/runner/work/framework/framework/src/Illuminate/Database/Concerns/ManagesTransactions.php:120
/home/runner/work/framework/framework/src/Illuminate/Database/Concerns/ManagesTransactions.php:100
/home/runner/work/framework/framework/tests/Database/DatabaseConnectionTest.php:158
```

driesvints:
No worries about the above. We fixed the test which probably was incorrect to begin with. Thanks for your work!

jaapio:
This was definitely a broken test. The mocked method in question was called to many times internally. Although phpunit didn't report this.

Issue #4229, opened 2020-05-15
Actual size is wrong in assertCount with generators


spaceemotion:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 9.1.4
| PHP version         | 7.4.0
| Installation Method | Composer

#### Summary

The original sample at https://github.com/sebastianbergmann/phpunit/issues/3302#issuecomment-420984415

```php
   public function testGenerator()
    {
        $generator = static function () {
            yield 1;
            yield 2;
            yield 3;
        };

        static::assertCount(3, $generator());
        static::assertCount(4, $generator());
    }
```

no longer seems to work when trying to assert the count of four. The error message spits out the following instead:

`Failed asserting that actual size 0 matches expected size 4`

#### Current behavior

I initially came across this in my own code, but found the example linked above. My generator definitely holds some values, and is being iterated in the assertCount() check, but the error message is just wrong.

#### How to reproduce

Just execute the example linked above.

#### Expected behavior

Something along the lines of:

`Failed asserting that actual size 3 matches expected size 4`

amenning:
I found this was because the generator cannot be cloned and can only be iterated over once.  The actual count was correctly determined in the [matches](https://github.com/sebastianbergmann/phpunit/blob/8.5/src/Framework/Constraint/Count.php#L44) function.  However, then the [failureDescription](https://github.com/sebastianbergmann/phpunit/blob/8.5/src/Framework/Constraint/Count.php#L119) attempts to iterate over the generator a second time to provide the failure details, resulting in the null count for the message.  

One solution could be to store the actual count result inside the matches method for use in the failureDescription to provide the expected behavior.

amenning:
#4271 is a potential solution PR

sebastianbergmann:
Superseded by #4567 and #4568.

PR #4271, opened 2020-06-07
Fix failure description of count constraint for generators

amenning:


sebastianbergmann:
Can you please provide a **minimal**, **self-contained**, reproducing test case that shows the problem you are trying to solve?

amenning:
> Can you please provide a **minimal**, **self-contained**, reproducing test case that shows the problem you are trying to solve?

This is for the [open issue related to count constraint of generators](https://github.com/sebastianbergmann/phpunit/issues/4229) 

I have added [this test case](https://github.com/sebastianbergmann/phpunit/blob/55fb3d88e41135536bd13c821728191e8563a213/tests/unit/Framework/Constraint/CountTest.php#L177) that shows the minimal, self-contained, reproducing test case based on the example given in the issue.

sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it because I believe #4567 and #4568 to be right course of action.

PR #4294, opened 2020-06-15
Enhancement: Enable phpdoc_summary fixer

localheinz:
This PR

* [x] enables the `phpdoc_summary` fixer
* [x] runs `php-cs-fixer`
* [x] sorts out a few issues manually

Slightly related to https://github.com/sebastianbergmann/phpunit/pull/3936.

💁‍♂️ For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/tree/v2.16.3#usage:

> **phpdoc_summary**
>
> PHPDoc summary should end in either a full stop, exclamation mark, or question mark.

### DIFF
--- a/src/Framework/TestSuite.php
+++ b/src/Framework/TestSuite.php
@@ -105,7 +105,7 @@ class TestSuite implements \IteratorAggregate, SelfDescribing, Test
     private $warnings = [];
 
     /**
-     * Constructs a new TestSuite:
+     * Constructs a new TestSuite:.
      *
      *   - PHPUnit\Framework\TestSuite() constructs an empty TestSuite.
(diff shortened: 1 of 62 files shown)


src/Framework/TestSuite.php
```diff
@@ -100,7 +100,7 @@ class TestSuite implements \IteratorAggregate, SelfDescribing, Test
     private $declaredClasses;
 
     /**
-     * Constructs a new TestSuite:
+     * Constructs a new TestSuite:.
```
sebastianbergmann, line 103: `:.` is wrong.

localheinz:

PR #4300, opened 2020-06-17
use current directory for default cache path

remicollet:
As phpunit command path seems a poor default:

- will result on vendor/bin/phpunit in most case
- will result on some global directory when global phpunit is used (phar) thus being used for "all" projects.

Open against master, but will be nice to have a better default in 8.x and 9.x

P.S. seems minor, as will only affects project without configuration file, as in most case path to configuration will be used.

### DIFF
--- a/src/TextUI/TestRunner.php
+++ b/src/TextUI/TestRunner.php
@@ -156,9 +156,9 @@ final class TestRunner extends BaseTestRunner
                 if (isset($arguments['configuration'])) {
                     \assert($arguments['configuration'] instanceof Configuration);
 
-                    $cacheLocation = $arguments['configuration']->filename();
+                    $cacheLocation = \dirname($arguments['configuration']->filename());
                 } else {
-                    $cacheLocation = $_SERVER['PHP_SELF'];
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
`master` is what will become PHPUnit 9.3 in August. I am open to have this is in `8.5` for PHPUnit 8.5 as well.

remicollet:
May result in some error such as:

`Warning: file_put_contents(/usr/bin/.phpunit.result.cache): failed to open stream: Permission denied in ...
`

sebastianbergmann:
@remicollet I agree that we should try to address this, and thank you for sending this pull request. But please note that the PHPUnit documentation explicitly advises against installing PHPUnit globally:

> Please note that it is not recommended to install PHPUnit globally, as `/usr/bin/phpunit` or `/usr/local/bin/phpunit`, for instance.

epdenouden:
@remicollet thanks for looking into this, I will have a good look. I'm in favour of cleaner solutions for the cache file management.

sebastianbergmann:

PR #4301, opened 2020-06-17
Remove whitespace terminology

simPod:
Following https://github.com/sebastianbergmann/phpunit/commit/8e9c76d33dab4095c9066072076f368193e4166d, `whitespace` might be considered as an offensive terminology and we should replace it with more appropriate word.

### DIFF
--- a/src/Util/Color.php
+++ b/src/Util/Color.php
@@ -17,7 +17,7 @@ final class Color
     /**
      * @var array<string,string>
      */
-    private const WHITESPACE_MAP = [
+    private const EMPTYSPACE_MAP = [
         ' '  => '·',
         "\t" => '⇥',
     ];
@@ -25,7 +25,7 @@ final class Color
     /**
      * @var array<string,string>
      */
-    private const WHITESPACE_EOL_MAP = [
+    private const EMPTYSPACE_EOL_MAP = [
         ' '  => '·',
         "\t" => '⇥',
         "\n" => '↵',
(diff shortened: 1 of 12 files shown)


sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it because in the conversations I had with "non-white" developers over the years, the term "whitespace" was never an issue whereas "blacklist" and "whitelist" were.

grongor:
Yeah, we need to wait few more years for this, the world is not ready yet. Such bullshit.

blacklistisnotracist:
Hi, I'm Indian. Believe or not, I'm NOT a troll.

I am really disappointed at this issue being closed without much consideration. I suppose @sebastianbergmann's non-white friends did not represent the true sentiment of most of the Indian folks.

The terminology whitespace comes from the assumption that paper is white by default. This could be interpreted as white people are the norm in this society.

Paper was somewhat brown in the ancient times. Humans' pursuit of purely white paper is based on our sense of values of white being "better", which also can offend brown people. "White paper is 'better' than brown paper" can easily be translated into the fact that you white people think that white people are better than brown people. This is such an arrogant attitude.

@grongor said this is bullshit, but this is not bullshit. Why a white person such as yourself, can say anything about this matter? Do you only care about white/black racism?

007hacky007:
> the term "whitespace" was never an issue whereas "blacklist" and "whitelist" were.
Why? I find that discriminating. Also I'm yellow. Hopefully I don't find anything offensive in the code or I'm gonna cry a lot.

simPod:
> because in the conversations I had with "non-white" developers over the years, the term "whitespace" was never an issue whereas "blacklist" and "whitelist" were.

But I'm non-white developer (as can be seen in my profile picture, not trolling). My father is  Belarusian and mother Egyptian, I inherited the skin colour from her.

Must admit I feel kind of disappointed. We have not had a conversation in past, though I supposed we could catch up now through this medium.

TomHetmer:
I realized my computer displays the comments here as black on white text.
I recompiled Mac OS X kernel to show everything in mulberry-orchid.
Now I can't see anything when I type this.

Issue #4307, opened 2020-06-20
TestCase::$backupGlobals is declared as boolean but its initial value is NULL


morozov:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 9.2.3
| PHP version         | 7.4.7
| Installation Method | Composer

#### Summary

In a project that uses PHPUnit and Psalm, Psalm detects the issue in PHPUnit that cannot be fixed in the project and therefore have to be suppressed.

#### Current behavior

```
ERROR: PropertyNotSetInConstructor - tests/ExampleTest.php:7:7 - Property 
Test\ExampleTest::$backupGlobals is not defined in constructor of Test\ExampleTest and in any 
methods called in the constructor (see https://psalm.dev/074)
class ExampleTest extends TestCase
```
The above is caused by a boolean property being non-initialized:
1. https://github.com/sebastianbergmann/phpunit/blob/61f7c75a11eb1c6908c87235757a428de2bae6d1/src/Framework/TestCase.php#L63-L66
2. https://github.com/sebastianbergmann/phpunit/blob/61f7c75a11eb1c6908c87235757a428de2bae6d1/src/Framework/TestCase.php#L389-L397

#### How to reproduce

```php
<?php declare(strict_types=1);

namespace Test;

use PHPUnit\Framework\TestCase;

class ExampleTest extends TestCase
{
    /**
     * @test
     */
    public function testExample() : void
    {
        $this->assertTrue(true);
    }
}
```

See the full demo project in the [gist](https://gist.github.com/morozov/189e817c9ca8eee4d225f1d3475bc230):
```
$ composer install
$ vendor/bin/psalm
```

Since the project's tests extend and use the PHPUnit test classes, suppressing errors in the `vendor/` directory doesn't help.

#### Expected behavior

The value of `$backupGlobals` is always within its declared type constraints.

sebastianbergmann:
Can you please open separate issues? Thanks!

morozov:
Please see #4309.

sebastianbergmann:
> Please see #4309.

Thanks!

PR #4308, opened 2020-06-20
Mark TestCase::$backupGlobals as nullable

morozov:
`TestCase::$backupGlobals` is a nullable boolean: https://github.com/sebastianbergmann/phpunit/blob/a4baf33f0c2cffb6de207cd4ed95be3bb8193103/src/Framework/TestCase.php#L1200

Fixes https://github.com/sebastianbergmann/phpunit/issues/4307.

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -60,7 +60,7 @@ abstract class TestCase extends Assert implements SelfDescribing, Test
     private const LOCALE_CATEGORIES = [\LC_ALL, \LC_COLLATE, \LC_CTYPE, \LC_MONETARY, \LC_NUMERIC, \LC_TIME];
 
     /**
-     * @var bool
+     * @var null|bool
      */
     protected $backupGlobals;
 


sebastianbergmann:
@Ocramius `roave/backward-compatibility-check` considers the change from `bool` to `bool|null` (which should be `?bool`) to be a breaking change:

```
[BC] CHANGED: Type documentation for property PHPUnit\Framework\TestCase#$backupGlobals changed from bool to bool|null
```

I am not 100% convinced that this is correct.

Ocramius:
@sebastianbergmann any reads from that property can no longer operate under the assumption that a boolean is contained in there: I am aware that phpunit documented it as `bool` while it stores a `null` when not initialized, but that's not something the tool can understand, as it only relies on the documented type.

Ocramius:
Note: the tool is saying (in the message) that the inferred type (from docblock) changed, which is why the wording "documentation" appears in the report.

It cannot understand if this is an actual change in type, so it assumes that the **actual** type changed.

morozov:
As per the above, it's not a BC break. The property documented as `bool` could also contain `NULL`. Now, this is documented but there's no change in the behavior observed by the API consumers.

sebastianbergmann:
@morozov When (not if) I merge this then I'll update the `--from` parameter for `roave/backward-compatibility-check` in the CI pipeline.

Issue #3236, opened 2018-08-03
Public API


sebastianbergmann:
In order to define a [Backward Compatibility Promise](http://symfony.com/doc/current/contributing/code/bc.html) we need to make it explicit which parts of PHPUnit's API are considered to be public.

We should start by adding `@internal` to all classes, interfaces, and traits. Then we need to figure which units of code must be public (and remove `@internal` from them) for being able to use PHPUnit.

keradus:
just in case - i can help with adding @internal everywhere with PHP CS Fixer

sebastianbergmann:
I know, thanks.

DavertMik:
Hi, Sebastian. 
There are quite a lot o projects which may use PHPUnit via its current API and it would be nice to keep it stable. Codeception is one of them. I'd kindly ask to make all APIs used here https://github.com/Codeception/phpunit-wrapper to be considered as public API. 

Also, I know that JetBrains also heavily relies on PHPUnit internals to make their integration working. 
Not sure of issues on their side but I think we can invite them for the discussion.

sebastianbergmann:
JetBrains does not rely on PHPUnit internals, not since the TeamCity test listener was added to PHPUnit.

keradus:
Nice thing with public API is that if sth is not marked as part of it, it can always become part of it.
But if it's already marked as public, taking it down is BC breaker.

Saying that - if we will make public API too narrow, we can always make it wider with any release (minor or even patch)

wbars:
I confirm, starting with PHPUnit 5.0 PhpStorm integration relies only on _output_ of TeamCity.php atm

keradus:
nice, @sebastianbergmann !

PR #4310, opened 2020-06-21
Make InvocationMocker part of the public API

### LINKED ISSUE #4309: PHPUnit exposes an internal API to the userland
| Q                   | A
| PHPUnit version     | 9.2.3
| PHP version         | 7.4.7

#### Summary

In a project that uses PHPUnit and Psalm, Psalm detects the issue in PHPUnit that cannot be fixed in the project and therefore has to be suppressed.

#### Current behavior

```
ERROR: InternalMethod - tests/ExampleTest.php:17:12 - The method 
PHPUnit\Framework\MockObject\Builder\InvocationMocker::method has been marked as internal 
(see https://psalm.dev/175)
        )->method('test');
```
The above is caused by the interface returning a class marked as `@internal`:
1. https://github.com/sebastianbergmann/phpunit/blob/7f664ffd558ba68031963f63d3e23a951ba441fc/src/Framework/MockObject/MockObject.php#L24
2. https://github.com/sebastianbergmann/phpunit/blob/cb135993b214d87d1cc10359201d23b8ab2057bc/src/Framework/MockObject/Builder/InvocationMocker.php#L29-L32

This issue was already brought up in https://github.com/sebastianbergmann/phpunit/issues/3742#issuecomment-559682132.

#### How to reproduce

```php

namespace Test;

class ExampleTest extends TestCase
{
    /**
     * @test
     */
    public function testExample() : void
    {
        $t = $this->createMock(T::class);
        $t->expects(
            $this->never()
        )->method('test');
    }
}

interface T
{
    public function test() : void;
}
```

See the full demo project in the [gist](https://gist.github.com/morozov/189e817c9ca8eee4d225f1d3475bc230):
```
$ composer install
$ vendor/bin/psalm
```

Since the project's tests extend and use the PHPUnit test classes, suppressing errors in the `vendor/` directory doesn't help.

#### Expected behavior

PHPUnit doesn't expose the internal parts of its API to the userland.

morozov:
The `InvocationMocker` interface is exposed to the userland via `MockObject::expects()`:
https://github.com/sebastianbergmann/phpunit/blob/7f664ffd558ba68031963f63d3e23a951ba441fc/src/Framework/MockObject/MockObject.php#L24

According to #3236, the `@internal` annotation should be removed.

Fixes #4309.

### DIFF
--- a/src/Framework/MockObject/Builder/InvocationMocker.php
+++ b/src/Framework/MockObject/Builder/InvocationMocker.php
@@ -26,9 +26,6 @@ use PHPUnit\Framework\MockObject\Stub\ReturnStub;
 use PHPUnit\Framework\MockObject\Stub\ReturnValueMap;
 use PHPUnit\Framework\MockObject\Stub\Stub;
 
-/**
- * @internal This class is not covered by the backward compatibility promise for PHPUnit
- */
 final class InvocationMocker implements InvocationStubber, MethodNameMatch
 {
     /**


morozov:
@sebastianbergmann would it be possible to tag another `8.5.x` release with this change? Thank you.

sebastianbergmann:
AFAICS, the change is in the `8.5` branch. It will be part of the next PHPUnit 8.5 release.

morozov:
Yeah. Is there ETA for it? Not that it’s urgent, I just saw a couple of 9.2 releases since this issue was resolved.

sebastianbergmann:
Apart from this change, there is currently no substantial change in the `8.5` branch that would warrant a release.

PR #4314, opened 2020-06-22
Add option to exit with exit code 1 when no tests are executed

### LINKED ISSUE #4313: Optionally exit with exit code 1 when no tests are executed

localheinz:
This PR

* [x] adds a `failOnEmptyTestSuite` option which - when set - will fail a test run when no tests were found

Fixes #4313.

💁‍♂️ I have tried adding an end-to-end test, but I'm unsure how to assert the exit code from a `.phpt` test. Is that even possible?

### Disabled

Screen Shot 2020-06-23 at 21 28 56

### Enabled

Screen Shot 2020-06-23 at 21 29 21

### DIFF
--- a/src/TextUI/CliArguments/Configuration.php
+++ b/src/TextUI/CliArguments/Configuration.php
@@ -185,6 +185,11 @@ final class Configuration
     /**
      * @var ?bool
      */
+    private $failOnEmptyTestSuite;
+
+    /**
+     * @var ?bool
+     */
     private $failOnIncomplete;
 
     /**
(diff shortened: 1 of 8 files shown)


src/TextUI/CliArguments/Configuration.php
```diff
@@ -956,6 +962,20 @@ public function executionOrderDefects(): int
...
     }
 
+    public function hasFailOnEmptyTestSuite(): bool
+    {
+        return $this->failOnEmptyTestSuite !== null;
+    }
+
+    public function failOnEmptyTestSuite(): bool
+    {
+        if ($this->failOnEmptyTestSuite === null) {
```
Ayesh, line 972:
```suggestion
        if (!$this->hasFailOnEmptyTestSuite()) {
```

I think this will improve the intent.

localheinz: You are right, but this will then look very different from the other methods in this class - that's why I stuck with the same design.

localheinz:

PR #4311, opened 2020-06-21
Refactor dependency infrastructure

epdenouden:
Implements part of the infrastructure needed for #4302.

It only contains internal code clean up to make the core runner and supporting code more expressive and robust.

Changes:
- introduce new `interface Reorderable` for all sortable internal test objects
- introduce `ExecutionOrderDependency` value object
- refactor `@depends` annotation handling; consolidate string juggling and parse into value objects
- reimplement `provides()/requires()` to use `ExecutionOrderDependency`
- refactor dependency handling during test run
- remove obsolete `[TestSuite|TestCase]::requiredTests`
- remove obsolete `getTestSorterUID()`
- additional test coverage for internals

PR #4317, opened 2020-06-23
Enhancement: Enable self_static_accessor fixer

localheinz:
This PR

* [x] enables the `self_static_accessor`
* [x] runs `php-cs-fixer fixer`

Not related to #4311.

💁‍♂️ For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/tree/v2.16.3#usage:

> **self_static_accessor**
>
>Inside a `final` class or anonymous class `self` should be preferred to `static`.


localheinz:

PR #4318, opened 2020-06-24
Fix: Continue running jobs in matrix

localheinz:
This PR

* [x] continues to run all jobs in a matrix when a job has failed

💁‍♂️ For reference, see https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast.


localheinz:

PR #4322, opened 2020-06-25
Add compatibility with the InternalIterator change

mbeccati:
There are no bundled classes extending `InternalIterator`, but now `Traversable` cannot be implemented directly and `DatePeriod` implements `IteratorAggregate`, so testing for it becomes moot.


mbeccati:
For reference:

https://revive.beccati.com/bamboo/browse/PHP-PHPUN-PM-2255/test/case/49602478
and:
https://github.com/php/php-src/commit/ff19ec2df3ef0dca8c29be83eddcde58234f4095

sebastianbergmann:
Cherry-picked into `8.5` and merged to `9.2` and `master` from there. Thanks!

PR #4323, opened 2020-06-25
Add support for PHP8

WouterSioen:
This will allow this library to be installed on PHP8 as well. Since all tests are ran on CI against PHP 8, I think this should be releasable.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
     "prefer-stable": true,
     "minimum-stability": "dev",
     "require": {
-        "php": "^7.3",
+        "php": "^7.3||^8.0",
         "ext-dom": "*",
         "ext-json": "*",
         "ext-libxml": "*",


sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it. I will not release any of my packages with `^8.0` in `composer.json` before PHP 8.0.0 RC1 has been released.

WouterSioen:
Okay, looking forward to be able to run my unit tests against PHP8!

WouterSioen:
It's a bit of a pity that this is blocking upgrading a huge amount of the PHP ecosystem to start running their CI against PHP8 to prepare for the new version, but I understand that you're not totally comfortable yet with already supporting an alpha version.

sebastianbergmann:
If you use Composer to install PHPUnit then you can use the new `--ignore-platform-req=php` CLI option that was introduced in Composer 2. If you use the PHAR, which is the recommended way to use PHPUnit anyway, you do not need to do anything.

WouterSioen:
Nice, thanks!

PR #4326, opened 2020-06-27
Add InvocationResolver and use in ReturnValueMap

0leksandr:
I'd like to use strict mode in my tests: if a mock called method, that I didn't mock (or didn't think of, or method was added after the test was written), test should fail. For that I use `MockBuilder::disableAutoReturnValueGeneration`.
But `ReturnValueMap` breaks this constraint - if arguments list didn't match to any of mocked, method just silently returns `null`. With this PR I'm trying to fix it.


sebastianbergmann:
Superseded by #4332.

PR #4329, opened 2020-06-27
Fix: Remove unused private method

localheinz:
This PR

* [x] removes an unused `private` method

💁‍♂️ Running 

```
$ git grep -i parseAnnotationContent
```

on current `9.2` yields no results.


localheinz:

PR #4330, opened 2020-06-27
Fix: Turn comments into DocBlocks

localheinz:
This PR

* [x] turns comments into DocBlocks


localheinz:

PR #4332, opened 2020-06-27
Add InvocationResolver and use in ReturnValueMap

0leksandr:
I'd like to use strict mode in my tests: if a mock called method, that I didn't mock (or didn't think of, or method was added after the test was written), test should fail. For that I use MockBuilder::disableAutoReturnValueGeneration.
But ReturnValueMap breaks this constraint - if arguments list didn't match to any of mocked, method just silently returns null. With this PR I'm trying to fix it.

### DIFF
--- a/src/Framework/Assert/Functions.php
+++ b/src/Framework/Assert/Functions.php
@@ -50,6 +50,7 @@ use PHPUnit\Framework\Constraint\TraversableContainsEqual;
 use PHPUnit\Framework\Constraint\TraversableContainsIdentical;
 use PHPUnit\Framework\Constraint\TraversableContainsOnly;
 use PHPUnit\Framework\ExpectationFailedException;
+use PHPUnit\Framework\MockObject\InvocationResolver;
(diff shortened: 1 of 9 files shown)


sebastianbergmann:
This pull request has failing tests.

0leksandr:
There is an alternative way to fix this problem (at least one I can think of) - add method `Stub::matches($invocation): bool`, and not even execute `ReturnValueMap::invoke` if it didn't match. I think, it depends on your vision of `ReturnValueMap`, which way to choose. Should it always fail when arguments didn't match (like when `InvocationMocker::with` didn't match), or only in "strict mode" (when `returnValueGeneration = false`)?
If alternative way is preferred, I'd like to do that, and not make "hack" in your code, which will only make it harder to deal with later.

0leksandr:
Sorry, I think I chose wrong approach. Main problem is that static method `TestCase::returnValueMap` knows nothing about `$autoReturnValueGeneration` set by user. A possible solution might be to make `TestCase::returnValueMap` non-static, but then there still will be needed some refactoring to pass `$autoReturnValueGeneration` (or other flag) to it.

Issue #3834, opened 2019-09-07
Trailing slash in XdebugFilterScriptGenerator.php breaks coverage in windows


shivammathur:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 8.3
| PHP version         | 7.2.10
| Installation Method | Composer

Trailing slash after `%s` in XdebugFilterScriptGenerator results in 0% coverage on windows as it has backslash as separator.
https://github.com/sebastianbergmann/phpunit/blob/c2fa2fc0ad5862544c7d6788685df5ed0200ec0f/src/Util/XdebugFilterScriptGenerator.php#L59

Current code on windows generates this `xdebug-filter.php`
```php
<?php declare(strict_types=1);
if (!\function_exists('xdebug_set_filter')) {
    return;
}

\xdebug_set_filter(
    \XDEBUG_FILTER_CODE_COVERAGE,
    \XDEBUG_PATH_WHITELIST,
    [
        'C:\path\to\dir/'
    ]
);
```

Correct `xdebug-filter.php` should be 

```php
<?php declare(strict_types=1);
if (!\function_exists('xdebug_set_filter')) {
    return;
}

\xdebug_set_filter(
    \XDEBUG_FILTER_CODE_COVERAGE,
    \XDEBUG_PATH_WHITELIST,
    [
        'C:\path\to\dir\\'
    ]
);
```

So constant `DIRECTORY_SEPARATOR` can be used with proper escaping.

shivammathur:
Hi @sebastianbergmann I will submit a PR in a bit.

shivammathur:
@sebastianbergmann Your patch gives a syntax error. See https://github.com/sebastianbergmann/phpunit/pull/3835#issuecomment-529090813

PR #4335, opened 2020-06-28
Fix: Run commands through PHP binary to ensure they are run on Windows

### LINKED ISSUE #4333: Windows CI does not run tests
Example: https://github.com/sebastianbergmann/phpunit/runs/815235702?check_suite_focus=true

```
Run ./phpunit --coverage-clover=coverage.xml
  ./phpunit --coverage-clover=coverage.xml
  shell: C:\Program Files\PowerShell\7\pwsh.EXE -command ". '{0}'"
  env:
    COMPOSER_ROOT_VERSION: 9.3-dev
    PHP_EXTENSIONS: dom, json, libxml, mbstring, pdo_sqlite, soap, xml, xmlwriter
    PHP_INI_VALUES: assert.exception=1, zend.assertions=1
    COMPOSER_CACHE_DIR: ~\AppData\Local\Composer
```

That's it, no actual output from PHPUnit.

localheinz:
This PR

* [x] explicitly runs commands through PHP binary to ensure they are run on `windows-latest`
* [x] configures `git` to avoid issues with line endings when running tests on `windows-latest`
* [x] adds slashes
* [x] skips a test related to file permissions on Windows

Fixes #4333.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -141,17 +141,17 @@ jobs:
 
       - name: Install lowest dependencies with composer
         if: matrix.dependencies == 'lowest'
-        run: ./tools/composer update --no-ansi --no-interaction --no-progress --prefer-lowest
+        run: php ./tools/composer update --no-ansi --no-interaction --no-progress --prefer-lowest
 
       - name: Install highest dependencies with composer
         if: matrix.dependencies == 'highest'
-        run: ./tools/composer update --no-ansi --no-interaction --no-progress
+        run: php ./tools/composer update --no-ansi --no-interaction --no-progress
 
       - name: Run sanity check
         run: bash ./build/scripts/sanity-check
 
       - name: Run tests with phpunit
-        run: ./phpunit --coverage-clover=coverage.xml
+        run: php ./phpunit --coverage-clover=coverage.xml
 
       - name: Send code coverage report to Codecov.io
         uses: codecov/codecov-action@v1
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -112,6 +112,10 @@ jobs:
           - highest
 
     steps:
+      - name: Configure git to avoid issues with line endings
+        if: matrix.os == 'windows-latest'
+        run: git config --global core.autocrlf false
+
       - name: Checkout
         uses: actions/checkout@v2
 
--- a/tests/unit/Framework/AssertTest.php
+++ b/tests/unit/Framework/AssertTest.php
@@ -646,6 +646,10 @@ XML;
 
     public function testAssertFileIsNotReadable(): void
     {
+        if (\PHP_OS_FAMILY === 'Windows') {
+            self::markTestSkipped('Cannot test this behaviour on Windows');
+        }
+
         $tempFile = \tempnam(
             \sys_get_temp_dir(),
             'unreadable'
--- a/tests/unit/Util/XDebugFilterScriptGeneratorTest.php
+++ b/tests/unit/Util/XDebugFilterScriptGeneratorTest.php
@@ -19,7 +19,7 @@ final class XDebugFilterScriptGeneratorTest extends TestCase
 {
     public function testReturnsExpectedScript(): void
     {
-        $expectedDirectory = __DIR__ . \DIRECTORY_SEPARATOR;
+        $expectedDirectory = \sprintf(\addslashes('%s' . \DIRECTORY_SEPARATOR), __DIR__);
         $expected          = <<<EOF
 <?php declare(strict_types=1);
 if (!\\function_exists('xdebug_set_filter')) {


.github/workflows/ci.yml
```diff
@@ -141,17 +141,17 @@ jobs:
 
       - name: Install lowest dependencies with composer
         if: matrix.dependencies == 'lowest'
-        run: ./tools/composer update --no-ansi --no-interaction --no-progress --prefer-lowest
+        run: php ./tools/composer update --no-ansi --no-interaction --no-progress --prefer-lowest
```
localheinz, line 144: @shivammathur

Do you have an idea why it would be necessary to specify the `php` binary here for running these commands on `windows-latest`?

Without it, they are not run, see #4333.

shivammathur: @localheinz 
`#!/usr/bin/env php` does not work on Windows, so you will have to call local `phar` tools with `php`. 

When tools are installed with `setup-php` on Windows, it creates a `batch` script to call the tools with php.

localheinz: Thank you, @shivammathur!

localheinz:
@sebastianbergmann 

Tests are running on `windows-latest` now, but a bunch of them are failing. Looking into it!

dvdoug:
> Tests are running on `windows-latest` now, but a bunch of them are failing.
> 
> Looks like most of them are related to line-endings.

Most of the failures I think will go away if the checkout uses LF, rather than CRLF. Try the workaround on https://github.com/actions/checkout/issues/38 or forcing LF via `.gitattributes`.

localheinz:
Thank you, @dvdoug - that helped! 👍

localheinz:
Does anyone have an idea how to fix the remaining test failures?

I haven't used Windows in the last 10 years.

shivammathur:
@localheinz To fix `XDebugFilterScriptGeneratorTest`

https://github.com/sebastianbergmann/phpunit/blob/081dcb49199f89cf0056a08f174d574f787fa1ac/tests/unit/Util/XDebugFilterScriptGeneratorTest.php#L31

```diff
- $expectedDirectory = __DIR__ . DIRECTORY_SEPARATOR; 
+ $expectedDirectory = addslashes(__DIR__ . DIRECTORY_SEPARATOR); 
```

to match

https://github.com/sebastianbergmann/phpunit/blob/081dcb49199f89cf0056a08f174d574f787fa1ac/src/Util/XdebugFilterScriptGenerator.php#L68

Related issue - https://github.com/sebastianbergmann/phpunit/issues/3834

localheinz:
@shivammathur 

I applied a fix with b2ef482e3, but I think it did not help.

shivammathur:
@localheinz Try this, I missed `sprintf`

https://github.com/sebastianbergmann/phpunit/blob/081dcb49199f89cf0056a08f174d574f787fa1ac/src/Util/XdebugFilterScriptGenerator.php#L67-L70

```diff
- $expectedDirectory = __DIR__ . DIRECTORY_SEPARATOR; 
+ $expectedDirectory = sprintf(addslashes('%s'.DIRECTORY_SEPARATOR), __DIR__); 
```

dvdoug:
On Windows, `chmod()` can set the read only flag but that's about it, the filesystem is simply too different in the way it works. 

Thus unless you want to try and use shell scripting to set up ACLs, I'd suggest simply skipping `PHPUnit\Framework\AssertTest::testAssertFileIsNotReadable` on Windows.

dvdoug:
The last failing test tests columns=max which is an implicitly variable number, but has a hard-coded result? I don't think that's a safe expectation as this is now showing.

tests/end-to-end/cli/columns-max.phpt
```diff
@@ -8,6 +8,14 @@ $_SERVER['argv'][3] = __DIR__ . '/../../_files/BankAccountTest.php';
 
 require __DIR__ . '/../../bootstrap.php';
 PHPUnit\TextUI\Command::main();
+--SKIPIF--
+<?php
+
+if (PHP_OS_FAMILY !== 'Windows') {
+    die('Maximum column length resolves to different value on Windows systems');
```
localheinz, line 15: Perhaps 120 characters?

sebastianbergmann: Might be solved by 82a15e69b2088822b3e8afac24e21ff611b8fa53.

sebastianbergmann: Because only the `--prefer-lowest` builds were failing.

dvdoug: 120 is the default, and probably safe for CI although it is configurable on a developer system

image

localheinz: That's it, @dvdoug and @sebastianbergmann!

👍

localheinz:
Thank you, @dvdoug, @sebastianbergmann, and @shivammathur!

PR #4337, opened 2020-06-29
Enhancement: Enable explicit_string_variable fixer

localheinz:
This PR

* [x] enables the `explicit_string_variable` fixer
* [x] runs `php-cs-fixer`
* [x] regenerates the baseline for `psalm`

>**explicit_string_variable** [`@PhpCsFixer`]
>
>Converts implicit variables into explicit ones in double-quoted strings or heredoc syntax.


localheinz:

PR #4338, opened 2020-06-29
Fix: Use constant_case instead of deprecated lowercase_constants fixer

localheinz:
This PR

* [x] uses the `constant_case` fixer instead of the deprecated `lowercase_constants` fixer

💁‍♂️ Running

```
$ ./tools/php-cs-fixer describe lowercase_constants
```

yields

```
Description of lowercase_constants rule.
The PHP constants `true`, `false`, and `null` MUST be in lower case. DEPRECATED: use `constant_case` instead.

Fixing examples:
 * Example #1.
   --- Original
   +++ New
   @@ -1,4 +1,4 @@
    <?php
   -$a = FALSE;
   -$b = True;
   -$c = nuLL;
   +$a = false;
   +$b = true;
   +$c = null;

```


localheinz:

PR #4339, opened 2020-06-29
Enhancement: Enable increment_style fixer

localheinz:
This PR

* [x] enables the `increment_style` fixer
* [x] runs `php-cs-fixer fix`

>**increment_style** [`@Symfony`, `@PhpCsFixer`]
>
>Pre- or post-increment and decrement operators should be used if possible.
>
>Configuration options:
>
>* `style` (`'post'`, `'pre'`): whether to use pre- or post-increment and decrement operators; defaults to `'pre'`

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -75,6 +75,9 @@ return PhpCsFixer\Config::create()
         'full_opening_tag' => true,
         'function_declaration' => true,
         'header_comment' => ['header' => $header, 'separate' => 'none'],
+        'increment_style' => [
+            'style' => PhpCsFixer\Fixer\Operator\IncrementStyleFixer::STYLE_POST,
+        ],
         'indentation_type' => true,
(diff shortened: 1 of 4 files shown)


src/Framework/Assert.php
```diff
@@ -2830,7 +2830,7 @@ public static function assertEqualXMLStructure(DOMElement $expectedElement, DOME
                 )
             );
 
-            for ($i = 0; $i < $expectedElement->attributes->length; $i++) {
+            for ($i = 0; $i < $expectedElement->attributes->length; ++$i) {
```
localheinz, line 2833: When preferred, I could also configure to consistently use the post-increment style.

sebastianbergmann: I think I would prefer that, yes.

localheinz:

PR #4340, opened 2020-06-29
Make ReturnValueMap fail when invoked arguments did not match the mocked ones

0leksandr:
When a method is stubbed with `ReturnValueMap`, and actual invoked arguments did not match to any of mocked ones' lists, method silently returns `void`.
IMHO, this is an issue, since user might expect different behavior, but test still may pass. The assumption is that user expects only stubbed argument lists, i.e. defined in `ReturnValueMap`.
This PR makes `ReturnValueMap::invoke` throw exception in such cases.

### DIFF
--- a/src/Framework/MockObject/Stub/ReturnValueMap.php
+++ b/src/Framework/MockObject/Stub/ReturnValueMap.php
@@ -26,7 +26,7 @@ final class ReturnValueMap implements Stub
 
     public function __construct(array $valueMap)
     {
-        $this->valueMap = \array_filter($valueMap, '\is_array');
+        $this->valueMap = \array_values(\array_filter($valueMap, '\is_array'));
     }
 
     public function invoke(Invocation $invocation)
@@ -45,7 +45,7 @@ final class ReturnValueMap implements Stub
             }
         }
 
-        throw $this->getExpectationFailedException($invocation->getParameters());
+        throw $this->getExpectationFailedException($invocation);
     }
 
     public function toString(): string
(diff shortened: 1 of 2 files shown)


sebastianbergmann:
Please explain the problem you are trying to solve here.

0leksandr:
Hi! @sebastianbergmann, is this (potentially) going to be merged/pushed?
If yes, but anything has to be done additionally, please let me know.
If no, I can close this PR.

sebastianbergmann:
I am still waiting for a reply to my [question](https://github.com/sebastianbergmann/phpunit/pull/4340#issuecomment-651592079):
> Please explain the problem you are trying to solve here.

0leksandr:
Oh, sorry, i thought you meant to describe purpose of commit in description, so updated that one (thought you were notified about it, sorry :disappointed:).
Following the description (see above), this PR comes from my particular user-case: when I mock with `ReturnValueMap`, i expect only arguments, that were mocked. Anything else i'd like to treat as fail.
Not sure if it's only me, or it might be considered as general approach.

0leksandr:
Trying to explain as simple as i can:
The following code
```lang=php
$mock->method('m')->willReturnValueMap([['a', 'b']]);
$mock->m('c');
```
should (may?) be treated as test failure, because `ReturnValueMap` is a stub _and_ an expectation (like `ParametersRule`) at the same time (IMHO). And this expectation was not met, since user did not expect `['c']` as arguments for `m` (they expected only `['a']`).

@sebastianbergmann, should i try to provide additional explanation/reasoning, or this may be enough for you to decide whether this functionality might be comited?

P.S. If it will be accepted, there is one other questionable moment left - what should be done in case if arguments did not match. Here, i am throwing an `ExpectationFailedException` with invoked and first stubbed (if present) arguments' list, because it is already integrated in PhpStorm. If you have better ideas, which i could try to implement, please let me know.

0leksandr:
At the moment, to achieve this functionality in my code, I use hackish wrappers over PHPUnit.
@sebastianbergmann, if this is a problem not only for me, I'd be glad to commit this into your framework. If you agree that the problem exists, but think that it should be solved/implemented in different way, please let me know.
Do I have to provide something more (explanation, reasoning, justification etc.)?

0leksandr:
@sebastianbergmann , if this PR is stuck because of breaking backward compatibility ([MockObjectTest::testStubbedReturnValueMap](https://github.com/sebastianbergmann/phpunit/pull/4340/files#diff-4da0d137c73f6befdfba3e3d890c4ff0c738d1d8e18e3192642b8e60c933ce8cL234)), I'd be glad to refactor it according to new rules and re-open in branch 9.4 (or whichever you think is appropriate).
If this functionality is not desired and should not be committed in any branch/version, I shall close this PR.
If I can do anything else, please let me know.

0leksandr:
Sigh :pensive:

PR #4344, opened 2020-06-29
Add flag for simulation of strict typing on mock methods

0leksandr:
Currently it is possible to change/break signature of mocked methods.
Possible bad scenarios:
- a method is mocked with incorrect signature (expects argument types, that are different from declared in original signature); incorrect value type is being passed to mocked method; test passes
- user didn't bind any expectation on arguments (neither correct, nor wrong ones); wrong parameter type is passed; test ignores that and passes

Although in real environment these methods will throw exceptions (because of argument types or number).

This PR allows to set flag "strictTypesCheck", which dynamically simulates static checks in real application (with declared `strict_types=1`).

### DIFF
--- a/src/Framework/MockObject/Api/Api.php
+++ b/src/Framework/MockObject/Api/Api.php
@@ -33,6 +33,11 @@ trait Api
     private $__phpunit_returnValueGeneration = true;
 
     /**
+     * @var bool
+     */
+    private $__phpunit_strictTypesCheck = false;
+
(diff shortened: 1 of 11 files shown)


sebastianbergmann:
Please explain the problem you are trying to solve here.

0leksandr:
I'm wrong, "problem" from description is not real, and already solved by your library. Sorry.

PR #4345, opened 2020-06-29
Add InvokedWithArgsCount rule

mblakley:
If you have a single function that you call with multiple different arguments, but you only care that the function got called with a specific argument, there doesn’t seem to be an easy way to set an expectation on it.  Example:
```
$this->myObject->expects($this->once())->method('notice')->with('A specific argument');
```
In this example, myObject->notice would be called multiple times with other arguments, but in the test I only care that it was called with this one specific argument value.  It doesn't seem to be possible to configure a mock object to check the expectations correctly in this case unless you enumerate all of the argument values that are passed in and use returnValueMap.

My solution to this was to set both the expectation and the expected argument in the same call, and also add a check for how many times the specified method is called with the specified arguments.  Example of my new class in use:
```
$this->myObject->expects(new InvokedWithArgsCount(1, ['A specific argument'])->method('notice');
```

I also allow other rules, to allow applying different levels of checks on different arguments:
```
$this->myObject->expects(new InvokedWithArgsCount(1, ['A specific argument', $this->anything()])->method('notice');
```

### DIFF
--- a/src/Framework/MockObject/Rule/InvokedWithArgsCount.php
+++ b/src/Framework/MockObject/Rule/InvokedWithArgsCount.php
@@ -66,9 +66,9 @@ class InvokedWithArgsCount extends InvocationOrder
     {
         parent::invoked($invocation);
 
-        $count = $this->incrementInvocationWithArgsCount($invocation);
+        $this->incrementInvocationWithArgsCount($invocation);
 
-        if ($count > $this->expectedCount) {
+        if ($this->currentCount > $this->expectedCount) {
             $message = $invocation->toString() . ' ';
 
             switch ($this->expectedCount) {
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it as I do not believe that there is a general use case for this.

mblakley:
Mockery supports this use case (sort of the default behavior), and we have used it extensively in our codebase.  We are attempting to use base phpunit as the standard, but not supporting this makes it much more difficult to convert our existing tests.

If there's another suggestion on how to handle this case with existing code, I would gladly take your advice for how to do that.  Or, if there would be a better way to add or modify a different class in the fluent chain order to accomplish this, I could work on that.

chadbarbedatto:
@sebastianbergmann I would like to argue in favor of this pull request. I've been a software engineer for 18 years and there have been numerous occasions where I have written code that makes multiple non-critical calls to some API but one critical call that needs to be tested. For example, perhaps you have code that uses an API to save some internal state along the way of accomplishing a task but then one final call to that API that save the final state and THAT is the thing you want to test. Requiring all of those internal calls to be mocked is tedious and verbose when all you really care about is that one final call.

PR #4352, opened 2020-07-03
Fix: Remove duplicated configuration

localheinz:
This PR

* [x] removes a duplicated configuration 

Related to #4338.

💁‍♂️ Running

```
$ /tools/php-cs-fixer describe lowercase_constants
```

yields

```
Description of lowercase_constants rule.
The PHP constants `true`, `false`, and `null` MUST be in lower case. DEPRECATED: use `constant_case` instead.

Fixing examples:
 * Example #1.
   --- Original
   +++ New
   @@ -1,4 +1,4 @@
    <?php
   -$a = FALSE;
   -$b = True;
   -$c = nuLL;
   +$a = false;
   +$b = true;
   +$c = null;

```

❓ Perhaps something went wrong when resolving conflicts?

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -85,7 +85,6 @@ return PhpCsFixer\Config::create()
         'line_ending' => true,
         'list_syntax' => ['syntax' => 'short'],
         'logical_operators' => true,
-        'lowercase_constants' => true,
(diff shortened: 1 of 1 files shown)


localheinz:

PR #4353, opened 2020-07-03
Enhancement: Enable magic_method_casing fixer

localheinz:
This PR

* [x] enables the `magic_method_casing` fixer

>**magic_method_casing** [`@Symfony`, `@PhpCsFixer`]
>
>Magic method definitions and calls must be using the correct casing.


localheinz:

PR #4354, opened 2020-07-03
Enhancement: Enable native_function_type_declaration_casing fixer

localheinz:
This PR

* [x] enables the `native_function_type_declaration_casing` fixer

>**native_function_type_declaration_casing** [`@Symfony`, `@PhpCsFixer`]
>
>Native type hints for functions should use the correct case.


localheinz:

PR #4355, opened 2020-07-03
Enhancement: Enable no_spaces_around_offset fixer

localheinz:
This PR

* [x] enables the `no_spaces_around_offset` fixer

>**no_spaces_around_offset** [`@Symfony`, `@PhpCsFixer`]
>
>There MUST NOT be spaces around offset braces.
>
>Configuration options:
>
>* `positions` (a subset of `['inside', 'outside']`): whether spacing should be fixed inside and/or outside the offset braces; defaults to `['inside', 'outside']`


localheinz:

PR #4356, opened 2020-07-03
Fix: Remove useless DocBlocks

localheinz:
This PR

* [x] removes useless DocBlocks

💁‍♂️ I intended to enable the `phpdoc_inline_tag` fixer, but figured we could do without these DocBlocks instead.


localheinz:

PR #4358, opened 2020-07-03
Enhancement: Enable strict_param fixer

localheinz:
This PR

* [x] enables the `strict_param` fixer
* [x] runs `php-cs-fixer fix`

>**strict_param** [`@PhpCsFixer:risky`]
>
>Functions should be used with `$strict` param set to `true`.
>
>Risky rule: risky when the fixed function is overridden or if the code relies on non-strict usage.


localheinz:

PR #4359, opened 2020-07-03
Fix: Avoid use of empty()

localheinz:
This PR

* [x] avoids the use of `empty()` where possible

### DIFF
--- a/src/Framework/Assert.php
+++ b/src/Framework/Assert.php
@@ -2825,7 +2825,7 @@ abstract class Assert
                 \sprintf(
                     '%s%sNumber of attributes on node "%s" does not match',
                     $message,
-                    !empty($message) ? "\n" : '',
+                    '' !== $message ? "\n" : '',
                     $expectedElement->tagName
                 )
             );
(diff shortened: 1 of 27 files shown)


sebastianbergmann:
Why do you want to avoid `empty()`?

localheinz:
@sebastianbergmann 

Perhaps it is a matter of taste, but with `empty()`, I find that the author's intent is unclear.

```php
var_dump([
    'array' => empty([]),
    'bool-false' => empty(false),
    'float' => empty(0.0),
    'int' => empty(0),
    'null' => empty(null),
    'string' => empty(''),
    'undefined' => empty($foo),
]);
```

All of these evaluate to `true`.

By not using `empty()`, we can clarify the intent.

epdenouden:
I like empty() more because that want to express, but I also like [] as it's more precise.

Yes I'll go make myself useful somewhere else now.

localheinz:
@epdenouden @sebastianbergmann 

Ha, no problem!

🤓

epdenouden:
Well you didn't have to close it, is what I was saying. Even if you do, I will be using more specific conditionals where appropriate anyway. :)

I also get @sebastianbergmann: with increasingly type-strict code the meaning of empty()

Question @localheinz, is it possible to teach Psalm to complain about ambiguous use of empty and leave it alone when it can deduce the type?

localheinz:
@epdenouden 

That’s probably a question @muglug can answer!

☺️

sebastianbergmann:
Does Psalm "complain" about `empty()`?

epdenouden:
Currently it doesn't. I don't mind either way :)

PR #4362, opened 2020-07-03
Fix: Remove condition that always evaluates to true

localheinz:
This PR

* [x] removes a condition that always evaluates to `true`

### DIFF
--- a/src/TextUI/TestRunner.php
+++ b/src/TextUI/TestRunner.php
@@ -470,34 +470,30 @@ final class TestRunner extends BaseTestRunner
             if (isset($arguments['configuration'])) {
                 $filterConfiguration = $arguments['configuration']->getFilterConfiguration();
 
-                if (!empty($filterConfiguration['whitelist'])) {
-                    $whitelistFromConfigurationFile = true;
+                $whitelistFromConfigurationFile = true;
+
+                foreach ($filterConfiguration['whitelist']['include']['directory'] as $dir) {
+                    $this->codeCoverageFilter->addDirectoryToWhitelist(
+                        $dir['path'],
+                        $dir['suffix'],
+                        $dir['prefix']
+                    );
                 }
 
-                if (!empty($filterConfiguration['whitelist'])) {
-                    foreach ($filterConfiguration['whitelist']['include']['directory'] as $dir) {
-                        $this->codeCoverageFilter->addDirectoryToWhitelist(
-                            $dir['path'],
-                            $dir['suffix'],
-                            $dir['prefix']
-                        );
-                    }
+                foreach ($filterConfiguration['whitelist']['include']['file'] as $file) {
+                    $this->codeCoverageFilter->addFileToWhitelist($file);
+                }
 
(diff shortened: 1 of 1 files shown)


src/TextUI/TestRunner.php
```diff
@@ -472,9 +472,7 @@ public function doRun(Test $suite, array $arguments = [], array $warnings = [],
 
                 if (!empty($filterConfiguration['whitelist'])) {
```
localheinz, line 473: Not sure, but I think this condition always evaluates to `true`.

👀

src/TextUI/TestRunner.php
```diff
@@ -470,34 +470,30 @@ public function doRun(Test $suite, array $arguments = [], array $warnings = [],
             if (isset($arguments['configuration'])) {
                 $filterConfiguration = $arguments['configuration']->getFilterConfiguration();
```
localheinz, line 471: Running 

```
$ git grep getFilterConfiguration
```

yields

```
src/TextUI/TestRunner.php:471:                $filterConfiguration = $arguments['configuration']->getFilterConfiguration();
src/Util/Configuration.php:138:    public function getFilterConfiguration(): array
tests/unit/Util/ConfigurationTest.php:196:            $this->configuration->getFilterConfiguration()
tests/unit/Util/ConfigurationTest.php:554:        $filter = $emptyConfiguration->getFilterConfiguration();
tests/unit/Util/ConfigurationTest.php:617:            $expectedConfiguration->getFilterConfiguration(),
tests/unit/Util/ConfigurationTest.php:618:            $actualConfiguration->getFilterConfiguration()
```

The only class that declares this method is `final` and does not implement an interface:

https://github.com/sebastianbergmann/phpunit/blob/0d3b3170a38f876773df670bdf612ea33875f9a2/src/Util/Configuration.php#L24

The method `getFilterConfiguration()` returns an array where the value for key `whitelist` is never empty:

https://github.com/sebastianbergmann/phpunit/blob/0d3b3170a38f876773df670bdf612ea33875f9a2/src/Util/Configuration.php#L138-L199

src/TextUI/TestRunner.php
```diff
@@ -470,34 +470,30 @@ public function doRun(Test $suite, array $arguments = [], array $warnings = [],
...
 
-                if (!empty($filterConfiguration['whitelist'])) {
-                    $whitelistFromConfigurationFile = true;
-                }
+                $whitelistFromConfigurationFile = true;
```
localheinz, line 473: In light of https://github.com/sebastianbergmann/phpunit/pull/4362/files#r449600526, the question is whether this variable makes sense.

What do you think, @sebastianbergmann?

sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it as I expect merging this from `8.5` to `9.2` and `master` to be a nightmare. I am open, of course, to a pull request that targets `master`.

localheinz:
@sebastianbergmann 

Looking at `master`, I can understand! As far as I can see, it's not an issue there anymore!

👍

PR #4363, opened 2020-07-06
Fix: Do not run tests with PHP 8 and JIT on windows-latest

localheinz:
This PR

* [x] stops running tests on PHP 8 with JIT on `windows-latest`


localheinz:

PR #4374, opened 2020-07-10
Remove Match interface

### LINKED ISSUE #4373: "match" is a semi-reserved keyword in PHP 8
` ParseError: syntax error, unexpected 'Match' (T_MATCH), expecting identifier (T_STRING) or '{'`

https://github.com/sebastianbergmann/phpunit/blob/master/src/Framework/MockObject/Builder/Match.php

GrahamCampbell:
From what I can see, this interface was not used, other than for defining ParameterMatch interface.

---

Closes https://github.com/sebastianbergmann/phpunit/issues/4373.


GrahamCampbell:
I guess Codecov is not waiting for all builds to complete.

GrahamCampbell:
🚀

jdrieghe:
Would it be okay to merge this to the 9.2 branch and ship it in a release? The interface is not covered by the BC promise due to its internal nature and plenty of people are already running their CI suite against the latest releases of  PHP 8.

sebastianbergmann:
PHPUnit 9.2 is not supported on PHP 8, PHPUnit 9.3 will be.

jdrieghe:
@sebastianbergmann understandable. According to your suggestion here (https://github.com/sebastianbergmann/phpunit/pull/4323#issuecomment-649555557) we're already running it in CI, that's why I was asking. Not sure if that's enough of a reason to quickly backport this PR?

jdrieghe:
Alternatively, we could always uese the latest `master` branch of phpunit to run the CI pipeline, but that also seems like a dangerous path to go down.

GrahamCampbell:
You can use `^9.3` for a version and in your composer.json, set minimum-stability dev and prefer-stable true. That way you will get 9.3.x dev versions until 9.3.0 is tagged, then stable versions from there on out.

jdrieghe:
Thanks @GrahamCampbell. Never thought of that.

PR #4378, opened 2020-07-15
Simplify array access

muglug:
Psalm doesn't know whether or not `$i` is `null` (it's not smart enough to understand that `false !== $arg = current($args)` implies `$args` is non-empty).

But, regardless, the array access is not necessary there.


localheinz (approved):
👍

sebastianbergmann:
`Workflow runs completed with no jobs`?

PR #3921, opened 2019-10-31
Enhancement: Run tests using GitHub actions, but keep collecting code coverage using Travis CI

localheinz:
This PR

* [x] lints XML files using GitHub actions
* [x] runs tests and collects code coverage using GitHub actions, and collects and sends code coverage to Codecov.io using Travis CI for pull requests

Follows https://github.com/sebastianbergmann/phpunit/pull/3833.

PR #4384, opened 2020-07-24
CI: remove double Dockerfile compiling

Slamdunk:
As far as I can tell from the GA logs https://github.com/sebastianbergmann/phpunit/runs/905995415?check_suite_focus=true the Docker image is build on every step, so the first one is useless.

Ping @localheinz as original proposer in https://github.com/sebastianbergmann/phpunit/pull/3921

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -85,9 +85,6 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
 
-      - name: Build Docker image
-        uses: ./.docker/lint-xml-configuration
-
       - name: Lint XML configuration files
         uses: ./.docker/lint-xml-configuration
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -85,9 +85,6 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
 
-      - name: Build Docker image
-        uses: ./.docker/lint-xml-configuration
```
localheinz, line 89: @Slamdunk 

This step is here only to separate building of the Docker image from using it.

sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. The task the Docker image was used for is no longer performed.

Issue #4288, opened 2020-06-11
Migration of PHPUnit XML Configuration file


sebastianbergmann:
[PHPUnit 9.3](https://github.com/sebastianbergmann/phpunit/milestone/38) introduces significant changes to PHPUnit's XML configuration file (`phpunit.xml`) for code coverage and logging:

#### PHPUnit 9.2
```xml
<?xml version="1.0" encoding="UTF-8"?>
<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/9.2/phpunit.xsd"
         cacheTokens="true"
         disableCodeCoverageIgnore="true"
         ignoreDeprecatedCodeUnitsFromCodeCoverage="true">
    <!-- ... -->

    <filter>
        <whitelist addUncoveredFilesFromWhitelist="true"
                   processUncoveredFilesFromWhitelist="true">
            <directory suffix=".php">src</directory>

            <exclude>
                <directory suffix=".php">src/generated</directory>
                <file>src/autoload.php</file>
            </exclude>
        </whitelist>
    </filter>

    <logging>
        <log type="coverage-html" target="html-coverage" lowUpperBound="50" highLowerBound="90"/>
        <log type="coverage-clover" target="clover.xml"/>
        <log type="coverage-crap4j" threshold="50" target="crap4j.xml"/>
        <log type="coverage-text" showUncoveredFiles="false" showOnlySummary="true" target="coverage.txt"/>
        <log type="junit" target="junit.xml"/>
        <log type="teamcity" target="teamcity.txt"/>
        <log type="testdox-html" target="testdox.html"/>
        <log type="testdox-text" target="testdox.txt"/>
        <log type="testdox-xml" target="testdox.xml"/>
        <log type="plain" target="logfile.txt"/>
    </logging>
</phpunit>
```
#### PHPUnit 9.3
```xml
<?xml version="1.0" encoding="UTF-8"?>
<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/9.3/phpunit.xsd">
    <!-- ... -->

    <coverage includeUncoveredFiles="true"
              processUncoveredFiles="true"
              ignoreDeprecatedCodeUnits="true"
              disableCodeCoverageIgnore="true">
        <include>
            <directory suffix=".php">src</directory>
[...]

PR #4385, opened 2020-07-26
Tentatively implements #4288

theseer:
This should implement all the migration steps required to move from the 9.2 sample configuration xml to the 9.3 expected output format. As per discussion with @sebastianbergmann, the Migrator code is ready to easily deal with more migrations coming in the future and allows for adding code to migrate configuration files from before 9.2.

__Pending issues__:

* [x] There's an attribute `cacheTokens="true"` in the root element of the `input-9.2.xml`. It is missing in the `output-9.3.xml` but I was not convinced simply removing it would be the correct migration.

* [ ] The test for the migration is failing as the XML, while being semantically identical, is not from a formatting perspective so the string comparison based assertion fails. I don't see a point in matching the formatting to make the test happy but didn't want to simply change it. We should adjust the assertion.

* [ ] There is some minor code duplication (ensuring the newly created `<coverage>` node exists) in some of the migration classes. This could potentially be refactored to avoid it. Not sure if it's worth the effort.

* [ ] As stated before, no integration is included, e.g. nothing calls this migration code except for the test

### DIFF
--- a/src/TextUI/XmlConfiguration/Migration/MigrationBuilder.php
+++ b/src/TextUI/XmlConfiguration/Migration/MigrationBuilder.php
@@ -1,0 +1,54 @@
+<?php declare(strict_types=1);
+/*
+ * This file is part of PHPUnit.
+ *
+ * (c) Sebastian Bergmann <sebastian@phpunit.de>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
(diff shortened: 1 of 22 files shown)


sebastianbergmann:
> There's an attribute `cacheTokens="true"` in the root element of the `input-9.2.xml`. It is missing in the `output-9.3.xml` but I was not convinced simply removing it would be the correct migration.

`cacheTokens="true"` has been removed without replacement.

Issue #4024, opened 2020-01-21
Make PHPUnit\TextUI\ResultPrinter an interface


jaydiablo:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 8.x+ (8.5.2 for sure)
| PHP version         | any
| Installation Method | n/a

#### Summary

The `PHPUnit\TextUI\ResultPrinter` class is marked as internal (https://github.com/sebastianbergmann/phpunit/blob/master/src/TextUI/ResultPrinter.php#L29) which is causing notices to be emitted during test runs that have an external result printer configured:

https://github.com/diablomedia/phpunit-pretty-printer/issues/27
https://github.com/mikeerickson/phpunit-pretty-result-printer/issues/155
https://github.com/sempro/phpunit-pretty-print/issues/28

```
/var/www/html $ bin/phpunit --filter ArraysTest
PHPUnit 8.5.2 by Sebastian Bergmann and contributors.

Testing Project Test Suite

                       App\Tests\Common\ArraysTest ...

Time: 216 ms, Memory: 8.00 MB

OK (3 tests, 6 assertions)

Other deprecation notices (1)

  1x: The "PHPUnit\TextUI\ResultPrinter" class is considered internal This class is not covered by the backward compatibility promise for PHPUnit. It may change without further notice. You should not use it from "DiabloMedia\PHPUnit\Printer\PrettyPrinter".
```

Is it possible to have this class be part of the public API as mentioned in #3236 ?

If not, what is the recommended way to implement result printers?

Thanks!

sebastianbergmann:
Let me start by saying that

```
Other deprecation notices (1)

  1x: The "PHPUnit\TextUI\ResultPrinter" class is considered internal This class is not covered by the backward compatibility promise for PHPUnit. It may change without further notice. You should not use it from "DiabloMedia\PHPUnit\Printer\PrettyPrinter".
```

is not emitted by PHPUnit. I assume that you are using a wrapper around PHPUnit, probably something like Symfony's PHPUnit bridge.

With that out of the way, yes, the fact that `PHPUnit\TextUI\ResultPrinter` is marked `@internal` is a problem. Its implementation is, and should stay, internal. However, for legacy reasons, it is also a public API as extending this class is an intended way for augmenting or changing the output printed by PHPUnit.

I think the best way forward is to

[...]

PR #4390, opened 2020-07-28
Update default printerClass

echojoshchen:
Update the default printerClass to DefaultResultPrinter.

This matches the change in #4024


sebastianbergmann:
Merged manually, thanks.

Issue #4414, opened 2020-08-10
Cache PHPUnit's cache(s) on GitHub Actions


sebastianbergmann:
As of d5720810075eb16b9f8b5862f6cf344133f8b34c, running PHPUnit's own test suite uses the `.phpunit.cache` directory for PHPUnit's test result cache as well as php-code-coverage's static analysis cache.

It would be nice if we could keep this directory between GitHub Actions runs.

PR #4415, opened 2020-08-10
Enhancement: Update actions/cache

localheinz:
This PR

* [x] updates `actions/cache`

Slightly related to #4414.

💁‍♂️ For reference, see https://github.com/actions/cache/compare/v1...v2.


localheinz:

Issue #4411, opened 2020-08-10
Running PHPUnit's own test suite creates two test result cache files


sebastianbergmann:
Running PHPUnit's own test suite creates two test result cache files: one in the location configured in https://github.com/sebastianbergmann/phpunit/blob/9.3.4/phpunit.xml#L7 (`.phpunit.cache/test-results`) and one in the default location (`.phpunit.result.cache`):

```
$ ls -lh .phpunit.cache
ls: cannot access '.phpunit.cache': No such file or directory

$ ls -lh .phpunit.result.cache
ls: cannot access '.phpunit.result.cache': No such file or directory
```

```
$ ./phpunit
.
.
.
```

```
$ ls -lh .phpunit.cache       
total 260K
-rw-r--r--. 1 sb sb 259K Aug 10 08:14 test-results

$ ls -lh .phpunit.result.cache
-rw-r--r--. 1 sb sb 27K Aug 10 08:14 .phpunit.result.cache
```

I am almost certain that `.phpunit.result.cache` is created by a test that does not clean up after itself (and/or uses the wrong working directory).

PR #4416, opened 2020-08-10
Enhancement: Cache .phpunit.cache directory between builds

### LINKED ISSUE #4414: Cache PHPUnit's cache(s) on GitHub Actions

localheinz:
This PR

* [x] caches the `.phpunit.cache` directory between builds

Fixes #4414.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -157,6 +157,15 @@ jobs:
       - name: Run sanity check
         run: bash ./build/scripts/sanity-check
 
+      - name: Cache phpunit cache directory
+        uses: actions/cache@v2
+        with:
+          path: .phpunit.cache/
+          key: php${{ matrix.php-version }}-phpunit-cache-${{ github.ref }}-${{ github.sha }}
+          restore-keys: |
+            php${{ matrix.php-version }}-phpunit-cache-${{ github.ref }}-
+            php${{ matrix.php-version }}-phpunit-cache-
+
       - name: Run tests with phpunit
         run: php ./phpunit --coverage-clover=coverage.xml
 


.github/workflows/ci.yml
```diff
@@ -157,6 +157,15 @@ jobs:
       - name: Run sanity check
         run: bash ./build/scripts/sanity-check
 
+      - name: Cache phpunit result cache
+        uses: actions/cache@v2
+        with:
+          path: .phpunit.result.cache
+          key: php${{ matrix.php-version }}-phpunit-result-cache-${{ github.ref }}-${{ github.sha }}
```
localheinz, line 164: Does that make sense?

Screen Shot 2020-08-10 at 11 30 09


💁‍♂️ For reference, see https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#github-context.

sebastianbergmann:
`.phpunit.result.cache` must not be cached. It is created "by accident" due to #4411.

localheinz:
@sebastianbergmann 

Ha, hang on - changing the cache target!

sebastianbergmann:
> Ha, hang on - changing the cache target!

Thanks! I misclicked and did not want to close this. Sorry!

localheinz:
@sebastianbergmann 

> Thanks! I misclicked and did not want to close this. Sorry!

No problem, I misread the ticket!

sebastianbergmann:
Can you change the target to `9.3` please? Thanks!

localheinz:
@sebastianbergmann 

Done!

sebastianbergmann:
Thanks!

Only now do I realize that this might not work the way I expect it to work. php-code-coverage's cache looks at `filemtime()` of the source code files and compares that to `filemtime()` of the cache file. If GitHub Actions makes a fresh clone/checkout for each build then we will never get a cache hit because `filemtime()` will always be newer than `filemtime()` for the cache files.

localheinz:
@sebastianbergmann 

Reminds me of https://github.com/phpstan/phpstan/issues/2122.

localheinz:
@sebastianbergmann 

Perhaps this helps?

https://github.com/phpstan/phpstan-src/commit/be3e4fc84b2cca3e06bb52d6031c5ca4ff12a8a9

/cc @ondrejmirtes

sebastianbergmann:

PR #4418, opened 2020-08-10
Fix invalid reference to ArrayAccess

muglug:


sebastianbergmann:
`src/Framework/Assert/Functions.php` is a generated file. Any change must be implemented in `build/scripts/generate-global-assert-wrappers.php`, the script that generates the file.

sebastianbergmann:
@localheinz Thanks! Can you make this against `9.3`, please? Thanks!

localheinz:
@muglug @sebastianbergmann 

I think this should be good!

muglug:
Thanks, sorry - last couple of days have been very busy

PR #4420, opened 2020-08-11
Enhance Exclude Migration to support multiple occurrences

### LINKED ISSUE #4419: Migration failed: whitelist element has unexpected children
| Q                   | A
| PHPUnit version     | 9.3.5
| PHP version         | 7.4.9

#### Summary

Running `vendor/bin/phpunit --migrate-configuration` results in a `Migration failed` message with the following XML file which is valid on PHPUnit 9.2:

```
<phpunit
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
    xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/9.2/phpunit.xsd"
>
    <filter>
        <whitelist>
            <directory suffix=".php">lib</directory>
            <exclude>
                <file>bundles/MyBundle/MyBundle.php</file>
            </exclude>
            <exclude>
                <directory>bundles/*Bundle/Controller</directory>
            </exclude>
        </whitelist>
    </filter>
</phpunit>
```

#### Current behavior

Migration fails. Additionaly, no detailed error message is given when it happens. I tried using `vendor/bin/phpunit --migrate-configuration -v` but it didn't help.

#### How to reproduce

Run the migration command with the above XML config.

#### Expected behavior

I expect the migration to complete.

theseer:
Closes #4419

This also introduces a SnapshotNodeList helper since we can't use live
lists as we are modifying the DOM tree while iterating

### DIFF
--- a/src/TextUI/XmlConfiguration/Migration/Migrations/MoveWhitelistExcludesToCoverage.php
+++ b/src/TextUI/XmlConfiguration/Migration/Migrations/MoveWhitelistExcludesToCoverage.php
@@ -30,6 +30,7 @@ final class MoveWhitelistExcludesToCoverage implements Migration
         }
 
         $excludeNodes = SnapshotNodeList::fromNodeList($whitelist->getElementsByTagName('exclude'));
+
         if ($excludeNodes->count() === 0) {
             return;
         }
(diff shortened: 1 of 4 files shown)


sebastianbergmann:
Cherry-picked into `9.3`.

Issue #4422, opened 2020-08-11
Regression in PHPUnit 9.3 when using libxml_disable_entity_loader in a data provider


bcremer:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 9.3
| PHP version         | n/a
| Installation Method | Composer / PHAR (both)

#### Summary

PHPUnits bootstrapping behavior changed with Version 9.3.

#### Current behavior

PHPUnit fails with an Fatal error when `libxml_disable_entity_loader(true)` was called in a data provider.

#### How to reproduce

```bash
cat <<'EOF' > phpunit.xml
<?xml version="1.0" encoding="UTF-8"?>

<phpunit>
    <testsuites>
        <testsuite name="Functional">
            <directory>.</directory>
        </testsuite>
    </testsuites>
</phpunit>
EOF


cat <<'EOF' > DummyTest.php
<?php
class DummyTest extends PHPUnit\Framework\TestCase
{
    /** @dataProvider provideData */
    public function testDummy() {}

    public function provideData()
    {
        libxml_disable_entity_loader(true);
        set_error_handler(function ($erno, $message) {
            throw new \RuntimeException($message);
        });
    }
}
EOF


wget -O phpunit9.2 https://phar.phpunit.de/phpunit-9.2.phar
wget -O phpunit9.3 https://phar.phpunit.de/phpunit-9.3.phar
```

```bash
$ php phpunit9.2 --debug
PHPUnit 9.2.6 by Sebastian Bergmann and contributors.

Test 'DummyTest::testDummy' started
Test 'DummyTest::testDummy' ended


Time: 00:00, Memory: 12.00 MB

OK, but incomplete, skipped, or risky tests!
Tests: 1, Assertions: 0, Skipped: 1.
```

```
$ php phpunit9.3 --debug
PHP Fatal error:  Uncaught RuntimeException: DOMDocument::schemaValidate(): Invalid Schema in /home/apidev/current/phpunit-repro/DummyTest.php:11
Stack trace:
#0 [internal function]: DummyTest->{closure}()
#1 phar:///home/apidev/current/phpunit-repro/phpunit9.3/phpunit/Util/Xml/Validator.php(26): DOMDocument->schemaValidate()
#2 phar:///home/apidev/current/phpunit-repro/phpunit9.3/phpunit/TextUI/XmlConfiguration/Loader.php(81): PHPUnit\Util\Xml\Validator->validate()
[...]

PR #4423, opened 2020-08-11
Fix config loading when libxml_disable_entity_loader(true) is used

bcremer:
Fixes https://github.com/sebastianbergmann/phpunit/issues/4422

### DIFF
--- a/tests/unit/Util/Xml/ValidatorTest.php
+++ b/tests/unit/Util/Xml/ValidatorTest.php
@@ -52,4 +52,23 @@ final class ValidatorTest extends TestCase
         $this->assertTrue($result->hasValidationErrors());
         $this->assertSame(PHP_EOL . '  Line 17:' . PHP_EOL . '  - Element \'filter\': This element is not expected.' . PHP_EOL, $result->asString());
     }
+
+    public function testDoesNotThrowExceptionWithDisabledEntityLoader(): void
+    {
+        libxml_disable_entity_loader(true);
+
+        $result = (new Validator())->validate(
+            (new Loader)->loadFile(
+                __DIR__ . '/../../../end-to-end/migration/possibility-to-migrate-from-92-is-detected/phpunit.xml',
+                false,
+                true,
+                true
+            ),
+            (new SchemaFinder)->find(Version::series())
+        );
+
+        $this->assertTrue($result->hasValidationErrors());
+        $this->assertSame(PHP_EOL . '  Line 17:' . PHP_EOL . '  - Element \'filter\': This element is not expected.' . PHP_EOL, $result->asString());
+        $this->assertTrue(libxml_disable_entity_loader(true), 'libxml_disable_entity_loader must be set back to it\'s original value');
+    }
 }
--- a/src/Util/Xml/Validator.php
+++ b/src/Util/Xml/Validator.php
@@ -22,12 +22,16 @@ final class Validator
     public function validate(DOMDocument $document, string $xsdFilename): ValidationResult
     {
         $originalErrorHandling = libxml_use_internal_errors(true);
+        $originalEntityLoader = libxml_disable_entity_loader(false);
 
-        $document->schemaValidate($xsdFilename);
-
-        $errors = libxml_get_errors();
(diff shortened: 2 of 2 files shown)


tests/unit/Util/Xml/ValidatorTest.php
```diff
@@ -52,4 +52,23 @@ public function testDoesNotValidateInvalidXmlFile(): void
...
+                false,
+                true,
+                true
+            ),
+            (new SchemaFinder)->find(Version::series())
+        );
+
+        $this->assertTrue($result->hasValidationErrors());
+        $this->assertSame(PHP_EOL . '  Line 17:' . PHP_EOL . '  - Element \'filter\': This element is not expected.' . PHP_EOL, $result->asString());
+        $this->assertTrue(libxml_disable_entity_loader(true), 'libxml_disable_entity_loader must be set back to it\'s original value');
```
theseer, line 72: To just "read" the state, we do not need to set "true".

bcremer: I'm not aware of a read-only function to get the current libxml_disable_entity_loader-state. Die functions default value is `true`.

bcremer:
```Deprecated: Function libxml_disable_entity_loader() is deprecated in /home/runner/work/phpunit/phpunit/src/Util/Xml/Validator.php on line 25

File: tests/unit/Util/Xml/ValidatorTest.php (hunk 1 above)
Comment (theseer) [reply tests/unit/Util/Xml/ValidatorTest.php:72]: At least according to the manual, the parameter is optional. So I at least assume that would be for "reading" the current state. I of course didn't try that ;)

**Comment by bcremer** [comment]
The PHP 8.0 runs are reporting the function call as deprecated: 

```
Deprecated: Function libxml_disable_entity_loader() is deprecated in /home/runner/work/phpunit/phpunit/src/Util/Xml/Validator.php on line 25
```

Please advise how to proceed.

**Comment by sebastianbergmann** [comment]
> Deprecated: Function libxml_disable_entity_loader() is deprecated in /home/runner/work/phpunit/phpunit/src/Util/Xml/Validator.php on line 25

This is due to http://git.php.net/?p=php-src.git;a=commitdiff;h=e0fa48f69dd14b52c8f1b2904ac7bd30472849a8.

The way I understand it, any call to `libxml_disable_entity_loader()` on PHP 8 can/should be skipped.

**Comment by theseer** [comment]
> The PHP 8.0 runs are reporting the function call as deprecated:
> 
> ```
> Deprecated: Function libxml_disable_entity_loader() is deprecated in /home/runner/work/phpunit/phpunit/src/Util/Xml/Validator.php on line 25
> ```
> 
> Please advise how to proceed.

I guess this will be getting a bit messy. If we care, we could add a condition to only call this function when PHP < 8.0 is in use and only if the current state is not what we want it to be.

If run under PHP 8, we'd have to eat the deprecation warning.

**Comment by theseer** [comment]
> The way I understand it, any call to `libxml_disable_entity_loader()` on PHP 8 can/should be skipped.


If we merely skip it the "bug" will be back, won't it? Or will PHPUnit also complain about it when it's used in a test suite?

**Comment by sebastianbergmann** [comment]
My assumption that any code, for instance @bcremer's test code and/or tested code, does not call `libxml_disable_entity_loader()` when used on PHP 8 was obviously wrong. Looks like we have to live with the deprecation warning for as long as PHPUnit supports a version of PHP that still has `libxml_disable_entity_loader()`. :-/

File: tests/unit/Util/Xml/ValidatorTest.php (hunk 1 above)
Comment (bcremer) [reply tests/unit/Util/Xml/ValidatorTest.php:72]: Unfortunately not, see: https://3v4l.org/NJprJ
 
```php
var_dump(libxml_disable_entity_loader(false)); // init with false
var_dump(libxml_disable_entity_loader()); // "read" value
var_dump(libxml_disable_entity_loader()); // value was modified
```

theseer: \o/ Okay, never mind. *sigh*

sebastianbergmann:

PR #4424, opened 2020-08-11
Use Snapshot list helper to avoid possible incomplete migrations

theseer:


sebastianbergmann:
Your patch is not up-to-date with 413a76edb3b421059d6ba5b7e4bad89d9a6e6490.

PR #4425, opened 2020-08-12
Enhancement: Add job to assert that generated files are up-to-date

localheinz:
This PR

* [x] adds a job to assert that generated files are up-to-date

Related to #4418.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -163,3 +163,52 @@ jobs:
         uses: codecov/codecov-action@v1
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
+
+  test-generated-code:
+    name: Test generated code
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        php-version:
+          - 7.2
+
+        dependencies:
+          - highest
+
+    steps:
+      - name: Checkout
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
We already have https://github.com/sebastianbergmann/phpunit/blob/master/tests/unit/Framework/Assert/FunctionsTest.php which *should* prevent the most common case that updating the generated file was forgotten when a new assertion method was added.

sebastianbergmann:
I do not like the "Generate Code" name for the job. Yes, we generate code as part of that job. But that is not what the job is about. It is about whether the generated file is up-to-date or not. I do not have a better name just yet.

localheinz:
@sebastianbergmann 

We could make this part of the **Tests** job - after all, it tests something. What do you think?

sebastianbergmann:
> We could make this part of the **Tests** job - after all, it tests something. What do you think?

Makes sense to me.

sebastianbergmann:
> > We could make this part of the **Tests** job - after all, it tests something. What do you think?
> 
> Makes sense to me.

Sorry for the noise, after a long time of silence no less, but I do think that if we do this then it should be a separate job. Let us call it "Test generated code"?

localheinz:
@sebastianbergmann 

I have renamed the job, please take a look!

PR #4431, opened 2020-08-15
Enhancement: Use trait-string where possible

localheinz:
This PR

* [x] uses `trait-string` where possible

💁‍♂️ For reference, see https://psalm.dev/docs/annotating_code/type_syntax/scalar_types/#trait-string.

### DIFF
--- a/src/Framework/MockObject/Generator.php
+++ b/src/Framework/MockObject/Generator.php
@@ -249,6 +249,8 @@ final class Generator
      * of the trait mocked. Concrete methods to mock can be specified with the
      * `$mockedMethods` parameter.
      *
+     * @psalm-param trait-string $traitName
+     *
      * @throws RuntimeException
      */
     public function getMockForTrait(string $traitName, array $arguments = [], string $mockClassName = '', bool $callOriginalConstructor = true, bool $callOriginalClone = true, bool $callAutoload = true, array $mockedMethods = null, bool $cloneArguments = true): MockObject
@@ -286,6 +288,8 @@ final class Generator
 
     /**
      * Returns an object for the specified trait.
+     *
+     * @psalm-param trait-string $traitName
      *
      * @throws RuntimeException
      */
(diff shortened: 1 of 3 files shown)


src/Framework/MockObject/Generator.php
```diff
@@ -249,6 +249,8 @@ interface_exists($originalClassName, $callAutoload)) {
      * of the trait mocked. Concrete methods to mock can be specified with the
      * `$mockedMethods` parameter.
      *
+     * @param trait-string $traitName
```
sebastianbergmann, line 252: For Psalm-specific annotations we use `@psalm-param`, at least so far. I think for this pull request it would be best to stick with that. But we should think about moving from `@psalm-param`, `@psalm-return`, and `@psalm-var` to `@param`, `@return`, and `@var`. How do other projects handle this? Pinging @Ocramius and @muglug out of the blue (sorry!).

Ocramius: Yep, using `@psalm-param` would be best to not annoy other static analysis tools that don't recognize this format.

What I generally do is combining `@psalm-param` for the actual type (if complex enough that only psalm can understand it) plus `@param` with a simpler type that is, for example, IDE friendly.

Ocramius: I think this patch is fine as-is, @sebastianbergmann

localheinz: @sebastianbergmann 

Adjusted!

I remember @Ocramius mentioning that he doesn't use the`@psalm-` prefix.

Ocramius: I do, but I am also trying to push the ecosystem forward. Moving from `@psalm-` to unprefixed will crash some tools somewhere, so be careful 😬

sebastianbergmann: As long as it does not crash any tool we are using ... 😬

localheinz:
Thank you, @Ocramius and @sebastianbergmann!

sebastianbergmann:
@muglug @Ocramius Can the `psalm-` prefix be omitted for `@psalm-template`, `@psalm-assert`, etc., too?

Ocramius:
Risky: the prefix  prevents  tools like `doctrine/annotations` from picking them up, so I suggest keeping them for the short term

PR #4439, opened 2020-08-27
Fix global assert wrapper

ngyuki:
#4418 fixes ArrayAccess, but It's not just ArrayAccess that has problems.
The same applies to the following global classes.

* [Countable](https://github.com/sebastianbergmann/phpunit/blob/9.3.6/src/Framework/Assert/Functions.php#L186)
* [DOMDocument](https://github.com/sebastianbergmann/phpunit/blob/9.3.6/src/Framework/Assert/Functions.php#L1828)
* [DOMElement](https://github.com/sebastianbergmann/phpunit/blob/9.3.6/src/Framework/Assert/Functions.php#L1903)


ngyuki:
Related: https://github.com/vimeo/psalm/pull/3985

PR #4441, opened 2020-08-29
Improvements when expecting exception objects

thiagocordeiro:
This PR changes how exception objects are expected, instead of getting exception values (class, message and code) the instance itself gets tested, which means any additional value on the exception might hold will also be tested


sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it as I think that this may be confusing. To be honest, I already think that `expectExceptionObject()` is superfluous at best and harmful at worst and should have never been added.

PR #4442, opened 2020-08-29
Code quality improvements

fpesch:


src/Util/Log/JUnit.php
```diff
@@ -51,7 +51,7 @@ final class JUnit extends Printer implements TestListener
     /**
      * @var bool
      */
-    private $reportRiskyTests = false;
+    private $reportRiskyTests;
```
Ayesh, line 54: This value is immediately set in the constructor, but I think it's OK to keep the default property value here for clarity.

Ayesh (approved):

sebastianbergmann:

PR #4444, opened 2020-08-31
some properties in TestCase may be null

fullbl:
Psalm warned me about `ERROR: PropertyNotSetInConstructor - tests/XxxTest.php:14:13 - Property Xxx\Tests\XxxTest::$backupStaticAttributes is not defined in constructor of Xxx\Tests\XxxTest and in any methods called in the constructor (see https://psalm.dev/074)
`.

Same issue with runClassInSeparateProcess and runTestInSeparateProcess. Notice that inside the class they are often compared against null, so I changed the annotation so they allow null, instead of setting a default!

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -167,7 +167,7 @@ abstract class TestCase extends Assert implements Reorderable, SelfDescribing, T
     protected $providedTests = [];
 
     /**
-     * @var bool
+     * @var ?bool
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it as the properties in question have been renamed and marked private.

PR #4447, opened 2020-09-07
Restore spy functionality

afilina:
Although I understand the reluctance to let people rely on an internal class, many people have come to rely on this functionality already.

- I can't upgrade test suites for some of my clients because the feature is completely gone.
- I have an online course where people keep complaining that spies no longer work in PHPUnit.
- I see numerous comments about this being needed: on this repo, on blog posts, etc.

It's a really easy win with nothing but a getter. The underlying `Invocation` class is a value object, so there's no risk in exposing it.

### DIFF
--- a/src/Framework/MockObject/Rule/InvocationOrder.php
+++ b/src/Framework/MockObject/Rule/InvocationOrder.php
@@ -29,6 +29,15 @@ abstract class InvocationOrder implements SelfDescribing, Verifiable
         return count($this->invocations);
     }
 
+    /**
+     * @return BaseInvocation[]
+     * @psalm-suppress PossiblyUnusedMethod
+     */
+    public function getInvocations(): array
+    {
+        return $this->invocations;
+    }
+
     public function hasBeenInvoked(): bool
     {
         return count($this->invocations) > 0;
--- a/tests/unit/Framework/MockObject/Matcher/SpyTest.php
+++ b/tests/unit/Framework/MockObject/Matcher/SpyTest.php
@@ -1,0 +1,36 @@
+<?php declare(strict_types=1);
+/*
+ * This file is part of PHPUnit.
+ *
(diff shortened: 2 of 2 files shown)


sebastianbergmann:
> It's a really easy win with nothing but a getter. The underlying `Invocation` class is a value object, so there's no risk in exposing it.

Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it as I do not believe that introducing the `getInvocations()` method you propose is a good solution. It would be an `@internal` method that returns a collection of objects of classes that are `@internal`. We cannot, and will not, give a backward compatibility promise for this code.

afilina:
It doesn't _have_ to be internal. In plenty of testing framework, these invocations are accessible. That's pretty much how spies work, unless you want to force every user to write custom classes for something that can come out-of-the-box at no cost. This makes me really sad, as now I can't introduce spies in my next course on PHPUnit without adding unnecessary complexity.

afilina:
Some more places where the community documented this feature:
https://stackoverflow.com/a/30067362/544894
https://lyte.id.au/2014/03/01/spying-with-phpunit/
https://matthiasnoback.nl/2014/07/test-doubles/
https://www.youtube.com/watch?v=_LeCyU1X5bY&lc=UgxmEiMMr2sbzAsFQtF4AaABAg&ab_channel=AnnaFilina

It's worth noting that the object returned by `$this->any()` was not always internal. Granted, it was on a different major version, but it used to be an actual feature, so people used it.

It's hard as it is to get people to write tests and upgrade their libraries. Why make it any harder?

rulatir:
What alternatives do we have? How can an assertion about arguments passed to an invocation of a stubbed method be currently achieved with a one-liner, rather than by writing a separate class residing in a separate file?

PR #4448, opened 2020-09-07
Allow DocBlock class to be called with built-in classes

derrabus:
When being called on built-in classes like `DateTime` or `ReflectionClass`, the `DocBlock` parser fails with obscure type errors. This happens because the reflection methods `getStartLine()`, `getEndLine()` and `getFileName()` return `false` for those classes.

Since there are no doc blocks to parse on those classes, I'd expect the parser to behave the same way as for a userland class without any doc blocks. This is what I've tried to achieve with this PR.

### DIFF
--- a/src/Util/Annotation/DocBlock.php
+++ b/src/Util/Annotation/DocBlock.php
@@ -102,13 +102,13 @@ final class DocBlock
      */
     private $parsedRequirements;
 
-    /** @var int */
+    /** @var null|int */
     private $startLine;
 
-    /** @var int */
+    /** @var null|int */
     private $endLine;
 
-    /** @var string */
+    /** @var null|string */
     private $fileName;
 
     /** @var string */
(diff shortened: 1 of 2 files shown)


sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it because the code in questions does not need support for built-in classes.

The `DocBlock` class is `@internal`, not covered by the backward compatibility promise for PHPUnit, and should not be used by code outside of PHPUnit itself. Its methods such as `ofClass()` and `ofMethod()` are not meant to be called for built-in classes. If they are, then that would be the bug I would like to see fixed. Please open a new ticket if you can show that this code is called for built-in classes.

derrabus:
> should not be used by code outside of PHPUnit itself

I agree, but PHPUnit uses that class internally quite often without prior check for built-in classes.

In my case, the DeprecationErrorHandler from Symfony's PhpUnitBridge tries to find out if a triggered error originates from a test case of a certain group. It uses the `PHPUnit\Util\Test` class for this purpose, which passes the given class to `DocBlock`. To be fair, `PHPUnit\Util\Test` is also flagged as internal, but there isn't really an official API for the kind of information that we need, is there?

Anyway, I think I can catch this case and prevent the call. Thanks for the quick heads-up.

PR #4451, opened 2020-09-11
Fix generating signatures with nullable unions

nicolas-grekas:
Found while debugging the failure on https://github.com/composer/composer/runs/1101133265

I'd happily accept help to finish the PR (add tests especially).

### DIFF
--- a/src/Framework/MockObject/MockMethod.php
+++ b/src/Framework/MockObject/MockMethod.php
@@ -292,7 +292,7 @@ final class MockMethod
             }
 
             if ($type !== null) {
-                if ($typeName !== 'mixed' && $parameter->allowsNull()) {
+                if ($typeName !== 'mixed' && $parameter->allowsNull() && !$type instanceof ReflectionUnionType) {
                     $nullable = '?';
                 }
 


sebastianbergmann:
Can you provide a **minimal**, **self-contained**, reproducing example that shows the problem you want to solve? Thanks! I will then use your example to create a test case for PHPUnit's test double code generator.

Seldaek:
Essentially I think what triggers it, is mocking the ZipArchive class, namely the extractTo method.

I believe this is the code triggering the issue in our test suite:

```php
        $zipArchive = $this->getMockBuilder('ZipArchive')->getMock();
        $zipArchive->expects($this->any())
            ->method('open')
            ->will($this->returnValue(true));
        $zipArchive->expects($this->any())
            ->method('extractTo')
            ->will($this->returnValue(false));
```

@nicolas-grekas said it was generating a mock with signature `public function extractTo(string $pathto, ?array|string|null $files = NULL)`, and the `?` is superfluous here.

sebastianbergmann:
Thanks, @Seldaek.

sebastianbergmann:
Merged manually, thanks.

Seldaek:
Thanks for the quick merge & release. I can confirm it's working well now. Down to 1 actual failure and then we're green on php 8 :)

sebastianbergmann:
Glad I could help.

ste93cry:
@sebastianbergmann would you be favourable to release this fix for version `8.5` as well? In `getsentry/sentry-symfony` we're still using that version in the build of the lowest deps due to the need of supporting PHP `7.2` and when installing it alongside with Symfony `6.x` our tests fail 😢

sebastianbergmann:
PHPUnit 8.5 must work for PHP code compatible with PHP 7.2 on PHP 8. Are you saying that this is not the case? I do have to say, though, that I am confused why "installing it alongside with Symfony" should have an effect on how PHPUnit behaves. PHPUnit does not depend on any Symfony component. Or is Symfony's PHPUnit bridge causing your problem?

nicolas-grekas:
When running "old" code base (like phpunit 8.5 in this case), one should expect the some deprecations are triggered. Either by the bridge, or the latest php, etc
Check the doc of the bridge to silence some deprecations @ste93cry

nicolas-grekas:
( if that is related to the bridge because I also don't get the link with Symfony yet )

Seldaek:
I did experience the same issue (I think) trying to make Composer tests pass with PHPUnit 8.5 on PHP 7.2 to 8.1 in https://github.com/composer/composer/pull/10343

The problem I saw was that some Symfony 6 code broke PHPUnit MockObject when running the PHP 8.x build (with PHPUnit 8.5). I did see a backport PR which was closed by @sebastianbergmann so I gave up on that project and went back to using symfony/phpunit-bridge to run multiple PHPUnit versions based on PHP version, which is working well now.

ste93cry:
>When running "old" code base (like phpunit 8.5 in this case), one should expect that some deprecations are triggered. Either by the bridge, or the latest php, etc

Sorry, I think that my message has been misinterpreted. Sentry needs to support PHP `7.2` as minimum version, which forces me to use PHPUnit `8.5` for the build that tests against the lowest version of the deps. Since this version of PHPUnit can be installed on PHP 8 and Symfony 6 supports only PHP 8, this combination downloads the Symfony Console component where the `InputInterface` interface uses a `nullable` union type as [argument](https://github.com/symfony/symfony/blob/80f4e7628511fafb38c5b2e4fbd85e264c247793/src/Symfony/Component/Console/Input/InputInterface.php#L56) of one of its methods. When trying to mock such interface, I get a fatal error because the typehint gets transformed to `?$default = false` which is an invalid syntax

PR #4457, opened 2020-09-13
Update warning when using expectExceptionMessageRegExp()

Nyholm:
This will show that you should use `expectExceptionMessageMatches` instead.


Nyholm:
Thank you for merging

PR #4472, opened 2020-09-27
Implement --until-success CLI option

tindarid:
`phpunit` has option `--repeat`, but sometimes we want something called `attempt` for unstable tests (tests run in parallel and sometimes fail, because of other tests)

`--until-success` option uses repeated testsuites provided by `--repeat` option and if testsuite successfully completed, doesnt' run another suites

### DIFF
--- a/src/Framework/TestResult.php
+++ b/src/Framework/TestResult.php
@@ -1206,6 +1206,14 @@ final class TestResult implements Countable
     }
 
     /**
+     * Returns number of defects.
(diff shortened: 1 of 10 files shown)


Tindarid:
I fixed some bugs, but idk what to do with exit code. I want to get 0 (not 1), if there were failures in previous `attempts`, but not in the last

sebastianbergmann:

PR #4474, opened 2020-09-28
New inheritance constraints and assertions

ptomulik:
I'd like to propose some new constraints and assertions for inheritance. They're inspired on ``isInstanceOf`` but are more specific. The new constraints are:

  - ``ImplementsInterface`` (whether the given class implements specific interface),
  - ``ExtendsClass`` (whether the given class extends specific class),
  - ``UsesTrait`` (whether the given class/trait uses specific trait) ,

and the accompanying assertions:

  - ``assertImplementsInterface``, ``assertNotImplementsInterface``,
  - ``assertExtendsClass``, ``assertNotExtendsClass``,
  - ``assertUsesTrait``, ``assertNotUsesTrait``.

I have also prepared some ``rst`` user docs for these functions (English), but not sure where to put them (shall they go to [phpunit-documentation-english](https://github.com/sebastianbergmann/phpunit-documentation-english)?).

### DIFF
--- a/src/Framework/Assert.php
+++ b/src/Framework/Assert.php
@@ -44,8 +44,10 @@ use PHPUnit\Framework\Constraint\ClassHasStaticAttribute;
 use PHPUnit\Framework\Constraint\Constraint;
 use PHPUnit\Framework\Constraint\Count;
 use PHPUnit\Framework\Constraint\DirectoryExists;
+use PHPUnit\Framework\Constraint\ExtendsClass;
 use PHPUnit\Framework\Constraint\FileExists;
 use PHPUnit\Framework\Constraint\GreaterThan;
+use PHPUnit\Framework\Constraint\ImplementsInterface;
 use PHPUnit\Framework\Constraint\IsAnything;
 use PHPUnit\Framework\Constraint\IsEmpty;
 use PHPUnit\Framework\Constraint\IsEqual;
@@ -81,6 +83,7 @@ use PHPUnit\Framework\Constraint\StringStartsWith;
 use PHPUnit\Framework\Constraint\TraversableContainsEqual;
 use PHPUnit\Framework\Constraint\TraversableContainsIdentical;
 use PHPUnit\Framework\Constraint\TraversableContainsOnly;
+use PHPUnit\Framework\Constraint\UsesTrait;
 use PHPUnit\Util\Type;
 use PHPUnit\Util\Xml;
 use PHPUnit\Util\Xml\Loader as XmlLoader;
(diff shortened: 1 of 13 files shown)


sebastianbergmann:
May I ask what your use case is for these assertions and constraints? I have to say that I would much rather deprecate and later remove the existing `assertClassHasAttribute()` etc. than to add more assertions like it.

ptomulik:
They were  created as part of one of my projects, but at some point I realized they're "quite generic" so maybe could land here. Example use cases:

- ``assertImplementsInterface(ContainerInterface::class, MyContainer::class)`` - ``MyContainer`` is quite good candidate to be [PSR-11](https://www.php-fig.org/psr/psr-11/) compliant,
- ``assertExtendsClass(AbstractWellTestedClass::class, MyClass::class)`` - if someone else tested ``AbstractWellTestedClass``, and I don't want repeat tests for inherited public functions,
- ``assertUsesTrait(WellTestedTrait::class, MyClass::class)`` - same as with abstract class, but this time with trait.

Yes, they can be considered as little value when the value is measured with generated code coverage. ;)

ptomulik:
> [...] I have to say that I would much rather deprecate and later remove the existing `assertClassHasAttribute()` etc. than to add more assertions like it.
So I ask one more question, because I also have in queue constraints/assertions for introspecting object properties/class static properties, like ``{object,class}PropertiesIdenticalTo(array $expected)``, ``{object,class}PropertiesEqualTo(array $expected)``. Do you think such assertions could be useful for PHPUnit? They're already coded, so migrating them to phpunit is a matter of few hours I guess (I'd be glad to make PR and share this code here).

sebastianbergmann:
> 
> 
> > [...] I have to say that I would much rather deprecate and later remove the existing `assertClassHasAttribute()` etc. than to add more assertions like it.
> 
> So I ask one more question, because I also have in queue constraints/assertions for introspecting object properties/class static properties, like `{object,class}PropertiesIdenticalTo(array $expected)`, `{object,class}PropertiesEqualTo(array $expected)`. Do you think such assertions could be useful for PHPUnit? They're already coded, so migrating them to phpunit is a matter of few hours I guess (I'd be glad to make PR and share this code here).

No, sorry. We had these in the past and they were recently removed.

ptomulik:
> > > [...] I have to say that I would much rather deprecate and later remove the existing `assertClassHasAttribute()` etc. than to add more assertions like it.
> > 
> > 
> > [...] like `{object,class}PropertiesIdenticalTo(array $expected)`, `{object,class}PropertiesEqualTo(array $expected)`.[...]
> 
> No, sorry. We had these in the past and they were recently removed.

Good to know. Is there any archived documentation from previous versions for these removed assertions? I'm curious I was reinvented wheel during last days. :)

sebastianbergmann:
> Yes, they can be considered as little value when the value is measured with generated code coverage. ;)

Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it because I believe that these assertions are not necessary. They assert what should be checked using a type checker such as Psalm or by PHP's type system. The exception here is `assertUsesTrait()`, of which there should be notion at runtime:

> There is no notion about Traits at runtime. They are used to group methods and reuse code and are totally flattened into the classes composed from them. It is almost like a language supported and failsafe copy'n'paste mechanism to build classes. Even though, there is no runtime notion of Traits, since they are part of the source code and thus, define the structure of the system, reflection about Traits still is possible, but they do not influence the runtime behavior of the system.

(https://wiki.php.net/rfc/horizontalreuse)

sebastianbergmann:
> Good to know. Is there any archived documentation from previous versions for these removed assertions? I'm curious I was reinventing wheel during last days. :)

Have a look at `assertAttributeEquals()`, for instance on https://phpunit.readthedocs.io/en/7.0/assertions.html#assertequals

ptomulik:
> > Good to know. Is there any archived documentation from previous versions for these removed assertions? I'm curious I was reinventing wheel during last days. :)
> 
> Have a look at `assertAttributeEquals()`, for instance on https://phpunit.readthedocs.io/en/7.0/assertions.html#assertequals

Oh I see. Mine are little bit different:

- they introspect multiple attributes at once (array of expectations is used),
- can be used recursively (check attributes of nested object being an attribute of top-level object),
- can examine values of attributes as well as values returned by accessor methods.

Issue #4368, opened 2020-07-08
Assertion assertNotEquals Provides Wrong Error Message On Failure


markrogoyski:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 8.5.8, 9.3-g33d19b2aa, 
| PHP version         | 7.3.3, 7.4.7
| Installation Method | Composer

#### Summary

The `assertNotEquals` test assertion will produce an incorrect error message on failure when asserting a string is not equal to another string in the following cases:
* String has double quotes in it
* String has "positive" words in it

In this case, failure of `assertNotEquals` means that the strings are equal.
"Positive" words are those that `PHPUnit\Framework\Constraint\LogicalNot::negate` will translate to "negative" words.

#### Current behavior

In the case that the string has double quotes in it, the error message will say: _Failed asserting that '$STRING' **is equal to** '$STRING'._, when the expected error message would be _Failed asserting that '$STRING' is **not** equal to '$STRING'._

In the case that the string has "positive" words in it, you get nonsense error messages where the string is "negated."

#### How to reproduce

Here is a basic test case that works as expected.
```php
class NotEqualsTest extends \PHPUnit\Framework\TestCase
{
    public function testNotEqualString(): void
    {
        // Given
        $string = 'a b c';

        // Then
        $this->assertNotEquals($string, $string);
    }
}
```
And it produces the expected error message "is not equal":
```
Failed asserting that 'a b c' is not equal to 'a b c'.
```

However, now we do the same simple test case but with double quotes appearing in it:
```php
class NotEqualsTest extends \PHPUnit\Framework\TestCase
{

    public function testNotEqualStringWithDoubleQuotes(): void
    {
        // Given
        $string = 'a "b" c';

        // Then
        $this->assertNotEquals($string, $string);
    }
}
```
And it produces an unexpected error message that it failed to assert that it "is equal":
```
Failed asserting that 'a "b" c' is equal to 'a "b" c'.
```

Here is another example with a JSON string.
```php
class NotEqualsTest extends \PHPUnit\Framework\TestCase
{
    public function testNotEqualJsonString(): void
    {
        // Given
        $string = '{"key":"value"}';

        // Then
        $this->assertNotEquals($string, $string);
    }
}
```
And the unexpected error message:
```
Failed asserting that '{"key":"value"}' is equal to '{"key":"value"}'.
```

Things get really strange when the string has "positive" words in it. They all become "negated" in the error message. Here is an example test case:
```php
[...]

PR #4479, opened 2020-10-03
Fix for #4368: Add tests for LogicalNot::negate, fixes to negate regex

kamioftea:
This is a partial fix for the issue raised in #4368 that

- Adds the tests suggested by @markrogoyski 
- Fixes the existing regex to match single quotes for both strings
- Allows the match of the first single quote enclosed strings to be optional, so this also works for the __toString() use case

The above fixes to the regex allow the suggested tests to pass, however test strings with single quotes in them still show incorrect behaviour, i.e. the following test still fails:

```php
public function testConstraintIsNotEqualStringContainsSingleQuotes(): void
{
    $string     = "a 'b' c";
    $other      = "a ''b'' c";
    $constraint = Assert::logicalNot(
        Assert::equalTo($string)
    );

    $this->assertTrue($constraint->evaluate($other, '', true));
    $this->assertFalse($constraint->evaluate($string, '', true));
    $this->assertEquals("is not equal to '{$string}'", $constraint->toString());
    $this->assertCount(1, $constraint);

    try {
        $constraint->evaluate($string);
    } catch (ExpectationFailedException $e) {
        $this->assertEquals(
            <<<EOF
Failed asserting that '{$string}' is not equal to '{$string}'.

EOF
            ,
            TestFailure::exceptionToString($e)    // Fails here
        );

        return;
    }

    $this->fail();
}
```


sebastianbergmann:

PR #4487, opened 2020-10-23
Changed order on filter, can give little speed improve

proggga:
Small speed improve,
when you got broken code and many exceptions, speed is really important
I've sorted checks in this change from fastest-easiest to slowest


proggga:
I was late

PR #4488, opened 2020-10-23
Changed order of checks to make this function a little bit faster

proggga:
Changed order of filter to a little speedup this code, is_file should be last, access to disk is slow by definition,
and we have few checks before, that can give as speed, especially in case when we have many errors and alot of traces


proggga:
Please add hacktoberfest-accepted label

PR #4494, opened 2020-10-26
Fix: Do not use deprecated set-env

localheinz:
This PR

* [x] stops using the deprecated `set-env`

💁‍♂️  For reference, see https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -120,7 +120,7 @@ jobs:
 
       - name: Override PHP ini values for JIT compiler
         if: matrix.compiler == 'jit'
-        run: echo "::set-env name=PHP_INI_VALUES::assert.exception=1, zend.assertions=1, opcache.enable=1, opcache.enable_cli=1, opcache.optimization_level=-1, opcache.jit=1255, opcache.jit_buffer_size=32M"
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -132,11 +132,11 @@ jobs:
 
       - name: Determine composer cache directory on Linux
         if: matrix.os == 'ubuntu-latest'
-        run: echo "::set-env name=COMPOSER_CACHE_DIR::$(./tools/composer config cache-dir)"
+        run: echo "COMPOSER_CACHE_DIR=$(./tools/composer config cache-dir)" >> $GITHUB_ENV
 
       - name: Determine composer cache directory on Windows
         if: matrix.os == 'windows-latest'
-        run: ECHO "::set-env name=COMPOSER_CACHE_DIR::~\AppData\Local\Composer"
+        run: ECHO "COMPOSER_CACHE_DIR=~\AppData\Local\Composer" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
```
localheinz, line 139: Not sure about this, let's see if it passes.

See https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#environment-files:

CleanShot 2020-10-26 at 17 01 42

localheinz:

PR #4499, opened 2020-10-28
allow for case insensitive class names #4498

### LINKED ISSUE #4498: Please make the class matching case insenstive
Recent versions of PHP Unit have added the requirement that the filename for a test be a prefix of the classname for the test.  The current requirement does a case sensitive compare.  My current preference is for mixed case classnames and all lower case filenames (though we do use mytestTest.php to comply with previous phpunit conventions).  It's not the end of the world to have to rename all of the tests, but it's a pain.  If there isn't a compelling reason to have a case sensitive compare, it will break fewer existing test setups to do it case insensitive.

For instance
some/path/mytestTest.php

should allow
```
<?php
class some_path_myTestTest extends PHPUnit\Framework\TestCase
{
    public function testTrue(): void
    {
        $this->assertTrue(true);
    }
}
```

realFlowControl:
Hey there :vulcan_salute:

this PR fixes #4498 by adding support for case insensitive namespaced class names when calling PHPUnit with the test case file name.

/Flo

### DIFF
--- a/src/Runner/StandardTestSuiteLoader.php
+++ b/src/Runner/StandardTestSuiteLoader.php
@@ -15,7 +15,7 @@ use function basename;
 use function class_exists;
 use function get_declared_classes;
 use function sprintf;
-use function str_replace;
+use function stripos;
 use function strlen;
 use function substr;
 use PHPUnit\Framework\TestCase;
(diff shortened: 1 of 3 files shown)


sebastianbergmann:
Can you send this against `9.4`, please? Thanks!

flow-control:
@sebastianbergmann rebased onto `9.4` and changed the base branch

flow-control:
IDK why codecov is failing :disappointed:

PR #4504, opened 2020-10-30
Enhancement: Use static::class instead of get_class($this)

localheinz:
This PR

* [x] uses `static::class` instead of `get_class($this)`

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -836,7 +836,7 @@ abstract class TestCase extends Assert implements SelfDescribing, Test
     public function getAnnotations(): array
     {
         return TestUtil::parseTestMethodAnnotations(
-            \get_class($this),
+            static::class,
             $this->name
         );
     }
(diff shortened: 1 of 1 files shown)


src/Framework/TestCase.php
```diff
@@ -1999,7 +1999,6 @@ private function verifyMockObjects(): void
     private function handleDependencies(): bool
     {
         if (!empty($this->dependencies) && !$this->inIsolation) {
-            $className  = \get_class($this);
```
localheinz, line 2002: Is it ok to inline this variable, @sebastianbergmann?

src/Framework/TestCase.php
```diff
@@ -836,14 +836,14 @@ public function setGroups(array $groups): void
...
-            \get_class($this),
+            static::class,
             $this->name
         );
     }
 
     /**
      * @throws \SebastianBergmann\RecursionContext\InvalidArgumentException
-     *
+     *handleDependencieshandleDependencies
```
sebastianbergmann, line 846: This looks wrong.

src/Framework/TestCase.php
sebastianbergmann: Why not?

localheinz: 👍

localheinz:

Issue #4091, opened 2020-02-12
Deprecate assertEqualXMLStructure()


sebastianbergmann:
I have never understood the use case for this assertion and have never seen it used in the wild.

povetina:
That actually doesn't mean people don't use it. 
For example, I maintain a project which among other things exports XML data to remote server. 
To check that my code generates correct XML file I use precisely assertEqualXMLStructure() function. 
Also, it doesn't strike me as the best practice to remove the function completely without providing a new one.

LlZARD:
Also using it.  Now have to work it around. 😔

tshafer:
I am using it as well.

theseer:
I'm not sure whether or not this very method or its implementation are the best way to solve the use case it at least for me did cover.

But the general use case is actually rather simple, given that the definition of "equal" with XML documents isn't exactly clear:

```xml
<?xml version="1.0" ?>
<root>
    <node />
</root>
```

### ./.


```xml
<?xml version="1.0" ?>
<root><node /></root>
```

Are these equal?

If you parse an XML document into a DOM and compare the resulting object graph, they won't be because of additional DOMText-Elements representing the whitespaces for the indentation and the line breaks in the first XML.

In many (most?) contexts though, the whitespace is actually pretty irrelevant. So how would one assert that two structures are "otherwise" equal?

Maybe the method name doesn't (didn't) really transport that intention, but that's what at least I have been using it for.

theseer:
Another example:

```xml
<?xml version="1.0" ?>
<root a="true" b="true" />
```

### ./.

```xml
<?xml version="1.0" ?>
<root b="true" a="true" />
```

Logically (semantically?), the two XML structures are equal. On a string and DOM-Basis, they are not.

theseer:
Talked with @sebastianbergmann, and we decided that I'll re-implement this functionality.

A PR will be provided for it hopefully soon, not sure yet as to when exactly I'll have time to do though.

theseer:
The PR mentioned re-implements the main aspect of the original, deprecated assertion `assertEqualXMLStructure`: The comparison of two XML trees for structural identicality.

Two DOM trees are to be considered identical when, after canonicalization, the following holds true:

    The same child - parent relationship is maintained
    The same attributes are set on a node
    The attribute values are identical
    The textual content of nodes are identical

[...]

PR #4507, opened 2020-11-01
Reimplement main aspect of assertEqualXMLStructure

theseer:
As discussed in https://github.com/sebastianbergmann/phpunit/issues/4091, there is a valid use case for having an assertion ensuring two DOM sub trees are structurally identical.

This PR re-implements the main aspect of the original, deprecated `assertEqualXMLStructure`: The comparison of two XML trees for structural identicality. 

Two DOM trees are to be considered identical when, after canonicalization, the following holds true:

- The same child - parent relationship is maintained
- The same attributes are set on a node
- The attribute values are identical
- The textual content of nodes are identical

Looking at the options of the original, deprecated method, this appears to be stricter. The original method for instance ignored namespace defining attributes as well as textual content.

As I didn't see a use case for that, I did not reimplement that aspect. My implementation basically ignores any whitespace and, due to canonicalization, reorders attributes as needed so the order of attributes is not relevant.

The name `assertDOMTreesEqualStructurally` is subject to change, only pending a better naming suggestion. It should probably reflect the actual assertion better.

### DIFF
--- a/src/Framework/Assert.php
+++ b/src/Framework/Assert.php
@@ -2253,14 +2253,20 @@ abstract class Assert
 
     public static function assertDOMTreesEqualStructurally(DOMElement $expectedElement, DOMElement $actualElement, bool $keepComments = false, string $message = ''): void
     {
-        $ed                     = new DOMDocument();
+        $ed = new DOMDocument();
+        $ed->appendChild($ed->importNode($expectedElement, true));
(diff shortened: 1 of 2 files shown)


newblaque (approved):

sebastianbergmann:
This branch has conflicts that must be resolved.

WinterSilence:
you use troll's solutions..

sebastianbergmann:

PR #4517, opened 2020-11-13
Fix: Evaluate condition before looping

localheinz:
This PR

* [x] evaluates conditions before looping

### DIFF
--- a/src/Framework/TestSuite.php
+++ b/src/Framework/TestSuite.php
@@ -602,56 +602,56 @@ class TestSuite implements IteratorAggregate, Reorderable, SelfDescribing, Test
 
         $test = null;
 
-        try {
-            foreach ($hookMethods['beforeClass'] as $beforeClassMethod) {
-                if ($this->testCase &&
-                    class_exists($this->name, false) &&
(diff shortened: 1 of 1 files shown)


src/Framework/TestSuite.php
```diff
@@ -602,56 +602,56 @@ public function run(TestResult $result = null): TestResult
...
 
-        try {
-            foreach ($hookMethods['beforeClass'] as $beforeClassMethod) {
-                if ($this->testCase &&
-                    class_exists($this->name, false) &&
-                    method_exists($this->name, $beforeClassMethod)) {
-                    if ($missingRequirements = TestUtil::getMissingRequirements($this->name, $beforeClassMethod)) {
-                        $this->markTestSuiteSkipped(implode(PHP_EOL, $missingRequirements));
-                    }
+        if ($this->testCase && class_exists($this->name, false)) {
```
localheinz, line 605: Unless I'm missing something, there does not appear to be any code executed within the loops that could change the outcome of this condition, or is there?

sebastianbergmann:
Unless this fixes a bug, please make this against `master`. Thanks!

localheinz:

PR #4518, opened 2020-11-13
Enhancement: Extract methods

localheinz:
This PR

* [x] extracts methods for recording test failures


src/Framework/TestResult.php
```diff
@@ -1254,4 +1261,29 @@ public function setRegisterMockObjectsFromTestArgumentsRecursively(bool $flag):
     {
         $this->registerMockObjectsFromTestArgumentsRecursively = $flag;
     }
+
+    private function recordError(Test $test, Throwable $t): void
```
localheinz, line 1265: These methods could be useful for @theseer and I, so we can emit events here (instead of in two places).

What do you think?

localheinz:

PR #4549, opened 2020-12-18
Enhancement: Enable static_lambda fixer

localheinz:
This PR

* [x] enables the `static_lambda` fixer
* [x] runs `tools/php-cs-fixer fix`

💁‍♂️ For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.17.1/doc/rules/function_notation/static_lambda.rst.


localheinz:

PR #4552, opened 2020-12-18
Fix: Enable echo_tag_syntax instead of deprecated no_short_echo_tag fixer

localheinz:
This PR

* [x] enables the `echo_tag_syntax` fixer instead of the deprecated `no_short_echo_tag` fixer

💁‍♂️ For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.17.1/doc/rules/php_tag/no_short_echo_tag.rst.


localheinz:

PR #4561, opened 2020-12-25
Fix: Use on_multiline instead of deprecated ensure_fully_multiline option for method_argument_space fixer

localheinz:
This PR

* [x] uses the `on_multiline` option instead of the deprecated `ensure_fully_multiline` option for the `method_argument_space` fixer

💁‍♂️ For reference, see 

- https://github.com/ergebnis/php-cs-fixer-config/pull/319
- https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.17.3/doc/rules/function_notation/method_argument_space.rst

CleanShot 2020-12-25 at 23 42 31


localheinz:

PR #4569, opened 2020-12-29
Fix: Remove unused private method

localheinz:
This PR

* [x] removes an unused `private` method


localheinz:

Issue #4566, opened 2020-12-28
Remove code for deprecated functiontionality from TestSuite


sebastianbergmann:
In `TestSuite.php` we have this code:

```php
if (count($this->tests) > ++$numTests) {
    $this->addWarning(
        sprintf(
            "Multiple test case classes per file is deprecated\n               in %s",
            $filename
        )
    );
}
```

It is not clear to me from reading the method's code where this fragment is from what code needs to be removed.

sebastianbergmann:
@flow-control Can you help me out here? Thanks!

flow-control:
@sebastianbergmann I can have a look and send a PR remove the functionality

sebastianbergmann:
Thank you!

flow-control:
Hey @sebastianbergmann,

I had a look and found out the following. For the things we deprecated already (multiple test classes per file, test case class not matching file name) I gave deprecation warnings only in `Runner/TestSuiteLoader::load()`, not in `Framework/TestSuite::addTestFile()`, but both functions basically do the same thing. I would like to prepare a PR to remove this duplication and use the `TestSuiteLoader` from the `TestSuite::addTestFile()` method if this okay for you.
Only question left for me is: It would be the polite thing to give a deprecation warning about this also in the the `TestSuite::addTestFile()` method, should I provide an additional PR to PHPUnit 9 for this deprecation warning?
If you like to, we could also have a quick chat about this.

/Florian

sebastianbergmann:
> I had a look and found out the following. For the things we deprecated already (multiple test classes per file, test case class not matching file name) I gave deprecation warnings only in `Runner/TestSuiteLoader::load()`, not in `Framework/TestSuite::addTestFile()`, but both functions basically do the same thing. I would like to prepare a PR to remove this duplication and use the `TestSuiteLoader` from the `TestSuite::addTestFile()` method if this okay for you.

Of course I am okay with that.

> Only question left for me is: It would be the polite thing to give a deprecation warning about this also in the the `TestSuite::addTestFile()` method, should I provide an additional PR to PHPUnit 9 for this deprecation warning?
> If you like to, we could also have a quick chat about this.

Yes, I think we should add that deprecation warning for PHPUnit 9.

Thanks!

flow-control:
[...]

PR #4570, opened 2020-12-29
Remove deprecated test suite code #4566

realFlowControl:
Hey @sebastianbergmann,

this PR will remove the deprecated multiple classes per single test file from the `Framework/TestSuite` class and relates to #4566. There are two things I would like to put your attention on:

#### Tests are being executed multiple times, when referenced multiple times in `phpunit.xml`

You can see this behaviour in `./phpunit tests/end-to-end/dataprovider-issue-2859.phpt` which uses a `phpunit.xml` file like this:
```xml

<phpunit>
    <testsuites>
        <testsuite name="Test Suite">
            <directory>./tests/</directory>
            <directory>./tests/*/</directory>
        </testsuite>
    </testsuites>
</phpunit>
```

There is exactly one PHP test file in the `tests` directory, but due to this configuration, this test is being execute two times with this PR applied. I personally would say, that it is to be expected, but I could also understand if you would like this to be "fixed" and the test only be executed one time.

#### Test case class name matching file name with suffix including a `.`

If you use a `.` (or basically any other char in the suffix that is not allowed to be used in a PHP class name), what should the resulting class name be?
There is one case in the tests to showcase this. Test file is  `tests/_files/BankAccountTest.test.php` when you execute this with `./phpunit --test-suffix ".test.php" tests/_files/` without the "fix" in the `BankAccountTest.test.php` file, phpunit fails as it looks for a test class called `BankAccountTest.test` which can not be found (and is invalid). So I changed the creation if the class name from file name to search for the first `.` char and just use this as the class name. In this case the class name is expected to be `BankAccountTest`.

I hope these things are okay with you, if not let`s talk :smile: 

There might be still some things in this PR, that`s why this is just a draft for the moment, will complete this in the next days.

Kind regards
Florian


sebastianbergmann:
> this PR will remove the deprecated multiple classes per single test file from the `Framework/TestSuite` class and relates to #4566.

Thank you so much!

> I personally would say, that it is to be expected, but I could also understand if you would like this to be "fixed" and the test only be executed one time.

I would say so, too. Let us keep it this way. I imagine any solution to this would increase complexity and with questionable gain.

> If you use a `.` (or basically any other char in the suffix that is not allowed to be used in a PHP class name), what should the resulting class name be?
> There is one case in the tests to showcase this. Test file is `tests/_files/BankAccountTest.test.php` when you execute this with `./phpunit --test-suffix ".test.php" tests/_files/` without the "fix" in the `BankAccountTest.test.php` file, phpunit fails as it looks for a test class called `BankAccountTest.test` which can not be found (and is invalid). So I changed the creation if the class name from file name to search for the first `.` char and just use this as the class name. In this case the class name is expected to be `BankAccountTest`.

Makes sense to me.

flow-control:
Only thing left from my point of view is that I need to wrap my head around Codecov. Honestly I do not understand how to read this and get the information of what lines miss code coverage now that where covered earlier ...

sebastianbergmann:
Thank you!

PR #4579, opened 2021-01-11
Enhancement: Add error context on failing hook

ruudvanderweijde:
When a test fails because of a Throwable the error was not helping to
locate the problem.


ruudvanderweijde:
@sebastianbergmann any thoughts on the suggested change?

sebastianbergmann:

PR #4585, opened 2021-01-15
Fix: Use short array syntax

localheinz:
This PR

* [x] uses short array syntax


sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it because this file is supposed to still work with PHP 5.3 (maybe older).

PR #4589, opened 2021-01-23
Fix: Create exporter only when needed

localheinz:
This pull request

* [x] creates an `Exporter` only when it is needed


localheinz:

PR #4593, opened 2021-01-26
Documented `callable` parameter structure for `Assert::callback()`

Ocramius:
`Assert::callback()` accepts a `$callable` that, **AT MOST**, accepts **1** parameter and
returns `bool`.

The number of times I've written `->with(self::callback(function ($a, $b, $c) : void {`
is very high:

 * I generally forget that `->with()` is variadic (that's OK)
 * I forget that `self::callback()` applies to a single parameter
 * I forget that the `callable` must return a `bool` in order to operate

In order to avoid these mistakes, the `Callback` constraint is now templated, and the
given `callable` is enforced to require at most one parameter, and to always return
`bool`.

As a counter-example to verify that this type signature works, I've run this against
`tests/static-analysis/TestUsingCallbacks.php` with following snippet:

```php
    public function testInvalid(): void
    {
        $mock = $this->createMock(SayHello::class);

        $mock
            ->expects(self::once())
            ->method('hey')
            ->with(self::callback(static function (string $a, string $b): bool {
                return true;
            }))
            ->willReturn('Hey Joe!');

        self::assertSame('Hey Joe!', $mock->hey('Joe'));
    }
```

The above now raises:

```
ERROR: InvalidArgument - tests/static-analysis/TestUsingCallbacks.php:62:35 - Argument 1 of PHPUnit\StaticAnalysis\TestUsingCallbacks::callback expects callable(mixed):bool, pure-Closure(string, string):true provided (see https://psalm.dev/004)
            ->with(self::callback(static function (string $a, string $b): bool {
                return true;
            }))
```

This produces

### DIFF
--- a/src/Framework/Assert.php
+++ b/src/Framework/Assert.php
@@ -2225,6 +2225,13 @@ abstract class Assert
         return new IsTrue;
     }
 
+    /**
+     * @psalm-template CallbackInput of mixed
+     *
+     * @psalm-param callable(mixed $callback): bool $callback
+     *
+     * @psalm-return Callback<CallbackInput>
+     */
     public static function callback(callable $callback): Callback
     {
         return new Callback($callback);
--- a/src/Framework/Constraint/Callback.php
+++ b/src/Framework/Constraint/Callback.php
@@ -9,18 +9,21 @@
  */
 namespace PHPUnit\Framework\Constraint;
 
-use function call_user_func;
-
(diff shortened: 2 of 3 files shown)


sebastianbergmann:
Merged manually, thanks.

Ocramius:
Excellent, thanks for the quick reaction, @sebastianbergmann!

PR #4598, opened 2021-02-03
Use template in param phpdoc

greg0ire:
It looks like Psalm is able to resolve it thanks to Callback.php, but
PHPStan isn't. To a human reading this phpdoc, it makes more sense to
see the template in both the param and return phpdoc anyway IMO.

This fixes the following PHPStan issue:

```
 ------ ----------------------------------------------------------------------- 
  Line   tests/Suggestion/Infrastructure/AlgoliaSuggestionIndexerTest.php       
 ------ ----------------------------------------------------------------------- 
  92     Unable to resolve the template type CallbackInput in call to method    
         static method PHPUnit\Framework\Assert::callback()                     
         💡 See:                                                                 
         https://phpstan.org/blog/solving-phpstan-error-unable-to-resolve-temp  
         late-type                                                              
 ------ ----------------------------------------------------------------------
 ```

I'm experiencing it on the following piece of code:

```
                    self::callback(static function (array $input) use ($expectedNumberOfQueryRules): bool {
                        self::assertCount($expectedNumberOfQueryRules, $input);                        return true;
                    })
```

### DIFF
--- a/src/Framework/Assert.php
+++ b/src/Framework/Assert.php
@@ -3117,7 +3117,7 @@ abstract class Assert
     /**
      * @psalm-template CallbackInput of mixed
      *
-     * @psalm-param callable(mixed $callback): bool $callback
+     * @psalm-param callable(CallbackInput $callback): bool $callback
      *
      * @psalm-return Callback<CallbackInput>
      */


sebastianbergmann:
Can you please send this against `8.5`? Thanks!

greg0ire:
> Can you please send this against 8.5? Thanks!

Yes sorry, I should know better :sweat_smile:

ondrejmirtes:
FYI I added a check to PHPStan that reports this situation: https://github.com/phpstan/phpstan-src/commit/18bdd34a808763764ea281d4f1b62fc0e7968873

See this example: https://phpstan.org/r/dd811d47-d1d1-4c5f-baad-1e5bf73d3e9b

greg0ire:
Nice!

PR #4639, opened 2021-03-31
Test classes loaded in data provider are not executed

kubawerlos:
When there is test class loaded in data provider then tests from it are not executed.

- ✅ tests with classes loaded not in data provider: https://github.com/sebastianbergmann/phpunit/commit/71c04c4156559d41d5756e761fc55b8c39fcef03 (all check are passing except code coverage, which is really weird as the is no single line changed in `src` directory)
- ❎ when test is updated to load class `BarTest` in `FooTest` then tests from `BarTest` are not executed: https://github.com/sebastianbergmann/phpunit/pull/4639/commits/5c79e30ebcbd1c56ae8bdd01e7e44104470ded92

The idea for a fix would be to resign from using [get_declared_classes](https://www.php.net/manual/en/function.get-declared-classes.php) ~hack~ solution and discover the classes based on file content, something like in https://github.com/sebastianbergmann/phpunit/pull/4639/commits/38d421f9943e67004cf802c7a73130a98864f516 (obviously whithout using `dev-main` branch from a silly library.

@sebastianbergmann WDYT? Do you agree with replacing `get_declared_classes` with discovering the classes based on file content? If so, how would you like to have it? Somewhere in PHPUnit codebase (in `Util` package?) or as separate repository?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,6 +29,7 @@
         "ext-xml": "*",
         "ext-xmlwriter": "*",
         "doctrine/instantiator": "^1.3.1",
+        "kubawerlos/class-finder": "dev-main",
         "myclabs/deep-copy": "^1.10.0",
         "phar-io/manifest": "^2.0.1",
         "phar-io/version": "^3.0.2",
--- a/src/Framework/TestSuite.php
+++ b/src/Framework/TestSuite.php
@@ -13,22 +13,19 @@ use const PHP_EOL;
 use function array_diff;
 use function array_keys;
 use function array_merge;
-use function basename;
 use function call_user_func;
 use function class_exists;
 use function count;
 use function dirname;
 use function file_exists;
-use function get_declared_classes;
 use function implode;
 use function is_bool;
 use function is_object;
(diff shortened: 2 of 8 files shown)


mfn:
Guess it makes sense to also check the library code part of this review, for the lazy: https://github.com/kubawerlos/class-finder/blob/main/src/ClassFinder.php

sebastianbergmann:

> When there is test class loaded in data provider then tests from it are not executed.

A data provider should/must not load test classes. I do not believe that the problem exists which you are trying to solve.

kubawerlos:
> A data provider should/must not load test classes.

I know it should/must not, but it can.

> I do not believe that the problem exists which you are trying to solve.

The test added in this PR shows the problem, when you revert the changes in `src/Framework/TestSuite.php` it will fail.

Tests added in `composer.json` via `autoload-dev` / `files` also will not be executed.

Is it possible you reconsider your decision, @sebastianbergmann?

sebastianbergmann:
No.

PR #4610, opened 2021-02-24
Mark TestCase::$backupStaticAttributes and TestCase::$runTestInSeparateProcess as nullable

morozov:
Similarly to the changes made in https://github.com/sebastianbergmann/phpunit/pull/4308, `TestCase::$backupStaticAttributes` and `TestCase::$runTestInSeparateProcess` should be marked as nullable:
1. https://github.com/sebastianbergmann/phpunit/blob/c651a94c1ead15c83de97581bb154ee5e7ae4630/src/Framework/TestCase.php#L1210
2. https://github.com/sebastianbergmann/phpunit/blob/c651a94c1ead15c83de97581bb154ee5e7ae4630/src/Framework/TestCase.php#L1220

Otherwise, Psalm reports the corresponding issues on level 2:
```php
<?php

namespace Tests;

use PHPUnit\Framework\TestCase;
use stdClass;

class ObjectTest extends TestCase
{
    /** @var stdClass */
    private $object;

    protected function setUp(): void
    {
        $this->object = new stdClass();
    }

    public function testObject(): void
    {
        self::assertIsNotNull($this->object);
    }
}
```
```
ERROR: PropertyNotSetInConstructor - tests/ObjectTest.php:8:7 - Property Doctrine\DBAL\Tests\ObjectTest::$backupStaticAttributes is not defined in constructor of Doctrine\DBAL\Tests\ObjectTest and in any private or final methods called in the constructor (see https://psalm.dev/074)
class ObjectTest extends TestCase

ERROR: PropertyNotSetInConstructor - tests/ObjectTest.php:8:7 - Property Doctrine\DBAL\Tests\ObjectTest::$runTestInSeparateProcess is not defined in constructor of Doctrine\DBAL\Tests\ObjectTest and in any private or final methods called in the constructor (see https://psalm.dev/074)
class ObjectTest extends TestCase

PR #4643, opened 2021-04-08
Declared two properties in TestCase as nullable

addiks:
Currently the properties `TestCase::runTestInSeparateProcess` and `TestCase::backupStaticAttributes` are declared as non-nullable but are not being set in the constructor. This can cause psalm issues (see https://psalm.dev/074) in projects using phpunit:

```
ERROR: PropertyNotSetInConstructor - tests/unit/Foo/Bar/BazTest.php:19:13 - Property Foo\Bar\BazTest::$runTestInSeparateProcess is not defined in constructor of Foo\Bar\BazTest and in any methods called in the constructor (see https://psalm.dev/074)
final class BazTest extends TestCase
```
```
ERROR: PropertyNotSetInConstructor - tests/unit/Foo/Bar/BazTest.php:16:13 - Property Foo\Bar\BazTest::$backupStaticAttributes is not defined in constructor of Foo\Bar\BazTest and in any methods called in the constructor (see https://psalm.dev/074)
final class BazTest extends TestCase
```

This pull request solves this issue (for these two properties) by declaring these properties as nullable booleans.

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -116,20 +116,14 @@ abstract class TestCase extends Assert implements Reorderable, SelfDescribing, T
      */
     protected $backupGlobalsExcludeList = [];
 
-    /**
-     * @var bool
-     */
-    protected $backupStaticAttributes;
+    protected ?bool $backupStaticAttributes;
 
     /**
      * @var array<string,array<int,string>>
      */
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
Duplicate of #4610.

Issue #4055, opened 2020-02-12
Tests for assertStringNotEqualsFileCanonicalizing()


sebastianbergmann:

Diraneh:
Try to use TestNG with  assertStringEquals to see if that is  false and then use your  assertStringNotEquals in your file

sebastianbergmann:
> Try to use TestNG with assertStringEquals to see if that is false and then use your assertStringNotEquals in your file

This repository is for PHPUnit (a PHP testing framework) and not about TestNG (a Java testing framework). And this is issue is about adding tests to PHPUnit's own test suite and not a question, for instance.

alecsmrekar:
#4645

PR #4645, opened 2021-04-12
Fix for #4055: Added testAssertStringNotEqualsFileCanonicalizing

alecsmrekar:
In AssertTest.php I have added a function that tests assertStringNotEqualsFileCanonicalizing() by comparing the contents of a test xml file with some test strings.


sebastianbergmann:
Cherry-picked into `10.5`, will merge to `main` from there. Thanks!

PR #4647, opened 2021-04-13
Issue #4051: Added test for assertFileNotEqualsCanonicalizing

alecsmrekar:
Linked to issue: #4051 

I first test if the method checks the existence of both the expected and actual files. I then test if the method correctly asserts that the files have different contents.


sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it: the test method you propose does too much.

PR #4649, opened 2021-04-14
Assert file content equals expected string

fefas:
I have a use case where I need to ensure that a file content is equal to an expected string. Since there is already a method for asserting the opposite case (string is equal to a file content), I just used the same code to create `assertFileEqualsString`.

@sebastianbergmann if you consider this assertion useful, I can work to get all of the following ones done:
- [x] `assertFileEqualsString`
- [ ] `assertFileNotEqualsString`
- [ ] `assertFileNotEqualsStringIgnoringCase`
- [ ] `assertFileNotEqualsStringCanonicalizing`

### DIFF
--- a/src/Framework/Assert.php
+++ b/src/Framework/Assert.php
@@ -625,6 +625,21 @@ abstract class Assert
     }
 
     /**
+     * Asserts that the contents of one file is equal to the string.
+     *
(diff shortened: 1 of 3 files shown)


sebastianbergmann:

PR #4653, opened 2021-04-19
Enhancement: Support string ending with empty suffix

CristianLlanos:
When asserting a suffix with an empty string it is failing when it should probably assume it does end wit the suffix
```php
assertStringEndsWith('', 'Some Text');
```
The way it is now, it throws an error:
```
Failed asserting that 'Some Text' ends with "".
```
I hope this contribution gets accepted. It would make some test configurations easier. Thanks.
If accepted, this commit should probably bubble up to the newer version.


sebastianbergmann:

Issue #4588, opened 2021-01-21
Feature Request: Add `--exclude-testsuite` option


phanan:
## Problem

As of current, PHPUnit has a `testsuite` option to specify which suite to run, but none for the reversed effect: to specify which suite _not_ to. For example, we have several test suites like this:

```bash
➜  phpunit --list-suites
PHPUnit 9.5.1 by Sebastian Bergmann and contributors.

Available test suite(s):
 - Feature
 - Integration
 - Unit
```

Due to the nature of our project, we'd like to extract `Feature` into its own CI pipeline. This is, of course, already doable with 

```bash
phpunit --testsuite=Feature
``` 

However, the other (main) pipeline still _includes_ Feature by default superfluously. To exclude it, we have to use `--testsuite` with a comma-separate list of included suites, like this

```bash
phpunit --testsuite Unit,Integration
```

which is not exactly terrible, but is pretty verbose and harder to maintain should there be any changes to the list of suites in the future.

## Suggestion

To add an `exclude-testsuite` option, with which the above command can be replaced with:

```bash
phpunit --exclude-testsuite Feature
```

Thoughts?

PR #4657, opened 2021-04-22
Add --exclude-testsuite option

alecsmrekar:
Covers issues #4588 
I add the logic for a new parameter, "--exclude-testsuite"


alecsmrekar:
I'm a bit of out my depth about why this specific PHP 8.1 test is failing, any clues?

sebastianbergmann:
Merged manually, thanks.

PR #4662, opened 2021-05-03
Implement __serialize() on DefaultTestResultCache

derrabus:
The `Serializable` interface is being phased out as described in https://wiki.php.net/rfc/phase_out_serializable

The consequence is that PHP 8.1 will trigger a deprecation notice if a class implements `Serializable` unless it also implements the magic methods `__serialize()` and `__unserialize()`. Those have been introduced with PHP 7.4 and are meant to replace the old `Serializable` mechanism.

PHPUnit's `DefaultTestResultCache` is affected by this change. This PR proposes to implement `__serialize()` and `__unserialize()` to avoid the deprecation.

Because this changes the output of the serialization slightly, I had to adjust a couple of tests.

### DIFF
--- a/src/Runner/DefaultTestResultCache.php
+++ b/src/Runner/DefaultTestResultCache.php
@@ -93,6 +93,36 @@ final class DefaultTestResultCache implements Serializable, TestResultCache
         $this->cacheFilename = $filepath ?? $_ENV['PHPUNIT_RESULT_CACHE'] ?? self::DEFAULT_RESULT_CACHE_FILENAME;
     }
 
+    public function __serialize(): array
+    {
+        return [
+            'defects' => $this->defects,
+            'times'   => $this->times,
+        ];
+    }
+
+    public function __unserialize(array $data): void
+    {
+        if (isset($data['times'])) {
+            foreach ($data['times'] as $testName => $testTime) {
+                assert(is_string($testName));
+                assert(is_float($testTime));
+                $this->times[$testName] = $testTime;
+            }
+        }
+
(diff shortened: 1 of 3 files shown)


sebastianbergmann:
Thank you for working on this, Alexander.

However, it may not be as simple as "just" updating the tests. Is the new code able to load from a cache file created by the old code? It is is: great! If it's not, we need to adapt the loader to not crash when loading a file it does not understand and treat that situation as a cache miss.

derrabus:
> Is the new code able to load from a cache file created by the old code?

It should be. There is a test in this PR that performs an `unserialize()` with the old format.

sebastianbergmann:
> It should be. There is a test in this PR that performs an `unserialize()` with the old format.

Thank you for pointing that out, I did not have time yet to (closely) look at the proposed changes.

derrabus:
However, it does not work the other way round: PHP 7.3 won't be able to unserialize the format generated by PHP 7.4.

sebastianbergmann:
Good to know, thanks. Only goes to show that using `serialize()` was not a great choice to begin with.

sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it. Instead, I have [changed](https://github.com/sebastianbergmann/phpunit/commit/3e3aecdeb1160de5500fa7715436e73689cc11a3) the implementation of `DefaultTestResultCache` to not use `serialize()` / `unserialize()` at all.

derrabus:
No worries. Your way is certainly the better solution in the long run!

PR #4668, opened 2021-05-13
Update LICENSE

exussum12:
The licence file has been changed so its not recognised by github. This puts the licence back to a more standard template


sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it as I do not believe this to be necessary.

PR #4681, opened 2021-05-21
Fix: Do not use set-env

localheinz:
This pull request

- [x] stops using `set-env` to set the `COMPOSER_CACHE_DIR` environment variable

Follows #4425.

🙈 Sorry, my bad!


localheinz:

PR #4682, opened 2021-05-21
Fix: Remove condition

localheinz:
This pull request

* [x] removes a condition that does not make sense

Follows #4681.

🙊


localheinz:

PR #4683, opened 2021-05-21
Fix: Use PHP 7.3

localheinz:
This pull request

* [x] uses PHP 7.3 instead of PHP 7.2 for a job


localheinz:

PR #4687, opened 2021-05-25
update meta to properly infer type of string parameters

isfedorov:
After the fix of https://youtrack.jetbrains.com/issue/WI-60544 type() meta directive won't treat string literals contents as types anymore, prefered way is to use (map([""=>"$0"])). This PR updates meta file according to preferred way.


sebastianbergmann:
Please send this against `8.5`, I will merge it from there to `9.5` and `master`. Thanks!

PR #4695, opened 2021-06-04
Fix: Continue on error when running tests on PHP 8.1

localheinz:
This pull requests

* [x] continues on error when running tests on PHP 8.1

💁‍♂️ For reference, see https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error.


sebastianbergmann:

PR #4713, opened 2021-06-15
Implement --no-output

Slamdunk:
Hi, https://github.com/sebastianbergmann/phpunit/commit/916cfb94382fcb3cd2ef43713a036af2d950ca29 removed the possibility to have a custom printer, which is fine.

But in [ParaTest](https://github.com/paratestphp/paratest) we need to have no output from PHPUnit process to be able to gather erroneous outputs sent to `STDOUT` and also be able to have a clear overview of the eventual exception thrown in case it happens.

Only `--no-output` has been implemented here: no `phpunit.xml` entry added, on purpose, as I see no point for disabled the output by default by configuration.

### DIFF
--- a/tests/end-to-end/loggers/no-output.phpt
+++ b/tests/end-to-end/loggers/no-output.phpt
@@ -1,0 +1,17 @@
+--TEST--
+phpunit --teamcity ../../_files/BankAccountTest.php
+--FILE--
+<?php declare(strict_types=1);
+
+print 'placeholder' . PHP_EOL;
+
+$_SERVER['argv'][] = '--do-not-cache-result';
+$_SERVER['argv'][] = '--no-configuration';
+$_SERVER['argv'][] = '--no-output';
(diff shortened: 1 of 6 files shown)


tests/end-to-end/loggers/no-output.phpt
```diff
@@ -0,0 +1,17 @@
+--TEST--
+phpunit --teamcity ../../_files/BankAccountTest.php
+--FILE--
+<?php declare(strict_types=1);
+
+print 'placeholder' . PHP_EOL;
```
Slamdunk, line 6: `PhptTestCase` doesn't allow empty `EXPECTF`, see https://github.com/sebastianbergmann/phpunit/blob/187407c87fbe7ee1455a1ef10e46e23fbd80860f/src/Runner/PhptTestCase.php#L350-L353

sebastianbergmann:
You beat me to it, I also wanted to implement something like this.

PR #4716, opened 2021-06-22
Compare PHP versions by string first

barryvdh:
Issue:
 - MacOS has a weird PHP version by default (`7.3.24-(to be removed in future macOS)`)
 - For some reason, checking this version against itself with `>=` is `false` (https://3v4l.org/EVYsT)
 - This allows PHPUnit to run (it passes against the 7.3 check) but fails against the directory check, which results in `No tests executed!`
 - See for a more extensive read: https://bbqsoftwares.com/blog/phpunit-big-sur by @lcharette

Proposed fixes:
 - Normalize PHP version instead of PHP_VERSION constants
 - Use internal version_compare method to normalize results
 - Check if strings are equal and skip version checks in that case (this PR)

### DIFF
--- a/src/TextUI/TestSuiteMapper.php
+++ b/src/TextUI/TestSuiteMapper.php
@@ -51,7 +51,7 @@ final class TestSuiteMapper
                 $testSuiteEmpty = true;
 
                 foreach ($testSuiteConfiguration->directories() as $directory) {
-                    if (!version_compare(PHP_VERSION, $directory->phpVersion(), $directory->phpVersionOperator()->asString())) {
+                    if ($directory->phpVersion() !== PHP_VERSION && !version_compare(PHP_VERSION, $directory->phpVersion(), $directory->phpVersionOperator()->asString())) {
                         continue;
                     }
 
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
I would rather `die();` on startup if `-(to be removed in future macOS)` is part of `PHP_VERSION`, to be honest.

barryvdh:
Yeah uhm I guess that's one possibility. It works otherwise just like normal PHP but it's very confusing that the phpunit runs, but can't find any tests. So an error like 'install a proper PHP version with Brew' is an alternative if you don't want any 'fixes' to work around this.

sebastianbergmann:
I fear that any workaround for this problem has the potential to create problems for developers not affected by the original problem.

While I appreciate the time you invested in preparing this pull request, I have decided not to merge it.

barryvdh:
No problem! Do you want me to submit an PR with the die() on startup, or do you prefer to add it yourself?

sebastianbergmann:
I'll do that myself, thanks.

lcharette:
> I fear that any workaround for this problem has the potential to create problems for developers not affected by the original problem.

That was my thought too. Note that once Apple actually remove PHP, "installing it with Brew" will be the only solution. So long term, might be best not to merge this and add the warning indeed.

sebastianbergmann:
See https://github.com/sebastianbergmann/phpunit/commit/2b0a4bb65811306a09d62d4c406b9d060baa8480

barryvdh:
Thanks!

PR #4719, opened 2021-06-22
Handle invalid PHP version strings

shalvah:
A better workaround for https://github.com/sebastianbergmann/phpunit/commit/2b0a4bb65811306a09d62d4c406b9d060baa8480. This way, PHPUnit will error for _any_ invalid PHP version strings (like "PHP 7.4.1-(custom)"), not just the one macOS currently has.

### DIFF
--- a/build/templates/binary-phar-autoload.php.in
+++ b/build/templates/binary-phar-autoload.php.in
@@ -1,6 +1,6 @@
 #!/usr/bin/env php
 <?php
-if (substr(PHP_VERSION, 0 - strlen('-(to be removed in future macOS)')) === '-(to be removed in future macOS)') {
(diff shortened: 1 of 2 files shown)


shalvah:
This works because `version_compare`'s equality check fails for invalid PHP versions:

```php
version_compare("7.4.1", "7.4.1", "=")
// => true
version_compare("7.3.22-(to be removed in future macOS)", "7.3.22-(to be removed in future macOS)", '=')
// => false
version_compare("7.3.22-(custom)", "7.3.22-(custom)", '=')
// => false
```

sebastianbergmann:
Thank you, this is indeed better.

PR #4743, opened 2021-07-23
Make PHARs created during build available

jrfnl:
I wanted to try and test something with the latest `dev-9.5` version (against PHP 8.1) and realized I would have to build the phar myself.

To make it easier for people to test proposed changes or test the latest dev version of PHPUnit with a phar, I'd like to propose to automatically make the phar(s) generated during the CI builds available as artifacts.

Each phar test build will generate a zip file for each uploaded file - in this case for PHPUnit 8, that would be 8 different zips containing phars.
The zips can be downloaded from the Actions Summary page after each CI run.

The default retention period for build artifacts is 90 days, but as this repo is under active development, I'd like to suggest limiting this to 14 days to conserve resources.

Also see: https://github.com/actions/upload-artifact

Note: while I didn't want to go outside the scope of this PR, I do wonder why the phar is being build on each PHP version.
I imagine building the phar once (against the lowest supported PHP version ?) and then using that phar for all the test runs against the various supported PHP versions would make the script more efficient as well as reflect the reality more closely as the Phar which is released is not differentiated for each PHP version either.

I'd happily adjust this PR to make that change (and reduce the artifact uploads to two phars - scoped and unscoped) if so desired.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -256,3 +256,10 @@ jobs:
 
       - name: Run PHAR-specific tests with scoped PHAR
         run: ./build/artifacts/phpunit-snapshot.phar --configuration tests/phar/phpunit.xml
+
+      - uses: actions/upload-artifact@v2
+        if: ${{ matrix.php-version == 7.4 }}
+        with:
+          name: phpunit-snapshot-phar
+          path: ./build/artifacts/phpunit-snapshot.phar
+          retention-days: 7


sebastianbergmann:
> To make it easier for people to test proposed changes or test the latest dev version of PHPUnit with a phar, I'd like to propose to automatically make the phar(s) generated during the CI builds available as artifacts.

Makes sense. In the past, a nightly build was built and published on `phar.phpunit.de`. According to the `access.log`, this was so rarely downloaded that I removed it.

> Each phar test build will generate a zip file for each uploaded file

Can creating a ZIP be avoided? I think having to unpack a ZIP before being able to use the PHAR is an inconvenience that should be avoided if possible.

> The default retention period for build artifacts is 90 days, but as this repo is under active development, I'd like to suggest limiting this to 14 days to conserve resources.

For something that used to be nightly (with no retention), I think that 14 days still too much. One day should be enough, IMO.

> Note: while I didn't want to go outside the scope of this PR, I do wonder why the phar is being build on each PHP version.

To discover issues such as PHP-Scoper not being able to build the PHAR as early as possible.

> I imagine building the phar once (against the lowest supported PHP version ?) and then using that phar for all the test runs against the various supported PHP versions would make the script more efficient as well as reflect the reality more closely as the Phar which is released is not differentiated for each PHP version either.

That makes sense, too, though. I think we should do what you propose here. Feel free to implement this as part of this PR.

> I'd happily adjust this PR to make that change (and reduce the artifact uploads to two phars - scoped and unscoped) if so desired.

Please only make the unscoped PHAR available for download.

I have renamed "nightly" to "snapshot" in 5f1b587879a25fddf1c777e6bfe48f2330c5623b because I think that is more accurate.

jrfnl:
> According to the access.log, this was so rarely downloaded that I removed it.

I get that, but with it being in Actions now, it can help if there is, for instance, a PR which needs testing by the original OP, as you can now just point them to the artifact created when the PR build ran. I imagine this can be useful at times.

Example case in point: I know that's how I've been asked to test things a number of times for bugs I reported for my Git GUI, for instance, where the dev env is so different that building the exe for that myself was not so straight forward.
It really helped to be able to give feedback on the PR in those cases and I just imagine that this principle may be useful here once in a while too.

> Can creating a ZIP be avoided? I think having to unpack a ZIP before being able to use the PHAR is an inconvenience that should be avoided if possible.

No, unfortunately it can't be per https://github.com/actions/upload-artifact#zipped-artifact-downloads

> For something that used to be nightly (with no retention), I think that 14 days still too much. 7 days should be more than enough, IMO.

✔️ 

> To discover issues such as PHP-Scoper not being able to build the PHAR as early as possible.
> That makes sense, too, though. I think we should do what you propose here. Feel free to implement this as part of this PR.
> Please only make the unscoped PHAR available for download.

We could choose to do both: the Phar could still be build on each PHP version during the test run, but only the _unscoped_ phar for _**one**_ particular PHP version would be uploaded as an artifact.

Against which PHP version is the phar build when PHPUnit is released ? Against the lowest supported PHP version ?
I'd suggest using that phar as the artifact as that would most closely match what people would download during a release.
 

> I have renamed "nightly" to "snapshot" in 5f1b587 because I think that is more accurate.
 ✔️

sebastianbergmann:
> > Please only make the unscoped PHAR available for download.
> 
> We could choose to do both: the Phar could still be build on each PHP version during the test run, but only the _unscoped_ phar for _**one**_ particular PHP version would be uploaded as an artifact.

I do not want to make unscoped PHARs available as they are only generated for internal testing purposes.

jrfnl:
@sebastianbergmann No problem, build against which PHP version ?

sebastianbergmann:
> Against which PHP version is the phar build when PHPUnit is released?

The version of PHP that is used to build the PHAR does not matter (unless it does, because one of the tools used is not compatible with the version of PHP used, but that is not relevant here).

jrfnl:
Ah, I'd just adjusted it to use the lowest + 1 version.

New commit description per the updated commit:

> #### CI: make phars created during build available
> 
> I wanted to try and test something with the latest `dev-9.5` version (against PHP 8.1) and realized I would have to build the phar myself.
> 
> To make it easier for people to test proposed changes or the latest dev version of PHPUnit with a phar, I'd like to propose to automatically make the phar(s) generated during the CI builds available as artifacts.
> 
> Per the discussion in the PR:
> * For each build one (1) phar file will be uploaded, which will be available as an artifact in a zip file.
> * This is the `scoped` phar.
> * And will be retained for 7 days.
> 
> The zips can be downloaded from the Actions Summary page after each CI run.
> 
> Also see: https://github.com/actions/upload-artifact

jrfnl:
Oops, made a type, should be fixed now. You can verify that it works by checking that the artifact can be seen on the PR CI build Summary page.

sebastianbergmann:
> > I imagine building the phar once (against the lowest supported PHP version ?) and then using that phar for all the test runs against the various supported PHP versions would make the script more efficient as well as reflect the reality more closely as the Phar which is released is not differentiated for each PHP version either.
> 
> That makes sense, too, though. I think we should do what you propose here. Feel free to implement this as part of this PR.

Are you still planning to implement this? If so, will that be a separate/new pull request? Wondering whether I should merge this as-is or wait. Thanks! for everything you have done so far.

jrfnl:
> > > I imagine building the phar once (against the lowest supported PHP version ?) and then using that phar for all the test runs against the various supported PHP versions would make the script more efficient as well as reflect the reality more closely as the Phar which is released is not differentiated for each PHP version either.
> > 
> > 
> > That makes sense, too, though. I think we should do what you propose here. Feel free to implement this as part of this PR.
> 
> Are you still planning to implement this? If so, will that be a separate/new pull request? Wondering whether I should merge this as-is or wait. Thanks! for everything you have done so far.

As far as I'm concerned this is ready for merge, unless you would like other changes.

I've changed it to work as proposed in [this comment](https://github.com/sebastianbergmann/phpunit/pull/4743#issuecomment-885521918):

> We could choose to do both: the Phar could still be build on each PHP version during the test run, but only the unscoped phar for one particular PHP version would be uploaded as an artifact.

So, the Phar is still build against each supported PHP version to allow for early detection of issues, like the PHP 8.1 PHP_Scoper issue.
But only one of the build Phar files is made available as a build artifact (per this PR the one build against PHP 7.4 scoped).

jrfnl:
@sebastianbergmann If you I'd be happy to still do another PR to split the "phar building test" from the "phar running test", where the _phar running test_ would then always use the artifact phar.

sebastianbergmann:
Yes, please. Thanks!

localheinz (commented):
👍

jrfnl:
@sebastianbergmann Got a initial setup done, but finding some issues.

One PR fixing an issue has been merged, though no new release tagged (yet): https://github.com/theseer/Autoload/pull/99

The other issue seems to be that the phar test runs as they were, look to have been broken to begin with. They only happened to succeed because the building of the Phar - including the `composer install` to get the dependencies in place to build the Phar - and the test run were being done in one job.

Once I split the job into two, with the second job not doing a `composer install`, but just running the tests using the pre-build phar file, the test run fails to start with a `PHP Fatal error:  require_once(): Failed opening required '/home/runner/work/phpunit/phpunit/vendor/autoload.php' (include_path='.:/usr/share/php') in /home/runner/work/phpunit/phpunit/tests/bootstrap.php on line 21` error.

See: https://github.com/jrfnl/phpunit/runs/3145719478?check_suite_focus=true
Relevant commit: https://github.com/jrfnl/phpunit/commit/e01256011f3d0df5ebdafa98acd914c9248aa405

If I add a `composer install` to the "test run against Phar" job, the builds pass:
See: https://github.com/jrfnl/phpunit/runs/3146983296?check_suite_focus=true
Relevant commit: https://github.com/jrfnl/phpunit/commit/b7fd7f30205330845f387361cc57c8bda3a7aa8f

That would indicate that there is a hick-up somewhere in the Phar build process.

I'll try and dig deeper, but if either you or @localheinz have any insights, they'd be welcome.

sebastianbergmann:
> If I add a `composer install` to the "test run against Phar" job, the builds pass:
> See: https://github.com/jrfnl/phpunit/runs/3146983296?check_suite_focus=true
> Relevant commit: [jrfnl@b7fd7f3](https://github.com/jrfnl/phpunit/commit/b7fd7f30205330845f387361cc57c8bda3a7aa8f)

This is due to `autoload-dev` (https://github.com/sebastianbergmann/phpunit/blob/master/composer.json#L76) and expected. The test suite fixture contains code units that need to be autoloaded.

jrfnl:
@sebastianbergmann Of course, that explains it. In that case, I'll tidy those commits up and will send in a PR soonish.

theseer:
I made a new phpab release (1.26.2). I'm not sure though if that was actually required: Even after merging @jrfnl 's PR (https://github.com/theseer/Autoload/pull/99), a local `composer update` didn't change anything on my box which is also used for creating the phar.

There's also a new release of `phive` in case that is relevant.

sebastianbergmann:
> I made a new phpab release (1.26.2).

Updated: https://github.com/sebastianbergmann/phpunit/commit/ae8359ed5e3acbf9138b55f5c50f58f5f5154923

jrfnl:
@theseer There was a whole chain of notices, mostly deprecations, in the CI, these were just a few of the fixes. All stepping stones to get a clean build on PHP 8.1 in the end. It all adds up.

https://twitter.com/jrf_nl/status/1418671832518045701

PR #4753, opened 2021-08-10
get away from SHA-1, switch to SHA-512

remicollet:
Available for years in phar extension.


sebastianbergmann:
Cherry-picked into `8.5` and merged to `9.5` and `master` from there.

PR #4772, opened 2021-09-06
TestDox HTML report not displayed correctly when browser has custom colour settings

DanielSiepmann:
Browsers support custom colour settings.
Some people might change default text and background colour.

The report defines custom colour for successful and failed tests.
It should therefore ensure that those colours work as expected.
Background colour and default text colour are set in addition.
This ensures an expected end result, no matter what colours might be set as defaults.


DanielSiepmann:
Example before:
image
after:
image

sebastianbergmann:
Please send this against the `8.5` branch. Thanks!

DanielSiepmann:
Done. Will try to use the oldest available branch in the future.

PR #4777, opened 2021-09-13
Allows running a single test by specifying line no

gquemener:
## tl; dr;
This PR intends to introduce possiblity to run a single test through its line number : `phpunit test/Foo/BarTest.php:42`

---

Hello,

Being able to quickly run a single test is probably the feature I miss the most in phpunit.

Currently, I know of 2 ways to do it, either using `--filter <testName>` or `--groups`.

Unfortunately, I personnaly feel this is not easy, as it requires me to copy/paste values from my editor to the command line.

[PhpSpec has a neat option to execute a single test by specifying its line number.](https://github.com/phpspec/phpspec/blob/main/docs/cookbook/console.rst#run-command)

This PR ports this feature into PHPUnit (check the [filter-line.phpt](https://github.com/sebastianbergmann/phpunit/pull/4777/files#diff-cc8d1051a0dc9a2b4dab76c1340e31585b59fb3a3963ddbeca5b24b32c31e1f8) test case)*.

The testsuite is not passing yet, as I wanted to first gather feedbacks about such feature.

I haven't been able to find any related issue/PR, so feel free to add any if you know that one already exists!

*The port is actually not that identical. PhpSpec allows to specify any line between the first and end line of the test method. This port allows to only specify the first line.

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -161,6 +161,8 @@ abstract class TestCase extends Assert implements Reorderable, SelfDescribing, T
 
     private string $name;
 
+    private ?int $line;
+
     /**
      * @psalm-var list<string>
      */
@@ -335,9 +337,10 @@ abstract class TestCase extends Assert implements Reorderable, SelfDescribing, T
     /**
      * @internal This method is not covered by the backward compatibility promise for PHPUnit
      */
-    public function __construct(string $name)
+    public function __construct(string $name, ?int $line = null)
     {
         $this->setName($name);
+        $this->setLine($line);
 
         $this->status = TestStatus::unknown();
     }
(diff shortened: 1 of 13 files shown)


gquemener:
I'm wondering how this should behave with `--filter` :thinking:

src/Framework/TestCase.php
```diff
@@ -1356,6 +1364,11 @@ public function wasPrepared(): bool
         return $this->wasPrepared;
     }
 
+    public function line(): int
```
gquemener, line 1367:
```suggestion
    public function line(): ?int
```

sebastianbergmann:

gquemener:
You're welcome, have a great day, and thanks for your work on PHPUnit :pray:

PR #4792, opened 2021-09-30
Add `RequiresClass` attribute

kbond:
Similar to `RequiresMethod` but just check the class (possibly interface too?). If this would be acceptable, I can complete this PR.


sebastianbergmann:

Issue #4798, opened 2021-10-13
Memory leaks in TestSuite class


kandy:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 9.2.6
| PHP version         | 7.4.12
| Installation Method | Composer 

#### Summary
PHPUnit requires a significant amount of memory on a large codebase due to memory leaks in TestSuite class

#### Current behavior
PHPUnit use ~2Gb for 30k tests

<img width="1121" alt="Screen Shot 2021-10-13 at 14 30 35" src="https://user-images.githubusercontent.com/100289/137203874-94f8f230-cdf4-43dd-9345-207d6c8e9a03.png">

#### How to reproduce

- git clone git@github.com:magento/magento2.git
- composer install
- vendor/bin/phpunit -c dev/tests/unit/phpunit.xml
- see memory usage  ( or use php-spx extension: SPX_ENABLED=1 SPX_SAMPLING_PERIOD=500 SPX_FP_LIVE=1 SPX_FP_FOCUS=zm  vendor/bin/phpunit -c dev/tests/unit/phpunit.xml)

#### Expected behavior
After applying patch below, memory usage dropped to 300MB

<img width="1117" alt="Screen Shot 2021-10-13 at 14 30 56" src="https://user-images.githubusercontent.com/100289/137205592-640d5356-48bb-452a-8ce7-15b0fde17779.png">

```diff

Index: vendor/phpunit/phpunit/src/Framework/TestSuite.php
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/vendor/phpunit/phpunit/src/Framework/TestSuite.php b/vendor/phpunit/phpunit/src/Framework/TestSuite.php
--- a/vendor/phpunit/phpunit/src/Framework/TestSuite.php	
+++ b/vendor/phpunit/phpunit/src/Framework/TestSuite.php	
@@ -134,7 +134,7 @@
             );
         }
 
-        $this->declaredClasses = \get_declared_classes();
+        // $this->declaredClasses = \get_declared_classes();
 
         if (!$theClass instanceof \ReflectionClass) {
             if (\class_exists($theClass, true)) {
@@ -354,7 +354,7 @@
         if (\file_exists($filename) && \substr($filename, -5) === '.phpt') {
             $this->addTest(new PhptTestCase($filename));
 
-            $this->declaredClasses = \get_declared_classes();
[...]

PR #4799, opened 2021-10-14
ISSUE-4798: Memory leaks in TestSuite class

kandy:
- Do not store list of classes in property
  fix: #4798

### DIFF
--- a/src/Framework/TestSuite.php
+++ b/src/Framework/TestSuite.php
@@ -121,11 +121,6 @@ class TestSuite implements IteratorAggregate, SelfDescribing, Test
     private $iteratorFilter;
 
     /**
-     * @var string[]
-     */
-    private $declaredClasses;
-
-    /**
      * Constructs a new TestSuite:.
      *
      *   - PHPUnit\Framework\TestSuite() constructs an empty TestSuite.
@@ -154,8 +149,6 @@ class TestSuite implements IteratorAggregate, SelfDescribing, Test
                 'ReflectionClass object or string'
             );
         }
-
-        $this->declaredClasses = get_declared_classes();
 
         if (!$theClass instanceof ReflectionClass) {
             if (class_exists($theClass, true)) {
@@ -379,10 +372,11 @@ class TestSuite implements IteratorAggregate, SelfDescribing, Test
             return;
         }
 
+        $declaredClasses = get_declared_classes();
         // The given file may contain further stub classes in addition to the
         // test class itself. Figure out the actual test class.
         $filename   = FileLoader::checkAndLoad($filename);
-        $newClasses = array_diff(get_declared_classes(), $this->declaredClasses);
+        $newClasses = array_diff(get_declared_classes(), $declaredClasses);
 
         // The diff is empty in case a parent class (with test methods) is added
         // AFTER a child class that inherited from it. To account for that case,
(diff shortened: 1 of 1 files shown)


IonBazan:
This has been introduced in #2861 on purpose for some edge case. Let's make sure this PR does not introduce a regression. 

cc @kubawerlos.

villfa:
In order to reduce the memory usage I propose to store only the number of declared classes instead of the whole list of declared classes:
**BEFORE**
```php
$declaredClasses = get_declared_classes();
// some code
$newClasses = array_diff(get_declared_classes(), $declaredClasses);
```

**AFTER**
```php
$numberOfDeclaredClasses = count(get_declared_classes());
// some code
$newClasses = array_slice(get_declared_classes(), $numberOfDeclaredClasses);
```

kandy:
> In order to reduce the memory usage I propose to store only the number of declared classes instead of the whole list of declared classes:

Based  on https://www.php.net/manual/en/function.get-declared-classes.php
> No particular order is guaranteed for the get_declared_classes() return value.

It's maybe not the best idea to depend on behavior that may be changed, but let's try

villfa:
I've missed this change. Thanks

kandy:
@villfa I have implemented your suggestion

sebastianbergmann:
When I apply this patch to `9.5` then I get the following test failure:

```
1) /usr/local/src/phpunit/tests/end-to-end/regression/GitHub/2972.phpt
Failed asserting that string matches format description.
--- Expected
+++ Actual
@@ @@
 
 Warning:       Test case class not matching filename is deprecated
                in /usr/local/src/phpunit/tests/end-to-end/regression/GitHub/2972/unconventiallyNamedIssue2972Test.php
+               Class name was 'PhptTestCase', expected 'unconventiallyNamedIssue2972Test'
+Warning:       Test case class not matching filename is deprecated
+               in /usr/local/src/phpunit/tests/end-to-end/regression/GitHub/2972/unconventiallyNamedIssue2972Test.php
                Class name was 'Issue2972Test', expected 'unconventiallyNamedIssue2972Test'
 
-Time: %s, Memory: %s
+Time: 00:00.051, Memory: 4.00 MB
 
 OK (2 tests, 2 assertions)
```

kandy:
@sebastianbergmann For 9.x there is one more usage of `declaredClasses` property in the class that should be replaced with `declaredClassesPointer`.  see  https://github.com/sebastianbergmann/phpunit/compare/master...kandy:Issue-4798-9x?expand=1#diff-ce07bbf9be7fe14e3102ab699c8a2d6c2a995bc23f8f93871fb4f3dde098dfaaR392 (line: 392)

sebastianbergmann:
>For 9.x there is one more usage of `declaredClasses` property

And by "9.x" you mean "8.x", right? Because that is the target of this pull request.

kandy:
> > For 9.x there is one more usage of `declaredClasses` property
> 
> And by "9.x" you mean "8.x", right? Because that is the target of this pull request.

no 9.5, see your previous comment https://github.com/sebastianbergmann/phpunit/pull/4799#issuecomment-964834321

sebastianbergmann:
> no 9.5, see your previous comment [#4799 (comment)](https://github.com/sebastianbergmann/phpunit/pull/4799#issuecomment-964834321)

My bad, too early in the morning :)

kandy:
@sebastianbergmann What is the status of this OR? Is any further action is needed to merge it?

sebastianbergmann:
I merged these changes into `8.5`. I merged them from there to `9.5` and made the change you suggested in https://github.com/sebastianbergmann/phpunit/pull/4799#issuecomment-987163544 in e16985050d5d98a84b7123058325f451f89051eb. The code in question does not exist in `master` anymore.

PR #4800, opened 2021-10-15
Make it easier to register multiple subscribers at once

IonBazan:
This change makes it much easier to register multiple subscribers with a single method call.


sebastianbergmann:
I am sorry, this fell of my radar. I am also sorry that I did not pick your commit, Ion, but doing it myself was quicker than solving the conflicts.

We still do not have any test for the facade, as Arne already pointed out. We will deal with that when we review the tests for the event system.

IonBazan:
No problem @sebastianbergmann, this PR serves more as a RFC rather than the actual change to be committed.

Issue #4775, opened 2021-09-11
Deprecated MockBuilder::setMethods does not throw a warning in PHPUnit 9


mondrake:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 9.5
| PHP version         | n/a
| Installation Method | n/a

#### Summary

Working on adjusting Drupal for PHPUnit 10, I realized `MockBuilder::setMethods` is removed from PHPUnit 10. Unlike other methods deprecated in PHPUnit 8 and 9, there's no warning thrown about the deprecation.

Since in Drupal we rely on deprecation errors being triggered, so far this deprecation was missed from any conversion.

#### Current behavior

n/a

#### How to reproduce

n/a

#### Expected behavior

I suppose a Warning should be thrown in PHPUnit 9, on top of the annotated deprecation.

sebastianbergmann:
Unfortunately, "soft deprecation" (annotating the methods in question with `@deprecated` is the only feasible way to do this.

PR #4803, opened 2021-10-16
Add warning to MockBuilder::setMethods() about deprecated method

adrmrn:
It's my very first PR to the PHPUnit project, so any suggestions are welcomed.

These changes are referring to the #4775 issue. Until now, there is no warning about the deprecated `setMethods()` method.

Problem I had was that `TestCase::createPartialMock()` still using `MockBuilder::setMethods()` and I'm not sure if it's possible to refactor code to use `onlyMethods()` & `addMethods()` instead. I decided to exclude warning for this specific use case.


sebastianbergmann:

PR #4818, opened 2021-11-10
Added assertArrayIsList()

kukulich:


kukulich:
@sebastianbergmann Would you accept this when I add more tests etc?

sebastianbergmann:
> @sebastianbergmann Would you accept this when I add more tests etc?

Yes.

src/Framework/Assert.php
```diff
@@ -113,6 +114,15 @@ public static function assertArrayNotHasKey(int|string $key, array|ArrayAccess $
         static::assertThat($array, $constraint, $message);
     }
 
+    public static function assertArrayIsList(array $array, string $message = ''): void
+    {
+        static::assertThat(
+            $array,
+            new ArrayIsList(),
```
sebastianbergmann, line 121: Please no superfluous `()` when no constructor arguments are passed.

kukulich:
It looks the fail of type checker is not related to this patch.

sebastianbergmann:
> It looks the fail of type checker is not related to this patch.

I'll look into this.

src/Framework/Constraint/Traversable/ArrayIsList.php
```diff
@@ -0,0 +1,64 @@
...
+    protected function matches(mixed $other): bool
+    {
+        if (!is_array($other)) {
+            return false;
+        }
+
+        if ($other === []) {
+            return true;
+        }
+
```
derrabus, line 40:
```suggestion

        if (function_exists('array_is_list')) {
            return array_is_list($other);
        }

```

On PHP 8.1, we have [a native function](https://www.php.net/array_is_list) for that. I suggest we leverage it if it's available.

sebastianbergmann:
Merged manually. Thanks!

PR #4831, opened 2021-11-25
Fixed missing the assert diff message of phpt in a teamcity output

benbor:
Hi! 

1. I've added missed diff output in case of failure phpt test in `--teamcity` configuration, 
when phpt scenario contains `--EXPECT--` section.
2. I've put the diff output in case of failure phpt test with section `--EXPECTF--` into the same style as `--EXPECT--`.

## Background

PHPUnit 9.5, **failed** test scenarious in `*.phpt` files, output in `--teamcity` 
Note:  `--teamcity` output is also used by PHPStorm, so all screenshot examples are taken from it.

image

As you can see:
1.  `--EXPECT--` and `--EXPECTF--` output is different. 
2. `<Click to see difference>` tool is absent

## First attempt

In the first attempt  https://github.com/sebastianbergmann/phpunit/commit/1086fedd4ea3df6ba6320a7d93cd33afecff4b2b I connected missed `--EXPECT--` diff to the message, in the same style, as it had been done for `--EXPECTF--` 

Pros: 
- That were small code changes 

Cons: 
- Diff was printed as plained text, i.e. the tool `<Click to see difference>` still hadn't work

### First attempt result, the `<Click to see difference>` was absent:
image

## Second attempt (current version)

I  founded, that  PHPStorm can work with diff in prettier way (you can see how common `PHPUnit\Framework\TestCase` works on top of screenshots), and I started to investigate. 
I've discovered the following: 
* `PHPTAssertionFailedError` is not consistent with `ExpectationFailedException`, which works perfect in `PHPUnit\Framework\TestCase` test cases. 
* `PHPTAssertionFailedError` displayed information in message for `--EXPECTF--`, because of  https://github.com/sebastianbergmann/phpunit/pull/3659 
* Difference in `--EXPECT--` and `--EXPECTF--` exists because  phpunit uses different  constraints: `StringMatchesFormatDescription`  and `IsEqual`. The first one used it's own `attach diff in message` implementation. 

So, I fixed all of them together, because they are coupled. What has been done:
1. Now `StringMatchesFormatDescription` works with expected diff as other classes: manipulates with `ComparisonFailure` object. I used the same idea of implementation as in `JsonMatches` constraint. 
2. Interface `ComparisonFailureContains` has been created, to have possibility to accept `ComparisonFailure` object from both classes: `PHPTAssertionFailedError` and `ExpectationFailedException`
3. Added two end-to-end logger tests  for both `--EXPECT--` and `--EXPECTF--` cases. 

### In comparison with First attempt fix: 
Pros: 
- both phpt scenarios `--EXPECT--` and `--EXPECTF--` have `<Click to see difference>`
- framework class `src/Framework/TestFailure.php`  became  independed from `PHPTAssertionFailedError` 
- `StringMatchesFormatDescription` works with `ComparisonFailure`  as all others.
 
Cons: 
- More changes

## Second attempt result:
image

### Questions to maintainers: 
* Please, let me know which solution is okay for PHPUnit, the First or the Second.
* Please, let me know whether I should do additional patches? And for which branches? This is 9.5 branch. Are patches in 8.5 enough? Or master also is required? Sorry, it is my very first PR in this repo, so I can't figure out whether this pr is a bugfix or a  new feature

Thanks!

### DIFF
--- a/src/Framework/Constraint/String/StringMatchesFormatDescription.php
+++ b/src/Framework/Constraint/String/StringMatchesFormatDescription.php
@@ -16,8 +16,7 @@ use function preg_match;
 use function preg_quote;
 use function preg_replace;
 use function strtr;
-use SebastianBergmann\Diff\Differ;
-use SebastianBergmann\Diff\Output\UnifiedDiffOutputBuilder;
+use SebastianBergmann\Comparator\ComparisonFailure;
 
 /**
  * @no-named-arguments Parameter names are not covered by the backward compatibility promise for PHPUnit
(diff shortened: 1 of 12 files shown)


benbor:
@sebastianbergmann Hi! 
May I wonder, do I violated some contribution guidelines or everything is ok from my side, and I just need wait some time? 
Thanks!

PR #4833, opened 2021-11-28
Fix: Unwrap data set name

localheinz:
This pull request

- [x] asserts that the name of `TestMethod` is assembled as expected
- [x] unwraps the data set name before passing it to a condition and `sprintf()`


sebastianbergmann:
Thanks!

localheinz:

PR #4838, opened 2021-12-12
Use spl_object_id instead of spl_object_hash where possible.

TysonAndre:
spl_object_id was added in php 7.2.
array_push allows calling with only one parameter since php 7.3.
Modify array in place rather than copying all values to new arrays
repeatedly.

Integers are slightly faster to compare than strings and use less
memory.

This in particular isn't performance sensitive, but update this as an
example for future additions or code based on it.


sebastianbergmann:
Thanks!

PR #4857, opened 2022-01-08
Fix access to optional array key when using `debug_backtrace()` function

ste93cry:
The `class`, `file` and `line` array keys are optional for a stacktrace frame, so to silence PHPStan when the signature of the `debug_backtrace()` function will be narrowed down to an array shape with optional keys these little fixes are needed


ste93cry:
@sebastianbergmann sorry to be "that guy", but since this PR is required to let the build in `phpstan/phpstan-src` pass, do you have any ETA on when the next `9.5` version containing it will be released?

sebastianbergmann:
I just released PHPUnit 9.5.12.

PR #4864, opened 2022-01-14
Patch xdebug_info

theofidry:
Here's the patch we discussed about. I'm not certain where to look at to check that it's working correctly though

### DIFF
--- a/build/config/php-scoper.php
+++ b/build/config/php-scoper.php
@@ -8,8 +8,21 @@
  * file that was distributed with this source code.
  */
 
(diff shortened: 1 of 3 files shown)


sebastianbergmann:
Looks like it does not work, as I still see `use function PHPUnit\xdebug_info;` in the scoped PHAR.

theofidry:
@sebastianbergmann should be fine now

sebastianbergmann:
Thank you so much!

PR #4865, opened 2022-01-15
update deprecation information

dbu:
Following the discussion in #3390

Imho there is no use in having `@deprecated` annotation if the new system is not available, as people can not avoid using this interface. And to my understanding, for version 8 and 9, implementing this interface is the way to go.

### DIFF
--- a/src/Framework/TestListener.php
+++ b/src/Framework/TestListener.php
@@ -12,49 +12,37 @@ namespace PHPUnit\Framework;
 use Throwable;
 
 /**
- * @deprecated Use the `TestHook` interfaces instead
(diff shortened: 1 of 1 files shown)


sebastianbergmann:

dbu:
that was my understanding of the discussion in #3390. 

do you not want to address the problem that people are mislead by deprecation notices and try to migrate to a thing that is not fully working and will be replaced in phpunit 10 anyways? or did you have a different solution in mind?

sebastianbergmann:
I hopefully addressed this in 9b5a2e5cc9d615ab9b1418753c42a2f8c0d5285e just now. Thank you for bringing this to my attention again.

dbu:
thanks!

PR #4870, opened 2022-01-21
Allow including the PHPUnit phar from Composer bin proxies

Seldaek:
I also added a hint in the output when execution is not allowed, because this was pretty confusing :)

Fixes https://github.com/composer/composer/issues/10471 where someone tried to package up the PHPUnit phar in a package and declares it as a Composer binary.

### DIFF
--- a/build/templates/binary-phar-autoload.php.in
+++ b/build/templates/binary-phar-autoload.php.in
@@ -46,7 +46,8 @@ foreach (['dom', 'json', 'libxml', 'mbstring', 'tokenizer', 'xml', 'xmlwriter'] 
     die(1);
 }
 
-if (__FILE__ === realpath($_SERVER['SCRIPT_NAME'])) {
+// allowing execution if the file is executed directly or included by a Composer bin proxy (which sets the _composer_bin_dir global)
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
This breaks existing tests, see [here](https://github.com/sebastianbergmann/phpunit/runs/4898840802?check_suite_focus=true#step:5:264).

Seldaek:
Yes, the question is whether this is an acceptable change to you? I think what breaks is the extra output, which is not strictly needed. But if you think it makes sense I will happily update the tests.

sebastianbergmann:
The extra output needs to go, I think. That being said, I am reluctant to merge this as I do not think what is described in https://github.com/composer/composer/issues/10471 is a good idea / should be supported.

Seldaek:
Yes I also find it a rather strange usage pattern but then again I'm not sure why the phar prevents being used as a library either. Up to you though, I don't have a strong agenda either way here :)

sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it because I do not think that this use case should be supported.

PR #4873, opened 2022-01-23
Use predefined PHP float epsilon

marijnvanwezel:
Instead of using an arbitrary epsilon, use the predefined PHP constant PHP_FLOAT_EPSILON, which is defined to be the smallest representable positive number "x", such that "x + 1.0 != 1.0".

See https://www.php.net/manual/en/reserved.constants.php#constant.php-float-epsilon for additional information.

### DIFF
--- a/src/Framework/Constraint/IsIdentical.php
+++ b/src/Framework/Constraint/IsIdentical.php
@@ -27,11 +27,6 @@ use SebastianBergmann\Comparator\ComparisonFailure;
 final class IsIdentical extends Constraint
 {
     /**
-     * @var float
-     */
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
Can you send this against `8.5`, please? Thanks!

marijnvanwezel:
@sebastianbergmann Thanks, I'll make a new PR for this, since this one is based on `marijnvanwezel/9.5` and it would be confusing to merge it into 8.5. I should have used a separate branch to create this pull request, but I didn't.

marijnvanwezel:
Closed in favour of #4874 (see comment above).

PR #4874, opened 2022-01-24
Use predefined PHP float epsilon

marijnvanwezel:
Instead of using an arbitrary epsilon, use the predefined PHP constant PHP_FLOAT_EPSILON, which is defined to be the smallest representable positive number "x", such that "x + 1.0 != 1.0".

See https://www.php.net/manual/en/reserved.constants.php#constant.php-float-epsilon for additional information.

### DIFF
--- a/src/Framework/Constraint/IsIdentical.php
+++ b/src/Framework/Constraint/IsIdentical.php
@@ -35,11 +35,6 @@ use SebastianBergmann\Comparator\ComparisonFailure;
 final class IsIdentical extends Constraint
 {
     /**
-     * @var float
-     */
-    private const EPSILON = 0.0000000001;
-
-    /**
      * @var mixed
      */
     private $value;
@@ -67,7 +62,7 @@ final class IsIdentical extends Constraint
         if (is_float($this->value) && is_float($other) &&
             !is_infinite($this->value) && !is_infinite($other) &&
             !is_nan($this->value) && !is_nan($other)) {
-            $success = abs($this->value - $other) < self::EPSILON;
+            $success = abs($this->value - $other) < PHP_FLOAT_EPSILON;
         } else {
             $success = $this->value === $other;
         }


sebastianbergmann:
Thanks!

src/Framework/Constraint/IsIdentical.php
```diff
@@ -67,7 +62,7 @@ public function evaluate($other, string $description = '', bool $returnResult =
         if (is_float($this->value) && is_float($other) &&
             !is_infinite($this->value) && !is_infinite($other) &&
             !is_nan($this->value) && !is_nan($other)) {
-            $success = abs($this->value - $other) < self::EPSILON;
+            $success = abs($this->value - $other) < PHP_FLOAT_EPSILON;
```
localheinz, line 65: Nice, @marijnvanwezel!

alfredbez:
In case someone stumbles upon this: This was breaking one assertion in our project which looked like this:
```php
self::assertSame(1.2345, (12.345/10));
```
Here's the failure message:
```
Failed asserting that 1.2345000000000002 is identical to 1.2345.
```

Quote from @sebastianbergmann:

> `assertEquals()` should not be used for floating point values either, at least not without its optional `$delta` parameter (in case the PHPUnit version that is used still has that). This optional parameter is deprecated in current versions and will be removed in the future. `assertEqualsWithDelta()` should be used in its stead.

([source](https://github.com/sebastianbergmann/phpunit/issues/3159#issuecomment-528908608), take also a look at [the docs](https://phpunit.readthedocs.io/en/stable/assertions.html#assertequalswithdelta))

Changing this like so fixed the problem:
```php
self::assertEqualsWithDelta(1.2345, (12.345/10), 0.0001);
```

Another way would be to convert that to a string using `sprintf`, depends on the use-case of course:
```php
self::assertSame('1.2345', sprintf('%.4f',12.345/10));
```

PR #4886, opened 2022-02-18
Fix: Invoke method as instance method

localheinz:
This pull request

- [x] invokes instance methods as instance methods


localheinz:

PR #4888, opened 2022-02-19
Fix: Do not declare test classes in global namespace

localheinz:
This pull request

- [x] stops declaring test classes in the global namespace

💁‍♂️ This is in preparation for moving test fixtures into the `PHPUnit\TestFixture` namespace.

### DIFF
--- a/tests/unit/Framework/Constraint/LogicalXorTest.php
+++ b/tests/unit/Framework/Constraint/LogicalXorTest.php
@@ -7,12 +7,10 @@
  * For the full copyright and license information, please view the LICENSE
  * file that was distributed with this source code.
  */
-namespace Framework\Constraint;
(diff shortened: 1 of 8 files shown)


sebastianbergmann:
Thanks! I cannot merge this to `9.5`, though. That means that you/we need to check `9.5` for similar problems.

localheinz:

Will follow up with a separate pull request!

sebastianbergmann:
I was able to merge most of the changes (see 30bb6523b213295de105dc33b8d7082c26c020f5 for `9.5` and 4b1edef9386f73e0c20852a35527d7d8cc9f20d6 for `master`).

Thank you for working on this, Andreas!

localheinz:
Any time, @sebastianbergmann!

PR #4890, opened 2022-02-19
Fix: Do not declare test classes in global namespace

localheinz:
This pull request

- [x] stops declaring test classes in the global namespace

Follows https://github.com/sebastianbergmann/phpunit/pull/4888#issuecomment-1046023847.


localheinz:

PR #4891, opened 2022-02-20
Fix: Enable `ext-curl`

localheinz:
This pull request

- [x] enables `ext-curl`

💁‍♂️ See https://github.com/sebastianbergmann/phpunit/runs/5263718117?check_suite_focus=true#step:5:7.


localheinz:

PR #4894, opened 2022-02-23
Declare generic types for Iterators

staabm:


src/Framework/TestSuite.php
```diff
@@ -513,6 +514,10 @@ public function run(TestResult $result): void
                     $message = "Exception in {$this->name}::{$afterClassMethod}" . PHP_EOL . $t->getMessage();
                     $error   = new SyntheticError($message, 0, $t->getFile(), $t->getLine(), $t->getTrace());
 
+                    if ($test === null) {
```
staabm, line 517: with the generics psalm thinks `$test` can be null at this point.
as the logic would hard error on null values it seems this cannot happen.

just added this if-case so psalm also knows `$test` cannot be null.
never contributed to phpunit.. I guess this should be done differently, but I don't know how.. feedback welcome.

PR #4900, opened 2022-03-01
fix: colors help

shaunthegeek:
colors help is wrong, it notice: no need equal sign, but actually need.

```shell
./vendor/bin/phpunit --help
PHPUnit 9.5.14 by Sebastian Bergmann and contributors.

  --coverage-php <file>       Export PHP_CodeCoverage object to file
  --coverage-text=<file>      Generate code coverage report in text format [default: standard output]

  --colors <flag>             Use colors in output ("never", "auto" or "always")
```

```shell
XDEBUG_MODE=coverage php vendor/bin/phpunit --coverage-text --colors never
```

> Cannot open file "never".

```shell
XDEBUG_MODE=coverage php vendor/bin/phpunit --coverage-text --colors=never
```

> OK (2 tests, 2 assertions)

### DIFF
--- a/src/TextUI/Help.php
+++ b/src/TextUI/Help.php
@@ -88,7 +88,7 @@ final class Help
             ['arg'    => '--static-backup', 'desc' => 'Backup and restore static properties for each test'],
             ['spacer' => ''],
 
-            ['arg'    => '--colors <flag>', 'desc' => 'Use colors in output ("never", "auto" or "always")'],
+            ['arg'    => '--colors=<flag>', 'desc' => 'Use colors in output ("never", "auto" or "always")'],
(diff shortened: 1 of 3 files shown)


sebastianbergmann:
Please send this against `8.5`. However, do note that `--colors` without any option is valid, too.

sinkcup:
`8.5` is right, but `9.5` and `master` is wrong, maybe need cherry-pick?

https://github.com/sebastianbergmann/phpunit/blob/e93feeab376fef7dc81cbaa48273f61df991849f/src/TextUI/Help.php#L87

https://github.com/sebastianbergmann/phpunit/blob/5ff8c545a50226c569310a35f4fa89d79f1ddfdc/src/TextUI/Help.php#L93

PR #4901, opened 2022-03-03
Test associative arrays comparison with canonicalization

benbankes:
One of these assertions will likely fail, resulting in a bug report.

### DIFF
--- a/tests/unit/Framework/AssertTest.php
+++ b/tests/unit/Framework/AssertTest.php
@@ -1507,6 +1507,13 @@ XML;
         $this->assertNotEqualsCanonicalizing([3, 2, 1], [2, 3, 4]);
     }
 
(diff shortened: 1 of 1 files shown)


benbankes:
I was relying on the ability to run the tests in CI/CD.  I guess I'll have to fire up a PHP 8 environment, or hope that someone else values this use case enough to get these tests running.

sebastianbergmann:
I am sorry, but I do not understand what you're trying to achieve with this pull request.

benbankes:
@sebastianbergmann 
Basically, when two associative arrays have different keys, I expect to be able to assert that they are not equal, with canonicalizing, but this is not the way PHPUnit currently behaves.

Hopefully the [bug report](https://github.com/sebastianbergmann/phpunit/issues/4902) I just added is clear, or at least repetitious enough to adequately communicate my perspective.

sebastianbergmann:

PR #4905, opened 2022-03-05
GH Actions: version update for various predefined actions

jrfnl:
A number of predefined actions have had major releases, which warrant an update to the workflow(s).

These updates don't actually contain any changed functionality, they are mostly just a change of the Node version used by the action itself (from Node 14 to Node 16).

Refs:
* https://github.com/actions/checkout/releases
* https://github.com/actions/download-artifact/releases
* https://github.com/actions/upload-artifact/releases

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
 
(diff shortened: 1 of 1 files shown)


jrfnl:
Note: The `setup-java` action could also do with an update, but as that would be skipping from `v1` to `v3`, I think someone who knows the details of the use of Java in the workflow should probably have a look at the changelog before doing that.

https://github.com/actions/setup-java/releases

PR #4907, opened 2022-03-06
Fix: Use `PHPUnit\TestFixture` as namespace for test fixtures

localheinz:
This pull request

- [x] uses `PHPUnit\TestFixture` as namespace for test fixtures

PR #4910, opened 2022-03-07
Fix: Remove conditions from test

localheinz:
This pull request

- [x] removes conditions from a test and shifts the responsibility of assembling expected data to the data provider

💁‍♂️ This will simplify the changes required for #4907 a lot (also better to understand what is going on and why)!


localheinz:

PR #4914, opened 2022-03-07
Fix: Import

localheinz:
This pull request

- [x] imports a class

### DIFF
--- a/tests/unit/Util/TestClassTest.php
+++ b/tests/unit/Util/TestClassTest.php
@@ -28,6 +28,7 @@ use ReflectionMethod;
 use RequirementsClassDocBlockTest;
 use RequirementsTest;
 use Test3194;
+use VariousDocblockDefinedDataProvider;
 use VariousIterableDataProviderTest;
 
 /**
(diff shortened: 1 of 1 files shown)


tests/unit/Util/TestClassTest.php
```diff
@@ -1024,7 +1025,7 @@ public function testWithDuplicateKeyDataProviders(): void
     public function testTestWithEmptyAnnotation(): void
     {
         $result = DocBlock::ofMethod(new ReflectionMethod(
-            \VariousDocblockDefinedDataProvider::class,
+            VariousDocblockDefinedDataProvider::class,
             'anotherAnnotation'
         ), VariousDocblockDefinedDataProvider::class)->getProvidedData();
```
localheinz, line 1030: Not sure why this class has not been imported before.

sebastianbergmann:
Merged, thanks. However, I cannot merged this from `8.5` to `9.5` without conflicts. Can you send a PR for `9.5`? Thanks!

localheinz:
Thank you, @sebastianbergmann - follow up is coming in a moment!

localheinz:
@sebastianbergmann 

Fix is not necessary for `9.5`, already applied in one way or another!

PR #4916, opened 2022-03-08
Fix: Use class keyword

localheinz:
This pull request

- [x] makes use of the class keyword where possible

Somewhat related to #4907.
Follows https://github.com/sebastianbergmann/phpunit/pull/4912#issuecomment-1061446659.

PR #4917, opened 2022-03-08
Fix: Use class keyword

localheinz:
This pull request

- [x] makes use of the class keyword where possible

Somewhat related to #4907.
Follows https://github.com/sebastianbergmann/phpunit/pull/4916#issuecomment-1061512453.


localheinz:

PR #4921, opened 2022-03-09
Fix: Remove unnecessary `require_once`

localheinz:
This pull request

- [x] removes unnecessary `require_once` statements

Follows https://github.com/sebastianbergmann/phpunit/pull/4920#issuecomment-1062587292.

PR #4922, opened 2022-03-09
Fix: Use `PHPUnit\TestFixture` as namespace for test fixtures

localheinz:
This pull request

- [x] uses `PHPUnit\TestFixture` as namespace for test fixtures

Spotted in #4921.


localheinz:

PR #4943, opened 2022-03-29
[XSD schema] Add force attribute to env tag

dmitryuk:
From issue https://github.com/sebastianbergmann/phpunit/issues/2353 it is possible to set `force` boolean attribute to `env` tag like
```
 <env name="APP_ENV" value="test" force="true" />
```
The PR adds the hint to Xsd schema to fix XML validator error (PHPStorm IDE for example)


sebastianbergmann:
Can you please send this against `8.5`? Thanks!

dmitryuk:
[d](https://github.com/sebastianbergmann/phpunit/pull/4944)

PR #4951, opened 2022-04-06
Fix `atMost(...)` invocation rule throws error on zero invocation

### LINKED ISSUE #4950: False error on `atMost(...)` invocation rule without call
| Q                   | A
| PHPUnit version     | 9.5.20 & 8.5.26
| PHP version         | 8.1.4
| Installation Method | PHAR

#### Summary

When using the `atMost(...)` invocation rule, and the UUT does not invoke the mocked function, then I get the following error:
```
Expectation failed for method name is "mockedMethodName" when invoked at most 1 times.
Mocked method does not exist.
```

#### Current behavior

Test fails with error, when mocked method never called with `atMost(...)` invocation rule.

#### How to reproduce

1. Mock a class.
2. Make an invocation rule with `atMost(1)` for example, but do not call the mocked method from the tested unit.
3. Run the test.

My example call:
```PHP
$this->storesRepository->expects(self::atMost(1))
  ->method('findOneBy')
  ->with(['name' => $storeName])
  ->willReturn($storeEntityWithSameName);
```

#### Expected behavior

Test passes as zero calls is smaller then the value passed to the `atMost(...)` invocation rule.

T-bond:
Closes #4950

### DIFF
--- a/src/Framework/MockObject/Matcher.php
+++ b/src/Framework/MockObject/Matcher.php
@@ -17,6 +17,7 @@ use PHPUnit\Framework\ExpectationFailedException;
 use PHPUnit\Framework\MockObject\Rule\AnyInvokedCount;
 use PHPUnit\Framework\MockObject\Rule\AnyParameters;
 use PHPUnit\Framework\MockObject\Rule\InvocationOrder;
+use PHPUnit\Framework\MockObject\Rule\InvokedAtMostCount;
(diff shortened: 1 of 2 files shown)


T-bond:
Self review done.

I hope I could help, and thanks for this great project. Tell me if you do need anything else from me. :)

PR #4964, opened 2022-04-26
Enable named data sets with the `#[TestWith*]` attributes

HypeMC:
Adds an optional data set `name` argument to the `#[TestWith]` & `#[TestWithJson]` attributes:

```php
#[TestWith(['a', 'b'], 'Name1')]
#[TestWith(['c', 'd'], 'Name2')]
public function testFoo(string $one, string $two)
{}
```

### DIFF
--- a/src/Framework/Attributes/TestWith.php
+++ b/src/Framework/Attributes/TestWith.php
@@ -20,14 +20,21 @@ use Attribute;
 final readonly class TestWith
 {
     private array $data;
+    private ?string $name;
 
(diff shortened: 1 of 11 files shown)


HypeMC:
@sebastianbergmann Hi, any thoughts on this one? Judging by the few likes I got, there seems to be at least some interest in this feature. If there's anything else I can do here, please let me know.

sebastianbergmann:
Merged manually, thanks!

Issue #4966, opened 2022-05-07
`TestCase::assertSame` (and related exact comparisons) must compare `float` exactly


mvorisek:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 9.5.10
| PHP version         | 7.4
| Installation Method | Composer

#### Summary

Currently test case like:

```
 $this->assertSame(8.20234376757473, 8.20234376757474);
```

pass even if the numbers are different. They are really stored differently, test case:

```
$this->assertSame(bin2hex(pack('e', 8.20234376757473)), bin2hex(pack('e', 8.20234376757474)));
```

fails correctly.

It seems phpunit converts float numbers using default php float to string conversion which depends on php.ini `precision` configuration, test case like:

```
ini_set('precision', '20');
$this->assertSame(8.20234376757473, 8.20234376757474);
```

fails as expected.

Also, the float diff output has currently limited precision.

To fix exact comparisons & diff output, I propose to convert float numbers to string using function like:

```
/**
 * Cast float to string with lossless precision.
 */
final public static function castFloatToString(float $value): string
{
    $precisionBackup = ini_get('precision');
    ini_set('precision', '-1');
    try {
        return (string) $value;
    } finally {
        ini_set('precision', $precisionBackup);
    }
}
```

sebastianbergmann:
Feel free to send a pull request for the `8.5` branch with your proposed solution.

By the way, your example works for me with `precision=14`:

```
$ cat Test.php                             
<?php declare(strict_types=1);
use PHPUnit\Framework\TestCase;

final class Test extends TestCase
{
    public function testOne(): void
    {
        $this->assertSame(8.20234376757473, 8.20234376757474);
    }
}
```
```
$ phpunit Test.php
PHPUnit 9.5.20-35-gf53bae505 #StandWithUkraine

F                                                                   1 / 1 (100%)

Time: 00:00.037, Memory: 6.00 MB

There was 1 failure:

1) Test::testOne
Failed asserting that 8.20234376757474 is identical to 8.20234376757473.

/home/sb/Test.php:8

FAILURES!
Tests: 1, Assertions: 1, Failures: 1.
```
```
$ php -v               
PHP 8.1.5 (cli) (built: Apr 12 2022 16:19:58) (NTS gcc x86_64)
Copyright (c) The PHP Group
Zend Engine v4.1.5, Copyright (c) Zend Technologies
```
```
$ php -i | grep precision
precision => 14 => 14
[...]

PR #4972, opened 2022-05-21
Drop comparison /w `PHP_FLOAT_EPSILON`

mvorisek:
fixes https://github.com/sebastianbergmann/phpunit/issues/4966

The removed code is residual from https://github.com/sebastianbergmann/phpunit/commit/136fae39306c8e6cda3fb5ec8fcc0ff8cff8cf4e and even a difference by `PHP_FLOAT_EPSILON` should be reported as a failure.

The `PHP_FLOAT_EPSILON` constant equals to `2.220446049250313E-16` (see 3v4l link in a post below) and is usable only when the `abs(a - b)` is done on normalized values, but when done so, the comparison is useless or satisfied because of the normalization rounding error solely. On the other side, imagine comparing very small numbers like `1e-100` and `1e-200`. The difference between these two normalized values is astronomically large, but currently the phpunit handles such numbers as same!


mvorisek:
marking as draft,
```
/**
 * Cast float to string with lossless precision.
 */
final public static function castFloatToString(float $value): string
{
    $precisionBackup = ini_get('precision');
    ini_set('precision', '-1');
    try {
        return (string) $value;
    } finally {
        ini_set('precision', $precisionBackup);
    }
}
```

must be put everywhere where float to string is casted (for print, for weak comparison, ...)

mvorisek:
@sebastianbergmann any idea why
```
1) PHPUnit\Framework\AssertTest::testAssertSameSucceeds with data set #4 (0.33333333333333, 0.33333333333333)
Failed asserting that 0.33333333333333 is identical to 0.33333333333333.
```
is output?

It seems is come from https://github.com/sebastianbergmann/exporter/blob/3.1/src/Exporter.php#L305 but used `var_export` should output the full precision - https://3v4l.org/KDZeJ

mvorisek:
PR is done

PR #4973, opened 2022-05-23
Enhancement: Use `WyriHaximus/github-action-composer-php-versions-in-range` to determine supported PHP version matrix

localheinz:
This pull request

- [x] uses [`WyriHaximus/github-action-composer-php-versions-in-range`](https://github.com/WyriHaximus/github-action-composer-php-versions-in-range) to determine the supported PHP version matrix

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,6 +30,24 @@ jobs:
       - name: Run friendsofphp/php-cs-fixer
         run: ./tools/php-cs-fixer fix --dry-run --show-progress=dots --using-cache=no --verbose
 
+  supported-php-version-matrix:
+    name: Supported PHP Version Matrix
+
+    runs-on: ubuntu-latest
+
+    outputs:
+      php-version: ${{ steps.supported-php-version-matrix.outputs.version }}
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Determine supported PHP version matrix
+        id: supported-php-version-matrix
+        uses: WyriHaximus/github-action-composer-php-versions-in-range@v1
+        with:
+          upcomingReleases: true
+
   type-checker:
     name: Type Checker
 
@@ -59,6 +77,9 @@ jobs:
   unit-tests:
     name: Unit Tests
 
+    needs:
+      - supported-php-version-matrix
+
     runs-on: ${{ matrix.os }}
 
     env:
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
If this works for us, then we should start with it in `8.5` and merge to `9.5` and `main` from there.

.github/workflows/ci.yml
```diff
@@ -306,10 +317,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php-version:
-          - "7.2"
-          - "7.3"
-          - "7.4"
+        php-version: ${{ fromJson(needs.supported-versions-matrix.outputs.version) }}
```
localheinz, line 320: @sebastianbergmann 

Looks like we will be running additional jobs here, for

- "8.0
- "8.1"
- "8.2"

.github/workflows/ci.yml
```diff
@@ -361,10 +369,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php-version:
-          - "8.0"
-          - "8.1"
-          - "8.2"
+        php-version: ${{ fromJson(needs.supported-versions-matrix.outputs.version) }}
```
localheinz, line 372: @sebastianbergmann 

Looks like we will be running additional jobs here, for

- "7.2
- "7.3"
- "7.4"

localheinz:
@sebastianbergmann 

Looks like this is blocked by https://github.com/WyriHaximus/github-action-composer-php-versions-in-range/pull/5.

WyriHaximus:
@localheinz @sebastianbergmann Just tagged [`v1.5.0`](https://github.com/WyriHaximus/github-action-composer-php-versions-in-range/releases/tag/v1.5.0) on that action to unblock this PR. Since you're using `@v1` that new version will be used on the next run.

WyriHaximus:
No, thank you @localheinz. Until a few weeks ago I never shared about this project, so having someone pick it up in their projects is exciting. Someone proposing it to a high profile project like this is very exciting. Thank you ❤️ .

localheinz:
@sebastianbergmann suggested to take a look, @WyriHaximus, so that's why I opened a pull request!

🤓

sebastianbergmann:

WyriHaximus:
For my own insights, what made you decide not to merge this?

sebastianbergmann:
On the `8.5` branch we have the jobs `test-phar` and `test-phar-without-code-coverage` that require different version ranges than the other jobs:

grafik

We could use your action for the other jobs, of course, but that would be inconsistent. That is why we decided against using it, at least for now.

WyriHaximus:
Thank you for sharing. And I fully agree, that action was built with keeping things simple in mind.

PR #4977, opened 2022-05-27
Fix: Reference `main` instead of `master`

localheinz:
This pull request

- [x] references `main` instead of `master` in pull request templates


localheinz:

localheinz:
@sebastianbergmann 

Looks like I missed one occurrence:

https://github.com/sebastianbergmann/phpunit/blob/afc53191f373a451e176cdaa1affe0aac8bc9ee8/build/test-extension/manifest.xml#L13

PR #5001, opened 2022-07-04
Regular callable call

MathiasReker:
Callables must be called without using call_user_func* when possible.


sebastianbergmann:

PR #5002, opened 2022-07-04
Assign null coalescing to coalesce equal

MathiasReker:
Use the null coalescing assignment operator ??= where possible.


sebastianbergmann:

PR #5012, opened 2022-07-08
Fix memory leak in ExceptionWrapper

mvorisek:
Currently `ExceptionWrapper` stores the original exception forever and prevents it to be GCed.

This is causing not only memory to grow/leak, but also if the original exception has a reference to an object, the object is not released as well. Such scenario is not uncommon, as exception stack trace/frame can contain object arguments.

In consequence when the unreleased object is a some limited resource like DB connection, with a lot of wrapped/unreleased exceptions, the resource can be exhausted by a few failing tests making the other tests fail because of them.

This fix maintains the shallow backtrace as WeakReference property is not expanded during backtrace dump.

No BC break introduced.

### DIFF
--- a/src/Framework/ExceptionWrapper.php
+++ b/src/Framework/ExceptionWrapper.php
@@ -9,11 +9,13 @@
  */
 namespace PHPUnit\Framework;
 
+use const PHP_VERSION_ID;
 use function array_keys;
 use function get_class;
 use function spl_object_hash;
 use PHPUnit\Util\Filter;
 use Throwable;
+use WeakReference;
 
 /**
  * Wraps Exceptions thrown by code under test.
@@ -37,6 +39,11 @@ final class ExceptionWrapper extends Exception
      * @var null|ExceptionWrapper
      */
     protected $previous;
+
+    /**
+     * @var null|WeakReference<Throwable>
+     */
+    private $originalException;
 
     public function __construct(Throwable $t)
     {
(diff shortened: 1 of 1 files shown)


src/Framework/ExceptionWrapper.php
```diff
@@ -109,14 +116,23 @@ public function getOriginalException(): ?Throwable
      */
     private function originalException(Throwable $exceptionToStore = null): ?Throwable
     {
-        static $originalExceptions;
+        // drop once PHP 7.3 support is removed
+        if (PHP_VERSION_ID < 70400) {
```
mvorisek, line 120: drop this code when merging to phpunit 10.x

mvorisek:
@sebastianbergmann can this be merged and released?

mvorisek:
I have updated the PR description to understand the real issue better. Please merge this fix into at least phpunit 9.5.

For next major release, is there any reason to not drop `ExceptionWrapper` completely?

sebastianbergmann:
> For next major release, is there any reason to not drop `ExceptionWrapper` completely?

The `ExceptionWrapper` has been removed in the `main` branch.

mvorisek:
> The `ExceptionWrapper` has been removed in the `main` branch.

Thank you ❤️ 

To fix the memory leaks in phpunit 9.5, can this PR be merged into 9.5.x branch? No BC break should be involved.

PR #5015, opened 2022-07-14
Update Ukraine word with the black foreground color

maurobonfietti:
Hi Sebastian :wave: Hi PHPUnit contributors:

**Summary**

The message `#StandWithUkraine` shows the Ukraine word with white text on a yellow background (when it uses a terminal with black background).

**Example of Current Behavior**

Screen Shot 2022-07-13 at 21 02 42

**Example of Expected Behavior**

Screen Shot 2022-07-13 at 21 03 46

With these minor changes could use similar colors as Packagist/Composer.

### DIFF
--- a/src/TextUI/TestRunner.php
+++ b/src/TextUI/TestRunner.php
@@ -325,7 +325,7 @@ final class TestRunner extends BaseTestRunner
                 Version::id() .
                 ' ' .
                 Color::colorize('bg-blue,fg-white', '#StandWith') .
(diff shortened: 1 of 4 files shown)


thomas-a-smith:
I can't believe you're spending development time on this kind of thing.

ppp0:
I would prefer if we keep (constantly changing) political issues out of the code. While I understand the urge, there's a few current affairs besides Ukraine to care about

Issue #5020, opened 2022-07-19
PHPUnit 9.x breaks loading of PSR-0/PEAR style classes


totten:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 9.5.21
| PHP version         | 7.4.26
| Installation Method | PHAR

#### Summary

PHPUnit 9.x breaks support for loading classes that use `_`-based naming.

For example (https://github.com/totten/phpunit-underscore-namespace), if you have a class `CRM_Utils_String`, then you would want a corresponding test-class `CRM_Utils_StringTest`. 

As is typical in PSR-0 and earlier PEAR-based conventions, the underscores are mapped to folders. The example project has typical filenames like:

* `CRM/Utils/String.php`
* `tests/phpunit/CRM/Utils/StringTest.php`

#### Current behavior

| Command | Result |
| -- | -- |
| `phpunit9` | Runs without error or warning |
| `phpunit9 path/to/test/folder/` | Runs with warning |
| `phpunit9 path/to/test/file.php` | Fails |

Full output reproduced at https://github.com/totten/phpunit-underscore-namespace#running-with-phpunit-9

#### How to reproduce

Checkout the example https://github.com/totten/phpunit-underscore-namespace

#### Expected behavior

| Command | Result |
| -- | -- |
| `phpunit9` | Runs without error or warning |
| `phpunit9 path/to/test/folder/` | Runs without error or warning |
| `phpunit9 path/to/test/file.php` | Runs without error or warning |

You can similar/expected output when running PHPUnit 8.x, eg https://github.com/totten/phpunit-underscore-namespace#running-with-phpunit-8

#### Discussion

The changelog (https://phpunit.de/announcements/phpunit-9.html) reports some known breaking changes:

* (Removed) Support for using class name as CLI argument
* (Removed) Support for declaring more that one test case class in test source file

However, it's important to note that the example here does not implicate either of these:

* The tests are executed based on file-names (not class-names).
* There is only one test-class in each source file. The file-names are consistent with well-known conventions (supported by PSR-0 and earlier PEAR conventions).

PR #5021, opened 2022-07-20
Restore support for PSR-0-style test-classes (#5020)

totten:
As per #5020, this PR fixes loading of test-classes which have underscore-based namespacing (e.g. loading `CRM_Utils_StringTest` in file `tests/phpunit/CRM/Utils/StringTest.php` from example repo https://github.com/totten/phpunit-underscore-namespace). The PR also adds an end-to-end regression test.

Note: The patch for `main` is very slightly different (and slightly shorter). I've only opened the PR for 9.x (since that's the older version with the bug), but you can see `main` patches at https://github.com/totten/phpunit/commits/main-psr0

### DIFF
--- a/src/Runner/StandardTestSuiteLoader.php
+++ b/src/Runner/StandardTestSuiteLoader.php
@@ -61,6 +61,12 @@ final class StandardTestSuiteLoader implements TestSuiteLoader
 
                     break;
                 }
+
+                if (stripos(substr($loadedClass, $offset - 1), '_' . $suiteClassName) === 0) {
+                    $suiteClassName = $loadedClass;
+
+                    break;
+                }
             }
         }
 
--- a/src/Runner/StandardTestSuiteLoader.php
+++ b/src/Runner/StandardTestSuiteLoader.php
@@ -52,16 +52,18 @@ final class StandardTestSuiteLoader implements TestSuiteLoader
         }
 
         if (!class_exists($suiteClassName, false)) {
-            // this block will handle namespaced classes
+            // Perhaps this file defines a class inside a namespace? Let's check...
             $offset = 0 - strlen($suiteClassName);
 
             foreach ($loadedClasses as $loadedClass) {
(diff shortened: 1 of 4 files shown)


sebastianbergmann:
@realFlowControl Do you have feedback on this? Thanks!

src/Runner/StandardTestSuiteLoader.php
```diff
@@ -61,6 +61,12 @@ public function load(string $suiteClassFile): ReflectionClass
 
                     break;
                 }
+
+                if (stripos(substr($loadedClass, $offset - 1), '_' . $suiteClassName) === 0) {
```
realFlowControl, line 65:
```suggestion
                // Support for PSR-0 namespaces (#5020)
                if (stripos(substr($loadedClass, $offset - 1), '_' . $suiteClassName) === 0) {
```

I would suggest adding a comment here, otherwise (at least speaking for myself) we will forget what this is for :wink:

realFlowControl (commented):
Looks good to me.

I don't know if we actually _need_ my suggested comment, but I would at least help the future @realFlowControl :wink:

src/Runner/StandardTestSuiteLoader.php
totten: Yep, good idea. I pushed up parallel comments (7b49c195641e994f9d7ba5b21c12dccfdde8af56) for the parts that detect `\` and `_`.

totten:
@sebastianbergmann - Comments were updated following @realFlowControl's suggestion. (If it helps, I also added comments on the [main-psr0](https://github.com/totten/phpunit/commits/main-psr0) variant of this branch.)

It should be OK to re-run the test?

sebastianbergmann:
Thank you! I have merged this into the `9.5` branch.

However, the patch cannot simply be merged into `main` for PHPUnit 10. I will accept a pull request for `main` that restores support for PEAR-style test-classes, but I will not work on this myself.

totten:
Thank you @sebastianbergmann! I've submitted the port as #5030.

PR #5030, opened 2022-08-12
Restore support for PSR-0-style test-classes (main/10.x)

totten:
This is a forward-port of #5021 from `9.5` to `main`/`10.x`.


sebastianbergmann:
Thank you!

PR #5060, opened 2022-09-24
Enhancement: Enable and configure `native_function_invocation` fixer

localheinz:
This pull request

- [x] enables and configures the `native_function_invocation` fixer
- [x] runs `tools/php-cs-fixer fix`

💁‍♂️ For reference, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v3.11.0/doc/rules/function_notation/native_function_invocation.rst.

### DIFF
--- a/tests/end-to-end/regression/4663/Issue4663Test.php
+++ b/tests/end-to-end/regression/4663/Issue4663Test.php
@@ -15,6 +15,6 @@ final class Issue4663Test extends TestCase
     {
         $this->expectError();
 
-        implode('', [new stdClass]);
+        \implode('', [new stdClass]);
     }
 }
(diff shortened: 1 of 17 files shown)


tests/end-to-end/regression/4663/Issue4663Test.php
```diff
@@ -15,6 +15,6 @@ public function testTestThatExpectsAnErrorPassesWhenTheErrorOccurs(): void
     {
         $this->expectError();
 
-        implode('', [new stdClass]);
+        \implode('', [new stdClass]);
```
localheinz, line 18: Not sure if this is a bug (class resides in root namespace).

localheinz:

PR #5080, opened 2022-10-13
Enable proxying to methods with static return type

hoogi91:
### Problem
Using method `createTestProxy` or creating a mock by using `enableProxyingToOriginalMethods` on mock builder leads to following issues when a method uses `static` in its return type - consider following:

```php
# MyClass
public function method(): static
{
    ...
    return $this;
}
```
and the following code to create the proxy class in test
```php
$class = $this->createTestProxy(MyClass::class);
$class->method();
```
which will result in something like
> TypeError: Mock_MyClass_26f4ca80::method(): Return value must be of type Mock_MyClass_26f4ca80, Vendor\MyClass returned

### Solution
Inside of [MockMethod](src/Framework/MockObject/MockMethod.php) class we check if original methods should be called and the return type contains `static` type. If that's true we use another template to render. The template always calls the original method as before but instead always returns the test proxy mock object.

This will fix the issue for all methods that are returning `static` but may break existing test expectations which are using `static` in a union return type due to the fact that the return value is no longer the return value of the real object but the test proxy itself.

### DIFF
--- a/src/Framework/MockObject/Generator/proxied_method_static.tpl
+++ b/src/Framework/MockObject/Generator/proxied_method_static.tpl
@@ -1,0 +1,23 @@
+
+    {modifier} function {reference}{method_name}({arguments_decl}){return_declaration}
+    {
+        $__phpunit_arguments = [{arguments_call}];
+        $__phpunit_count     = func_num_args();
+
+        if ($__phpunit_count > {arguments_count}) {
+            $__phpunit_arguments_tmp = func_get_args();
+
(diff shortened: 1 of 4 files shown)


hoogi91:
See #5083 :)

hoogi91:
Oh gosh, totally missed this one in the last view weeks 😄  it should now be mergable

hoogi91:
I just re-opened the PR completly in #5137

PR #5087, opened 2022-10-25
Add Phan attributes to createMock

mro95:
Phan doesn't understand that `createMock()` returns a mock object. I noticed if we add the same attributes like we did with psalm, it works fine.


mro95:
Related: https://github.com/phan/phan/issues/338

sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it. This should be addressed in Phan.

PR #5094, opened 2022-11-09
Dataproviders: By PHP 8.1, reflection can invoke private methods, therefore dataproviders can be private

HonzaMatosik:
Thanks to PHP RFC "Make reflection setAccessible() no-op" (https://wiki.php.net/rfc/make-reflection-setaccessible-no-op) is possible in PHP 8.1 to invoke private methods with Reflection without mutating state (without calling `setAccessible(true)`).

For me it make sense to just start using `private` methods for data providers, as there isn't general reason to leave them `public`. However, there also isn't reason to not support it. 

So in this PR I'm:
1) Adding tests to ensure private and protected data providers works.
2) Internally use more restricted (private or protected) methods for data providers

Related documentation update to encurage using private methods: https://github.com/sebastianbergmann/phpunit-documentation-english/pull/267


sebastianbergmann:

PR #5106, opened 2022-11-29
Assert::assertThrows()

mabar:
Just an idea for alternative exceptions checking. I am using this pattern often for checked exceptions with extra methods which cannot be tested with `expectException*` methods.

Let me know if you like the idea and if I should finish the PR.

```php
final class Test extends \PHPUnit\Framework\TestCase
{
	public function test(): void
	{
		$e = self::assertThrows(fn () => throw new \Exception());
		self::assertInstanceOf(\Exception::class, $e);
		self::assertSame('', $e->getMessage());
	}
}
```


sebastianbergmann:

PR #5113, opened 2022-12-08
PHP error instead of PHPUnit error when trying to create test double for read-only class

michaljusiega:
Hi!

According to new changes delivered in PHP 8.2, we cannot mock `readonly` classes like `final` or `enum` by extending them.

See reference: https://github.com/php/php-src/blob/7850c103894c50c4af4c2632c5ab60480a0fbdd3/Zend/tests/readonly_classes/readonly_class_inheritance_error1.phpt

Or shall we to add `readonly` keyword into generated mocked class to skip Runtime exception from PHP like: https://github.com/php/php-src/blob/7850c103894c50c4af4c2632c5ab60480a0fbdd3/Zend/tests/readonly_classes/readonly_class_inheritance_success.phpt

### DIFF
--- a/src/Framework/MockObject/Exception/ClassIsReadonlyException.php
+++ b/src/Framework/MockObject/Exception/ClassIsReadonlyException.php
@@ -1,0 +1,28 @@
+<?php declare(strict_types=1);
+/*
+ * This file is part of PHPUnit.
+ *
+ * (c) Sebastian Bergmann <sebastian@phpunit.de>
+ *
+ * For the full copyright and license information, please view the LICENSE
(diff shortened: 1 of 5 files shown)


sebastianbergmann:
Cherry-picked into `9.5` and merged to `main` from there. Thank you!

zoltanka:
Is there any reason behind this? Or any way to bypass this? Because this seems to me sofar very self-judgemental, leading to have a lot of boilerplate code just to have something tested.

Aliance:
@sebastianbergmann hello, Sebastian. Can you share your thoughts about this?

PR #5116, opened 2022-12-20
Improve performance of TestCase::registerMockObjectsFromTestArguments() with large test arguments arrays

fancyweb:
`TestCase::registerMockObjectsFromTestArguments()` iterates recursively on arrays tests arguments (from data providers).

The "visited" strategy with `in_array` is super slow with large arrays.
I thought we could reuse the `Context` class from the `sebastian/recursion-context` package like we already do with the `Enumerator` class in the `if` block.
I'm not really sure if it makes sense and if we can use it though since I don't know much about internals :man_shrugging: 

I'm running a test with a data provider that provides one array argument of 50k items.
With the current strategy:
```
Time: 00:16.220, Memory: 76.02 MB
```

With the proposed change:
```
Time: 00:00.143, Memory: 76.02 MB
```

`sebastian/recursion-context` is already a transitive dependency through `sebastian/object-enumerator`.

If it looks good, should it be applied to lower branches?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -45,6 +45,7 @@
         "sebastian/exporter": "^5.0",
         "sebastian/global-state": "^6.0",
         "sebastian/object-enumerator": "^5.0",
+        "sebastian/recursion-context": "^5.0",
         "sebastian/type": "^4.0",
         "sebastian/version": "^4.0"
     },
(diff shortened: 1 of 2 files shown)


sebastianbergmann:
Thank you! I do not think (right now, at least) that this should be backported to `8.5` and `9.5`.

PR #5121, opened 2022-12-23
Fixes phpunit/phpunit#5120

razzed:
Basically, these options do work together and are what I would consider a valid use case. List only the tests in a single test suite.

Note that if you don't accept this PR, there's a copy-paste error on line 79 - should say "--testsuite is ignored"


sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I fixed this in `8.5`, `9.5`, and `main` for `--list-tests` as well as `--list-tests-xml` without merging your commit. Sorry!

razzed:
No problem, seemed like something that was a short oversight. I'll be back! ;)

PR #5134, opened 2023-01-14
Adds Assert::assertThrows() for easier exception checking

Sharom:
This pull request adds `Assert::assertThrows()` which reuses `Exception` constraint in a way different from `TestCase::expectException()`.

Usage of `Assert::assertThrows()` enables us to validate postconditions after the exception in a strain forward way:

```php
public function testExample(): void
{
    $component = new MyComponent();

    $this->assertThrows(MyException::class, fn () => $component->send('hello'));

    // and here we can write more assertions to validate postconditions..
}
```

instead of:

```php
public function testExample(): void
{
    $component = new MyComponent();

    try {
        $component->send('hello');
        $this->fail('Exception should be thrown.');
    } catch (MyException) {}
    
    // other assertions to validate postconditions..
}
```


sebastianbergmann:

Sharom:
Hello, @sebastianbergmann. May I ask you some details, please?

Don't you like the concept itself or do you have some questions to design/implementation?

sebastianbergmann:
I do not like the concept itself.

Sharom:
First I give an example from real life.
I often use spy implementation o PSR Logger and wish to check that some record has been made in case of exception.

With current API I need to use `try..catch` + `fail()` which is cumbersome.
My implementation was inspired by XCTest, this is xUnit family framework made by Apple.

Do you have any alternative idea?

Issue #3506, opened 2019-02-01
PHP options should be passed to child processes


remicollet:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 8.0.0
| PHP version         | 7.2.14
| Installation Method | Composer

Playing with pcov...

When enabled (in php.ini), everything OK
```
$ php vendor/bin/phpunit 
PHPUnit 8.0.0 by Sebastian Bergmann and contributors.

Runtime:       PHP 7.2.14 with PCOV 1.0.0
Configuration: /work/GIT/sql-parser/phpunit.xml.dist
```

When disabled globally, and only enabled in command line, it fails

```
$ php -d pcov.enabled=1 vendor/bin/phpunit 
PHPUnit 8.0.0 by Sebastian Bergmann and contributors.

Runtime:       PHP 7.2.14 with PCOV 1.0.0
Configuration: /work/GIT/sql-parser/phpunit.xml.dist

.....................................II.................EW.....  63 / 545 ( 11%)

```

With some errors like: "_No code coverage driver available_"
```
1) PhpMyAdmin\SqlParser\Tests\Parser\ComponentTest::testParse
PHPUnit\Framework\Exception: Fatal error: Uncaught SebastianBergmann\CodeCoverage\RuntimeException: No code coverage driver available in /work/GIT/sql-parser/vendor/phpunit/php-code-coverage/src/CodeCoverage.php:898
Stack trace:
#0 /work/GIT/sql-parser/vendor/phpunit/php-code-coverage/src/CodeCoverage.php(147): SebastianBergmann\CodeCoverage\CodeCoverage->selectDriver(Object(SebastianBergmann\CodeCoverage\Filter))
#1 Standard input code(40): SebastianBergmann\CodeCoverage\CodeCoverage->__construct(NULL, Object(SebastianBergmann\CodeCoverage\Filter))
#2 Standard input code(112): __phpunit_run_isolated_test()
#3 {main}
  thrown in /work/GIT/sql-parser/vendor/phpunit/php-code-coverage/src/CodeCoverage.php on line 898

Caused by
ErrorException: unserialize(): Error at offset 0 of 696 bytes in /work/GIT/sql-parser/vendor/phpunit/phpunit/src/Util/PHP/AbstractPhpProcess.php:254
Stack trace:
[...]

PR #5135, opened 2023-01-15
Enable xdebug manually

uuf6429:
>For XDebug, this is a bit different, as there is no "enable" option (may have to be enabled using -d zend_extension=xdebug, but this can be another issue because of extension load order).

_Originally posted by @remicollet in https://github.com/sebastianbergmann/phpunit/issues/3506#issuecomment-459941554_
            

It is very typical for developers to run PHP without xdebug and only enable it when necessary through `-d zend_extension=xdebug`.
PhpStorm, for example follows this approach once properly configured: running tests does not enable xdebug. Debugging or covering tests enables xdebug.

This however does not work for process-isolated-tests (`@runTestsInSeparateProcess`): even though settings are [now copied to the child process](https://github.com/sebastianbergmann/phpunit/issues/3506), xdebug will not be enabled. This is understandable since it's not trivial to find if xdebug is enabled by default (and I was also not able to find the location of the library programmatically, so PR comment below).

The objective of this PR is to do exactly that. 🙏 

----

@sebastianbergmann as soon as it's decided where to go, I can also open PRs for other versions.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -41,7 +41,7 @@
         "sebastian/code-unit": "^1.0.6",
         "sebastian/comparator": "^4.0.8",
         "sebastian/diff": "^4.0.3",
-        "sebastian/environment": "^5.1.3",
+        "sebastian/environment": "^5.1.4",
         "sebastian/exporter": "^4.0.5",
         "sebastian/global-state": "^5.0.1",
         "sebastian/object-enumerator": "^4.0.3",
--- a/src/Util/PHP/AbstractPhpProcess.php
+++ b/src/Util/PHP/AbstractPhpProcess.php
@@ -209,6 +209,10 @@ abstract class AbstractPhpProcess
                     array_keys(ini_get_all('xdebug'))
                 )
             );
+
+            if (!$this->runtime->isXdebugLoadedByDefault()) {
+                $settings[] = 'zend_extension=' . (getenv('TODO_USE_PHPUNIT_SETTINGS') ?: 'php_xdebug');
+            }
         }
 
         $command .= $this->settingsToParameters($settings);


composer.json
```diff
@@ -41,7 +41,7 @@
         "sebastian/code-unit": "^1.0.6",
         "sebastian/comparator": "^4.0.8",
         "sebastian/diff": "^4.0.3",
-        "sebastian/environment": "^5.1.3",
+        "sebastian/environment": "^5.1.4",
```
uuf6429, line 44: I'm assuming [this PR](https://github.com/sebastianbergmann/environment/pull/66) will be released as that version.

src/Util/PHP/AbstractPhpProcess.php
```diff
@@ -209,6 +209,10 @@ public function getCommand(array $settings, string $file = null): string
                     array_keys(ini_get_all('xdebug'))
                 )
             );
+
+            if (!$this->runtime->isXdebugLoadedByDefault()) {
+                $settings[] = 'zend_extension=' . (getenv('TODO_USE_PHPUNIT_SETTINGS') ?: 'php_xdebug');
```
uuf6429, line 214: getenv('TODO_USE_PHPUNIT_SETTINGS')
    
 Should I...
1. ...just use `php_xdebug` since it already solves most cases
2. ...use `getenv('XX')` to offer a bit of flexibility (what should it be named?)
3. ...use PHPUnit xml config functionality to "do it properly"

sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it. This should be handled outside of PHPUnit, sorry.

uuf6429:
I understand it's a very specific annoyance, but I can't see it solved outside PHPUnit.

Enabling xdebug globally is terrible for day-to-day use (think installing composer packages with xdebug enabled 😆).

Additionally, #3506 feels half-baked without this - one could also argue that it was not needed if PHP was properly configured.

I've also tried hackier approaches, by the way, and nothing worked well (`<php><ini...>...`, modifying PHPUnit at runtime).. so not a lot of options unless I missed something.

PR #5140, opened 2023-01-17
XSD Schema Location: add local reference to ease autocompletion and validation

Slamdunk:
By default most IDEs don't support autocompletion and validation upon remote schemas, but all support multiple schema locations to be specified.

Since I guess almost all PHPUnit users nowadays install it by composer, that means that `vendor/phpunit/phpunit/phpunit.xsd` is always available as a primary source of truth.


sebastianbergmann:
Is it safe to add this local reference when `vendor/phpunit/phpunit/phpunit.xsd` is not guaranteed to exist? For instance, there is no such file when PHPUnit is used from its PHAR.

sebastianbergmann:
Can you target the `8.5` branch, please? Thanks!

Slamdunk:
> all support multiple schema locations to be specified.

Apparently it was true, but that's not the case anymore, sadly :disappointed:

Issue #5139, opened 2023-01-16
StandardTestSuiteLoader regression in PHPUnit 9.5.28


enumag:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 9.5.28
| PHP version         | 8.2.1
| Installation Method | Composer

#### Summary

Earlier today I ran a full composer update and Paratest started failing. I reported the issue [here](https://github.com/paratestphp/paratest/issues/716) however after some more digging the bug isn't caused by Paratest but rather recent change in PHPUnit's StandardTestSuiteLoader. In particular this commit is causing the issue: https://github.com/sebastianbergmann/phpunit/commit/83eaffa12425bb512ba55f466ed9092dcb8e3ffe

#### Current behavior

Current implementation of PHPUnit's StandardTestSuiteLoader returns ReflectionClass for cases where it previously threw an exception. This change is causing Paratest to crash completely.

```
Fatal error: Uncaught TypeError: PHPUnit\Util\Annotation\DocBlock::__construct(): Argument #4 ($startLine) must be of type int, bool given, called in /app/vendor/phpunit/phpunit/src/Util/Annotation/DocBlock.php on line 145 and defined in /app/vendor/phpunit/phpunit/src/Util/Annotation/DocBlock.php:164
Stack trace:
#0 /app/vendor/phpunit/phpunit/src/Util/Annotation/DocBlock.php(145): PHPUnit\Util\Annotation\DocBlock->__construct('', true, Array, false, false, false, '__construct', 'Exception')
#1 /app/vendor/phpunit/phpunit/src/Util/Annotation/Registry.php(93): PHPUnit\Util\Annotation\DocBlock::ofMethod(Object(ReflectionMethod), 'Exception')
#2 /app/vendor/phpunit/phpunit/src/Util/Test.php(590): PHPUnit\Util\Annotation\Registry->forMethod('Exception', '__construct')
#3 /app/vendor/brianium/paratest/src/Parser/Parser.php(118): PHPUnit\Util\Test::isTestMethod(Object(ReflectionMethod))
#4 /app/vendor/brianium/paratest/src/Parser/Parser.php(94): ParaTest\Parser\Parser->getMethods()
#5 /app/vendor/brianium/paratest/src/Runners/PHPUnit/SuiteLoader.php(192): ParaTest\Parser\Parser->getClass()
[...]

PR #5141, opened 2023-01-17
Add failing test

enumag:
Failing test for #5139.

This test fails on PHPUnit 5.9.28 but passes if I revert https://github.com/sebastianbergmann/phpunit/commit/83eaffa12425bb512ba55f466ed9092dcb8e3ffe - thus demonstrating the difference in the behavior.

### DIFF
--- a/tests/unit/Runner/StandardTestSuiteLoaderTest.php
+++ b/tests/unit/Runner/StandardTestSuiteLoaderTest.php
@@ -1,0 +1,24 @@
+<?php declare(strict_types=1);
+/*
+ * This file is part of PHPUnit.
+ *
+ * (c) Sebastian Bergmann <sebastian@phpunit.de>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+namespace PHPUnit\Runner;
+
+use PHPUnit\Framework\TestCase;
+
+final class StandardTestSuiteLoaderTest extends TestCase
+{
+    public function testIssue5139(): void
+    {
+        $this->expectException(Exception::class);
+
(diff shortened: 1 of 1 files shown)


sebastianbergmann:

I am sorry, but this test case does not help me understand what a real world test would look like that breaks the new behaviour introduced in 83eaffa12425bb512ba55f466ed9092dcb8e3ffe. I need a real test case class in a sourcecode file that triggers the issue when run through PHPUnit's test runner on the CLI.

enumag:
I do not believe it's possible to trigger the issue through PHPUnit's test runner. As I said the problem only happens when I use Paratest which is internally using the StandardTestSuiteLoader.

enumag:
@sebastianbergmann Simply put the commit is a backwards compatibility break for Paratest. I do not think I can provide any better test than this one which demonstrates the behavior difference quite well.

sebastianbergmann:
`$file = (new \ReflectionClass(Exception::class))->getFileName();` evaluates to `false`:

```
php > var_dump((new ReflectionClass(Exception::class))->getFileName());
bool(false)
```

`(new StandardTestSuiteLoader())->load($file);` cannot work when `$file` is `false` because `StandardTestSuiteLoader::load()` expects a `string`.

I am sorry to say this, but this is a problem on ParaTest's end: they are using a private implementation detail of PHPUnit (`StandardTestSuiteLoader` is marked `@internal`). They need to make sure to not call `StandardTestSuiteLoader::load()` with an invalid argument.

enumag:
> $file = (new \ReflectionClass(Exception::class))->getFileName(); evaluates to false:

No it doesn't. You're in the `PHPUnit\Runner` namespace meaning that `Exception::class` is in fact `\PHPUnit\Runner\Exception::class`.

PR #5149, opened 2023-01-24
chore: Remove check on static attribute always set to false

codisart:
Signed-off-by: codisart <louis.celier@gmail.com>

It seems to me that the "self::$initialized" was never updated from its default value of false. 
So the check in the self::initialize() method was useless.

### DIFF
--- a/src/Util/ExcludeList.php
+++ b/src/Util/ExcludeList.php
@@ -137,7 +137,6 @@ final class ExcludeList
      * @psalm-var list<string>
      */
     private static array $directories = [];
-    private static bool $initialized  = false;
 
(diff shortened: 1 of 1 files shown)


sebastianbergmann:

codisart:
Alright, can I help in another way fot this ? Do we want the value $self::initialized to be memoized and passed at true in the self::initialize() method ?

codisart:
Ah it is done ^^! https://github.com/sebastianbergmann/phpunit/commit/e86998b40b9fcec99ee8c6f79dce325e9296d8de

Have a nice day !

PR #5152, opened 2023-01-30
Fix: Continue on error when running tests on PHP 8.3

localheinz:
This pull request

- [x] use a combination of [`if`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsif) and [`the continue-on-error`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error) to prevent unit tests failing on PHP 8.3 to fail the build

Somewhat related to https://github.com/shivammathur/setup-php/issues/692.


localheinz:

PR #5154, opened 2023-01-30
Fix: Continue entire job when running on PHP 8.3

localheinz:
This pull request

- [x] uses a combination of [`continue-on-error`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error) and [`include`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategymatrixinclude) to allow an entire job to fail when running it on an unstable PHP version

Follows #5152.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,6 +68,8 @@ jobs:
       PHP_EXTENSIONS: none, curl, dom, json, libxml, mbstring, openssl, phar, soap, tokenizer, xml, xmlwriter
       PHP_INI_VALUES: assert.exception=1, zend.assertions=1, error_reporting=-1, log_errors_max_len=0, display_errors=On
 
+    continue-on-error: ${{ matrix.experimental }}
+
     strategy:
       fail-fast: false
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -68,6 +68,8 @@ jobs:
       PHP_EXTENSIONS: none, curl, dom, json, libxml, mbstring, openssl, phar, soap, tokenizer, xml, xmlwriter
       PHP_INI_VALUES: assert.exception=1, zend.assertions=1, error_reporting=-1, log_errors_max_len=0, display_errors=On
 
+    continue-on-error: ${{ matrix.experimental }}
```
localheinz, line 71: Unfortunately, we currently can not use an expression here, for example

```suggestion
    continue-on-error: ${{ matrix.php-version }} == "8.3"
```

For reference, see https://github.com/actions/runner/issues/1492.

localheinz:

PR #5155, opened 2023-01-30
Fix: Continue entire job when running on PHP 8.3

localheinz:
This pull request

- [x] uses a combination of [`continue-on-error`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error) and [`include`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategymatrixinclude) to allow an entire job to fail when running it on an unstable PHP version

Follows #5152.
Follows #5154.


localheinz:

PR #5162, opened 2023-02-03
Allow subscriber to implement multiple interfaces

### LINKED ISSUE #5161: Cannot register subscriber that implements multiple interfaces
| Q                   | A
| PHPUnit version     | 10.0.1
| PHP version         | 8.1

#### Summary

A subscriber that implements multiple subscriber interfaces gets registered only for the first declared interface.

#### Current behavior
Only one kind of event is received.

#### How to reproduce
Write a subscriber like this:
```
class TestWarning extends AbstractTestHook implements WarningTriggeredSubscriber, PhpWarningTriggeredSubscriber
{
    public function notify(WarningTriggered|PhpWarningTriggered $event): void
    {
        // ...
    }
}
```
and register it.

#### Expected behavior
Both kind of events are received.

I'll try to follow up with a PR shortly.

Jean85:
This fixes #5161.

### DIFF
--- a/src/Event/Dispatcher/DirectDispatcher.php
+++ b/src/Event/Dispatcher/DirectDispatcher.php
@@ -54,13 +54,13 @@ final class DirectDispatcher implements SubscribableDispatcher
             );
         }
 
-        $eventClassName = $this->typeMap->map($subscriber);
+        foreach ($this->typeMap->map($subscriber) as $eventClassName) {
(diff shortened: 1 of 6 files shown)


sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it for the reason I laid out in https://github.com/sebastianbergmann/phpunit/issues/5161#issuecomment-1416671087.

PR #5167, opened 2023-02-04
Added TestData attribute with named data sets and variadic parameter support

### LINKED ISSUE #5163: Variadic support for #[TestWith] attribute
Loving PHPUnit 10, thank you so much for it!  I've switched a lot of my `@dataProvider` annotations to use the `#[TestWith]` attribute, but I have found the fact that it takes an array of parameters rather than variadic parameters to be awkward.  For example, I think this is cleaner:

```php
#[TestWith(20, 21, false)]
#[TestWith(22, 21, true)]
public function testAgeRequirement(int $age, int $minimumAge, bool $expectedToPass): void
{
    $this->assertSame($expectedToPass, $age >= $minimumAge);
}
```

...than this:

```php
#[TestWith([20, 21, false])]
#[TestWith([22, 21, true])]
public function testAgeRequirement(int $age, int $minimumAge, bool $expectedToPass): void
{
    $this->assertSame($expectedToPass, $age >= $minimumAge);
}
```

...because the former almost looks like an invocation of the test method, which to me feels intuitive.  I feel the latter looks like you're passing in array as a parameter.  Since `TestWith::__construct()` does not take in any other parameters, I think it would be a better developer experience to take in a variadic parameter instead.  Obviously, this would be a breaking change, which might mean we should use a new attribute instead.  Using a variadic parameter would also complicate supporting a feature such as [named data sets](https://github.com/sebastianbergmann/phpunit/pull/4964).  I'm curious what your thoughts are on this and whether we should support variadic parameters or introduce a new attribute to prevent the breaking change.  If so, I'll contribute a pull request.

davidbyoung:
This is meant to simplify how inline test data is passed into tests.  Rather than requiring everything to be wrapped in an array like 
```php
#[TestWith(['foo', 'bar'])]
public function testSomething(string $a, string $b): void
{
    // ...
}
```
You can now just pass in the parameters directly using the new `#[TestData]` attribute:
```php
#[TestData('foo', 'bar')]
public function testSomething(string $a, string $b): void
{
    // ...
}
```
which looks more like an invocation of the test method.  Additionally, this PR introduces support for naming data sets using a `name` parameter, eg
```php
#[TestData('foo', 'bar', name: 'Tests something with "foo" and "bar"')]
public function testSomething(string $a, string $b): void
{
    // ...
}
```
I have purposely not created an annotation version of this because PHPDoc is not a good place to define PHP values to pass into test methods.

### DIFF
--- a/src/Framework/Attributes/TestData.php
+++ b/src/Framework/Attributes/TestData.php
@@ -1,0 +1,55 @@
+<?php declare(strict_types=1);
+/*
+ * This file is part of PHPUnit.
+ *
+ * (c) Sebastian Bergmann <sebastian@phpunit.de>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+namespace PHPUnit\Framework\Attributes;
+
+use function is_string;
+use Attribute;
+use PHPUnit\Event\InvalidArgumentException;
+
+/**
+ * @psalm-immutable
+ *
+ * @no-named-arguments Parameter names are not covered by the backward compatibility promise for PHPUnit
+ */
+#[Attribute(Attribute::TARGET_METHOD | Attribute::IS_REPEATABLE)]
+final class TestData
+{
+    private readonly array $data;
+    private readonly ?string $name;
+
+    /**
+     * @throws InvalidArgumentException
+     */
+    public function __construct(mixed ...$data)
+    {
+        if (isset($data['name'])) {
+            if (!is_string($data['name'])) {
+                throw new InvalidArgumentException('Name must be of type string.');
+            }
+
+            $this->name = $data['name'];
+            unset($data['name']);
+        } else {
+            $this->name = null;
+        }
+
+        $this->data = $data;
+    }
+
+    public function data(): array
+    {
+        return $this->data;
+    }
+
+    public function name(): ?string
+    {
+        return $this->name;
+    }
+}
--- a/src/Framework/Attributes/TestData.php
+++ b/src/Framework/Attributes/TestData.php
@@ -9,9 +9,7 @@
  */
 namespace PHPUnit\Framework\Attributes;
 
-use function is_string;
 use Attribute;
-use PHPUnit\Event\InvalidArgumentException;
 
 /**
  * @psalm-immutable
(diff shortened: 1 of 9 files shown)


davidbyoung:
I have a few open questions:

- Should `TestWith` and `TestWithJson` be deprecated in favor of the new `TestData` attribute to unify how to pass inline data to a test?
- I built my implementation sort of off the `TestWith` functionality, eg marking the `TestData` metadata's `isTestWith()` as `true`. Should I completely separate this implementation from `TestWith`?
- This solution uses an optional named argument to name the data set. Typically, PHPUnit has avoided supporting named arguments. Is my approach acceptable, or should I consider an alternative?

sebastianbergmann:
> have purposely not created an annotation version

Good, as no new annotations will be added to PHPUnit.

sebastianbergmann:
> This solution uses an optional named argument to name the data set. Typically, PHPUnit has avoided supporting named arguments. Is my approach acceptable, or should I consider an alternative?

Please consider an alternative.

sebastianbergmann:
> I built my implementation sort of off the `TestWith` functionality, eg marking the `TestData` metadata's `isTestWith()` as `true`. Should I completely separate this implementation from `TestWith`?

I think we do not need a new `Metadata` object, only a new attribute. The attribute parser should create the same `Metadata` object from the two attributes.

sebastianbergmann:
> Should `TestWith` and `TestWithJson` be deprecated in favor of the new `TestData` attribute to unify how to pass inline data to a test?

I do not think so, at least not for now.

davidbyoung:
Thanks for the feedback!  Looking to xUnit and NUnit for inspiration, it looks like xUnit does not support named data sets in their `[InlineData]` attribute, but [NUnit does via named parameters](https://docs.nunit.org/articles/nunit/writing-tests/attributes/testcase.html?q=testcase), similar to my original approach.  As far as I can tell, we have the following options:

* Use named parameters, although this would be a break with how PHPUnit has typically worked.
* Have a separate attribute `#[NamedTestData('This is the name', 'foo', 'bar')]` whose first parameter is the name and subsequent parameters are passed to the test method.  We could still introduce the `#[TestData('foo', 'bar')]` attribute to prevent us from having to pass in the parameters in an array, just without supporting named data sets.
* Require data sets to always be named as the first parameter in `#[TestData('This is the name', 'foo', 'bar')]`.
* Continue to wrap parameters in an array, similar to [this approach](https://github.com/sebastianbergmann/phpunit/pull/4964) but with the negative of having to wrap all parameters in an array, which, at least to me, feels unintuitive.
* Scrap the concept of named data sets altogether.

I can implement a PR for any of these approaches, but wanted to get your input first.

davidbyoung:
@sebastianbergmann Did you have any preference on my list of proposed ways forward above?

davidbyoung:
Hi @sebastianbergmann, have you had an opportunity to review the different options I enumerated above? Do any sound preferable to you? Or should I abandon this feature altogether? Thanks!

sebastianbergmann:
The more I think about this, the more I think that things should stay as they are. Sorry.

Issue #5169, opened 2023-02-05
Parameter type for constructor of GreaterThan and LessThan constraints does not match parameter type for Assert::greaterThan() and Assert::lessThan()


drexlma:
TypeError: PHPUnit\Framework\Constraint\GreaterThan::__construct(): Argument #1 ($value) must be of type int|float, string given, called in phar:///usr/local/bin/phpunit/phpunit/Framework/Assert.php on line 1620

PHPUnit 10.0.3 by Sebastian Bergmann and contributors.
Runtime:       PHP 8.2.2

`
		$this->assertGreaterThanOrEqual('1',(int) count($body['kategorien']));`

PR #5170, opened 2023-02-05
Update Assert.php

drexlma:
#5169 
Typ passend zu https://github.com/sebastianbergmann/phpunit/blob/main/src/Framework/Constraint/Cardinality/GreaterThan.php: public function __construct(float|int $value)


sebastianbergmann:

PR #5185, opened 2023-02-09
Detail deprecations for 9.6.1 in the changelog

johnbillion:
The 9.6 branch adds several deprecations that aren't documented anywhere except in PRs from two years ago during planning for PHPUnit 11. This adds the method names to the changelog for 9.6.1 so at least they can be found there.


sebastianbergmann:
Cherry-picked into `9.6`.

PR #5188, opened 2023-02-09
Add test for bug: JUnit doesn't log error, and it doesn't appear in DefaultPrinter

Slamdunk:
This test passes on `9.6` (with proper, minor adjustments of the `--FILE--` section of course).

Basically in `10.0` when:

* The test errors in the `setUp`
* The `JUnit` logger is active

The following unexpected behaviours happen all at once:

1. The output is truncated after progress:
`E                                                                   1 / 1 (100%)No tests executed!`
2. The junit log file reports that the test passed
3. The exit status is `0`

### DIFF
--- a/tests/end-to-end/logging/_files/TypeErrorTest.php
+++ b/tests/end-to-end/logging/_files/TypeErrorTest.php
@@ -9,19 +9,21 @@
  */
 namespace PHPUnit\TestFixture;
 
+use DateTime;
+use DateTimeImmutable;
 use PHPUnit\Framework\TestCase;
 
 final class TypeErrorTest extends TestCase
 {
-    private \DateTimeImmutable $dateTime;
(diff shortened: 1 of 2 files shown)


sebastianbergmann:
Please add an `--XFAIL--` as long as this bug is not fixed. See https://github.com/sebastianbergmann/phpunit/commit/a3549f1f8dabcbc366b8139c0f71b694dc2e24fd#diff-857aef1a9952bb4d15370bbdc6684ec245135671e8c077c84f527b4f86720f23R13 for an example.

sebastianbergmann:
Thank you! Do you plan on working on a fix?

Slamdunk:
I am not, currently.

What worries me the most is the `0` exit code: silent errors produce the worst outcome.

Issue #5192, opened 2023-02-10
Running an empty test shows help screen


localheinz:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 10.0.7
| PHP version         | 8.1.4
| Installation Method | Composer

#### Summary

After upgrading from `phpunit/phpunit:9.6.3` to `phpunit/phpunit:10.0.7` running `phpunit` with a `phpunit.xml` that points to a directory that does not contain any tests succeeds, but shows the help screen.

#### Current behavior

Running `phpunit` with a `phpunit.xml` that points to a directory that does not contain any tests succeeds, but shows the help screen.

#### How to reproduce

See https://github.com/ergebnis/symfony-application-template/pull/1544, and give me a few moments to open a pull request.

#### Expected behavior

Running `phpunit` with a `phpunit.xml` that points to a directory that does not contain any tests succeeds with warnings.

PR #5193, opened 2023-02-10
Fix: Add test asserting that running empty test suite does not fail

localheinz:
This pull request

- [x] adds a test asserting that running an empty test suite does not fail

Related to #5192.


localheinz:

Issue #4026, opened 2020-01-24
Feature similar to withConsecutive(), but without checking order


dbrekelmans:
In [the docs](https://phpunit.readthedocs.io/en/8.5/test-doubles.html#test-doubles-mock-objects-examples-with-consecutive-php) we find this example for `withConsecutive()`:

```
<?php
use PHPUnit\Framework\TestCase;

class FooTest extends TestCase
{
    public function testFunctionCalledTwoTimesWithSpecificArguments()
    {
        $mock = $this->getMockBuilder(stdClass::class)
                     ->setMethods(['set'])
                     ->getMock();

        $mock->expects($this->exactly(2))
             ->method('set')
             ->withConsecutive(
                 [$this->equalTo('foo'), $this->greaterThan(0)],
                 [$this->equalTo('bar'), $this->greaterThan(0)]
             );

        $mock->set('foo', 21);
        $mock->set('bar', 48);
    }
}
```

If your class calls method `set` with parameter `foo` first and with parameter `bar` second, the test will pass. However, if you call it with `bar` first and `foo` second the test will fail. This leads to brittle tests (as is mentioned [in a note](https://phpunit.readthedocs.io/en/8.5/test-doubles.html#test-doubles-mock-objects-examples-enable-clone-object-parameters-php) under the `at()` docs).

It would be lovely to have a way to check `method 'foo' is called exactly twice. Once with parameter 'x' and once with parameter 'y'`, without checking the order.

A concrete example is testing that certain events are dispatched when it doesn't matter in what order they are dispatched (of course sometimes you do care about the order and in that case you can use `withConsecutive()`).

sebastianbergmann:
@dbrekelmans @jaapio Sorry for dropping the ball on this. Are you able to send a pull request with an implementation for this in time for PHPUnit 10 (February 5, 2021)?

jaapio:
No sorry, I had to spend more time on my own projects. So I didn't have the time to design something for this.

Unless I find a way to focus a longer period on the mocks. I don't think I will be able to provide a patch for this.

dbrekelmans:
Not likely within the timeframe. I can look into it, but I don't know how much time I'll have to work on this and I've never worked on the PHPUnit codebase before so it'll take more time to figure out a solution.

sebastianbergmann:
Thank you for getting back to me so quickly. I'll see what I can do.

sebastianbergmann:
Any update on this? Thanks!

[...]

PR #5203, opened 2023-02-15
MockObject can be used to test consecutive method calls.

biozshock:
Allow mock object to be used to mock methods with same name, but with different parameters.
Addresses #4255 and #4026.

This can be easily ported to 9.6. Just let me know if i need to target 9.6.

### DIFF
--- a/tests/unit/Framework/MockObject/Builder/InvocationMockerTest.php
+++ b/tests/unit/Framework/MockObject/Builder/InvocationMockerTest.php
@@ -21,6 +21,7 @@ use PHPUnit\Framework\TestCase;
 use PHPUnit\TestFixture\Foo;
 use PHPUnit\TestFixture\MockObject\ClassWithAllPossibleReturnTypes;
 use PHPUnit\TestFixture\MockObject\ClassWithImplicitProtocol;
+use PHPUnit\TestFixture\MockObject\ClassWithVariadicArgumentMethod;
 use stdClass;
 
 #[CoversClass(InvocationMocker::class)]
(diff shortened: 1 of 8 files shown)


sebastianbergmann:
> This can be easily ported to 9.6. Just let me know if i need to target 9.6.

PHPUnit 9 is closed for new functionality, the `main` target is correct.

sebastianbergmann:
I did not have time to review this yet. Please note that by now I am very reluctant when it comes to add new functionality to test doubles.

It would help if you could provide real world use cases for the functionality you propose. Tests for test double functionality are usually not really representative and are not helpful for understanding what the functionality they test should be used and how.

localheinz:
@biozshock 

Can you share how you would implement test cases for the examples that you shared in https://github.com/sebastianbergmann/phpunit/pull/5203#issuecomment-1479101454 

- with the current functionality in `phpunit/phpunit` and
- with the functionality you propose in this pull request

?

biozshock:
@localheinz Sure. Something along these lines:

```php

class ForwardTest
{
  public function testOld(): void
  {
    $handler = $this->createMock(Handler::class);
    $handler->method('handle')
      ->withConsecutive($event1, $event2)
      ->willReturnOnConsecutiveCalls($return1, $return2);

   $forward = new Forward();
   $forward->handle($handle);
  }

  public function testNew(): void
  {
    $handler = $this->createMock(Handler::class);
    $handler->expects(self::once())
      ->method('handle')
      ->with($event1)
      ->willReturn($return1)
      ->andThen(self:once())
      ->method('handle')
      ->with($event2)
      ->willReturn($return2);

   $forward = new Forward();
   $forward->handle($handle);
  }
}

File: tests/unit/Framework/MockObject/Builder/InvocationMockerTest.php [hunk 1]
```diff
@@ -243,4 +244,37 @@ public function testWillReturnAlreadyInstantiatedStubs(): void
...
+    public function testConsecutiveCallsInOrder(): void
+    {
+        $mock = $this->createMock(ClassWithVariadicArgumentMethod::class);
+        $mock->expects($this->exactly(2))
+            ->method('foo')
+            ->with(1)
+            ->willReturn('first')
+            ->andThen($this->once())
+            ->with(2)
+            ->willReturn('second');
```
lolli42, line 257: This is a bit weird, "strict" call-order is not possible when same arguments are given multiple times.
I would have expected, this would work as well, but it fails:

$mock->expects($this->once())
            ->method('foo')
            ->with(1)
            ->willReturn('first')
            ->andThen($this->once())
            ->with(2)
            ->willReturn('second')
            ->andThen($this->once())
            ->with(1)
            ->willReturn('first');

lolli42: In general, I like the idea of having a substitution of withConsecutive() using the fluent interface. The details are a bit nifty though, see my example above. It's pretty hard to end up with a simple syntax without surprises. andThen() suggests a "strict" order, while "orThen" is "may be called with this as well, and if so, then return that". That's not entirely clear to me with proposed solution.

albert361:
Hello everyone, 
Any updates or progress here?

X-Coder264:
@localheinz Has there maybe been any status update in regards to this PR?

sebastianbergmann:

xserrat:
Hello @sebastianbergmann , do you know if there's any plan on the roadmap to include this option for future releases? Thanks in advance!

PR #5211, opened 2023-02-18
Fix TestCase::onNotSuccessfulTest() PHP 8.0 and lower BC

mvorisek:
introduced in https://github.com/sebastianbergmann/phpunit/commit/b61f014f0896ac75b3f7711a95c6fd0a5fa575fa

In PHP 8.0 and lower there is no `never` native type.

PHP projects overriding `TestCase::onNotSuccessfulTest()` method like:

```php
<?php

class MyTest extends TestCase {
    protected function onNotSuccessfulTest(Throwable $t): void
    {
        throw $t;
    }
}
```

should still be able to run the tests on PHP 8.0 and lower (with phpunit 9.x).

This PR loosen the return type requirement from `never` to `void`. PHP 8.1 and higher only projects can use the native `never` return type already - https://3v4l.org/gjAI0.

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -1383,9 +1383,11 @@ abstract class TestCase extends Assert implements Reorderable, SelfDescribing, T
     /**
      * This method is called when a test method did not execute successfully.
      *
+     * @return never
+     *
      * @throws Throwable
      */
-    protected function onNotSuccessfulTest(Throwable $t): never
+    protected function onNotSuccessfulTest(Throwable $t): void
     {
         throw $t;
     }


sebastianbergmann:

PHPUnit 10.0 is not supported on PHP 8.0, so the issue you want to fix does not exist.

mvorisek:
The problem is present when PHPUnit 10.0 needs to be supported but the project still needs support PHP 8.0 and lower.

Currently, when PHPUnit 10.0 needs to be supported, `never` return type needs to be declared - https://3v4l.org/gjAI0 - which leads to fatal error in PHP 8.0.

Without this PR, fix like https://github.com/atk4/core/commit/1faf85e26cfbebbfbdaa114274f62c298479c5df is needed. Is it the best/recommended solution?

mvorisek:
@sebastianbergmann may I please ask you for a recommended way to solve the problem and/or reconsider this PR, thank you

sebastianbergmann:
> The problem is present when PHPUnit 10.0 needs to be supported but the project still needs support PHP 8.0 and lower.

Do not use PHPUnit 10 in projects that need to support PHP < 8.1.

PR #5223, opened 2023-02-20
rename objectHasAttribute to objectHasProperty

### LINKED ISSUE #5220: Add `assertObjectHasProperty()` and `assertObjectHasNotProperty()`

jasverix:
Fix #5220


sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it. This is too big of a change to introduce into the PHPUnit 9 version series.

jasverix:
I renamed the entire class, could maybe just add the method without renaming the class? If you think #5220 should be done, I can try to make a smaller PR for it.

PR #5225, opened 2023-02-20
Allow providing named arguments from a data provider

### LINKED ISSUE #5150: Allow providing named arguments from a data provider
Using a data provider you can easily run the same test for a different set of values. However, this construction currently requires the number of parameters to match the expected number, and them be in the exact same order as expected by the test method.

Since PHP 8.0 it is possible to pass argument by name to a method, which can be combined with array destructuring to allow passing arguments by name to a function. This would allow something like the very simple example below:

```php

final class DataTest extends TestCase
{
    /**
     * @dataProvider additionProvider
     */
    public function testEquals(int $a, int $b = 0, int $expected = 0): void
    {
        $this->assertSame($expected, $a + $b);
    }

    public function additionProvider(): array
    {
        return [
            [0, 1, 1], // already works
            [0], // already works
            [1, 'b' => 3, 'expected' => 4], // already works (array keys are discarded)
            [1, 'expected' => 4, 'b' => 3], // does not work (as intended) yet
        ];
    }
}
```

I got my local testsuite to behave like this by removing the `array_values` call that is applied to the test arguments before running, i.e. changing 
https://github.com/sebastianbergmann/phpunit/blob/a06635903e3f50fad271d30777660b08b6de6232/src/Framework/TestCase.php#L1016
to
```php
            $testResult = $this->{$this->name}(...$testArguments); 
```
but I'm not sure of possible consequences elsewhere, so have not yet found time to make a proper PR yet.

*Edit*: it seems that the `array_values` call is actually required to make `dependencyInput` function properly; if I change 
https://github.com/sebastianbergmann/phpunit/blob/a06635903e3f50fad271d30777660b08b6de6232/src/Framework/TestCase.php#L1011
to
```php
```
all existing tests seem to pass locally as before.

jnoordsij:
Resolves #5150

This PR aims to allow passing named arguments from a data provider, to have more flexibility in re-ordering arguments and skipping optional ones without having to provide a default value.

Note 1: I was not entirely sure on where to put the test; if there is a more suitable location to add this, let me know!

Note 2: Combining named arguments with input from a dependency is not supported, as providing positional arguments after named ones in an array is not supported, and data provider arguments will go first as [per doc description](https://docs.phpunit.de/en/10.0/writing-tests-for-phpunit.html#data-providers). I could not find a test to ensure this behavior, so made https://github.com/jnoordsij/phpunit/commit/ac028c8eb14da44268378a5a0fe0b9840aa67e2e this PR doesn't break existing behavior. I also created https://github.com/jnoordsij/phpunit/commit/f7e28c4c7a51c1f805d5ec94ae29feaa5acf63f4 to demonstrate the unsupported case. If wanted, I can add those tests to this PR or a new PR.

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -1176,10 +1176,10 @@ abstract class TestCase extends Assert implements Reorderable, SelfDescribing, T
      */
     final protected function runTest(): mixed
     {
-        $testArguments = array_merge($this->data, $this->dependencyInput);
+        $testArguments = array_merge($this->data, array_values($this->dependencyInput));
 
         try {
-            $testResult = $this->{$this->name}(...array_values($testArguments));
+            $testResult = $this->{$this->name}(...$testArguments);
         } catch (Throwable $exception) {
             if (!$this->shouldExceptionExpectationsBeVerified($exception)) {
                 throw $exception;
--- a/tests/_files/DataProviderNamedArgumentsTest.php
+++ b/tests/_files/DataProviderNamedArgumentsTest.php
@@ -1,0 +1,30 @@
+<?php declare(strict_types=1);
+/*
+ * This file is part of PHPUnit.
+ *
+ * (c) Sebastian Bergmann <sebastian@phpunit.de>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+namespace PHPUnit\TestFixture;
+
+use PHPUnit\Framework\Attributes\DataProvider;
+use PHPUnit\Framework\TestCase;
+
+class DataProviderNamedArgumentsTest extends TestCase
+{
+    public static function providerMethod()
+    {
+        return [
+            ['a' => 1, 'b' => 2, 'c' => 3],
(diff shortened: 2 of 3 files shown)


jnoordsij:
Hey there! Was wondering if there was anything left for me to do that can help getting this reviewed and/or merged?

sebastianbergmann:
This will not go into PHPUnit 10.1, sorry.

jnoordsij:
> This will not go into PHPUnit 10.1, sorry.

I understand that it was a bit of a short time to get it in. Is there anything I can do to help it go into the next minor version?

jnoordsij:
Rebased this on `main` and tests still seem to be fine locally.

Note: for anyone subscribed interested, testing these changes in your local setup to ensure everything works fine on your end and then leaving a review might be helpful in landing this PR.

sebastianbergmann:
> I also created https://github.com/jnoordsij/phpunit/commit/f7e28c4c7a51c1f805d5ec94ae29feaa5acf63f4 to demonstrate the unsupported case. If wanted, I can add those tests to this PR or a new PR.

Yes, please send a pull request for test(s) for the unsupported case(s). Thanks!

PR #5229, opened 2023-02-20
assertObjectHasProperty() and assertObjectNotHasProperty()

### LINKED ISSUE #5220: Add `assertObjectHasProperty()` and `assertObjectHasNotProperty()`

jasverix:
Added PR for PhpUnit 10.0. Still not sure if you accept the feature request at all, but then the PR is at least here.

If accepted, I can try to make a "small-as-possible" change for PhpUnit 9 so we can get rid of the deprecation warnings. Just add the new methods and let them call the old methods or something.

If you don't want the feature, then it is OK, but then I will probably add the code in our codebase instead, because we cannot live without them in our API tests 😅 

Fix #5220


sebastianbergmann:
> Still not sure if you accept the feature request at all, but then the PR is at least here.

I am open to the idea and will review this PR as soon as I find the time.

> I can try to make a "small-as-possible" change for PhpUnit 9 so we can get rid of the deprecation warnings.

PHPUnit 9 is closed for additions.

sebastianbergmann:
Did you add `ClassHasProperty` (https://github.com/sebastianbergmann/phpunit/pull/5229/files#diff-69d378dcbcd5862a44ba12f6649211842df4255a14a2e500b1de62f0275bd07d) by mistake?

jasverix:
Yeah, I actually just fetched the old classes from 9 and renamed them. Have gone through the code now and removed ClassHasProperty, and added the two new functions to Functions.php.

sebastianbergmann:
Can you please fix the coding guidelines and type checker issues and squash your changes into a single commit when you're done? Thanks!

PR #5231, opened 2023-02-21
assertObjectHasProperty() and assertObjectNotHasProperty() #5229

### LINKED ISSUE #5220: Add `assertObjectHasProperty()` and `assertObjectHasNotProperty()`

jasverix:
Fix #5220

### DIFF
--- a/src/Framework/Constraint/Object/ObjectHasProperty.php
+++ b/src/Framework/Constraint/Object/ObjectHasProperty.php
@@ -1,0 +1,76 @@
+<?php declare(strict_types=1);
+/*
+ * This file is part of PHPUnit.
+ *
+ * (c) Sebastian Bergmann <sebastian@phpunit.de>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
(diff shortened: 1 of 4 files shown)


src/Framework/Constraint/Object/ObjectHasProperty.php
```diff
@@ -0,0 +1,76 @@
...
+    protected function matches(mixed $other): bool
+    {
+        if (!is_object($other)) {
+            return false;
+        }
+
+        try {
+            return (new ReflectionObject($other))->hasProperty($this->propertyName);
+        } catch (ReflectionException) {
+            return false;
```
jasverix, line 55: Am getting a 'codecov' warning here. The problem is that psalm thinks that the above line can throw a ReflectionException, and is complaining if I don't have the catch. But it is not possible to write code that will cause this exception, so I cannot cover it with tests.

sebastianbergmann: I will take care of this when I merge this.

sebastianbergmann:
Merged manually, thanks.

PR #5267, opened 2023-03-02
Fix possibly negative str_repeat argument in ProgressPrinter

enumag:
```
  str_repeat(): Argument #2 ($times) must be greater than or equal to 0        
    at vendor/phpunit/phpunit/src/TextUI/Output/Default/ProgressPrinter/ProgressPrinter.php:256    
```

### DIFF
--- a/src/TextUI/Output/Default/ProgressPrinter/ProgressPrinter.php
+++ b/src/TextUI/Output/Default/ProgressPrinter/ProgressPrinter.php
@@ -54,8 +54,10 @@ final class ProgressPrinter
 
     public function testRunnerExecutionStarted(ExecutionStarted $event): void
     {
+        $this->numberOfTestsRun   = 0;
         $this->numberOfTests      = $event->testSuite()->count();
         $this->numberOfTestsWidth = strlen((string) $this->numberOfTests);
+        $this->column             = 0;
         $this->maxColumn          = $this->numberOfColumns - strlen('  /  (XXX%)') - (2 * $this->numberOfTestsWidth);
     }
 


sebastianbergmann:
How did you trigger this?

enumag:
Unfortunately I don't know. It happens randomly with low probability in my GitLab pipeline ever since I updated to PHPUnit 10.0. It only happens with Paratest and never locally nor in any kind of deterministic way. Still I don't think there is any harm in adding this fix.

sebastianbergmann:
I fear that this might be the wrong place to fix this, but I have not investigate this yet.

@Slamdunk Do you think this could be an issue specific to Paratest?

sebastianbergmann:
@enumag Can you please squash your two commits and give the resulting commit a message that is more meaningful than `Fix ProgressPrinter.php` or `Update ProgressPrinter.php`? Thank you!

enumag:
@sebastianbergmann @Slamdunk Updated.

sebastianbergmann:
> But I guess the correct fix should be the following

I have implemented that change in ffc9794e5a134040b188904d09cfc053ea4fcc5b.

@enumag Can you please test whether that change solves your problem?

enumag:
@sebastianbergmann I cannot since I have no way to reliably reproduce the issue.

sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it as I do not think that this is right place in the code to address this issue.

enumag:
Then what is the right place in your opinion?

sebastianbergmann:
> Then what is the right place in your opinion?

https://github.com/sebastianbergmann/phpunit/pull/5267#issuecomment-1453272483

enumag:
I see, you missed the fact that I updated this PR according to that comment. :D

sebastianbergmann:
Sorry that I missed this!

PR #5271, opened 2023-03-03
Correct composer suggestion description

AJenbo:
The value in suggestions is not for the version (like in required) but for a descriptive test that can help developers in deciding if they should install the suggested package.

This hint will be show to developers when running `composer suggests`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -54,9 +54,9 @@
         "sort-packages": true
     },
     "suggest": {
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
Please send this for the `8.5` branch. Thanks!

AJenbo:
@sebastianbergmann done, this now also adds a description for the 2 other packages that where suggested in PHPUnit 8.5

PR #5279, opened 2023-03-13
StringContains: bug fix for ignoring line endings

jrfnl:
As things were, at the time of the comparison, only the line endings of the `$haystack` were normalized, not the line endings of the `$needle`, which led to false negatives.

Fixed now.

Includes additional tests demonstrating the issue and safeguarding the fix.

Includes a similar additional test for the case-insensitivity check, which did not have this issue, but wasn't being safeguarded against it either.

### DIFF
--- a/src/Framework/Constraint/String/StringContains.php
+++ b/src/Framework/Constraint/String/StringContains.php
@@ -27,6 +27,10 @@ final class StringContains extends Constraint
 
     public function __construct(string $string, bool $ignoreCase = false, bool $ignoreLineEndings = false)
     {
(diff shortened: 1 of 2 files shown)


jrfnl:
Note: the CI / Coding style issue is unrelated to this PR and I believe it's a false positive anyway.

sebastianbergmann:
> Note: the CI / Coding style issue is unrelated to this PR and I believe it's a false positive anyway.

https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/issues/6828 :)

PR #5291, opened 2023-03-25
Fail run with invalid data providers

marmichalski:
Before continuing with this, I assume the current behaviour is indeed wrong, @sebastianbergmann? This test currently fails with:
```diff
-PHPUnit Finished (Shell Exit Code: 1)
+PHPUnit Finished (Shell Exit Code: 0)
```

What it means is, if you have a test with invalid/missing data provider and another one that is passing, the whole run is considered success.


sebastianbergmann:
Thank you for bringing this to my attention, and for providing a test case!

PR #5298, opened 2023-04-01
Constraint/logicalNot: bug fix - prevent weird replacement

jrfnl:
If a value passed to an assertion would have `exists` in it, it would get replaced by `does not exist` leading to weird error messages.

Fixed now.

I have a test for this, but as this code isn't tested in the 8.5 branch, I will pull the test separately and to the `main` branch.

PR #5299, opened 2023-04-01
Constraint/logicalNot: test for bug fix

jrfnl:
Sister-PR to PR #5298 containing the test for the fix.


jrfnl:
Note: the test won't pass without the fix from #5298.

sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it due the problem I described [here](https://github.com/sebastianbergmann/phpunit/pull/5298#issuecomment-1493228053).

Please send a new PR against `9.6` that does not break existing tests. The tests that break in PHPUnit 9.6 should probably not have been deleted when I cleaned up the mess that the constraint tests were.

I am sorry about the messy situation (the fragility of `LogicalNot::negate()` as well as that not all branches have the same tests for this).

PR #5323, opened 2023-04-13
Fix: Adjust order of values when assembling error message

### LINKED ISSUE #5322: Broken output on startup when required extensions are missing
```
$ php -n ./phpunit      
PHPUnit requires the "extensions are" extensions, but the "dom", "json", "libxml", "mbstring", "tokenizer", "xml", "xmlwriter" dom", "mbstring", "tokenizer", "xml", "xmlwriter not available.
```

localheinz:
This pull request

- [x] adjusts the order of values for assembling the error message when required extensions are unavailable

Fixes #5322.
Follows #5281.


sebastianbergmann:
This also needs to be fixed in `9.6` AFAICS.

sebastianbergmann:
Thank you!

PR #5330, opened 2023-04-15
Optionally ignore suppression of deprecations, notices, and warnings

### LINKED ISSUE #5328: Optionally ignore suppression of deprecations, notices, and warnings
Follows #5294 and supersedes #5302.

The events emitted by the error handler should be augmented with a `$suppressed` property. Then it is up to event handlers, result and progress printers, for instance, to decide whether or not they process suppressed `E_*`.

sebastianbergmann:
Closes #5328.


sebastianbergmann:
@sbuerk May I ask for your feedback on https://github.com/sebastianbergmann/phpunit/blob/58917f06d02c6ef052e9c027433ff0b5cb43406f/src/Runner/ErrorHandler.php#L42?

`ErrorHandler::__invoke()` now always triggers an event for `E_STRICT`, `E_NOTICE`, `E_USER_NOTICE`, `E_WARNING`, `E_USER_WARNING`, `E_DEPRECATED`, `E_USER_DEPRECATED`, and `E_USER_ERROR`. It also always returns `true` now after it emitted an event.

I am not sure whether always returning `true` (except in the `default` case) is the right thing to do here. [If it returned `false` then the normal error handler would continue](https://www.php.net/manual/en/function.set-error-handler.php).

The value objects representing the respective events know whether or now whether the `E_*` they represent was suppressed or not.

Right now, the events for suppressed `E_STRICT`, `E_NOTICE`, `E_USER_NOTICE`, `E_WARNING`, `E_USER_WARNING`, `E_DEPRECATED`, `E_USER_DEPRECATED`, and `E_USER_ERROR` are ignored by both the default progress printer and the default result printer. This will be configurable (separately for each type of `E_*`), but the default configuration should be that suppressed `E_*` are ignored.

The configuration options mentioned above have not been implemented yet.

sbuerk:
@sebastianbergmann Thank you for picking this up. :heart: I/we have closed my try/pull-request to make this on some way configurable. Which was the right thing to do!

I really appreciate this approach over the state some days ago, because of following reasons:

* PHPUnit is in charge of all errors
* PHPUnit allows the configuration how suppressed error levels should be handled for different use-cases :heart:
* At least for the transition having suppressed error levels ignored by default is the correct way, in my opionion (same behaviour stats at first). This can be eventually flipped in next major (11.x) I guess, but not sure.
* Put all errors with calcualted `suppressed` information to all related events is really nice
* Handling suppressed `E_*` through the event system, but configurable is the best outcome to get

This pull-request in current state keeps current behaviour, but opens for making it configurable. Which would really be a good thing.

Background:

We love to have tests as hard failing as possible, means `E_ALL` without custom error/exception handlers hiding issues and so on, and even taking notices,warnings and depreations as a error/failure. Through our deprecation policy, we must ensure that we do not call code before we deprecate things. We are using not suppressed `E_USER_DEPRACTED` for that. Using the silence operator `@` is generally discouraged in the code base, but there are valid reason for the one or other place to use it at least temporarly to avout warnings/notices/deprecations (php) or third party code.

From my side, I would say that we archieve all goals for all use-case.

> I am not sure whether always returning true (except in the default case) is the right thing to do here. [If it returned false then the normal error handler would continue](https://www.php.net/manual/en/function.set-error-handler.php).

I would say, with providing the suppressed state in the events for each error level and the upcoming option to configure it, PHPUnit is in full charge of handling errors and handles it. Therefore, return `true` in all cases now should be the absolute correct return value in my opinion (except default).

PHPUnit ErrorHandler (ExceptionHandler) should be the only one in charge for test execution, thus let it handle everything is the right way in my view. (We have ErrorHandler/ExceptionHandler checks in our base TestCase to ensure that no tests misses to cleanup correctly if ErrorHandler/ExceptionHanlder are registered - only allowing the PHPUnit onces. 

Mainly, because with PHPUnit we are in another context. This is not a real application, project or framework. This is test related, and there is no need that it should fallback to let php handle error levels. PHPUnit handles all of them, suppressed or not. Per options, it is controlled `what` should happen (display/fail/stop + ignore/not-ignore suppressed).

Even if not tested with `E_ALL` setup, suppressing stuff even if the `@` silence operator is not used, is now possible and could lead to fails if wanted.

I guess the plan is, to add tests for not-suppressed error-levels again if they are made configureable. May I ask what the planned configuration level is ? Configuration file as <PHPUnit> arguments and/or cli options ?


This along with having it configurable opens for all constellations. :+1:

sebastianbergmann:
> I guess the plan is, to add tests for not-suppressed error-levels again if they are made configureable. May I ask what the planned configuration level is ? Configuration file as arguments and/or cli options?

I am reluctant when it comes to adding new CLI options, so this will be configurable using the XML configuration only.

sebastianbergmann:
Thank you so much for your feedback!

PR #5343, opened 2023-04-23
Provide distinct messages when a test is considered risky due to output buffering level mismatch

cosmastech:
Currently, a test is marked risky if it didn't close its output buffer OR if it closed additional output buffers.

This PR will give the developer granularity to see which of the two cases caused the issue.


cosmastech:
> Can you please send this against `10.1` (instead of against `main`)? Thanks!

Done. Thank you!

PR #5360, opened 2023-05-11
make attribute CoverFunction available for setting on methods

benblub:
```php
    #[CoversFunction('DummyClass::foo')]
    public function testSomething()
    {
        // ..
    }
```

### DIFF
--- a/src/Framework/Attributes/CoversFunction.php
+++ b/src/Framework/Attributes/CoversFunction.php
@@ -16,7 +16,7 @@ use Attribute;
  *
  * @no-named-arguments Parameter names are not covered by the backward compatibility promise for PHPUnit
  */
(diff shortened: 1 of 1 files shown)


sebastianbergmann:

Going forward, code coverage metadata is only supported on the class-level.

brandonburkett:
I have been searching for way to long on this,  but is it simply not possible using attributes to achieve something like the follow for class method annotations?

```
/**
 * @coversDefaultClass \App\Services\SomeClass
 */
 class SomeClassTest extends TestCase
{
    /**
     * @test
     * @covers ::myMethod()
     */
    public function canMyMethodDoX(): void
    {
    }
    
    /**
     * @test
     * @covers ::myOtherMethod()
     */
    public function canMyOtherMethodDoX(): void
    {
    }
}
```

Thank you for your time.

PR #5365, opened 2023-05-19
Use different exit codes for different failuretypes when exiting for these is requested

theseer:
This PR implements an enhancement to ShellExitCodeCalculator to have different exit codes for different failure types


theseer:
I'm not currently sure why there is a test failure on ONE system with 8.1.x.

sebastianbergmann:

PR #5387, opened 2023-05-27
Enhancement: Pass all directories and names at once

localheinz:
This pull request

- [x] passes all directories and names at once instead of using multiple method calls

### DIFF
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -10,16 +10,20 @@ EOF;
 
 $finder = PhpCsFixer\Finder::create()
     ->files()
-    ->in(__DIR__ . '/src')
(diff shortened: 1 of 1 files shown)


.php-cs-fixer.dist.php
```diff
@@ -10,16 +10,20 @@
...
+        __DIR__ . '/src',
+        __DIR__ . '/tests/_files',
+        __DIR__ . '/tests/end-to-end',
+        __DIR__ . '/tests/unit',
+    ])
+    ->notName([
+        '*.phpt',
+        'ClassWithAllPossibleReturnTypes.php',
+        'ClassWithStaticReturnTypes.php',
+        'ClassWithUnionReturnTypes.php',
```
localheinz, line 23: Switched the order here - now they are ordered by value.

Issue #5400, opened 2023-05-31
Data collected during phar based extension loading get swallowed on failure


theseer:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 10.2-g1313279ad
| PHP version         | 8.2.6
| Installation Method | Source

#### Summary

The `PharLoader` collects a lot of messages when loading extensions from a phar archive.
These messages are not shown in the CLI output

#### Current behavior

No output of the loading process _on error_ <strike>, neither success nor failure.</strike>

#### How to reproduce

Enable any extension, which fails to load.

#### Expected behavior

Success or error output per extension loaded / attempted.

theseer:
I actually didn't check if that problem exists in PHPUnit 9. I'd assume it's not.

theseer:
Okay, found it. 

While we catch issues at bootstraping, we also do a hard exit in [Application::bootstrapExtensions](https://github.com/sebastianbergmann/phpunit/blob/main/src/TextUI/Application.php#L373), when any error occurs.

When the loading of the extension didn't work for any reason consequently the requested bootstrapping fails as well. So that much makes sense.

Unfortunately, the [bootstrapping](https://github.com/sebastianbergmann/phpunit/blob/main/src/TextUI/Application.php#L120) is called before the extension loading results are [shown](https://github.com/sebastianbergmann/phpunit/blob/main/src/TextUI/Application.php#L147), leading to a loss of information.

I'm not sure how to best approach this.

theseer:
The problem only seems to occur on load failures. I see extension information when the extension successfully loaded.
I updated the issue accordingly.

PR #5401, opened 2023-05-31
Catch catchable things from require when loading extensions from phar

theseer:
This fixes the crash mentioned in https://github.com/ergebnis/phpunit-slow-test-detector/pull/273#issuecomment-1570216686

### DIFF
--- a/src/Runner/Extension/PharLoader.php
+++ b/src/Runner/Extension/PharLoader.php
@@ -77,7 +77,7 @@ final class PharLoader
             }
 
             try {
-                /*
+                /**
                  * @psalm-suppress UnresolvableInclude
                  */
                 @require $file;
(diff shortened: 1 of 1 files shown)


theseer:
I don't think my change caused the test breakage on 8.3.

I'm also not sure how to better test the catching unless we have #5400 covered.

sebastianbergmann:
Cherry-picked 7fa0cde5985a0948c1546c9813f8101aa358ac05 into `10.1` and merged to `main` from there.

ab0f6c40ffbbba686f99254fcb68040290917363 is not needed (after f37fc18f5309c96fd2ff9e3747505d46dd4f49bd and 5a9577359cc90b65fe639dc923ceee79633e5124).

PR #5442, opened 2023-07-10
Baseline

### LINKED ISSUE #5441: Baseline for `E_(USER_)DEPRECATED`, `E_(USER_)NOTICE`, `E_STRICT`, and `E_(USER_)WARNING`
PHPUnit's test runner registers an [error handler](https://www.php.net/manual/en/function.set-error-handler.php) and processes `E_DEPRECATED`, `E_USER_DEPRECATED`, `E_NOTICE`, `E_USER_NOTICE`, `E_STRICT`, `E_WARNING`, and `E_USER_WARNING` errors, for instance, that are triggered while a test is executed. This error handler emits events that are, for instance, subscribed to and used by output printers and loggers.

Since #5293, the reporting of deprecations, notices, and warnings can be limited to specified directories. Since #5328, there is an option to ignore the suppression of deprecations, notices, and warnings. Since #5428, the `#[WithoutErrorHandler]` attribute can be used to disable PHPUnit's error handler for a test method.

Following the example of static analysis tools such as Psalm and PHPStan, PHPUnit should support the concept of a [baseline](https://phpstan.org/user-guide/baseline) to

> declare the currently reported list of errors as “the baseline” and cause it not being reported in subsequent runs. It allows you to be interested in [issues] only in new and changed code

- [x] Implement `baseline` attribute on the `<source>` element of PHPUnit's XML configuration file to configure the path to the baseline file to be used
- [x] Implement `--generate-baseline` CLI option to generate a baseline file (at the location configured in the XML configuration file, see above)
- [x] Make paths in `Issue` objects relative to `baseline.xml` location
- [x] Use the baseline (when it is configured) to not display information about `E_(USER_)DEPRECATED`, `E_(USER_)NOTICE`, `E_STRICT`, and `E_(USER_)WARNING` issues that are "baselined"
- [x] Change `--generate-baseline` CLI option (see above) to require the specification of the path to where the baseline file should be written
- [x] Implement `--use-baseline <path>` CLI option to specify the baseline file to be used
- [x] Implement `--ignore-baseline` CLI option to ignore the baseline

sebastianbergmann:


sebastianbergmann:
@localheinz @theseer If you (or anyone else, for that matter), have some time to spare (haha!, sorry), I would appreciate your review. I am not primarily looking for feedback on architecture/design of the implementation, but rather for feedback on UX/DX aspects.

One thing that I already know that I would like to implement next (and in time for the release) is a CLI option to ignore the baseline as well as printing the number of ignored issues when 1) there are ignored issues and 2) the baseline is not ignored.

ondrejmirtes:
@sebastianbergmann If I understand it correctly, you're talking about this logic:

Let's say that the user asks for a new baseline file that's going to be in `/foo/bar/baseline.neon`. And in the baseline file there's an pre-existing error that's reported in `/foo/baz/Lorem.php`.

This logic needs to figure out that the relative path mentioned in `baseline.neon` is `../baz/Lorem.php`, right?

If we're talking about the same thing, in PHPStan the logic is implemented here:

* Implementation: https://github.com/phpstan/phpstan-src/blob/1.11.x/src/File/ParentDirectoryRelativePathHelper.php
* Tests: https://github.com/phpstan/phpstan-src/blob/1.11.x/tests/PHPStan/File/ParentDirectoryRelativePathHelperTest.php

Maybe there's an easier way to do this, I'm not good at making these algorithms look elegant in code :)

sebastianbergmann:
> If we're talking about the same thing, in PHPStan the logic is implemented here:

We are!

>  * Implementation: https://github.com/phpstan/phpstan-src/blob/1.11.x/src/File/ParentDirectoryRelativePathHelper.php
>  * Tests: https://github.com/phpstan/phpstan-src/blob/1.11.x/tests/PHPStan/File/ParentDirectoryRelativePathHelperTest.php

Thank you for pointing me to these sources, I shall have a look ASAP.

tests/_files/baseline/expected.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0"?>
+<files>
+ <file path="FileWithIssues.php">
+  <line number="10" hash="6bf70f0b8c461415955e2d3a97cfbe664f5b957b">
```
localheinz, line 4: @sebastianbergmann 

If I understand correctly, we combine issues and caching in a single file. I believe it makes sense from the perspective of maintaining the feature (less complex), but may be a bit unexpected from the point of view of a user.

### Reasons for changes in the baseline

For example, in projects in which I use static code analysis, I typically only regenerate the baseline (single step) instead of running a static code analysis first, then fixing or regenerating the baseline second (two steps).

Let's ignore structural changes!

With `phpstan/phpstan`, the regenerated baseline has **two** reasons to change:

- an issue has been added to the baseline
- an issue has been removed from the baseline

With `vimeo/psalm`, the regenerated baseline has **three** reasons to change:

- an issue has been added to the baseline
- an issue has been removed from the baseline
- the version number of `vimeo/psalm` has changed

With `phpunit/phpunit`, the regenerated baseline has **four** reasons to change:

- an issue has been added to the baseline
- an issue has been removed from the baseline
- the line number in which an issue has been reported has changed
- the hash of a line in which an issue has been reported has changed

Not sure if that bothers a user, but you need to look more closely at the diff of the baseline if you care about it, to find out if you have added or removed an issue to the baseline.

What do you think?

sebastianbergmann:
@ondrejmirtes In edf06386b6691abe19b8d4d9beeaaaf591828043, I have copied your implementation and tests from PHPStan. I hope that is okay.

Maybe we (c|sh)ould extract this class into its own separate library?

ondrejmirtes:
>  I have copied your implementation and tests from PHPStan. I hope that is okay.

Yes, it's okay :)

I'm not a huge fan of ultra-small libraries although last time something similar was proposed to me, it turned out pretty okay, and I started using it in PHPStan :) https://github.com/theofidry/cpu-core-counter

But in this case I'm not sure, it's more likely I'll need some small adjustments and then it's hard to come up with a proper abstraction, instead of just editing the lines in PHPStan.

PR #5452, opened 2023-07-26
feat: make prophesize() static to ease migration to PHPUnit 10

dunglas:
Following the discussion in https://github.com/phpspec/prophecy-phpunit/pull/56#issuecomment-1651719897, I crafted a patch that makes `TestCase::prophesize()` static. This would allow `phpspec/phpunit-prophecy`  to provide a static implementation of `prophesize()` compatible with both PHPUnit 9 and 10.

I tried to keep the patch as small as possible, but as you pointed out @sebastianbergmann, `addWarning` and `recordTestDoubles` require some refactoring to be handled properly.

I tried this patch on API Platform, it works and allows running our whole unit test suite with both PHPUnit 9 and PHPUnit 10, using static data providers.

Feel free to close if the patch is too large.

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -313,7 +313,14 @@ abstract class TestCase extends Assert implements Reorderable, SelfDescribing, T
     /**
      * @var \Prophecy\Prophet
      */
-    private $prophet;
+    private static $prophet;
+
+    /**
+     * @internal
+     *
+     * @var string[]
+     */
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it as I am not comfortable with a change this large for a branch that is closed for all changes except for bug fixes and changes to make it compatible with future PHP versions. Sorry :-(

PR #5462, opened 2023-08-05
Support for multiple arguments

joke2k:
This work adds the support for providing multiple files or directories as arguments to PHPUnit:

```
./phpunit FooTest.php test/
```

It is very useful when you want to test multiple things together without having a Suite defined in the configuration, and most important (for me) is that it works for CircleCI's parallelisation command (apparently phpunit is one of the few test frameworks without this feature).

It could be easily extended for phpunit9.

**Point to discuss**
- naming the Test suite: a proposal is using a hash (md5) of all provided arguments (the list could be very long)

### DIFF
--- a/src/TextUI/Configuration/TestSuiteBuilder.php
+++ b/src/TextUI/Configuration/TestSuiteBuilder.php
@@ -10,6 +10,7 @@
 namespace PHPUnit\TextUI\Configuration;
 
 use function assert;
+use function count;
 use function is_dir;
 use function is_file;
 use function realpath;
@@ -38,16 +39,29 @@ final class TestSuiteBuilder
     public function build(Configuration $configuration): TestSuite
     {
         if ($configuration->hasCliArguments()) {
-            $argument = realpath($configuration->cliArguments()[0]);
+            $arguments = [];
 
-            if (!$argument) {
-                throw new TestFileNotFoundException($configuration->cliArguments()[0]);
+            foreach ($configuration->cliArguments() as $cliArgument) {
+                $argument = realpath($cliArgument);
+
+                if (!$argument) {
+                    throw new TestFileNotFoundException($cliArgument);
+                }
+
+                $arguments[] = $argument;
             }
 
-            $testSuite = $this->testSuiteFromPath(
-                $argument,
-                $configuration->testSuffixes(),
-            );
+            if (count($arguments) === 1) {
+                $testSuite = $this->testSuiteFromPath(
+                    $arguments[0],
+                    $configuration->testSuffixes(),
+                );
+            } else {
+                $testSuite = $this->testSuiteFromPathList(
+                    $arguments,
+                    $configuration->testSuffixes(),
+                );
+            }
         }
 
         if (!isset($testSuite)) {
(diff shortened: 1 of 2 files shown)


sebastianbergmann:
> It could be easily extended for phpunit9.

PHPUnit 9 is closed for changes other than bug fixes and adaptions to new versions of PHP.

joke2k:
Thank you for the reply and to take into consideration this work.

I applied the CSFixer rules, sorry about that.

I was wondering how to tell to psalm that `$configuration->cliArgument()` cannot raise a `NoCliArgumentException` in the `TestSuiteBuilder::build` because a few lines above `$configuration->hasCliArgument()` has been checked, should we add:
```
            // @psalm-suppress NoCliArgumentException
            foreach (explode(' ', $configuration->cliArgument()) as $cliArgument) {
```
or in the signature of the method?

joke2k:
@sebastianbergmann do you have any ideas on how to satisfy psalm in this case?

sebastianbergmann:
I do not have time to look into this right now.

joke2k:
@sebastianbergmann this is ready to be checked again. psalm issue should be fixed.

Thank you

src/TextUI/Configuration/TestSuiteBuilder.php
```diff
@@ -38,16 +42,27 @@ final class TestSuiteBuilder
...
+                }
+
+                $arguments[] = $argument;
             }
 
-            $testSuite = $this->testSuiteFromPath(
-                $argument,
-                $configuration->testSuffixes(),
-            );
+            $testSuite = count($arguments) === 1
```
sebastianbergmann, line 57: Please do not use the ternary operator here.

sebastianbergmann:
> naming the Test suite: a proposal is using a hash (md5) of all provided arguments (the list could be very long)

As of 51f93ba6af4ebdfdebc04de04163c055cd70d842, the name of the top-level test suite that is created when a directory or file path is passed as an argument to the test runner is now always `CLI Arguments`.

src/TextUI/Configuration/TestSuiteBuilder.php
joke2k: splitted using a more readable if-else.

sebastianbergmann:
Thank you for patience. I did not like the `implode()` ... `explode()` workaround that you had to use. As of d0103c021ebd4323abc2fec6a7d5d9c701c10da9, this is no longer necessary. Once you adapted your PR to that, I think we can merge this. Thanks!

joke2k:
Adapted the code with your main to use the new cliArgument methods.
I tested locally but I recently updated my php to 8.1.22 and I receive some wired error on some phpt tests like:
```
/open-source/phpunit/tests/end-to-end/regression/765.phpt
Failed asserting that string matches format description.
--- Expected
+++ Actual
@@ @@
 
 ERRORS!
 Tests: 1, Assertions: 1, Errors: 1.
+Standard input code:8:
 int(2)
```
but I can see that your pipeline is working on the main, so I think is an issue only on my laptop.

sebastianbergmann:
This branch cannot be rebased due to conflicts.

joke2k:
Hi @sebastianbergmann, I squashed all the commits and rebased the branch.

PR #5485, opened 2023-08-22
Allow pure enums as method parameters

ericfortmeyer:


sebastianbergmann:
Cherry-picked into `10.3` and merged to `main` from there. Thank you!

PR #5490, opened 2023-08-28
Actually run the tests on Windows

jrfnl:

image
Src: https://github.com/sebastianbergmann/phpunit/actions/runs/5989020062/job/16244808391

Note: I've double-checked and confirmed that the tests really aren't running on Windows by introducing a test failure - this shows that the [Ubuntu tests fail (as expected), but that the Windows tests do not fail (as they aren't running)](https://github.com/jrfnl/phpunit/actions/runs/6001033906).
A test build with the same test failure + this commit shows that the [Windows tests will now fail correctly (as they will now run)](https://github.com/jrfnl/phpunit/actions/runs/6001062381).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -101,10 +101,10 @@ jobs:
           tools: none
 
       - name: Install dependencies with Composer
-        run: ./tools/composer update --no-ansi --no-interaction --no-progress
+        run: php ./tools/composer update --no-ansi --no-interaction --no-progress
 
       - name: Run tests with PHPUnit
-        run: ./phpunit --testsuite unit
+        run: php ./phpunit --testsuite unit
 
   end-to-end-tests:
     name: End-to-End Tests
@@ -151,10 +151,10 @@ jobs:
           tools: none
 
       - name: Install dependencies with Composer
-        run: ./tools/composer update --no-ansi --no-interaction --no-progress
+        run: php ./tools/composer update --no-ansi --no-interaction --no-progress
 
       - name: Run tests with PHPUnit
-        run: ./phpunit --testsuite end-to-end
+        run: php ./phpunit --testsuite end-to-end
 
   code-coverage:
     name: Code Coverage


jrfnl:
Also note: I intended to pull this against 8.5 (yes, I've updated the base branch now), but accidentally pulled this against `main`. While the build will pass with this change for PHPUnit 8.5, it will [fail against `main`](https://github.com/sebastianbergmann/phpunit/actions/runs/6001155534) due to incompatibilities with Windows (either in the test code or the src code, I haven't investigated)

sebastianbergmann:
Thank you for working on this, Juliette.

However, it has been "ages" since I have come across developers that use native binaries on Windows. For years, developers that use PHP on Windows either use WSL or Docker containers. Against that background, I wonder whether we really need to care about running the tests on Windows.

localheinz:
@jrfnl 

I remember doing a bit of work related to running tests on GitHub Actions on Windows, but I stopped pursuing it for the following reasons:

- as a long-time Windows (and DOS) user, I switched to macOS in 2010, and I am not looking back
- as a macOS user, when you neither have nor ever again want to have a computer that runs Windows, debugging failing GitHub Actions jobs on Windows is painful

jrfnl:
> I wonder whether we really need to care about running the tests on Windows.

@sebastianbergmann Good question. I personally use the PHPUnit PHAR file and Windows PHP binaries, but then, I switch PHP versions multiple times a day, often even within the same command. Then again, I'm pretty sure I'm the exception, not the rule, though I imagine I'm not the only one working is open source who does this.

I can imagine multiple alternatives to this PR:
* Only run against Windows OS for high/low PHP - should safeguard things sufficiently.
* Remove the builds against Windows completely.

As I said above, against the PHPUnit 8.x branch, the test runs pass on Windows, I haven't checked against the 9.x branch, on the 10.x branch/main, the tests fail, but it's 24 failures on 2800 tests, so I imagine fixing those (or explicitly ignoring them) is a relatively small effort.

Seeing @localheinz's remark, I could probably debug those 24 next week if so desired.

sebastianbergmann:
Merged into `8.5` and to `9.6` from there. However, some end-to-end tests fail for [PHPUnit 9.6](https://github.com/sebastianbergmann/phpunit/actions/runs/6024715638/job/16344052227) on Windows. Reverted for now as of c7e801e2751a879f3c28aca102a4f4e44a773748.

> I could probably debug those 24 next week if so desired.

That would be appreciated. And no rush, it does not have to be next week :-) Thanks!

jrfnl:
> > I could probably debug those 24 next week if so desired.
> 
> That would be appreciated. And no rush, it does not have to be next week :-) Thanks!

Better for it to be off my plate than for it to end up on my never ending "to do when I find some time"-list.

PR for the 9,x branch is now open (#5500). Next up: check if any extra fixes are needed for the 10.x branch.

PR #5492, opened 2023-08-30
Combine filter options with test/group lists

### LINKED ISSUE #5552: Combine filter options with test/group lists
The options `--list-tests`, `--list-tests-xml` and `--list-groups` cannot be combined with most test-filtering options. I have not found why this was done, but possibly it was because `RecursiveIteratorIterator` could not work with PHPUnit's filter iterators*.

<details><summary>*) technical reason</summary>

PHPUnit filter iterators extend `RecursiveFilterIterator`. That class seems to make its child iterators like this: `$childIterator = new static($iterableMember);`. That fails when the child class adds parameters to the constructor, with an exception like this:

```
Message:  Too few arguments to function PHPUnit\Runner\Filter\NameFilterIterator::__construct(), 1 passed and exactly 2 expected
Location: src/Runner/Filter/NameFilterIterator.php:40

#0 [internal function]: PHPUnit\Runner\Filter\NameFilterIterator->__construct()
#1 src/TextUI/Command/Commands/ListTestsAsTextCommand.php(38): RecursiveFilterIterator->getChildren()
...
```
</details>

## Why

Allowing the filtering option to be combined with the listing functions has the following benefits:

- Outputting filtered lists can be useful for trying out filters.
- This will allow external tools to use phpunit's filtering functions.

Black-Mantha:
Closes #5552

A few questions/comments I have for the reviewer:
- Do I need to add a commit adding a line to `ChangeLog-10.4.md` or is that done by the merger? I would add this pull request's title under the "Changed" header.
- Combining these options used to produce a warning, not an error. Does that mean this is technically a breaking change? (I've put a comment on the issue to discuss this)
- I could have solved this by fixing the filter iterators, but in that case the responsibility of recursing the filters would have moved from `TestSuite::injectFilter` to the iterators, which would mean the filters don't work when you recurse manually, such as in `TestRunner::run()`.
- `--list-test-suites` uses a completely different set of classes, so I would have to alter quite a bit to make it work.
- In a previous version, I injected the filter in `Application::run()`. This meant I didn't have to change `executeCommandsThatRequireCliConfigurationAndTestSuite`, but also caused `Ignore*ForCodeCoverage` annotations to be ignored if they were in filtered-out tests. I believed this to be undesirable. Was this the right call?

### DIFF
--- a/src/Metadata/Api/CodeCoverage.php
+++ b/src/Metadata/Api/CodeCoverage.php
@@ -25,6 +25,7 @@ use PHPUnit\Metadata\Uses;
 use PHPUnit\Metadata\UsesClass;
 use PHPUnit\Metadata\UsesDefaultClass;
 use PHPUnit\Metadata\UsesFunction;
+use PHPUnit\Util\TestSuiteIteratorIterator;
 use SebastianBergmann\CodeUnit\CodeUnitCollection;
 use SebastianBergmann\CodeUnit\InvalidCodeUnitException;
 use SebastianBergmann\CodeUnit\Mapper;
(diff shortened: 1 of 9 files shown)


sebastianbergmann:

PR #5500, opened 2023-09-04
[9.6] GH Actions: actually run the tests on Windows (includes necessary fixes)

jrfnl:
Sister-PR to #5490 for the 9.x series.

----

### E2E Tests: use OS agnostic directory separators

This fixes a number of tests which would otherwise fail when running on Windows.

### E2E/DiffTest: fix line endings

As far as I can see, this test is about the display of the diff, not about the line endings, in which case, let's make the line endings explicit and not OS-dependent to allow this test to pass on Windows as well.

### E2E/diff-colorized: add separate test for Windows

The colorized diff display on Windows is different from *nix systems.

This commit:
* Adds a test skip for the pre-existing test when the tests are run on Windows.
* Adds a separate test for the colorized diff display on Windows, which is skipped when running on *nix systems.

### GH Actions: actually run the tests on Windows

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -101,10 +101,10 @@ jobs:
           tools: none
 
       - name: Install dependencies with Composer
-        run: ./tools/composer update --no-ansi --no-interaction --no-progress
+        run: php ./tools/composer update --no-ansi --no-interaction --no-progress
 
       - name: Run tests with PHPUnit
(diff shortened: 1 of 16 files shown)


jrfnl:
Note: I'd recommend for this PR to just be merged to the 9.x branch and not merged up to 10.x. I'll open a separate PR for 10.x.

PR #5501, opened 2023-09-04
[10.3] GH Actions: actually run the tests on Windows (includes necessary fixes)

jrfnl:
Sister-PR to #5490 and #5500 for the 10.x series.

Note: the build does not (yet) pass. This is related to a finding from the tests, which points to a potential bug - see commit https://github.com/sebastianbergmann/phpunit/commit/4c478fe9e7858ea631fc6f2f77571ff4f5c33240 for more info.

---

### E2E Tests: use OS agnostic directory separators

This fixes a number of tests which would otherwise fail when running on Windows.

### E2E/DiffTest: fix line endings

As far as I can see, this test is about the display of the diff, not about the line endings, in which case, let's make the line endings explicit and not OS-dependent to allow this test to pass on Windows as well.

### E2E/diff-colorized: add separate test for Windows

The colorized diff display on Windows is different from *nix systems.

This commit:
* Adds a test skip for the pre-existing test when the tests are run on Windows.
* Adds a separate test for the colorized diff display on Windows, which is skipped when running on *nix systems. (Note: this is an adjusted test compared to the 9.x version)

### Unit tests: compare multi-line expectations ignoring line endings

As the test files are saved using linux lined endings, while the actual output is generated using the `PHP_EOL` constant, the output will not match unless the line ending type is ignored.

### Unit Tests/SourceFilterTest: use OS based directory separators

Allow the `SourceFilter` class to actually find the file when included by using the correct OS based directory separator.

### Unit Tests/IsEqual*Tests: fix line endings

As far as I can see, these tests are about the "is equal" constraint, not about the line endings, in which case, let's make the line endings explicit and not OS-dependent to allow the failure message comparison test to pass on Windows as well.

### Unit Tests/ResultPrinterTest: allow for slight difference in resulting output on Windows

Not sure if this should be considered a bug, but the output does look different on Windows.

For now, I've added a separate expectation file for Windows to document the difference.

### Unit Tests/SourceMapperTest: use OS based directory separators

While this fixes the majority of the test failure diff, it does actually highlight something which could be considered a bug and for which I'd like a second opinion before investing time in this:
```
1) PHPUnit\TextUI\Configuration\SourceMapperTest::testDeterminesWhetherFileIsIncluded with data set "file included using directory"
--- Expected
+++ Actual
@@ @@
 Array &0 (
     'E:\path\to\PHPUnit\phpunit\tests\_files\source-filter\a\PrefixSuffix.php' => true
+    'E:\path\to\PHPUnit\phpunit\tests\_files\source-filter\a\c\.hidden\PrefixSuffix.php' => true
     'E:\path\to\PHPUnit\phpunit\tests\_files\source-filter\a\c\Prefix.php' => true
     'E:\path\to\PHPUnit\phpunit\tests\_files\source-filter\a\c\PrefixSuffix.php' => true
     'E:\path\to\PHPUnit\phpunit\tests\_files\source-filter\a\c\Suffix.php' => true

2) PHPUnit\TextUI\Configuration\SourceMapperTest::testDeterminesWhetherFileIsIncluded with data set "file included using directory, but excluded using file"
--- Expected
+++ Actual
@@ @@
 Array &0 (
+    'E:\path\to\PHPUnit\phpunit\tests\_files\source-filter\a\c\.hidden\PrefixSuffix.php' => true
     'E:\path\to\PHPUnit\phpunit\tests\_files\source-filter\a\c\Prefix.php' => true
     'E:\path\to\PHPUnit\phpunit\tests\_files\source-filter\a\c\PrefixSuffix.php' => true
     'E:\path\to\PHPUnit\phpunit\tests\_files\source-filter\a\c\Suffix.php' => true
```

### GH Actions: actually run the tests on Windows

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -98,10 +98,10 @@ jobs:
           tools: none
 
       - name: Install dependencies with Composer
-        run: ./tools/composer update --no-ansi --no-interaction --no-progress
+        run: php ./tools/composer update --no-ansi --no-interaction --no-progress
 
       - name: Run tests with PHPUnit
-        run: ./phpunit --testsuite unit
+        run: php ./phpunit --testsuite unit
 
   end-to-end-tests:
     name: End-to-End Tests
@@ -146,10 +146,10 @@ jobs:
           tools: none
 
       - name: Install dependencies with Composer
-        run: ./tools/composer update --no-ansi --no-interaction --no-progress
+        run: php ./tools/composer update --no-ansi --no-interaction --no-progress
 
       - name: Run tests with PHPUnit
-        run: ./phpunit --testsuite end-to-end
+        run: php ./phpunit --testsuite end-to-end
 
   code-coverage:
     name: Code Coverage
--- a/tests/end-to-end/cli/filter-error-handler/filter-for-error-handler-events-disabled.phpt
+++ b/tests/end-to-end/cli/filter-error-handler/filter-for-error-handler-events-disabled.phpt
@@ -16,7 +16,7 @@ require_once __DIR__ . '/../../../bootstrap.php';
 PHPUnit %s by Sebastian Bergmann and contributors.
 
 Runtime: %s
-Configuration: %s/filter-disabled.xml
+Configuration: %s%efilter-disabled.xml
 
 W                                                                   1 / 1 (100%)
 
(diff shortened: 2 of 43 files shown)


jrfnl:
> Unit Tests/SourceMapperTest: use OS based directory separators
>
> While this fixes the majority of the test failure diff, it does actually highlight something which could be considered a bug and for which I'd like a second opinion before investing time in this:

@sebastianbergmann Have you got an opinion on this one ? Is this a bug to be fixed in the src or something I should work around in the tests ?

sebastianbergmann:
> Have you got an opinion on this one?

Not yet, today is my first day back from vacation. I hope to get to this soon, sorry to keep you waiting.

jrfnl:
> > Have you got an opinion on this one?
> 
> Not yet, today is my first day back from vacation. I hope to get to this soon, sorry to keep you waiting.

No rush. I won't have time to look at this until after the weekend anyway (conference). Just wanted to make sure that it was clear that the reason the build isn't passing is a known one with a question attached ;-)

sebastianbergmann:
I have merged this PR (thank you!), but disabled the test in question on Windows. I have created #5503 for the investigation into this issue.

sebastianbergmann:
I will take care of the new [test failures](https://github.com/sebastianbergmann/phpunit/actions/runs/6106124959/job/16570710139) on Windows on the `main` branch.

jrfnl:
> I will take care of the new [test failures](https://github.com/sebastianbergmann/phpunit/actions/runs/6106124959/job/16570710139) on Windows on the `main` branch.

Thanks. I just had a quick look and those all look like they need the `%e` placeholder for the slashes like in a637e06, but I suppose you already sorted that ;-)

sebastianbergmann:
Yes, `main` was easy. The branch for #5442, however, requires quite some work. I'm on it, but it'll wait until after I am done with "real work" for the day.

PR #5507, opened 2023-09-12
Fix: Do not configure deprecated `function_typehint_space` fixer

localheinz:
This pull request

- [x] stops configuring the deprecated `function_typehint_space` fixer

Follows 187a9c4c8.

### DIFF
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -106,7 +106,6 @@ $config->setFinder($finder)
         'fully_qualified_strict_types' => true,
         'function_declaration' => true,
         'function_to_constant' => true,
-        'function_typehint_space' => true,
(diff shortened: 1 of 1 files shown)


.php-cs-fixer.dist.php
```diff
@@ -106,7 +106,6 @@
         'fully_qualified_strict_types' => true,
         'function_declaration' => true,
         'function_to_constant' => true,
-        'function_typehint_space' => true,
```
localheinz, line 109: The corresponding alternative has already been configured, see

https://github.com/sebastianbergmann/phpunit/blob/c70b8108ac4cab82f67c66459ac74b3cdaef563e/.php-cs-fixer.dist.php#L330-L334

PR #5572, opened 2023-11-21
Generate diff when comparing enums with assertSame

JanTvrdik:
PHPUnit's Exporter supports enums, but when comparing enums with `assertSame`, the diff is currently not generated and the only message is "Failing to assert that two variables reference the same object".

This PR improves the underlying `IsIdentical` constraint to generate human readable diff by utilizing existing enum support in `Exporter`.


sebastianbergmann:
This should target the `10.4` branch. Thanks!

sebastianbergmann:
Sorry about that, that was not intended. The `10.4` branch was deleted because there will be no more PHPUnit 10.4 releases. I am sorry for the inconvenience, but can you please send this against the `10.5` branch? Thanks! I'll merge it ASAP then.

JanTvrdik:
Resent as https://github.com/sebastianbergmann/phpunit/pull/5582

PR #5575, opened 2023-11-25
encode arguments ordering in equality assertion functions

miranalmehrab:
Hello, I would like to add some functions that are more of function aliases with arguments embedded in their names. For example, while working on a Laravel project I was calling the ``$this->assertEquals($actual, $expected)``, whereas the correct way was ``$this->assertEquals($expected, $actual)``. While reading the **Clean Code**  by **Robert C. Martin** the other day, I came across the same issue regarding the lack of natural ordering of the arguments. So I hope, these explicit function names will help someone in the future.

### DIFF
--- a/src/Framework/Assert.php
+++ b/src/Framework/Assert.php
@@ -277,6 +277,16 @@ abstract class Assert
     }
 
     /**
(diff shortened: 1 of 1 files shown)


sebastianbergmann:

PR #5591, opened 2023-11-30
Skip less E2E tests on Windows

mvorisek:
103 E2E tests were skipped on Windows, 93 were passing without any change needed.

The remaining 10 are adjusted to pass with `\` Windows directory separator.


sebastianbergmann:
Thank you!

Issue #5428, opened 2023-06-26
Attribute `#[WithoutErrorHandler]` to disable PHPUnit's error handler for a test method


jrfnl:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 10*
| PHP version         | 8.2 (though irrelevant)
| Installation Method | Composer / PHAR (either)

#### Summary

Code which relies on retrieving the last error message and acting on it can no longer be tested as `error_get_last()` will always return `null`.

This can also lead to additional PHP notices being reported by PHPUnit for errors which would never occur in a real-life situation.

#### Current behavior

Tests which are testing code which relies on `error_get_last()` are failing on PHPUnit 10 with notices along these lines:
```
There were 2 failures:

1) Jrf\PHPUnit10\Scenario\Tests\FooTest::testMethodA
Failed asserting that exception message '' contains 'Failed to open stream'.

2) Jrf\PHPUnit10\Scenario\Tests\FooTest::testMethodB
Failed asserting that null is identical to 'Triggering'.

path/to/bug-report-reproduction-scenarios/tests/FooTest.php:22

--

2 tests triggered 2 PHP warnings:

1) Jrf\PHPUnit10\Scenario\Tests\FooTest::testMethodA
Trying to access array offset on value of type null
path/to/bug-report-reproduction-scenarios/src/Foo.php:17

path/to/bug-report-reproduction-scenarios/tests/FooTest.php:12

2) Jrf\PHPUnit10\Scenario\Tests\FooTest::testMethodB
Trying to access array offset on value of type null
path/to/bug-report-reproduction-scenarios/tests/FooTest.php:22

path/to/bug-report-reproduction-scenarios/tests/FooTest.php:21

--

1 test triggered 1 PHP deprecation:

1) Jrf\PHPUnit10\Scenario\Tests\FooTest::testMethodA
Exception::__construct(): Passing null to parameter #1 ($message) of type string is deprecated
path/to/bug-report-reproduction-scenarios/src/Foo.php:17

path/to/bug-report-reproduction-scenarios/tests/FooTest.php:12
```

#### How to reproduce

Reproduction scenario available at: https://github.com/jrfnl/bug-report-reproduction-scenarios/commit/b3d93663ef8b07600b0d8029279f1bb94d56cac1

GH actions build for the reproduction branch demonstrating the issue: https://github.com/jrfnl/bug-report-reproduction-scenarios/actions/runs/5378708821

#### Expected behavior

[...]

PR #5592, opened 2023-11-30
Fix error handler to not silence `error_get_last()` result

### LINKED ISSUE #5587: `error_get_last()` is not set when errors are logged
related with #5428

| Q                   | A
| PHPUnit version     | 10.4.2
| PHP version         | 8.2.12

#### Current behavior

`error_get_last()` is not set when errors are logged (PHPUnit error handler is active). The errors are logged by default which is a good. But the current impl. of the error handler causes code that relies on `error_get_last()` set after a core php function has returned a failure like:
```
$dirHandle = opendir($dirPath);
if ($dirHandle === false) {
    $errorMessage = preg_replace('~^.+?:\s*~s', '', error_get_last()['message']);
}
```
to fail as `error_get_last()` is empty.

This is because the current impl. of the error handler returns `true` [1] when an error is handled. The php docs say [2]:
> It is important to remember that the standard PHP error handler is completely bypassed for the error types specified by error_levels unless the callback function returns false.

[1] https://github.com/sebastianbergmann/phpunit/blob/10.4.2/src/Runner/ErrorHandler.php#L144
[2] https://www.php.net/manual/en/function.set-error-handler.php

The issue is present since PHPUnit 10, in PHPUnit 9 and lower `error_get_last()` was set correctly.

#### How to reproduce

```
$dirHandle = opendir('/--non-existing-path--');
self::assertNotNull(error_get_last());
```

#### Expected behavior

The PHPUnit handler should:
1. log an error
2. suppress its output
3. not affect the `error_get_last()` result, the result must be the same /w and /wo PHPUnit error handler active

I tried to workaround about this issue https://3v4l.org/kS1vM/rfc#vgit.master but if the output is suppressed by lowering the reporting level when handling the error, it seems there is no way to restore the level back immediatelly after the error handling is finished.

I belive the solution is to modify the PHPUnit error handler:
1. to return `false` when handling the error - it will let php to handle the error natively
2. to backup the current `error_reporting()` level and set it to `error_reporting(error_reporting() & (E_ERROR | E_PARSE | E_CORE_ERROR | E_CORE_WARNING | E_COMPILE_ERROR | E_COMPILE_WARNING))` when the handler activates (the listed error levels cannot be handled by an user handler [2])
3. to restore the backed error level when the error handler deactivates (or better increase the current error level if lower than the backed one)
4. https://github.com/sebastianbergmann/phpunit/commit/e676667054ec44e55678a87dc7d8ee8b0a3637ba original fix and https://github.com/sebastianbergmann/phpunit/commit/c2c8dbbd80b6e5c5ec98c844b37661f575908c7f can be reverted then, no `WithoutErrorHandler` attribute needed

mvorisek:
fix #5587 (and older  #5428)

The updated PHPUnit error handler still supresses the output and still logs the errors as before.

This PR implies no BC break for existing PHPUnit 10.x tests /w and /wo `WithoutErrorHandler` attribute.

The `WithoutErrorHandler` attribute should be deprecated in the next minor version, as the new error handler impl. does not effectively imply any side effects so the behaviour /w and /wo PHPUnit error handler enabled is newly the same. In the next major version, the attribute should be completely removed and related https://github.com/sebastianbergmann/phpunit/commit/c2c8dbbd80b6e5c5ec98c844b37661f575908c7f and https://github.com/sebastianbergmann/phpunit/commit/e676667054ec44e55678a87dc7d8ee8b0a3637ba commits reverted.

In addition to fixing the linked issues:
- `error_clear_last()` is called before every test to clear `error_get_last()`
- `E_USER_ERROR` aborts script execution by default, so newly the updated PHPUnit error handled does the same

### DIFF
--- a/src/Runner/ErrorHandler.php
+++ b/src/Runner/ErrorHandler.php
@@ -17,8 +17,10 @@ use const E_DEPRECATED;
 use const E_ERROR;
 use const E_NOTICE;
 use const E_PARSE;
+use const E_RECOVERABLE_ERROR;
 use const E_STRICT;
 use const E_USER_DEPRECATED;
+use const E_USER_ERROR;
 use const E_USER_NOTICE;
 use const E_USER_WARNING;
 use const E_WARNING;
@@ -37,6 +39,7 @@ use PHPUnit\Util\ExcludeList;
 final class ErrorHandler
 {
     private const UNHANDLEABLE_LEVELS         = E_ERROR | E_PARSE | E_CORE_ERROR | E_CORE_WARNING | E_COMPILE_ERROR | E_COMPILE_WARNING;
+    private const UNSUPPRESSEABLE_LEVELS      = E_ERROR | E_PARSE | E_CORE_ERROR | E_COMPILE_ERROR | E_USER_ERROR | E_RECOVERABLE_ERROR;
     private static ?self $instance            = null;
     private ?Baseline $baseline               = null;
     private bool $enabled                     = false;
@@ -52,7 +55,7 @@ final class ErrorHandler
      */
     public function __invoke(int $errorNumber, string $errorString, string $errorFile, int $errorLine): bool
     {
-        $suppressed = !($errorNumber & error_reporting());
+        $suppressed = (error_reporting() & ~self::UNSUPPRESSEABLE_LEVELS) === 0;
 
         if ($suppressed && (new ExcludeList)->isExcluded($errorFile)) {
             return false;
--- a/src/Runner/ErrorHandler.php
+++ b/src/Runner/ErrorHandler.php
@@ -163,13 +163,7 @@ final class ErrorHandler
             return;
         }
 
-        $oldErrorHandler = set_error_handler($this);
-
-        if ($oldErrorHandler !== null) {
-            restore_error_handler();
-
-            return;
-        }
+        set_error_handler($this);
 
         $this->enabled = true;
 
--- a/src/Runner/ErrorHandler.php
+++ b/src/Runner/ErrorHandler.php
@@ -9,8 +9,14 @@
  */
 namespace PHPUnit\Runner;
 
+use const E_COMPILE_ERROR;
+use const E_COMPILE_WARNING;
+use const E_CORE_ERROR;
+use const E_CORE_WARNING;
 use const E_DEPRECATED;
+use const E_ERROR;
 use const E_NOTICE;
+use const E_PARSE;
 use const E_STRICT;
 use const E_USER_DEPRECATED;
 use const E_USER_NOTICE;
@@ -30,9 +36,11 @@ use PHPUnit\Util\ExcludeList;
  */
 final class ErrorHandler
 {
-    private static ?self $instance = null;
-    private ?Baseline $baseline    = null;
-    private bool $enabled          = false;
+    private const UNHANDLEABLE_LEVELS         = E_ERROR | E_PARSE | E_CORE_ERROR | E_CORE_WARNING | E_COMPILE_ERROR | E_COMPILE_WARNING;
+    private static ?self $instance            = null;
+    private ?Baseline $baseline               = null;
+    private bool $enabled                     = false;
+    private ?int $originalErrorReportingLevel = null;
 
     public static function instance(): self
     {
@@ -146,7 +154,7 @@ final class ErrorHandler
                 return false;
         }
 
-        return true;
+        return false;
     }
 
     public function enable(): void
@@ -164,6 +172,9 @@ final class ErrorHandler
         }
 
         $this->enabled = true;
+
+        $this->originalErrorReportingLevel = error_reporting();
+        error_reporting($this->originalErrorReportingLevel & self::UNHANDLEABLE_LEVELS);
     }
 
     public function disable(): void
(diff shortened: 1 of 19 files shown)


src/Runner/ErrorHandler.php
```diff
@@ -140,13 +152,13 @@ public function __invoke(int $errorNumber, string $errorString, string $errorFil
                     $suppressed,
                 );
 
-                break;
+                throw new ErrorException('Any E_*_ERROR must abort execution');
```
mvorisek, line 155: better throwable class is probably wanted, also, it should extend `\Error` not `\Exception`

sebastianbergmann:
This will not go into a minor version such as PHPUnit 10.5. Please target `main`, thank you.

sebastianbergmann:
@jrfnl What do you think?

sebastianbergmann:
This still targets the `10.5` branch.

sebastianbergmann:
Maybe I am wrong, but this change feels "too big" for me to land in a bugfix release. I will need to think about this.

mvorisek:
Rebase fixed to account for https://github.com/sebastianbergmann/phpunit/pull/5599 changes.

src/Runner/ErrorHandler.php
mvorisek: @sebastianbergmann Thank you for merging this fix. I did not add a custom class as I do not know how to name it and where to put it, so I would be grateful if you can help me and add a custom PHPUnit class for this exception and base it from `\Error` so `catch (\Exception $e)` clauses do not catch/handle it. This should be true for all PHPUnit exceptions.

driesvints:
Hi @sebastianbergmann @mvorisek. It seems this PR breaks Lumen's test suite: https://github.com/laravel/lumen-framework/actions/runs/7373447479/job/20062657751

<details><summary>Details</summary>
<p>

```
Run vendor/bin/phpunit
PHPUnit 10.5.5 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.1.27
Configuration: /home/runner/work/lumen-framework/lumen-framework/phpunit.xml.dist

........................................................FF....... 65 / 67 ( 97%)
..                                                                67 / 67 (100%)

Time: 00:00.109, Memory: 26.00 MB

There were 2 failures:

1) HandleExceptionsTest::testEnsuresDeprecationsDriver
null does not match expected type "array".

/home/runner/work/lumen-framework/lumen-framework/tests/HandleExceptionsTest.php:119

2) HandleExceptionsTest::testEnsuresNullDeprecationsDriver
Failed asserting that null matches expected 'NullHandler'.

/home/runner/work/lumen-framework/lumen-framework/tests/HandleExceptionsTest.php:144

FAILURES!
Tests: 67, Assertions: 129, Failures: 2.
Error: Process completed with exit code 1.
``` 

</p>
</details> 

I can confirm the changes in this PR cause this because if I revert them locally, tests pass again. At this time I can't provide a way to reproduce or exactly pinpoint why this PR causes the tests to fail as I'm too unfamiliar with PHPUnit's internals.

@sebastianbergmann would you maybe consider reverting this PR in a similar fashion as was done with https://github.com/sebastianbergmann/phpunit/pull/5619 ?

sebastianbergmann:
@driesvints Yes, I would consider that.

I am travelling for the next two weeks and will only be able to work on and off. Therefore, it would help me a lot to get this done quickly if you could send a pull request for `10.5` that restores the previous behaviour. Once I have that, I should be able to tag a release ASAP.

mvorisek:
Reverting this will cause issues for users relying on this bugfix.

@driesvints please provide a minimal repro as a testcase and I can look into it.

driesvints:
@sebastianbergmann Thank you, we appreciate that! I've sent in the PR here: https://github.com/sebastianbergmann/phpunit/pull/5637

> please provide a minimal repro as a testcase and I can look into it.

I don't know how sorry.

mvorisek:
Isolate https://github.com/laravel/lumen-framework/blob/v10.0.2/tests/HandleExceptionsTest.php#L119 test into a repro repository with minimal deps/sources.

mvorisek:
Isolate https://github.com/laravel/lumen-framework/blob/v10.0.2/tests/HandleExceptionsTest.php#L119 test into a repro repository with minimal deps/sources.

The issue is probably in https://github.com/laravel/lumen-framework/blob/v10.0.2/src/Concerns/RegistersExceptionHandlers.php#L73 as it relies on specific `error_reporting()` - set that prior your test like `error_reporting(error_reporting() | \E_USER_DEPRECATED)`.

driesvints:
I'm sorry but I don't understand why we should change our code. This has worked perfectly fine for us for years with using PHPUnit. If PHPUnit were to change this in a new major version we'd could look into adopting the new way of doing things. But right now I feel this is a breaking change, sorry.

mvorisek:
@driesvints see the two issues in this PR description, there was big breaking change introduced in PHPUnit 10.0.0 and this PR fixes it - your code relies on specific `error_reporting` error level. I understand you might argue this is breaking change - we cannot satisfy both worlds.

The reason is we MUST NOT HANDLE the php warnings but instead LOG them only. This is what this PR does. To prevent php printing the unhandled php warnings, the `error_reporting` must be set to a lower level. I do not think there is other way to do it. I spent more than a day on figuring the php internals to understand the error handling limitations - if you have any question about error handling or even have an idea how to do things better, feel free to share.

PR #5605, opened 2023-12-08
Implement `expectUserDeprecationMessage()` and `expectUserDeprecationMessageMatches()` to expect `E_USER_DEPRECATED` issues

sebastianbergmann:
This implements the `expectUserDeprecationMessage()` and `expectUserDeprecationMessageMatches()` methods for expecting `E_USER_DEPRECATED` issues.

The test method must be attributed with `#[PHPUnit\Framework\Attributes\IgnoreDeprecations]` in order to not have the expected deprecation(s) trigger PHPUnit's default handling of `E_USER_DEPRECATED` issues.

Multiple deprecations can be expected in the same test by calling `expectUserDeprecationMessage()` and `expectUserDeprecationMessageMatches()` multiple times.

@stof This is my first attempt to implement what we discussed yesterday. Would this help you?


stof:
The feature of the symfony/phpunit-bridge actually allows defining multiple expected deprecations in the same test. This does not seem to be supported by your feature.

stof:
Btw, is it actually true that PHP issues have non-empty-string as their message ? Isn't it possible in PHP to pass an empty string to `trigger_error` ?

sebastianbergmann:
> The feature of the symfony/phpunit-bridge actually allows defining multiple expected deprecations in the same test. This does not seem to be supported by your feature.

Would this help?

```php
$this->expectUserDeprecationMessage(
    [
        'deprecation message',
        'another deprecation message',
    ]
);
```

sebastianbergmann:
> Isn't it possible in PHP to pass an empty string to `trigger_error`?

That may be the case, but is outside the scope of this PR.

stof:
Well, I asked that question because this PR is adding those `@psalm-param non-empty-string` on all those APIs.

stof:
> Would this help?
> 
> ```
> $this->expectUserDeprecationMessage(
>     [
>         'deprecation message',
>         'another deprecation message',
>     ]
> );
> ```

This could work. But wouldn't it be possible to call the method multiple times instead ? This would more closely match our existing API (and maybe allow us to provide a compatibility layer for the migration)

sebastianbergmann:
> Well, I asked that question because this PR is adding those `@psalm-param non-empty-string` on all those APIs.

No, it is not. Those changes were done in `main`. No idea why GitHub thinks that they are part of this PR.

sebastianbergmann:
> But wouldn't it be possible to call the method multiple times instead?

I already started implemented that instead of what I proposed :)

sebastianbergmann:
> > But wouldn't it be possible to call the method multiple times instead?
> 
> I already started implemented that instead of what I proposed :)

Pushed.

sebastianbergmann:
> > Well, I asked that question because this PR is adding those `@psalm-param non-empty-string` on all those APIs.
> 
> No, it is not. Those changes were done in `main`. No idea why GitHub thinks that they are part of this PR.

Force-pushed the latest changes and now this PR only shows changes exclusive to the respective branch.

sebastianbergmann:
@jrfnl I see that you reacted with 🎉 to this. When this is merged, do you think that `#[WithoutErrorHandler]` should be deprecated and later removed as proposed by @mvorisek in #5592?

jrfnl:
@sebastianbergmann Thanks for asking.

> @jrfnl I see that you reacted with 🎉 to this.

Correct. I did so as I'm seeing that a number of code bases are not making the move to support PHPUnit 10 largely because they can no longer test their own deprecations (and to a much lesser extend notices and warnings).


> When this is merged, do you think that `#[WithoutErrorHandler]` should be deprecated and later removed as proposed by @mvorisek in #5592?

I don't see this as directly related to each other ?

The `#[WithoutErrorHandler]` was introduced for `error_get_last()` and yes, it _could_ be used to implement a custom error handler to allow for still testing deprecations and such, but the original use case for this, was code using `error_get_last()` in the source code of a code base to get an error message (from a stream error) to use in an Exception. Not to work around the removal of the `expect*()` methods.

src/Framework/TestCase.php
```diff
@@ -1535,6 +1564,58 @@ protected function onNotSuccessfulTest(Throwable $t): never
...
+     */
+    private function verifyDeprecationExpectations(): void
+    {
+        foreach ($this->expectedUserDeprecationMessage as $deprecationExpectation) {
+            $this->numberOfAssertionsPerformed++;
+
+            $expectedDeprecationTriggered = false;
+
+            foreach (DeprecationCollector::deprecations() as $deprecation) {
+                if (in_array($deprecationExpectation, DeprecationCollector::deprecations(), true)) {
```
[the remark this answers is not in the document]
sebastianbergmann: Because I made a stupid copy/paste mistake.

src/Framework/TestCase.php
```diff
@@ -1535,6 +1564,48 @@ protected function onNotSuccessfulTest(Throwable $t): never
...
 
+    /**
+     * @throws ExpectationFailedException
+     */
+    private function verifyDeprecationExpectations(): void
+    {
+        foreach ($this->expectedUserDeprecationMessage as $deprecationExpectation) {
+            $this->numberOfAssertionsPerformed++;
+
+            if (!in_array($deprecationExpectation, DeprecationCollector::deprecations(), true)) {
```
stof, line 1575: This requires an exact match on the deprecation message. For `expectExceptionMessage`, PHPUnit checks it with `IsOrContains` semantic. Shouldn't it be consistent here as well ?

sebastianbergmann: I prefer this inconsistency over repeating the mistake that was made with `expectExceptionMessage()`.

stof: If this is considered a mistake, then OK.

Maybe `expectExceptionMessage` should then mark a test as risky when it matches with Contains semantic instead of Is semantic (not sure this is doable in the current architecture)

sebastianbergmann: Yes, I want to deprecate sub-string matching in `expectExceptionMessage()` at some point with the goal removing support for it entirely later.

stof: Maybe you should add a test enforcing that `expectedUserDeprecationMessage('message')` will *not* match `expectedUserDeprecationMessage('another message')` in the tests of the feature. AFAICT, implementing an IsOrContains semantic in the check would still keep the current tests green.

sebastianbergmann:
@stof Thank you for reviewing this!

Issue #5550, opened 2023-10-27
GitHub Releases


sebastianbergmann:
Over time, some users have requested that this project uses [GitHub Releases](https://docs.github.com/en/repositories/releasing-projects-on-github/about-releases) in addition to "just tags". Most recently in #5548, for instance.

The only way to document changes that works for me is maintain a [ChangeLog](https://github.com/sebastianbergmann/phpunit/blob/10.4.2/ChangeLog-10.4.md) file that is under version control. PHPUnit's ChangeLog file follows the [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) standard format.

I am not opposed to duplicate the information from PHPUnit's ChangeLog file to a GitHub Release. I am opposed to manually do that.

I am sure that this process can be automated. When a tag is pushed, parse the relevant information for the version identified by that tag from that tag's ChangeLog file and create a GitHub Release with that information.

Unfortunately, I do not have the time to research and implement this automation.

I am hopeful that @localheinz will be able to implement this automation.

PR #5610, opened 2023-12-11
Enhancement: Create release when tag is pushed

localheinz:
This pull request

- [x] adds a release workflow that uses [`ncipollo/release-action`](https://github.com/ncipollo/release-action) to create releases

Related to #5550.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,39 @@
+# https://docs.github.com/en/actions
+
+on:
+  push:
+    tags:
+      - "**"
+
+name: Release
+
+jobs:
+  release:
+    name: Release
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Install PHP with extensions
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: 8.3
+          coverage: none
+          extensions: none
+          tools: none
+
+      - name: Determine tag
+        run: echo "RELEASE_TAG=${GITHUB_REF#refs/tags/}" >> $GITHUB_ENV
+
+      - name: Parse changelog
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -0,0 +1,48 @@
...
+                prerelease: false,
+                repo: context.repo.repo,
+                tag_name: process.env.RELEASE_TAG,
+              });
+            
+              core.exportVariable('RELEASE_ID', response.data.id);
+              core.exportVariable('RELEASE_UPLOAD_URL', response.data.upload_url);
+            } catch (error) {
+              core.setFailed(error.message);
+            }
```
localheinz, line 48: This reflects https://github.com/ergebnis/.github/blob/1.8.0/actions/github/release/create/action.yaml#L20-L51 and so far, has been tested "in production".

.github/workflows/release.yaml
```diff
@@ -0,0 +1,48 @@
...
+                generate_release_notes: true,
+                name: process.env.RELEASE_TAG,
+                owner: context.repo.owner,
+                prerelease: false,
+                repo: context.repo.repo,
+                tag_name: process.env.RELEASE_TAG,
+              });
+            
+              core.exportVariable('RELEASE_ID', response.data.id);
+              core.exportVariable('RELEASE_UPLOAD_URL', response.data.upload_url);
```
localheinz, line 45: This exports `RELEASE_UPLOAD_URL` as an environment variable for uploading release assets.

sebastianbergmann:
Does this already parse the ChangeLog file and use the information from there as release notes?

Alternatively, it could just link to `https://github.com/sebastianbergmann/phpunit/blob/X.Y.Z/ChangeLog-X.Y.md#XYZ---YYYY-MM-DD`.

I am sorry that I have to ask this, I just do not know how to test this without merging the PR and tagging a release.

PR #5619, opened 2023-12-18
Check and restore error/exception global handlers

mvorisek:
related with #5592 

no functional change to any test which does not add/remove error/exception handlers (or remove exactly the ones the test added)

inspired by output buffer `TestCase::stopOutputBuffering()` code

### DIFF
--- a/.psalm/baseline.xml
+++ b/.psalm/baseline.xml
@@ -510,6 +510,10 @@
     <MissingThrowsDocblock>
       <code>getMethod</code>
     </MissingThrowsDocblock>
+    <PossiblyNullArgument>
+      <code><![CDATA[$this->backupGlobalErrorHandlers]]></code>
+      <code><![CDATA[$this->backupGlobalExceptionHandlers]]></code>
+    </PossiblyNullArgument>
     <PropertyNotSetInConstructor>
       <code>$outputBufferingLevel</code>
     </PropertyNotSetInConstructor>
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -20,6 +20,7 @@ use const PHP_EOL;
 use const PHP_URL_PATH;
 use function array_keys;
 use function array_merge;
+use function array_reverse;
 use function array_values;
 use function assert;
 use function basename;
@@ -28,6 +29,7 @@ use function class_exists;
 use function clearstatcache;
 use function count;
 use function defined;
+use function error_clear_last;
 use function explode;
 use function getcwd;
 use function implode;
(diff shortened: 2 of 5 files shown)


mvorisek:
PR should be done, I run psalm locally, I am not sure if the errors are related with this PR, please help me to fix the CI.

sebastianbergmann:
Can you explain which use cases you have in mind? Thanks.

mvorisek:
Any code which sets & restore (remove) error/exception handlers. They are tricky to debug if some handlers are not restored (removed).

When PHPUnit is executed, there is usually no error/exception handler. Then one is set by the PHPUnit runner. So the performance is fast, as they are newer than a few handlers active.

nunomaduro:
@sebastianbergmann, this pull request introduced a BC issue that is currently impacting all existing Laravel 10 users: https://github.com/laravel/framework/issues/49502. I wonder if you could consider reverting this change and porting it to PHPUnit 11?

sebastianbergmann:
@nunomaduro Can youprovide a **minimal**, **self-contained** (without Laravel), reproducing test case that shows the problem?

@mvorisek What do you think? I am inclined to follow @nunomaduro's suggestion of reverting this change for PHPUnit 10.5.

nunomaduro:
@sebastianbergmann Here you go:

```bash
git clone git@github.com:nunomaduro/phpunit-issue-global-handlers.git
cd phpunit-issue-global-handlers
composer update
./vendor/bin/phpunit
```

I understand the reasoning behind this pull request. However, currently there are dozens of widely used packages (including Laravel itself) that statically register their own error|exception handler **just once,** to prevent them from being registered on each test method and avoid memory leaks.

Ideally, the best scenario for us would be to have this feature as an opt-in, so we don't even have to deal with it. If that's not possible, could you port it to PHPUnit 11? This would give us time to manage the changes.

sebastianbergmann:
This was reverted on both the `10.5` and `main` branches and PHPUnit 10.5.5 has been tagged.

nunomaduro:
Thank you!

mvorisek:
@sebastianbergmann this is expected when test code leave handlers in an unexpected state - can in be relanded in phpunit 10.6?

sebastianbergmann:
There will not be a PHPUnit 10.5, but this could go into PHPUnit 11.

sebastianbergmann:
> Ideally, the best scenario for us would be to have this feature as an opt-in, so we don't even have to deal with it. If that's not possible, could you port it to PHPUnit 11? This would give us time to manage the changes.

I plan to simply revert 23c16c1ce51fbaf05569321094532b428bf5fbd8 (reverting the revert of this PR) for PHPUnit 11.

PR #5628, opened 2023-12-27
Wrong backtrace line is reported

### LINKED ISSUE #5574: Wrong backtrace line is reported
| Q                   | A
| PHPUnit version     | 10.4.2
| PHP version         | 8.2.12

#### Summary

Below is minimal repro test with two exceptions created on a separate line. For some reasons, the same line for both exceptions is reported, but the decisive is where the exception was created.

native php dump: https://3v4l.org/EIXk2 (notice line 3 and 4 is reported)

#### How to reproduce

```
<?php

declare(strict_types=1);

namespace Repro;

use PHPUnit\Event\Code\ThrowableBuilder;

class ReproTest extends TestCase
{
    public function test(): void
    {
        $line = __LINE__;
        $innerException = new \Error('Inner Exception');
        $exception = new \Exception('My exception', 0, $innerException);

        $phpunitThrowable = ThrowableBuilder::from($exception);
        $phpunitThrowableStr = str_replace(\PHP_EOL, "\n", $phpunitThrowable->asString());

        self::assertSame(
            str_replace('[lineMain]', (string) ($line + 2), str_replace('[lineInner]', (string) ($line + 1), <<<'EOF'
                Exception: My exception

                self.php:[lineMain]

                Caused by
                Error: Inner Exception

                self.php:[lineInner]
                EOF)) . "\n",
            str_replace(__FILE__, 'self.php', $phpunitThrowableStr)
        );
    }
}
```

#### Current behavior

```diff
1) Atk4\Core\Tests\ReproTest::test
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
 'Exception: My exception

-self.php:16
+self.php:15

 Caused by
 Error: Inner Exception
```

#### Expected behavior

no diff

samlitowitz:
Resolves #5574

### DIFF
--- a/tests/end-to-end/regression/5574.phpt
+++ b/tests/end-to-end/regression/5574.phpt
@@ -19,15 +19,15 @@ Time: %s, Memory: %s MB
 
 There was 1 error:
 
-1) PHPUnit\TestFixture\Issue5574\Issue5574Test::testOne
+1) PHPUnit\TestFixture\Issue5574\Issue5574Test::testThrownWrappedThrowablesOutputsCorrectStackTraceForEach
 Exception: My exception
 
-%sIssue5574Test.php:14
+%sIssue5574Test.php:22
 
 Caused by
 Error: Inner Exception
 
-%sIssue5574Test.php:13
+%sIssue5574Test.php:21
 
 ERRORS!
 Tests: 1, Assertions: 0, Errors: 1.
--- a/tests/end-to-end/regression/5574.phpt
+++ b/tests/end-to-end/regression/5574.phpt
@@ -1,5 +1,5 @@
 --TEST--
-https://github.com/sebastianbergmann/phpunit/issues/5567
+https://github.com/sebastianbergmann/phpunit/issues/5574
 --FILE--
 <?php declare(strict_types=1);
 $_SERVER['argv'][] = '--do-not-cache-result';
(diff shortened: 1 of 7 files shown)


tests/end-to-end/regression/5574.phpt
```diff
@@ -0,0 +1,33 @@
...
+E                                                                   1 / 1 (100%)
+
+Time: %s, Memory: %s MB
+
+There was 1 error:
+
+1) PHPUnit\TestFixture\Issue5574\Issue5574Test::testOne
+Exception: My exception
+
+%sIssue5574Test.php:14
```
mvorisek, line 25: unit test like in https://github.com/sebastianbergmann/phpunit/issues/5574 issue description would be good too

samlitowitz: > unit test like in https://github.com/sebastianbergmann/phpunit/issues/5574 issue description would be good too

I do not understanding what you are suggesting. Would you please re-state it.

sebastianbergmann:
As a bug fix, this should target the `10.5` branch.

samlitowitz:
> As a bug fix, this should target the `10.5` branch.

Thank you, this has been corrected. This PR is now based on 10.5.

tests/end-to-end/regression/5574.phpt
mvorisek: I suggest to keep the E2E test and to add pure unit test like in issue description with `self::assertSame` asserting the whole redacted stack trace. The reason is the numbers in the E2E tests can be replaced by `%d` by some mass edit someday (unintentionally) and also because unit tests are easier to debug.

PR #5632, opened 2023-12-29
Fix Testdox colorization handling of wide lines

epdenouden:
# Summary
phpunit_screenshot_testdox_margin_fix_summary

# User story
**As a** developer who uses the testdox reporting as my default
**I want** the output to handle coloring failure messages more efficiently
**so that** I don't end up with a completely red terminal or memory allocation exceptions

# Considerations and fix
* Stop the terminal from scrolling unnecessarily.
  It makes relevant details very hard to find, see screenshots below. Crop the message box to always fit the width of the terminal, except for lines that already go past the right margin.
* Long output lines should not be broken up.
  This way soft-wrapped lines can be copied from the terminal without linebreaks. These long lines often contain URLs or tokens important to the test scenario.
* Improve resource efficiency
* Coverage for new functionality

### Screenshots of the issue during everyday usage of PHPUnit
A picture says more than a thousand `malloc()`s:

phpunit_testdox_memery_failed_api_assert

When viewing the output in an editor without color support the memory hunger easy to spot: the highlighted area is _not_ empty but filled with spaces, terminated by a terminal command and newline. The current implementation will apply this style of padding to every line shorter than the longest, without any safeguards.

phpunit_testdox_memery_failed_api_assert_raw

### DIFF
--- a/src/TextUI/Output/Facade.php
+++ b/src/TextUI/Output/Facade.php
@@ -211,6 +211,7 @@ final class Facade
             self::$testDoxResultPrinter = new TestDoxResultPrinter(
                 self::$printer,
                 $configuration->colors(),
+                $configuration->columns(),
             );
         }
 
--- a/src/TextUI/Output/TestDox/ResultPrinter.php
+++ b/src/TextUI/Output/TestDox/ResultPrinter.php
@@ -33,11 +33,13 @@ final readonly class ResultPrinter
 {
     private Printer $printer;
     private bool $colors;
+    private int $columns;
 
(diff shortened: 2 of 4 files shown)


epdenouden:
When trying to rebase on `10.5` I got some 'sometimes' test failures. Since the v11 release is in February already I have kept that as the target. Let me know if you'd like me to rebase and fix for v10.

This pull request fixes an annoying bug I kept tripping over at work. While digging around the Testdox code I sketched out an efficiency refactoring of the colorization internals. Tidy up a lot of the `split` + `implode` chaining without changes in user behaviour.

PR #5638, opened 2024-01-01
Implement xdebug.mode skip condition for end-to-end tests

epdenouden:
This is the alternative implementation pull request #5635 and implements the conditional skip for the `xdebug.mode=develop` check.


epdenouden:
Need to work on my rebase, here's the 10.5 pull: https://github.com/sebastianbergmann/phpunit/pull/5639

Issue #5455, opened 2023-08-01
`willReturnCallback()` does not pass unknown named variadic arguments to callback


stefanfisk:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 10.2.6
| PHP version         |  8.2.8
| Installation Method | Composer

#### Summary

Mocks for methods that accept var args do not passed unknown named variadic arguments to the callback in `willReturnCallback()`. AFAIK this should be supported since PHP 8.0.0.

I believe the issue exists because the mocks use `func_*()` for getting the passed arguments, but those functions no longer return all arguments. See [func_get_args()](https://www.php.net/manual/en/function.func-get-args.php#refsect1-function.func-get-args-notes) for details.

#### Current behavior

Only indexed arguments are passed.

#### How to reproduce

Here's two passing tests for unmocked cases and a failing test that uses a mock.

`tests/Support/VarArger.php`
```php
<?php

namespace Tests\Support;

class VarArger
{
    public function fn(mixed ...$args): mixed
    {
        return [$foo, ...$args];
    }
}
```

`tests/WithReturnCallbackTest.php`
```php
<?php

namespace Tests;

use PHPUnit\Framework\TestCase;
use Tests\Support\VarArger;
use stdClass;

class WithReturnCallbackTest extends TestCase
{
    public function testRealInstance(): void
    {
        $real = new VarArger();

        $args = ['foo', 'bar', 'baz' => 'qux'];

        $this->assertSame(
            $args,
            $real->fn(...$args),
        );
    }

    public function testCallback(): void
    {
        $callback = fn ($foo, ...$args) => [$foo, ...$args];

        $args = ['foo', 'bar', 'baz' => 'qux'];

        $this->assertSame(
            $args,
            $callback(...$args),
        );
    }

    public function testMock(): void
    {
        $mock = $this->createMock(VarArger::class);
        $mock
            ->expects($this->once())
            ->method('fn')
            ->withAnyParameters()
            ->willReturnCallback(fn ($foo, ...$args) => [$foo, ...$args]);

        $args = ['foo', 'bar', 'baz' => 'qux'];

        $this->assertSame(
            $args,
            $mock->fn(...$args),
        );
    }
}
```

The mock class looks like this:
```php
<?php

class Mock_VarArger_343a613f extends Tests\Support\VarArger implements PHPUnit\Framework\MockObject\MockObject
{
    use \PHPUnit\Framework\MockObject\Api;
[...]

PR #5648, opened 2024-01-07
Fix issue with variadic parameters not being passed like they should

marac19901990:
**Issue:**
  - #5455 variadic arguments not being passed to closure

**Solution:**
  - fixed the issue and changed the tests to reflect the changes

### DIFF
--- a/src/Framework/MockObject/Generator/MockMethod.php
+++ b/src/Framework/MockObject/Generator/MockMethod.php
@@ -316,7 +316,7 @@ EOT;
             }
 
             if ($parameter->isVariadic()) {
-                continue;
+                $name = '...' . $name;
             }
 
             if ($parameter->isPassedByReference()) {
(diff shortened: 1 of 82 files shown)


sebastianbergmann:
Does this problem also exist in PHPUnit 9.6?

marac19901990:
> Does this problem also exist in PHPUnit 9.6?

I forgot to check, I'll check and update if needed, my guess is that it doesn't work there either

marac19901990:
> Does this problem also exist in PHPUnit 9.6?

It can't be fixed in 9.6, I made a pull request but then realized that array unpacking for string indexed elements is not supported prior to php 8.1.  I'll update the 10.5 pull request with a better solution, but nothing to be done for 9.6

sebastianbergmann:
Cherry-picked into `10.5` and merged to `main` from there. Thanks!

marac19901990:
> Cherry-picked into `10.5` and merged to `main` from there. Thanks!

You didn't have to cherry pick, these are the 10.5 changes, there was another pull request for 9.6 :)

sebastianbergmann:
I know :) I cherry-picked because I wanted to make some additional changes locally before the merge.

Issue #5658, opened 2024-01-13
PHPUnit 10.5.6 is causing "abnormal termination" in ParaUnit


Wirone:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 10.5.6
| PHP version         | 8.1-8.4
| Installation Method | Composer

#### Summary

PHPCSFixer's CI started to [fail](https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/actions/runs/7513916670) today when PHPUnit 10.5.6 started to be installed for PHP 8.1+. I don't know the exact problem yet, just wanted to report the issue.

#### Current behavior

[Example PR with a failure](https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/pull/7723). Update: it looks like it's strictly related to our CI setup, where we configure PHP's extensions using strict list (using `none` as one of the items, so only explicitly listed ones are installed). PHPUnit started using `ctype` exception but does not list it in `composer.json` as an requirement, so 10.5.6 can be installed even if the extension was missing.

```
PARAUNIT v.2.2.3
by Francesco Panina, Alessandro Lai & Shark Dev Team @ Facile.it
XXXXXXXXXX                                                                    10


Execution time -- 00:00:00

Executed: 10 test classes, 0 tests

Abnormal Terminations (fatal Errors, Segfaults) output:

1) /home/runner/work/PHP-CS-Fixer/PHP-CS-Fixer/tests/AutoReview/CiConfigurationTest.php
Possible abnormal termination, last prepared test was /home/runner/work/PHP-CS-Fixer/PHP-CS-Fixer/tests/AutoReview/CiConfigurationTest.php
```

Error comes under ParaUnit, but there was not new version of it, and it has been working with it for a long time.

#### How to reproduce

See reproduction in PHPCSFixer's CI.

#### Expected behavior

No abnormal termination / segmentation faults.

sebastianbergmann:
Thank you for your report.

Please provide a **minimal**, **self-contained**, reproducing test case that shows the problem you are reporting. "Minimal" also means without third-party extensions for / wrappers around PHPUnit such as ParaUnit.

Without such a minimal, self-contained, reproducing test case I will not be able to investigate this issue.

sebastianbergmann:
@naderman @Seldaek Is there a best practice / easy way to use Composer in the context of git-bisect?

[...]

PR #5661, opened 2024-01-14
Sync PHP extensions

kubawerlos:
Fixes https://github.com/sebastianbergmann/phpunit/issues/5658

Synchronizing the extensions checked in the runtime: https://github.com/sebastianbergmann/phpunit/blob/10.5.6/phpunit#L41 with the ones in composer.json: https://github.com/sebastianbergmann/phpunit/blob/10.5.6/composer.json#L26-L31

Apparently, [paraunit](https://github.com/facile-it/paraunit) does not run the check from PHPUnit, in the 1st commit of https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/pull/7727 I've added the check and after adding `ctype` extension to CI the problem from the linked issue is resolved.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,10 +23,12 @@
     "prefer-stable": true,
     "require": {
         "php": ">=8.1",
+        "ext-ctype": "*",
         "ext-dom": "*",
         "ext-json": "*",
         "ext-libxml": "*",
         "ext-mbstring": "*",
+        "ext-tokenizer": "*",
         "ext-xml": "*",
         "ext-xmlwriter": "*",
         "myclabs/deep-copy": "^1.10.1",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "e06728e5442edec84af96f94a889b4a7",
+    "content-hash": "417d09bd26e28681a2d4effc3e023e19",
     "packages": [
         {
             "name": "myclabs/deep-copy",
@@ -1529,10 +1529,12 @@
     "prefer-lowest": false,
     "platform": {
         "php": ">=8.1",
+        "ext-ctype": "*",
         "ext-dom": "*",
         "ext-json": "*",
         "ext-libxml": "*",
         "ext-mbstring": "*",
+        "ext-tokenizer": "*",
         "ext-xml": "*",
         "ext-xmlwriter": "*"
     },


sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it as PHPUnit's own code does not depend on `ext/ctype` or `ext/tokenizer`. Including these extensions in the `require` section of PHPUnit's `composer.json` file would be incorrect.

PHPUnit depends on libraries such as `nikic/php-parser`, for example, that depend on `ext/ctype` and `ext/tokenizer`. These extensions are listed in the `require` section of the `composer.json` file of `nikic/php-parser`, for example.

Wirone:
Shouldn't the Composer resolve it somehow, so dependencies that require `ctype` are not installed on a runtime that doesn't have the extension? I did not dig deeper, but there must be a bug at some level (Composer, Fixer's deps, vendor's deps) which causes such invalid setup 🤔.

mvorisek:
@sebastianbergmann then these must not be in https://github.com/sebastianbergmann/phpunit/blob/10.5.6/phpunit#L41

sebastianbergmann:
I disagree: the extensions are required to run PHPUnit. This is what PHPUnit checks on startup. This has nothing to do with the dependencies that are listed in PHPUnit's `composer.json` file. If you install PHPUnit using Composer then Composer checks that all extensions are available, both from PHPUnit's own `composer.json` file as well as from the `composer.json files of PHPUnit's dependencies.

sebastianbergmann:
> Shouldn't the Composer resolve it somehow, so dependencies that require `ctype` are not installed on a runtime that doesn't have the extension? I did not dig deeper, but there must be a bug at some level (Composer, Fixer's deps, vendor's deps) which causes such invalid setup 🤔.

As I just wrote in https://github.com/sebastianbergmann/phpunit/pull/5661#issuecomment-1890989139, it is my understanding and expectation that Composer checks transitive dependencies on PHP extensions.

sebastianbergmann:
However, it appears that Composer does not check transitive dependencies on PHP extensions:

```
php -v                                
PHP 8.3.1 (cli) (built: Dec 20 2023 12:44:38) (NTS gcc x86_64)
Copyright (c) The PHP Group
Zend Engine v4.3.1, Copyright (c) Zend Technologies
    with Zend OPcache v8.3.1, Copyright (c), by Zend Technologies
```
```
php --re ctype
Exception: Extension "ctype" does not exist
```
```
composer require --dev phpunit/phpunit    
./composer.json has been created
Running composer update phpunit/phpunit
Loading composer repositories with package information
Updating dependencies
Lock file operations: 27 installs, 0 updates, 0 removals
  - Locking myclabs/deep-copy (1.11.1)
  - Locking nikic/php-parser (v5.0.0)
  - Locking phar-io/manifest (2.0.3)
  - Locking phar-io/version (3.2.1)
  - Locking phpunit/php-code-coverage (10.1.11)
  - Locking phpunit/php-file-iterator (4.1.0)
  - Locking phpunit/php-invoker (4.0.0)
  - Locking phpunit/php-text-template (3.0.1)
  - Locking phpunit/php-timer (6.0.0)
  - Locking phpunit/phpunit (10.5.6)
  - Locking sebastian/cli-parser (2.0.0)
  - Locking sebastian/code-unit (2.0.0)
  - Locking sebastian/code-unit-reverse-lookup (3.0.0)
  - Locking sebastian/comparator (5.0.1)
  - Locking sebastian/complexity (3.2.0)
  - Locking sebastian/diff (5.1.0)
  - Locking sebastian/environment (6.0.1)
  - Locking sebastian/exporter (5.1.1)
  - Locking sebastian/global-state (6.0.1)
  - Locking sebastian/lines-of-code (2.0.2)
  - Locking sebastian/object-enumerator (5.0.0)
  - Locking sebastian/object-reflector (3.0.0)
  - Locking sebastian/recursion-context (5.0.0)
  - Locking sebastian/type (4.0.0)
  - Locking sebastian/version (4.0.1)
  - Locking symfony/polyfill-ctype (v1.28.0)
  - Locking theseer/tokenizer (1.2.2)
Writing lock file
Installing dependencies from lock file (including require-dev)
Package operations: 27 installs, 0 updates, 0 removals
  - Installing sebastian/version (4.0.1): Extracting archive
  - Installing sebastian/type (4.0.0): Extracting archive
  - Installing sebastian/recursion-context (5.0.0): Extracting archive
  - Installing sebastian/object-reflector (3.0.0): Extracting archive
  - Installing sebastian/object-enumerator (5.0.0): Extracting archive
  - Installing sebastian/global-state (6.0.1): Extracting archive
  - Installing sebastian/exporter (5.1.1): Extracting archive
  - Installing sebastian/environment (6.0.1): Extracting archive
  - Installing sebastian/diff (5.1.0): Extracting archive
  - Installing sebastian/comparator (5.0.1): Extracting archive
  - Installing sebastian/code-unit (2.0.0): Extracting archive
  - Installing sebastian/cli-parser (2.0.0): Extracting archive
  - Installing phpunit/php-timer (6.0.0): Extracting archive
  - Installing phpunit/php-text-template (3.0.1): Extracting archive
  - Installing phpunit/php-invoker (4.0.0): Extracting archive
  - Installing phpunit/php-file-iterator (4.1.0): Extracting archive
  - Installing theseer/tokenizer (1.2.2): Extracting archive
  - Installing symfony/polyfill-ctype (v1.28.0): Extracting archive
  - Installing nikic/php-parser (v5.0.0): Extracting archive
  - Installing sebastian/lines-of-code (2.0.2): Extracting archive
  - Installing sebastian/complexity (3.2.0): Extracting archive
  - Installing sebastian/code-unit-reverse-lookup (3.0.0): Extracting archive
  - Installing phpunit/php-code-coverage (10.1.11): Extracting archive
  - Installing phar-io/version (3.2.1): Extracting archive
  - Installing phar-io/manifest (2.0.3): Extracting archive
  - Installing myclabs/deep-copy (1.11.1): Extracting archive
  - Installing phpunit/phpunit (10.5.6): Extracting archive
1 package suggestions were added by new dependencies, use `composer suggest` to see details.
Generating autoload files
24 packages you are using are looking for funding.
Use the `composer fund` command to find out more!
No security vulnerability advisories found.
Using version ^10.5 for phpunit/phpunit
```
```
./vendor/bin/phpunit --version               
PHPUnit requires the "ctype", "dom", "json", "libxml", "mbstring", "tokenizer", "xml", "xmlwriter" extensions, but the "ctype" extension is not available.
```

I think this issue should be raised with Composer.

Wirone:
> As I just wrote in https://github.com/sebastianbergmann/phpunit/pull/5661#issuecomment-1890989139, it is my understanding and expectation that Composer checks transitive dependencies on PHP extensions.

That's how I understand it too. If project requires package A, package A requires package B, and package B requires extension X, then it would not make sense to require X in package A, because if package B changes its implementation and does not require X anymore, then it would lead to dead requirement in package A.

@Seldaek sorry for pinging, but what's the Composer's approach in that regard? In our case it led to issue that was easy to spot only because of [diff between PHPUnit's versions](https://github.com/sebastianbergmann/phpunit/compare/10.5.5...10.5.6), where `ctype` was added to CI workflows. But from ParaUnit errors in our CI it was pretty hard to understand what happened.

kubawerlos:
There is no bug in Composer, see that in [the comment](https://github.com/sebastianbergmann/phpunit/pull/5661#issuecomment-1890990606) there is a line in logs:
```cli
  - Installing symfony/polyfill-ctype (v1.28.0): Extracting archive
```
and it [provides](https://github.com/symfony/polyfill-ctype/blob/v1.28.0/composer.json#L21-L23) `ctype` extension implementation. So `composer install` went fine.

Seldaek:
Right, composer definitely checks transitive dependencies, regardless if they're on ext or regular packages. The polyfill probably explains it but I'm not sure why it'd fail then if the polyfill is present. But i haven't looked in depth at the issue here.

sebastianbergmann:
*sigh*

So Composer automatically pulls in a polyfill for the `ctype` extension instead of erroring out? That is at least unexpected.

The output of `composer why` is also confusing:

```
$ composer why symfony/polyfill-ctype                         
There is no installed package depending on "symfony/polyfill-ctype"
```

sebastianbergmann:
> I'm not sure why it'd fail then if the polyfill is present. But i haven't looked in depth at the issue here

The "it" in question is PHPUnit, not Composer. PHPUnit does an `extension_loaded('ctype')` on startup that returns `false`.

Seldaek:
Or i guess more appropriate if you want to check for presence of required exts would be to check for functions that are needed and not the ext as a whole. That'd work with polyfills.

sebastianbergmann:
> And nope composer does not pull polyfills in automatically. Something must be requiring it.

Good to hear. However, I am unable to find a reference to `symfony/polyfill-ctype` in any `composer.json` file that is involved in installing PHPUnit 10.5.6.

Seldaek:
Yeah i see.. Is that running the require with a blank composer.json or did you have other things already? Any plugins loaded? Sorry just on my phone so cannot look too close into it right now but if you give me clear repro steps I'll take a look tomorrow.

sebastianbergmann:
https://github.com/sebastianbergmann/phpunit/pull/5661#issuecomment-1890990606 was in an empty directory that did not have a `composer.json` file before `composer require --dev phpunit/phpunit` was run. I do not use any Composer plugins.

sebastianbergmann:
While I hopefully worked around the issue now in 59b331b95fa0f81406cb7d1cebae9e690bada7e3 (PHPUnit 10.5.7 has been released with this change), I look forward to learn how/why `symfony/polyfill-ctype` is pulled in.

Thank you for looking into this, @Seldaek!

kubawerlos:
> I look forward to learn how/why `symfony/polyfill-ctype` is pulled in.

I did a similar thing (removing plugins, clearing Composer's cache as well) and run:
```cli
composer require --dev phpunit/phpunit -vvv --profile
```
and in part named "Loading composer repositories with package information" there is:
```
[20.8MiB/0.90s] Downloading https://repo.packagist.org/p2/symfony/polyfill-ctype.json
```
So, my guess is that when Composer downloads all the metadata it "finds" `symfony/polyfill-ctype` (as a dependency in some version that it might not even install) and then, when resolving dependencies it sees that "there is no ext-ctype", but "hey, I have something that provides it" and thus installs the polyfill.

Wirone:
I believe I read something about it in the past, but I can't remember what that was. Did look at the Composer's repo and I did not find anything about polyfills, but from mobile it's hard 🙄. I believe the auto-polyfill was somewhere on Symfony side, but I wouldn't bet on it 😅.

Seldaek:
Ok so good news is I can reproduce it using this:

```json
{
    "config": {
        "platform": {
            "ext-ctype": false
        }
    },
    "require-dev": {
        "phpunit/phpunit": "*"
    }
}
```

Then `composer update` selects `symfony/polyfill-ctype`, because some old phpunit version requires `symfony/yaml`, and that requires polyfill-ctype, so the polyfill becomes loaded in the package pool, and then the solver picks it to fulfill the missing ext-ctype requirement.

This does not happen if you do `composer require --dev phpunit/phpunit:^10`, limiting the scope of phpunit versions to ^10, because then it never looks at symfony/yaml.. And in that case it also downgrades nikic/php-parser to v4 instead of v5.

So yeah it all makes sense in the current context. We will need to see if this can be fixed. We have an idea how it might be fixable but we need to see what the perf impact is.

sebastianbergmann:
Thank you so much, @Seldaek, for looking into this!

Is there already/will there be an issue for Composer where this is tracked?

Wirone:
@Seldaek it makes sense, because only PHP 8.1-8.3 were affected and I believe the difference was PHPUnit and `nikic/php-parser`. Thank you for investigating this.

Seldaek:
Here is where it's tracked, but it's looking more complex than I'd like https://github.com/composer/composer/pull/11798

PR #5429, opened 2023-06-27
Allow doubling `readonly` classes

drealecs:
There is currently a problem with implementing mocks for readonly classes and the only impediment so far is that mocks have mutable internal state as explained here: https://github.com/sebastianbergmann/phpunit/issues/5353#issuecomment-1535696396

This PR is moving phpunit internal state associated with stub/mock classes/objects to separate global classes.

Moved state from stubs and mock objects:
- private static array $__phpunit_configurableMethods
- private bool $__phpunit_returnValueGeneration
- private ?InvocationHandler $__phpunit_invocationMocker
- private static array $__phpunit_deprecation_emitted_for_test
- private object $__phpunit_originalObject

to two singleton internal classes, `InternalStubState`  and `InternalMockObjectState` that store the state in
- an array with class name as key for previously static state
- a WeakMap with objects as key for previously instance state

not introducing new memory leaks issues.

And also implements readonly doubles for readonly classes that now is possible.

PR #5663, opened 2024-01-14
Allow doubling `readonly` classes

### LINKED ISSUE #5353: Class "App\Service\Service" is declared "readonly" and cannot be doubled
| Q                   | A
| PHPUnit version     | 10.0.16
| PHP version         | 8.2.5

Hello,

We've encountered an issue where mocking readonly classes is not working in the new PHP version. Is there any way to mock readonly classes? Shouldn't PHPUnit have this functionality built-in?

Thanks in advance for any help or guidance you can provide.

drealecs:
This implements readonly doubles for readonly classes (second commit) that it possible after migrating internal state associated with Stub/MockObject implementation to separate classes (first commit).

Implementation was discussed in #5429 and is continued here.
Fixes #5353.


src/Framework/MockObject/Runtime/Api/MockObjectInternalState.php
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * @internal This class is not covered by the backward compatibility promise for PHPUnit
+ */
+final class MockObjectInternalState
+{
+    /**
+     * @var array<class-string, array<string, true>>
+     */
+    private static array $deprecationEmittedForTest = [];
```
drealecs, line 20: This needs to be removed, it's just a small artifact from porting to 10.5

drealecs: done

tests/unit/Framework/MockObject/Creation/CreateMockTest.php
```diff
@@ -53,12 +52,13 @@ public function testCannotCreateMockObjectForFinalClass(): void
         $this->createMock(FinalClass::class);
     }
 
-    #[RequiresPhp('8.2')]
```
drealecs, line 56: The coverage pipeline was running using PHP 8.3 since release, but it was not working since this mentioned 8.2 exactly.

src/Framework/MockObject/Runtime/Api/DoubledCloneMethod.php
```diff
@@ -16,6 +16,6 @@ trait DoubledCloneMethod
 {
     public function __clone(): void
     {
-        $this->__phpunit_invocationMocker = clone $this->__phpunit_getInvocationHandler();
+        $this->__phpunit_stubInternalState->cloneInvocationHandler();
```
drealecs, line 19: Unfortunately, I identified a bug in the implementation, for a behavior that tests are not currently covering. I have added the tests now in #5665 that would fail with current code.
The correct code with using a readonly property for the internal state should be:
```suggestion
        $this->__phpunit_stubInternalState = clone $this->__phpunit_stubInternalState;
        $this->__phpunit_stubInternalState->cloneInvocationHandler();
```  
  
  
Cloning readonly properties does not work in PHP 8.1 and PHP 8.2, so this kind of implementation can only exist in a PHPUnit version that has a minim PHP supported version as PHP 8.3.

I think we can avoid the limitation by changing the internal state implementation to use a WeakMap for the invocation handler, with test doubles as keys, so we can properly clone it.

drealecs: I applied my suggested change to have the code without bugs and fix at least PHP 8.3.
If anyone finds a solution, feel free to suggest something.

sebastianbergmann: We should not try to `clone` the test double object when `readonly` properties are involved. Instead, before the `clone` statement is reached, we should raise a PHPUnit error with an explanation that cloning test double objects that have `readonly` properties requires PHP >= 8.3.

drealecs: Yes, thank you, the idea is fine and blocks only cloning when the mock is for a readonly class.
But I think this also implies that we should create readonly properties for internal state only when mocking readonly classes? With current implementation, we have the internal state as a readonly property for all cases.
I'm thinking that having a separate Trait for readonly class vs not readonly classes would not be nice, even if keeping differences at minimum only for the property definition. Do you have an idea for this?

mougrim:
Hi all!
Is there anything I can help to merge it?

sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it as I attempt a different implementation in #5804.

drealecs:
> Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it as I attempt a different implementation in #5804.

Sounds good and looks similar, looking a bit on the phone.  
Glad that we can end up with a something that will work, so thank you for that.  
Also sorry for not being able to allocate more time to finish this.

PR #5664, opened 2024-01-14
Allow doubling `readonly` classes (PHPUnit 11)

drealecs:
This implements readonly doubles for readonly classes (second commit) that it possible after migrating internal state associated with Stub/MockObject implementation to separate classes (first commit).

Implementation was discussed in #5429 and is continued in #5663 and for the PHPUnit 11 here.


sebastianbergmann:

If/when #5663 is merged into `10.5` then I will handle the merge to `main`.

drealecs:
Sure, no problem. There was a small part with the static `$__phpunit_deprecation_emitted_for_test` that was implemented that I had implemented, that does not exist in 10.5. Let me know if you need any help there, but I guess you can check the diff that exists here in case you want to use anything.

PR #5665, opened 2024-01-14
Add a test to validate cloning of InvocationHandler on test doubles

drealecs:
For each test double, there is a set of invocation expectations.
When cloning test double, there is a clone on the set, copying existing invocations.

The behavior is that the expectations done before cloning are shared between the original test doubles and the clone, while the new expectations added to each after cloning are not shared.


sebastianbergmann:
Thank you!

PR #5667, opened 2024-01-15
Add additional tests for data provider values

jnoordsij:
Follow-up to #5225.

This MR adds tests that assert the following behavior:
- When a test receives input from both a data provider method and from one or more tests it depends on, the arguments from the data provider will come before the ones from depended-upon tests. This is taken from [the doc description](https://docs.phpunit.de/en/10.5/writing-tests-for-phpunit.html#data-providers).
- Combining named arguments (as introduced in #5225) with input from a dependency is not supported.

### DIFF
--- a/tests/_files/DataProviderDependencyResultTest.php
+++ b/tests/_files/DataProviderDependencyResultTest.php
@@ -1,0 +1,41 @@
+<?php declare(strict_types=1);
+/*
+ * This file is part of PHPUnit.
+ *
+ * (c) Sebastian Bergmann <sebastian@phpunit.de>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+namespace PHPUnit\TestFixture;
+
+use PHPUnit\Framework\Attributes\DataProvider;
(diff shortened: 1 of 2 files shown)


tests/_files/DataProviderDependencyResultTest.php
```diff
@@ -0,0 +1,42 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+namespace PHPUnit\TestFixture;
+
+use PHPUnit\Framework\Attributes\DataProvider;
+use PHPUnit\Framework\Attributes\Depends;
+use PHPUnit\Framework\TestCase;
+
+class DataProviderDependencyResultTest extends TestCase
```
sebastianbergmann, line 16: Please declare the test case class `final`.

sebastianbergmann:
I cherry-picked 4e913e31619e1b5452489b8e930fb56af01add1b into `10.5` and merged it to `main` from there. Then I cherry-picked 0e37d5433f8a05c3a8fc9e393b65a3fa3066619d into `main`.

Issue #5595, opened 2023-12-01
Process isolation doesn't work starting from 10.2.0


l-alexandrov:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 10.2.0
| PHP version         | 8.3.0
| Installation Method | Composer

#### Summary
When enable processs isolation in any form (CLI argument, PHP doc annotation or attribute in `phpunit.xml`) the tests are failing with an error. 
I'm using PHPUnit in combination with Laravel 10.34.2. Prior to PHPUnit 10.2.0 I had no issues with proccess isolation.

#### Current behavior
 When run a test in a process isolation I get `Test was run in child process and ended unexpectedly` error.

#### How to reproduce

`vendor/bin/phpunit --process-isolation` and see the error.

#### Expected behavior

Getting no errors.

#### Configuration
```xml
<?xml version="1.0" encoding="UTF-8"?>
<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/10.4/phpunit.xsd"
         bootstrap="vendor/autoload.php"
         executionOrder="depends,defects"
         beStrictAboutOutputDuringTests="true"
         failOnRisky="true"
         failOnWarning="true"
         colors="true"
         enforceTimeLimit="true"
         defaultTimeLimit="3"
>
    <testsuites>
        <testsuite name="Unit">
            <directory>./tests/Unit</directory>
        </testsuite>
        <testsuite name="Feature">
            <directory>./tests/Feature</directory>
        </testsuite>
        <testsuite name="Integration">
            <directory>./tests/Integration</directory>
        </testsuite>
    </testsuites>
  <coverage/>
    <php>
        <server name="APP_ENV" value="testing"/>
        <server name="BCRYPT_ROUNDS" value="4"/>
    </php>
  <source>
    <include>
      <directory suffix=".php">./app</directory>
    </include>
  </source>
</phpunit>
```

sebastianbergmann:
Thank you for your report.

Please provide a **minimal**, **self-contained**, reproducing test case that shows the problem you are reporting.

Without such a minimal, self-contained, reproducing test case I will not be able to investigate this issue.

l-alexandrov:
It looks it's a Laravel issue, since I have no problems when extending PHPUnit's `TestCase` class.

arclw:
[...]

PR #5677, opened 2024-01-23
fix: Setting error handler inside an isolated test may cause test failure on PHP 8.3

siketyan:
## Related Issues

- https://github.com/sebastianbergmann/phpunit/issues/5595
- https://github.com/laravel/framework/issues/49237
- https://github.com/laravel/framework/issues/49593

## Summary

On PHP 8.3, calling `set_error_handler` in an isolated test (via `#[RunInSeparateProcess]`) may cause test failure, regardless of what the error handler does.

Some frameworks, including Laravel, enables their error handler on booting. After the test run, PHPUnit's temporary code tries to call `rewind(STDOUT)`. This may cause an error when STDOUT is not rewindable, as the comment says; this is expected. The error handler set by the framework is called even though `@` is prepended to the call.

After that `PHPUnit\Runner\ErrorHandler` is called somehow only on PHP 8.3; I don't know why this is called (a PHP bug?). ErrorHandler tries to find the call stack, but it's not possible. Then the test fails by ErrorHandler throwing `NoTestCaseObjectOnCallStackException`. In my application tests, this is shown as 'Test was run in child process and ended unexpectedly'.

This pull request sets a no-op error handler temporarily during doing fallible calls with `@`. This helps the tests calling `set_error_handler` but not calling `restore_error_handler` after the test.

## Workaround

Call `restore_error_handler` on `tearDown` or `tearDownAfterClass` depending on the situation.

### DIFF
--- a/src/Util/PHP/Template/TestCaseClass.tpl
+++ b/src/Util/PHP/Template/TestCaseClass.tpl
@@ -66,6 +66,8 @@ function __phpunit_run_isolated_test()
 
     ini_set('xdebug.scream', '0');
 
+    set_error_handler('__phpunit_error_handler');
+
     // Not every STDOUT target stream is rewindable
     @rewind(STDOUT);
 
@@ -78,6 +80,8 @@ function __phpunit_run_isolated_test()
             @rewind(STDOUT);
         }
     }
+
+    restore_error_handler();
 
     file_put_contents(
         '{processResultFile}',
--- a/src/Util/PHP/Template/TestCaseClass.tpl
+++ b/src/Util/PHP/Template/TestCaseClass.tpl
@@ -66,7 +66,7 @@ function __phpunit_run_isolated_test()
 
     ini_set('xdebug.scream', '0');
 
-    set_error_handler('__phpunit_error_handler');
+    set_error_handler(static fn() => false);
 
     // Not every STDOUT target stream is rewindable
     @rewind(STDOUT);
--- a/src/Util/PHP/Template/TestCaseMethod.tpl
+++ b/src/Util/PHP/Template/TestCaseMethod.tpl
@@ -31,11 +31,6 @@ if ($composerAutoload) {
     require $phar;
 }
 
-function __phpunit_error_handler($errno, $errstr, $errfile, $errline)
-{
-   return true;
-}
-
 function __phpunit_run_isolated_test()
 {
     $dispatcher = Facade::instance()->initForIsolation(
@@ -71,7 +66,7 @@ function __phpunit_run_isolated_test()
 
     ini_set('xdebug.scream', '0');
 
-    set_error_handler('__phpunit_error_handler');
+    set_error_handler(static fn() => false);
 
     // Not every STDOUT target stream is rewindable
     @rewind(STDOUT);
@@ -101,6 +96,11 @@ function __phpunit_run_isolated_test()
             ]
         )
     );
+}
+
+function __phpunit_error_handler($errno, $errstr, $errfile, $errline)
+{
+    return true;
 }
 
 set_error_handler('__phpunit_error_handler');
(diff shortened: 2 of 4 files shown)


sebastianbergmann:
CC @mvorisek

src/Util/PHP/Template/TestCaseClass.tpl
```diff
@@ -66,6 +66,8 @@ function __phpunit_run_isolated_test()
 
     ini_set('xdebug.scream', '0');
 
+    set_error_handler('__phpunit_error_handler');
```
mvorisek, line 69:
```suggestion
    set_error_handler(static fn() => true);
```

siketyan:
> before/after https://github.com/sebastianbergmann/phpunit/commit/f38fc88c76470f0f02cf2d662f4d50dda682fb70 (PHPUnit 10.5.9)

Nothing changed for the regression test.

> before/after https://github.com/sebastianbergmann/phpunit/pull/5619 (PHPUnit master (upcoming 11.x))?

Before 93aa92b9a1c77476e97ac1832e58907017f7ee28: Same as 10.5 branch.
After 93aa92b9a1c77476e97ac1832e58907017f7ee28: Test succeeds but marked as a risky test:

```
PHPUnit 10.5.3-33-g5da503ce1 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.3.2

Time: 00:00.069, Memory: 8.00 MB

There was 1 risky test:

1) PHPUnit\TestFixture\Issue5595Test::test
Test code or tested code did not remove its own error handlers

/Users/siketyan/.local/src/github.com/sebastianbergmann/phpunit/tests/end-to-end/regression/5595/Issue5595Test.php:25

OK, but there were issues!
Tests: 1, Assertions: 1, Risky: 1.
```

> Is it somehow related with https://github.com/php/php-src/commit/b3e33be44391bdd9816a7ec6e68394327d7da24a?

It can be, I couldn't find which part of the change cause this.

src/Util/PHP/Template/TestCaseClass.tpl
siketyan: Fixed in 12b1a26.

mvorisek:
> > before/after #5619 (PHPUnit master (upcoming 11.x))?
> 
> Before [93aa92b](https://github.com/sebastianbergmann/phpunit/commit/93aa92b9a1c77476e97ac1832e58907017f7ee28): Same as 10.5 branch. After [93aa92b](https://github.com/sebastianbergmann/phpunit/commit/93aa92b9a1c77476e97ac1832e58907017f7ee28): Test succeeds but marked as a risky test:

If so, it seems the error handler is set during a test not during a bootstrap. Such issue is detected by PHPUnit master (upcoming 11.x) correctly and the proper fix is to never end a test with modified global handlers as advised in https://github.com/laravel/framework/issues/49237#issuecomment-1905734819 and like reasoned in https://github.com/sebastianbergmann/phpunit/pull/5619#issue-2047224434.

siketyan:
I'll submit a pull request to Laravel later. Are we not going to add this change for the temporary fix in PHPUnit 10 anyway?

siketyan:
To be clear: "bootstrapping" which I wrote in the description is of Laravel, not PHPUnit. It's called in setUp for each tests.

mvorisek:
In PHPUnit 10.x is it only undetected, in (upcoming) PHPUnit 11.x is it detected. In PHPUnit 10.x the detection was not added for BC reasons. So I do not think there is much to fix outside backporting #5619 when process isolation is active (with empty merge commit to PHPUnit 11.x).

https://github.com/mvorisek/phpunit/blob/0647107202f42b13910ab4d382e3f0f801dac231/tests/end-to-end/regression/5592/Issue5592Test.php test should be improved to an error handler added in `setUp`/`tearDown` and /w process isolation.

siketyan:
Closing this as PHPUnit 11 has the error handler detector and the issue should be resolved in Laravel 11. Thank you for your help!

Issue #5681, opened 2024-01-24
Other unknown classes to clean up


janedbal:
Followup of #5680. There are still few references to unknown classes. Few look intentional, but at least that `CliTestDoxPrinter` is imo cleanup candidate, here is an output to verify:

```
  • C
      tests/end-to-end/regression/4376/tests/Test.php:19

  • CoveredClass
      tests/_files/NamespaceCoverageMethodTest.php:24

  • FooBarHandler
      tests/_files/RouterTest.php:24

  • PHPUnit\Logging\TestDox\CliTestDoxPrinter
      tests/_files/TestableCliTestDoxPrinter.php:14

  • ThisClassDoesNotExist
      tests/unit/Metadata/Parser/Annotation/RegistryTest.php:85
```

sebastianbergmann:
I am sorry, but this issue is not helpful. If you think that you have found a source file in the test fixture of PHPUnit's test suite (great! there are likely a few.) then please delete it and try to run the test suite. If it does not fail then please open a pull request.

PR #5682, opened 2024-01-24
Remove TestableCliTestDoxPrinter

janedbal:
As found in https://github.com/sebastianbergmann/phpunit/issues/5681, this class extend non-existing class


sebastianbergmann:
Is `main` the earliest branch where this file can be deleted?

janedbal:
I dont yet see into the branching here, I can try to find the commit where the parent class was deleted if that helps.

janedbal:
Deleted here: https://github.com/sebastianbergmann/phpunit/commit/4605ade88b2e91228fba1b34de766bb0008b8a4c

janedbal:
So it can be deleted even in 10.5 branch imo.

sebastianbergmann:
Cherry-picked into `10.5` and merged to `main` from there. Thanks!

PR #5698, opened 2024-02-09
disallowed trailing comma in the parameter list in PHP < 8.0

melaxon:
PHP Parse error:  syntax error, unexpected ')'

Prior to PHP 8.0, this would have caused a parse error due to the disallowed trailing comma in the parameter list


sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it as PHP 8.2 is required.

PR #5703, opened 2024-02-16
Support --group in --list-tests

### LINKED ISSUE #5702: combine `--list-tests` with `--group`
running `php vendor/bin/phpunit --group levels --list-tests` on the [PHPStan-src codebase](https://github.com/phpstan/phpstan-src) returns the same list of tests, as when not using `--group levels`: `php vendor/bin/phpunit --list-tests`

I would expect `--group levels` to apply the same filtering as then running phpunit itself with `--group levels`.

```
$ vendor/bin/phpunit --version
PHPUnit 9.5.23 #StandWithUkraine

$ php -v
PHP 8.2.12 (cli) (built: Oct 24 2023 21:15:35) (NTS Visual C++ 2019 x64)
Copyright (c) The PHP Group
Zend Engine v4.2.12, Copyright (c) Zend Technologies
```

what I want to achieve: I want to use phpunit test filtering capabilities to separate the huge PHPStan test-suite into groups which I can run in parallel in GitHub Actions. running it in several parallel GitHub Action jobs has the benefit of utilizing more CPUs and therefore a faster feedback loop

staabm:
early submit to get some functional questions answered. I will cleanup CS and type-checker after we have the same idea how the feature should work

closes https://github.com/sebastianbergmann/phpunit/issues/5702

### DIFF
--- a/src/TextUI/Command/Commands/ListTestsAsTextCommand.php
+++ b/src/TextUI/Command/Commands/ListTestsAsTextCommand.php
@@ -36,7 +36,10 @@ final readonly class ListTestsAsTextCommand implements Command
         $buffer .= 'Available test(s):' . PHP_EOL;
 
         $configuration = Registry::get();
-        $configuredGroups = $configuration->groups();
+        $configuredGroups = [];
+        if ($configuration->hasGroups()) {
+            $configuredGroups = $configuration->groups();
+        }
 
         foreach (new RecursiveIteratorIterator($this->suite) as $test) {
             if ($test instanceof TestCase) {
@@ -50,6 +53,10 @@ final readonly class ListTestsAsTextCommand implements Command
                     str_replace(' with data set ', '', $test->nameWithDataSet()),
                 );
             } elseif ($test instanceof PhptTestCase) {
+                if ($configuredGroups !== []) {
+                    continue;
+                }
+
                 $name = $test->getName();
             } else {
                 continue;
--- a/src/TextUI/Command/Commands/ListTestsAsTextCommand.php
+++ b/src/TextUI/Command/Commands/ListTestsAsTextCommand.php
@@ -15,6 +15,7 @@ use function str_replace;
 use PHPUnit\Framework\TestCase;
 use PHPUnit\Framework\TestSuite;
 use PHPUnit\Runner\PhptTestCase;
+use PHPUnit\TextUI\Configuration\FilterNotConfiguredException;
 use PHPUnit\TextUI\Configuration\Registry;
 use RecursiveIteratorIterator;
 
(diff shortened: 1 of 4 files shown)


src/TextUI/Command/Commands/ListTestsAsTextCommand.php
```diff
@@ -35,8 +35,15 @@ public function execute(): Result
 
         $buffer .= 'Available test(s):' . PHP_EOL;
 
+        $configuration = Registry::get();
+        $configuredGroups = $configuration->groups();
+
         foreach (new RecursiveIteratorIterator($this->suite) as $test) {
             if ($test instanceof TestCase) {
+                if (!$this->testContainsConfiguredGroups($test, $configuredGroups)) {
```
staabm, line 43: should a `PhptTestCase` be skipped when groups are passed via CLI flags, since a `PhptTestCase` cannot define a group (AFAICT) ?

I would skip them - do you agree?

sebastianbergmann:
I wrote down my thoughts on the proposed change in https://github.com/sebastianbergmann/phpunit/issues/5702#issuecomment-1950206237 and will only comment on the currently proposed implementation here.

If we change options such as `--list-tests` to apply filters such as `--group` then we should reuse the existing code (`TestSuiteFilterProcessor`) to perform the filtering.

After d2003b64f70acf59568dd18faea82369eab6500c, this (c|sh)ould be a simple as
```patch
diff --git a/src/TextUI/Application.php b/src/TextUI/Application.php
index 2bf6d3c78..5f2c8bb31 100644
--- a/src/TextUI/Application.php
+++ b/src/TextUI/Application.php
@@ -37,6 +37,7 @@
 use PHPUnit\Runner\Extension\ExtensionBootstrapper;
 use PHPUnit\Runner\Extension\Facade as ExtensionFacade;
 use PHPUnit\Runner\Extension\PharLoader;
+use PHPUnit\Runner\Filter\Factory;
 use PHPUnit\Runner\GarbageCollection\GarbageCollectionHandler;
 use PHPUnit\Runner\ResultCache\DefaultResultCache;
 use PHPUnit\Runner\ResultCache\NullResultCache;
@@ -416,6 +417,8 @@ private function executeCommandsThatDoNotRequireTheTestSuite(Configuration $conf
 
     private function executeCommandsThatRequireTheTestSuite(Configuration $configuration, CliConfiguration $cliConfiguration, TestSuite $testSuite): void
     {
+        (new TestSuiteFilterProcessor(new Factory))->process($configuration, $testSuite);
+
         if ($cliConfiguration->listGroups()) {
             $this->execute(new ListGroupsCommand($testSuite));
         }
```

sebastianbergmann:

sebastianbergmann:
> After [d2003b6](https://github.com/sebastianbergmann/phpunit/commit/d2003b64f70acf59568dd18faea82369eab6500c), this (c|sh)ould be a simple as

Alas, it is not: https://github.com/sebastianbergmann/phpunit/issues/5720#issuecomment-1975096608. But I'm working on it :)

PR #5716, opened 2024-02-29
Better respect PHP native array key handling for `assertArrayIs*ToArrayOnlyConsideringListOfKeys()`

jrfnl:
Related to sebastianbergmann/phpunit#5600

As things are, arrays in PHP can have either integer or string keys. Depending on the key input, PHP does some type juggling magic though, like auto-converting purely integer string keys to integers and flooring floating point keys to integers.

While experienced devs will know this pitfall, less experienced devs (who also write tests) may not be as aware and may provide the keys in `$keysToBeConsidered` the same way as the original array was defined, not realizing that the type of some of the keys will have auto-magically been changed by PHP.

The code in the new `assertArrayIs*ToArrayOnlyConsideringListOfKeys()` assertions, with its use of strict `in_array()` [did not respect the key juggling PHP does](https://3v4l.org/FdReu), while [the code for the `assertArrayIs*ToArrayIgnoringListOfKeys` assertions did](https://3v4l.org/AfHoc) (as `unset()` - and `isset()` for that matter - will do the same type juggling for the array keys).

This commit adjusts the code for the `assertArrayIs*ToArrayOnlyConsideringListOfKeys()` assertions to handle arrays keys passed in `$keysToBeConsidered` consistently in the same way PHP itself would do.

Includes tests.
Includes tests for the same for the `assertArrayIs*ToArrayIgnoringListOfKeys` assertions which were not affected by this bug.

### DIFF
--- a/src/Framework/Assert.php
+++ b/src/Framework/Assert.php
@@ -9,11 +9,9 @@
  */
 namespace PHPUnit\Framework;
 
-use function array_keys;
 use function class_exists;
 use function count;
 use function file_get_contents;
-use function in_array;
 use function interface_exists;
 use function is_bool;
 use ArrayAccess;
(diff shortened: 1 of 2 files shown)


jrfnl:
Thanks for accepting this fix @sebastianbergmann !

Just in case anyone is wondering why this could be an issue, here's an example usecase where this could come into play with these assertions:

Think an integration test where the code under test inserts some data into a database and the tests wants to verify that the data was correctly added.

In the test:
* The code under test would be run and returns an array of integer strings representing the inserted record ids.
* Next a DB query would be run to retrieve the full records in the database table as an array indexed by the record ids.
* Then the results of second query would be compared with the expected records for only the inserted record ids using the `assertArrayIs*ToArrayOnlyConsideringListOfKeys()` assertions using the return value of the code under test as `$keysToBeConsidered`.

By default (without an abstraction layer), most databases will return text strings, even for ids, so without this fix that comparison would have failed.

PR #5727, opened 2024-03-09
Prevent duplicate call of `NamePrettifier::prettifyTestMethodName()`

staabm:
stumbled over it in a recent blackfire profile

grafik


sebastianbergmann:

PR #5728, opened 2024-03-09
Lazy collorize method name

staabm:
looking into blackfire profiles I can see the TestDoc colorization takes 10-15% of runtime.

testing of this PR confirmed the thesis

before
```
php ./phpunit --testsuite unit     
…
Time: 00:00.466, Memory: 44.00 MB
```

after
```
php ./phpunit --testsuite unit     
…
Time: 00:00.420, Memory: 52.00 MB
```

blackfire diff

grafik

---

now the bad news :-). while this PR works with the unit test-suite, it does not with the end-to-end suite.

I can see that the changed TestDox class is no longer serializable, and therefore does not work with tests using process isolation or similar features.

since 10-15% percent is not that bad, I figured it would nevertheless make sense to bring this up for discussion.
maybe we can utilize the basic idea of doing coloring lazily for all tests which do not use process isolation..?

I would love anyones feedback

### DIFF
--- a/src/Event/Value/Test/TestDox.php
+++ b/src/Event/Value/Test/TestDox.php
@@ -14,17 +14,21 @@ namespace PHPUnit\Event\Code;
  *
  * @no-named-arguments Parameter names are not covered by the backward compatibility promise for PHPUnit
  */
-final readonly class TestDox
+final class TestDox
 {
-    private string $prettifiedClassName;
-    private string $prettifiedMethodName;
-    private string $prettifiedAndColorizedMethodName;
+    private readonly string $prettifiedClassName;
 
-    public function __construct(string $prettifiedClassName, string $prettifiedMethodName, string $prettifiedAndColorizedMethodName)
+    /** @var callable */
+    private $prettifiedMethodNameCallable;
+
+    /** @var callable */
+    private $prettifiedAndColorizedMethodNameCallable;
+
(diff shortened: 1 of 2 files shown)


sebastianbergmann:

`PHPUnit\Event\Code\TestDox` is a value object and must not do anything else than hold the values and provide access to it.

staabm:
yeah I realized it, after everything broke when changing it :-).

the only place we currently use colorizing of the pretty method name is here:
https://github.com/sebastianbergmann/phpunit/blob/3b2a45569d40bf7ca06926c1c3b23d473d880312/src/TextUI/Output/TestDox/ResultPrinter.php#L101

what do you think about moving the colorization part only out of the TestDox class and move it instead in the only class which uses it?

sebastianbergmann:
Makes sense, I think.

This would mean, though, that the optional `$colorize` parameter of `TestDox::prettifiedMethodName(bool $colorize = false)` would no longer have an effect as the value `TestDox` object would only have the prettified method name without colors. This would be a (very tiny) break of backward compatibility.

I am all for implementing this in the next major version. I am still pondering whether this break of backward compatibility could be considered negligible, allowing us to do it in the next patch release.

staabm:
did some research on github.com and I was not able to find a single project which call `TestDox->prettifiedMethodName(true)` (except phpunit forks)

PR #5739, opened 2024-03-12
Micro-optimize `NamePrettifier::prettifyTestMethodName()`

staabm:

grafik

after the change it no longer appears


sebastianbergmann:

PR #5740, opened 2024-03-12
Micro-optimize `TestRunner::runTestWithTimeout()`

staabm:

grafik


sebastianbergmann:

staabm:
should I target future PRs directly to 10.5 ?

sebastianbergmann:
If they also apply to `10.5` then yes, please. Thanks!

PR #5742, opened 2024-03-12
Prevent file IO when not strictly necessary

staabm:
file IO largely depends on hardware and performance can vary a lot. it needs to be avoided at all cost.
especially on windows IO is slow (in comparison to linux based systems).

this PR re-orders conditions to do the IO only at the very last.

### DIFF
--- a/.psalm/baseline.xml
+++ b/.psalm/baseline.xml
@@ -927,8 +927,6 @@
       <code><![CDATA[listTestsXml]]></code>
       <code><![CDATA[logEventsText]]></code>
       <code><![CDATA[logEventsText]]></code>
-      <code><![CDATA[logEventsText]]></code>
-      <code><![CDATA[logEventsVerboseText]]></code>
       <code><![CDATA[logEventsVerboseText]]></code>
       <code><![CDATA[logEventsVerboseText]]></code>
       <code><![CDATA[logfileJunit]]></code>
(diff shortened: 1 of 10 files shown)


Yurunsoft:
If I use `set_error_handler()` in my program to catch errors and turn warnings into exceptions, this PR will make it impossible to test.

```log
Message:  file_get_contents(/imi/tests/.phpunit.result.cache): Failed to open stream: No such file or directory
Location: /imi/vendor/phpunit/phpunit/src/Runner/ResultCache/DefaultResultCache.php:88
```

<https://github.com/imiphp/imi/actions/runs/8342624739/job/22831196788>

staabm:
Sounds like your error handler is not properly detect the `@` suppression. 

Could you provide more information and a full reproducer (please open a new issue)

sebastianbergmann:
@staabm This was now reported again as #5764. While I do think that this caused by custom error handlers that are too eager and/or interfere with PHPUnit's error handler, I am leaning towards (partially) reverting these changes.

staabm:
@sebastianbergmann let me try to clean this up a bit, so we are more generous in such cases

Yurunsoft:
This PR fixes my issue: https://github.com/imiphp/imi/pull/683

Use `0 === (error_reporting() & $errno)` in `set_error_handler`

staabm:
@Yurunsoft your PR looks good. 

The latest phpunit release relaxed this PRs changes/regressions with https://github.com/sebastianbergmann/phpunit/pull/5765

PR #5747, opened 2024-03-13
Cache result of `Groups::groups()`

staabm:

grafik

----

after this PR the calls are gone

grafik

### DIFF
--- a/.psalm/baseline.xml
+++ b/.psalm/baseline.xml
@@ -640,13 +640,6 @@
     </RedundantCondition>
   </file>
   <file src="src/Metadata/Api/Groups.php">
-    <LessSpecificReturnStatement>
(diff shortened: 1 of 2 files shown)


staabm:
I don't know what todo with these psalm errors. they don't make sense to me.

It looks like this method never returned a `list<string>`, because `array_unique` might return a array with holes (which contradics the list type)

staabm:
> You need to wrap it with `array_values(array_unique(…))`, this gets rid of potential holes and is usually accepted by psalm & friends.

yes, but that would change the logic of the function. I am not sure the types are wrong or the implementation :)

staabm:
I had a closer look at the logic involved and the callers. it seems the callers don't need a `list` and fixing the types also fixes some baseline errors. I think thats the way to go.

Issue #5230, opened 2023-02-21
`@runClassInSeparateProcess` has the same effect as `@runTestsInSeparateProcesses`


sebastianbergmann:
As previously reported in #3258, `@runClassInSeparateProcess` has the same effect as `@runTestsInSeparateProcesses`.

Unless I am mistaken (which I may well be), this can be proven by applying ...

```patch
diff --git a/tests/end-to-end/regression/2724/SeparateClassRunMethodInNewProcessTest.php b/tests/end-to-end/regression/2724/SeparateClassRunMethodInNewProcessTest.php
index c62c030be..4bbaf9c7d 100644
--- a/tests/end-to-end/regression/2724/SeparateClassRunMethodInNewProcessTest.php
+++ b/tests/end-to-end/regression/2724/SeparateClassRunMethodInNewProcessTest.php
@@ -46,4 +46,12 @@ public function testTestMethodIsRunInSeparateProcess(): void
         $this->assertNotSame(self::INITIAL_PARENT_PROCESS_ID, self::$parentProcessId);
         $this->assertNotSame(self::$processId, self::$parentProcessId);
     }
+
+    /**
+     * @depends testTestMethodIsRunInSeparateProcess
+     */
+    public function testTestMethodIsRunInSameProcessAsOtherTestMethodsOfThisTestClass(): void
+    {
+        $this->assertSame(self::$processId, \getmypid());
+    }
 }
```

... to augment an existing test:

```
./phpunit tests/end-to-end/regression/2724-diff-pid-from-parent-process.phpt                                         
PHPUnit 9.6.3-21-g9da10e7e66 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.2.3
Configuration: /usr/local/src/phpunit/phpunit.xml

F                                                                   1 / 1 (100%)

Time: 00:00.262, Memory: 4.00 MB

There was 1 failure:

1) /usr/local/src/phpunit/tests/end-to-end/regression/2724-diff-pid-from-parent-process.phpt
Failed asserting that string matches format description.
--- Expected
+++ Actual
@@ @@
 PHPUnit 9.6.3-21-g9da10e7e66 by Sebastian Bergmann and contributors.
 
-.                                                                   1 / 1 (100%)
+.F                                                                  2 / 2 (100%)
 
 Time: 00:00.156, Memory: 4.00 MB
 
-OK (1 test, 3 assertions)
+There was 1 failure:
+
[...]

Issue #5749, opened 2024-03-14
reduce process isolation overhead


staabm:
while looking more into phpunit performance, I was profilling subprocesses:

![grafik](https://github.com/sebastianbergmann/phpunit/assets/120441/2a34593c-7247-4c10-865c-8d6af52a8442)

I think it is pretty interessting that running a test in isolation can be dominated by e.g. script compile time.

I wonder whether we can - per default - set a few options for the subprocesses used by process-isolation, e.g. 
- `-d opcache.jit=disable)` to disable jit overhead
- `-d opcache.validate_timestamps=0`

---

another thing which came to mind: could we place the test beeing run separate from the "framework" code required to run in isolation, so the "bootstrapping" of the isolated processes don't need to re-compile everything but at best only the actual test-case?

staabm:
some more food for thinking: maybe we can utilize pcntl_fork() instead of creating subprocesses.
thesis is: a forked process already contains everything the parent contained, so when forked 'early enough'  all of phpunit would already be loaded and would not need to be loaded again

staabm:
bit of progress:

> I wonder whether we can - per default - set a few options for the subprocesses used by process-isolation, e.g.
> 
>     * `-d opcache.jit=disable)` to disable jit overhead
> 
>     * `-d opcache.validate_timestamps=0`
> 
>     * `-d zend.enable_gc=0` disable GC?

I tried these and did not find any measurable differences..



> maybe we can utilize pcntl_fork() instead of creating subprocesses.

started looking into a POC for subprocess forking. stay tuned.

MasonM:
I've also been investigating speeding up our test suite that makes heavy use of process isolation, but unfortunately haven't found anything that helps without major caveats.

You mentioned file-based opcaching. I tried that by setting `opcache.file_cache=/tmpfs`, but unfortunately this causes segfaults (both on `aarch64` and `amd64`):
```
$ php --version
PHP 8.2.10 (cli) (built: Sep  4 2023 08:13:17) (NTS)
Copyright (c) The PHP Group
Zend Engine v4.2.10, Copyright (c) Zend Technologies
    with Zend OPcache v8.2.10, Copyright (c), by Zend Technologies
    with Xdebug v3.2.1, Copyright (c) 2002-2023, by Derick Rethans

$  time phpunit -c ./phpunit.xml-dist tests/JsonHelperTest.php
PHPUnit 9.5.23 #StandWithUkraine

[...]

PR #5233, opened 2023-02-22
Fixed #5230: @runClassInSeparateProcess has the same effect as @runTestsInSeparateProcesses

WalterWoshid:
Did some refactoring and now it works :)

I also did some testing for my personal use case and it works!!!

PR #5751, opened 2024-03-16
POC: lightweight subprocess isolation via pcntl_fork()

staabm:
still work in progress.. 

refs https://github.com/sebastianbergmann/phpunit/issues/5749#issuecomment-1999325702

### DIFF
--- a/src/Framework/TestRunner.php
+++ b/src/Framework/TestRunner.php
@@ -9,6 +9,7 @@
  */
 namespace PHPUnit\Framework;
 
+use PHPUnit\TestRunner\TestResult\PassedTests;
 use const PHP_EOL;
 use function assert;
 use function class_exists;
@@ -249,6 +250,86 @@ final class TestRunner
      * @throws StaticAnalysisCacheNotConfiguredException
      */
     public function runInSeparateProcess(TestCase $test, bool $runEntireClass, bool $preserveGlobalState): void
+    {
+        if ($runEntireClass && $this->isPcntlForkAvailable()) {
+            // forking the parent process is a more lightweight way to run a test in isolation.
+            // it requires the pcntl extension though.
+            $this->runInFork($test);
+            return;
+        }
+
+        // running in a separate process is slow, but works in most situations.
+        $this->runInWorkerProcess($test, $runEntireClass, $preserveGlobalState);
+    }
+
+    private function isPcntlForkAvailable(): bool {
+        $disabledFunctions = ini_get('disable_functions');
+
+        return
+            function_exists('pcntl_fork')
+            && !str_contains($disabledFunctions, 'pcntl')
+            && function_exists('socket_create_pair')
+            && !str_contains($disabledFunctions, 'socket')
+        ;
+    }
+
+    private function runInFork(TestCase $test): void
+    {
+         if (socket_create_pair(AF_UNIX, SOCK_STREAM, 0, $sockets) === false) {
+             throw new \Exception('could not create socket pair');
+         }
+
+        $pid = pcntl_fork();
+        // pcntl_fork may return NULL if the function is disabled in php.ini.
+        if ($pid === -1 || $pid === null) {
+            throw new \Exception('could not fork');
+        } else if ($pid) {
+            // we are the parent
+
+            pcntl_waitpid($pid, $status); // protect against zombie children
+
+            // read child output
+            $output = '';
+            while(($read = socket_read($sockets[1], 2048, PHP_BINARY_READ)) !== false) {
+                $output .= $read;
+            }
+            socket_close($sockets[1]);
+
+            $php = AbstractPhpProcess::factory();
+            $php->processChildResult($test, $output, ''); // TODO stderr
+
+        } else {
+            // we are the child
+
+            $offset                  = hrtime();
+            $dispatcher = Event\Facade::instance()->initForIsolation(
+                \PHPUnit\Event\Telemetry\HRTime::fromSecondsAndNanoseconds(
+                    $offset[0],
+                    $offset[1]
+                )
+            );
+
+            $test->setInIsolation(true);
+            $test->runBare();
+
+            // send result into parent
+            socket_write($sockets[0],
+                serialize(
+                    [
+                        'testResult'    => $test->result(),
+                        'codeCoverage'  => CodeCoverage::instance()->isActive() ? CodeCoverage::instance()->codeCoverage() : null,
+                        'numAssertions' => $test->numberOfAssertionsPerformed(),
+                        'output'        => !$test->expectsOutput() ? $output : '',
+                        'events'        => $dispatcher->flush(),
+                        'passedTests'   => PassedTests::instance()
+                    ]
+                )
+            );
+            socket_close($sockets[0]);
+        }
+    }
+
+    private function runInWorkerProcess(TestCase $test, bool $runEntireClass, bool $preserveGlobalState): void
     {
         $class = new ReflectionClass($test);
 
--- a/src/Framework/TestRunner.php
+++ b/src/Framework/TestRunner.php
@@ -10,6 +10,7 @@
 namespace PHPUnit\Framework;
 
 use PHPUnit\TestRunner\TestResult\PassedTests;
+use PHPUnit\Util\PHP\PcntlFork;
 use const PHP_EOL;
 use function assert;
 use function class_exists;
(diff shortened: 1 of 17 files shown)


sebastianbergmann:
Keep in mind that isolation is more important than performance. I mention this because I do not know how much global state (included files, declared constants, loaded classes, global variables, ...) carries over from the parent process into the child process when forking is used.

staabm:
In my understanding the goal of process isolation is to isolate the tests from each other.

Is it also the goal to isolate phpunit from the test?

sebastianbergmann:
> Is it also the goal to isolate phpunit from the test?

No.

staabm:
perf comparison this PR:

```
➜  phpunit git:(fork) ✗ php ./phpunit tests/_files/TestWithClassLevelIsolationAttributes.php
PHPUnit 11.1-gadccd7efc by Sebastian Bergmann and contributors.

Runtime:       PHP 8.2.17

Time: 00:00.006, Memory: 8.00 MB

There was 1 risky test:

1) PHPUnit\TestFixture\TestBuilder\TestWithClassLevelIsolationAttributes::testOne
This test did not perform any assertions

/Users/staabm/workspace/phpunit/tests/_files/TestWithClassLevelIsolationAttributes.php:24

OK, but there were issues!
Tests: 1, Assertions: 0, Risky: 1.
```

vs. main-branch@4303b9eb2

```
➜  phpunit git:(main) php ./phpunit tests/_files/TestWithClassLevelIsolationAttributes.php
PHPUnit 11.1-g4303b9eb2 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.2.17

Time: 00:00.046, Memory: 8.00 MB

There was 1 risky test:

1) PHPUnit\TestFixture\TestBuilder\TestWithClassLevelIsolationAttributes::testOne
This test did not perform any assertions

/Users/staabm/workspace/phpunit/tests/_files/TestWithClassLevelIsolationAttributes.php:24

OK, but there were issues!
Tests: 1, Assertions: 0, Risky: 1.
```

POC looks 7-8x faster

src/Framework/TestRunner.php
```diff
@@ -249,6 +250,174 @@ public function run(TestCase $test): void
...
+            $dispatcher = Event\Facade::instance()->initForIsolation(
+                \PHPUnit\Event\Telemetry\HRTime::fromSecondsAndNanoseconds(
+                    $offset[0],
+                    $offset[1]
+                )
+            );
+
+            $test->setInIsolation(true);
+            try {
+                $test->run();
```
staabm, line 397: I think I need to check `$runEntireClass` at this point and run only a single method. not sure how yet.
(in a similar way it is done in  `runInWorkerProcess()`)

MasonM (commented):
One problem with `pcntl_fork()` is database connections. If the parent process has a connection to MySQL/PostgreSQL/etc, the child process will inherit that resource. When the child exits, the connection will be automatically closed, which means subsequent children will receive connection errors. More details at https://www.php.net/manual/en/function.pcntl-fork.php#70721

staabm:
If a test creates a resource which needs isolation, said test needs to run in isolation. 

As soon as a single test runs in the main process with something which has sideeffects on other tests, things break. But thats not unique to forking. You have the same problem with worker processes, as soon as the test which requires isolation runs not in a worker

MasonM:
@staabm You're correct that the current process-based isolation mechanism (which uses `proc_open()`) also has problems with side effects under certain circumstances, but database connections aren't one of them. Database connections aren't inherited when `proc_open()` is used, but they will with this change, which means it's backwards-incompatible. Creating database connections in the main process before running tests with `@runInSeparateProcess` is a common pattern in the codebases I work with.

You could prevent breaking backwards compatibility by making this opt-in using a new config option or attribute, but this is a pretty big footgun.

staabm:
I see. Thanks for mentioning. I guess if this PR propsal will land, we will definitely need a opt-in until we get some decent real world battle test experience

MasonM:
No problem. I'm happy to run any proposed changes on our codebase, which make heavy use of process isolation and would greatly benefit from any performance benefits. I came across this while doing my own research on optimizing such tests, but unfortunately I haven't found anything that helps without major caveats.

staabm:
I have just added annotation support, so one can opt-in to fork based isolation via attribute.

example test which uses fork based isolation:
```
/*
 *
 *
 */
namespace PHPUnit\TestFixture\Metadata\Attribute;

use PHPUnit\Framework\Attributes\RunClassInSeparateProcess;
use PHPUnit\Framework\Attributes\RunTestsInSeparateProcesses;

#[RunClassInSeparateProcess(true)]
#[RunTestsInSeparateProcesses]
final class ProcessIsolationForkedTest extends TestCase
{
    #[RunInSeparateProcess]
    {
    }
}
```

the `true` in `#[RunClassInSeparateProcess(true)]` indicates that fork should be used instead of a separate worker.
that way we can easily ship the feature and don't break existing assumptions of worker based isolation per default.

tbh I don't like bool attribute parameters, but I found it would be consistent with other attributes.
IMO something like `#[RunClassInSeparateProcess("forkIfPossible")]` would be easier to read.

this PR is also missing some more end-to-end tests, but I wanted to discuss the UX and design before going into full test-coverage 

I am open for feedback

staabm:
@sebastianbergmann any feedback on https://github.com/sebastianbergmann/phpunit/pull/5751#issuecomment-2026922512 ?

sebastianbergmann:
> Currently I don't plan to also add phpdoc annotation support, as I read this will be dropped in the next major

No new annotations are added since PHPUnit 10 came out.

> the `true` in `#[RunClassInSeparateProcess(true)]` indicates that fork should be used

I am not sure whether I like the boolean flag, I think I would prefer a separate attribute for this. Don't worry, I would implement the attribute and metadata-related changes after merging this.

I think we should resolve #5230 / #5233 before working on this, though. What do you think?

sebastianbergmann:
@staabm Heads-up: I have cleaned up the code a bit:

- Introduced [`PhpJob`](https://github.com/sebastianbergmann/phpunit/blob/0805e6f0d8b77ede3aa54c4ce6e0a045157f39c8/src/Util/PHP/PhpJob.php) value object that encapsulates all relevant information about a job
- Introduced [`DefaultPhpJobRunner`](https://github.com/sebastianbergmann/phpunit/blob/0805e6f0d8b77ede3aa54c4ce6e0a045157f39c8/src/Util/PHP/DefaultPhpJobRunner.php) class for running `PhpJob`s using the tried-and-true implementation using `proc_open()` etc.
- Migrated existing code to use `DefaultPhpJobRunner`

`DefaultPhpJobRunner` implements the [`PhpJobRunner`](https://github.com/sebastianbergmann/phpunit/blob/0805e6f0d8b77ede3aa54c4ce6e0a045157f39c8/src/Util/PHP/PhpJobRunner.php) interface. For now, this interface is `@internal` and no mechanism is in place to configuration an implementation to use instead of `DefaultPhpJobRunner`. This might change in the future if and when we identify a need for this (sooner if you tell me that you would like to implement your `pcntl_fork()` based approach that is the topic of this PR as a `PhpJobRunner` implementation.

staabm:
thanks for the groundwork. I am planning to bring this pctnl based process isolation into phpunit, if we find a way which is acceptable for you. I think it can be really useful to have a faster isolation alternative.

sebastianbergmann:
> thanks for the groundwork. I am planning to bring this pctnl based process isolation into phpunit, if we find a way which is acceptable for you. I think it can be really useful to have a faster isolation alternative.

Okay, then I'll try to find time ASAP to make this pluggable. I already how to do it, just need to do it.

sebastianbergmann:
As of 83e157ad71884051da758f4445cfbb469050f602, [`PhpJobRunnerRegistry`](https://github.com/sebastianbergmann/phpunit/blob/83e157ad71884051da758f4445cfbb469050f602/src/Util/PHP/PhpJobRunnerRegistry.php) can be used to configure a custom `PhpJobRunner` implementation.

staabm:
thx for your work. I tried rebasing the PR, but I am not sure yet, whether we have all pieces in the `JobRunner` to implement a pcntl variant.

I need a way to run the job in the current process like I did in 

https://github.com/sebastianbergmann/phpunit/pull/5751/files#diff-2843ca5c5620803be3ae210b4cd8f4f74eac269d29f3fef343ce212a3aed05f1R92-R119

but it seems `PhpJob` alone does not provide enough context so I can run the test-case. Maybe I missed something..?

sebastianbergmann:
> Maybe I missed something?

I doubt that you are missing something. The `PhpJob` value object currently only encapsulates the information that the default implementation of `PhpJobRunner` needs. What exactly do you need?

staabm:
in a `PcntlJobRunner` I would need to fork the current process. 

after forking
- in the main process I would need to wait for the forked worker process to end an report the result
- in the child process I would need to run the test itself, report the result to the main process and kill the child

currently I see no way to run the test-case from within a `JobRunner`.
I think therefore I need a reference to `TestCase` within my `PcntlJobRunner`.

staabm:
just realized that [psalm is also using workers via fork](https://blog.daniil.it/2025/05/11/psalm-v6-deep-dive-copy-on-write-dynamic-task-dispatching/)

the implementation is beeing [backported into amphp/parallel ](https://github.com/amphp/parallel/pull/212)

staabm:
superseded by https://github.com/sebastianbergmann/phpunit/pull/6288

Issue #5719, opened 2024-03-02
Prepare dependencies for deprecation of implicitly nullable parameter types in PHP 8.4


sebastianbergmann:
@kocsismate put forward an RFC to [deprecate implicitly nullable parameter types](https://wiki.php.net/rfc/deprecate-implicitly-nullable-types) for PHP 8.4.

PHPUnit >= 8.5 as well as its dependencies need to be updated to not trigger the respective deprecation warning.

I have already updated PHPUnit's own code on the `8.5`, `9.6`, `10.5`, `11.0`, and `main` branches as well as the code of PHPUnit's dependencies for which I am also responsible.

I used this [fork](https://gist.github.com/sebastianbergmann/38b138b5254cc39a7c1136dc5a0ce315) of @kocsismate's [script](https://gist.github.com/kocsismate/cf3bdfbf35eb10224ee5ecd29b39656b) to find implicitly nullable parameter types in PHPUnit's own code as well as its dependencies.

Below is a list of implicitly nullable parameter types in PHPUnit's dependencies, shown per PHPUnit version, that are not under my control.

@theseer and I discussed that I will create pull requests for `phar-io/manifest` and `theseer/tokenizer` that he will merge. New releases of these two packages without implicitly nullable parameter types should be tagged within the next few days.

@nikic Would you accept a pull request for `nikic/php-parser` version 4.18 that removes implicitly nullable parameter types? If so, I would prepare that.

#### PHPUnit 8.5 (`8.5` branch)

- [X] `vendor/phar-io/manifest/src/values/Manifest.php:78` (from `phar-io/manifest` version 2.0.3 by @theseer)
- [X] `vendor/theseer/tokenizer/src/XMLSerializer.php:22` (from `theseer/tokenizer` version 1.2.2 by @theseer)

#### PHPUnit 9.6 (`9.6` branch)

- [X] `vendor/phar-io/manifest/src/values/Manifest.php:78` (from `phar-io/manifest` version 2.0.3 by @theseer)
- [X] `vendor/theseer/tokenizer/src/XMLSerializer.php:22` (from `theseer/tokenizer` version 1.2.2 by @theseer)
- [ ] `vendor/nikic/php-parser/lib/PhpParser/ConstExprEvaluator.php:40` (from `nikic/php-parser` version 4.18.0 by @nikic)
[...]

PR #5753, opened 2024-03-18
Update nikic/php-parser

mbeccati:
Refs #5719


mbeccati:
Thanks @sebastianbergmann!

I can confirm my independent OSS test suite vs latest PHP branches has turned green again for PHPUnit: https://github.com/mbeccati/php-latest-builds/actions/runs/8338666315

Issue #3536, opened 2019-02-20
Add support for stubbing and mocking closures


HeahDude:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 8.1
| PHP version         | 7.x
| Installation Method | Composer / PHAR

Would you accept a PR to support mocking closure, to ease checking that one is called?

I.e : https://3v4l.org/Yaq5A

```php
public function testClosureIsCalled()
{
    $callback = function () {};

    $this->assertClosureWillBeCalled($callback);

    SomeBusiness:caller($callback);
}
```

sebastianbergmann:
I cannot promise acceptance of such a pull request but would, of course, consider it.

morozov:
This is already possible by mocking `__invoke()` on `stdClass` in one line:
```php
public function testClosureIsCalled() : void
{
        $closure = $this->createPartialMock(stdClass::class, ['__invoke']);

        $closure->expects(
            $this->exactly(2)
        )->method('__invoke');

        $closure();
        $closure();
}
```
Unlike this workaround, the proposed API is not really flexible, however, it'd be nice if the workaround was formalized as a feature.

veewee:
@morozov 
The provided solution results in:

> createPartialMock() called with method(s) __invoke that do not exist in stdClass. This will not be allowed in future versions of PHPUnit.

An API for this might be nice indeed!

ChrisHSandN:
After some experimentation there still seems to be a method which works:

```js
$mock = $this
  ->getMockBuilder(\stdclass::class)
  ->addMethods(['__invoke'])
  ->getMock();

$mock
  ->expects(self::once())
  ->method('__invoke'));
```

(Tested on PHPUnit 9 / PHP 7.3)

mnapoli:
Following up on this: https://github.com/sebastianbergmann/phpunit/pull/5759

PR #5759, opened 2024-03-20
Add closure mocks

mnapoli:
## The problem

Mocking closures is doable today by mocking `__invoke` on an invokable class. However, this is troublesome, as it requires writing an invokable class and mocking a class method (which is confusing, hard to figure out, and makes tests harder to understand afterward).

Here is another approach that is verbose and fragile:

```php
    public function test_validation(): void
    {
        $rule = new CustomLaravelValidationRule();
        
        $failed = false;
        $rule->validate('code', 'abcd', function () use (&$failed) {
            return $failed = true;
        });
        $this->assertTrue($failed);
    }
```

## The solution

This new helper makes things easier via a new `$this->createClosureMock()` method.

```php
$mock = $this->createClosureMock();

$mock->expectsClosure($this->once())
    ->willReturn(123);

$this->assertSame(123, $mock());
```

Here's the test above rewritten using the new helper:

```php
    public function test_validation(): void
    {
        $rule = new CustomLaravelValidationRule();
        
        $mock = $this->createClosureMock();
        $mock->expectsClosure($this->never());

        $rule->validate('code', 'abcd', $mock);
    }
```

This is essentially just a shortcut to the workaround that is doable today, so this PR should not introduce too much new code.

Related to #3536

Note that #3536 mentioned a workaround using `stdClass`. I was not able to make it work: PHPUnit refused to mock stdClass's `__invoke` method because it does not exist.

This is, to me, yet another reason to have such a helper.

### DIFF
--- a/src/Framework/MockObject/Runtime/Stub/ClosureMock.php
+++ b/src/Framework/MockObject/Runtime/Stub/ClosureMock.php
@@ -1,0 +1,45 @@
+<?php declare(strict_types=1);
+/*
+ * This file is part of PHPUnit.
+ *
+ * (c) Sebastian Bergmann <sebastian@phpunit.de>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+namespace PHPUnit\Framework\MockObject\Stub;
+
+use PHPUnit\Framework\InvalidArgumentException;
+use PHPUnit\Framework\MockObject\Builder\InvocationMocker;
+use PHPUnit\Framework\MockObject\MethodCannotBeConfiguredException;
+use PHPUnit\Framework\MockObject\MethodNameAlreadyConfiguredException;
+use PHPUnit\Framework\MockObject\Rule\InvocationOrder;
+
+/**
+ * @mixin \PHPUnit\Framework\MockObject\MockObject
+ *
+ * @method InvocationMocker method($constraint)
+ */
+class ClosureMock
+{
+    public function __invoke(): mixed
+    {
+        return null;
+    }
+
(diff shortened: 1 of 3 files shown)


src/Framework/MockObject/Runtime/Stub/ClosureMock.php
```diff
@@ -0,0 +1,45 @@
...
+use PHPUnit\Framework\InvalidArgumentException;
+use PHPUnit\Framework\MockObject\Builder\InvocationMocker;
+use PHPUnit\Framework\MockObject\MethodCannotBeConfiguredException;
+use PHPUnit\Framework\MockObject\MethodNameAlreadyConfiguredException;
+use PHPUnit\Framework\MockObject\Rule\InvocationOrder;
+
+/**
+ * @mixin \PHPUnit\Framework\MockObject\MockObject
+ *
+ * @method InvocationMocker method($constraint)
```
mnapoli, line 21: I don't understand why Psalm does not recognize this annotation. PhpStorm does, and I used the same annotation as in `MockObject`: https://github.com/sebastianbergmann/phpunit/blob/437af97b42ed0431819f44a45a85f3c6bef8fac9/src/Framework/MockObject/Runtime/Interface/MockObject.php#L16

mnapoli:
I rebased and the static analysis error was solved 🎉

This is ready to be reviewed/merged.

The code coverage is not happy that `__invoke()` is not covered, but this is an empty method and it's the one being mocked, so that makes sense, I think it's perfectly fine that it's not covered. If you want me to find a way to force coverage let me know.

PR #5766, opened 2024-03-23
Do not use a shell in `proc_open()` if not really needed

staabm:
refs https://github.com/sebastianbergmann/phpunit/issues/5749#issuecomment-2016012543

in my testing I can confirm on PHP 8.3 the process starts 1-2ms faster.
a hello world process isolation test on my machine runs in 46-47ms, so 1-2ms faster is not too bad actually.

requires https://github.com/sebastianbergmann/environment/pull/72

PR #5772, opened 2024-03-28
Cleanup process handling after dropping temp-file handling

staabm:
after https://github.com/sebastianbergmann/phpunit/pull/5766 we can remove some bits of unused code now.


src/Util/PHP/DefaultPhpProcess.php
```diff
@@ -40,7 +40,7 @@ class DefaultPhpProcess extends AbstractPhpProcess
      */
     public function runJob(string $job, array $settings = []): array
     {
-        if ($this->stdin || $this->useTemporaryFile()) {
+        if ($this->stdin) {
             if (!($this->tempFile = tempnam(sys_get_temp_dir(), 'phpunit_')) ||
```
staabm, line 44: we still need the temp-file for PHPT cases in which [STDIN is provided via test-case](https://github.com/sebastianbergmann/phpunit/commit/b9100692ac6a15c75ab2be06976c9d27de0f7bc5)

PR #5773, opened 2024-03-28
Make exporter configurable

sebastianbergmann:
PHPUnit uses `SebastianBergmann\Exporter\Exporter` from `sebastian/exporter` to export variables as strings, for instance when generating comparison failure messages or when creating immutable value objects representing tests and their data (from data providers) that are passed through the event system to internal as well as third-party subscribers.

In order to allow for the customization of the textual representation of comparison failures, to improve performance (using `SebastianBergmann\Exporter\Exporter` is expensive when large data structures are involved), and for other use cases it would be nice if `SebastianBergmann\Exporter\Exporter` could be swapped for a different implementation.

- [X] Introduce `PHPUnit\Util\Exporter` interface
- [X] Implement `PHPUnit\Util\DefaultExporter` as the default implementation of `PHPUnit\Util\Exporter` that simply wraps `SebastianBergmann\Exporter\Exporter`
- [X] Implement `PHPUnit\Util\ExporterFacade` that uses `PHPUnit\Util\DefaultExporter` by default, but can be configured to use an alternative implementation of `PHPUnit\Util\Exporter`
- [X] Implement `PHPUnit\Util\ExporterChain` for chaining `PHPUnit\Util\Exporter` implementations
- [X] Use `ExporterFacade::instance()->export()` instead of `SebastianBergmann\Exporter\Exporter::export()`
- [ ] Make alternative implementations of `PHPUnit\Util\Exporter` configurable via PHPUnit's XML configuration file
- [ ] Explore whether `SebastianBergmann\Exporter\Exporter::shortenedRecursiveExport()` and `SebastianBergmann\Exporter\Exporter::shortenedExport` should also be wrapped
- [ ] Explore whether the usage of `SebastianBergmann\Exporter\Exporter` in `sebastian/comparator` should also be configurable

### DIFF
--- a/src/Event/Emitter/DispatchingEmitter.php
+++ b/src/Event/Emitter/DispatchingEmitter.php
@@ -27,7 +27,7 @@ use PHPUnit\Event\TestSuite\Sorted as TestSuiteSorted;
 use PHPUnit\Event\TestSuite\Started as TestSuiteStarted;
 use PHPUnit\Event\TestSuite\TestSuite;
 use PHPUnit\TextUI\Configuration\Configuration;
-use SebastianBergmann\Exporter\Exporter;
+use PHPUnit\Util\ExporterFacade;
 
 /**
  * @internal This class is not covered by the backward compatibility promise for PHPUnit
@@ -582,7 +582,7 @@ final class DispatchingEmitter implements Emitter
             new Test\TestProxyCreated(
                 $this->telemetryInfo(),
                 $className,
-                (new Exporter)->export($constructorArguments),
+                ExporterFacade::instance()->export($constructorArguments),
             ),
         );
     }
--- a/src/Event/Value/Test/TestMethodBuilder.php
+++ b/src/Event/Value/Test/TestMethodBuilder.php
@@ -17,8 +17,8 @@ use PHPUnit\Event\TestData\DataFromTestDependency;
 use PHPUnit\Event\TestData\TestDataCollection;
 use PHPUnit\Framework\TestCase;
 use PHPUnit\Metadata\Parser\Registry as MetadataRegistry;
+use PHPUnit\Util\ExporterFacade;
 use PHPUnit\Util\Reflection;
-use SebastianBergmann\Exporter\Exporter;
 
 /**
  * @internal This class is not covered by the backward compatibility promise for PHPUnit
@@ -60,7 +60,6 @@ final readonly class TestMethodBuilder
 
     private static function dataFor(TestCase $testCase): TestDataCollection
     {
-        $exporter = new Exporter;
         $testData = [];
 
         if ($testCase->usesDataProvider()) {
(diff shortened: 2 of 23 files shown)


staabm:
while I agree that a configurable exporter can have benefit for specific use-cases, I would expect the "default exporter" to be fast enough, so I don't need to structure my unit test arround a "limitation" of phpunit, that objects/arrays in data-providers are not efficient enough.

no offense on this end - just my point of view.

sebastianbergmann:
> no offense on this end - just my point of view

None taken.

>  I would expect the "default exporter" to be fast enough

Sure!

As I said before, I am not doing this for performance. https://github.com/sebastianbergmann/exporter/pull/55 is likely a more appropriate solution for performance issues.

BladeMF:
Isn't it better architecture here to have a Factory where one can add multiple exporters (in a config file, probably? or a bootstrap file?) that are called for each type? E.g.
```php
class ExporterFactory 
{
  public function addExporter(ExporterInterface $exporter) { ... }
  
  public function getExporterForValue(mixed $value) {
    foreach($this->exporters as $exporter) {
      if($exporter->supports($value)) return $exporter;
    }
    return $this->defaultExporter;
  }
}
```
That way doctrine can create their own exporter, I can make my own, etc.

sebastianbergmann:
@BladeMF I wanted to keep this as simple as possible, but will ponder your suggestion.

sebastianbergmann:
>How will the `use` method be called? A configuration value (seems best, no)?

That is what I mean by "Make alternative implementation of PHPUnit\Util\Exporter configurable via PHPUnit's XML configuration file" in https://github.com/sebastianbergmann/phpunit/pull/5773#issue-2213075643.

>* Will the custom exporter have access to the defaul timplementation to defer some of the work to?

Currently not, but I will ponder that.

BladeMF:
Thanks for looking into it.

sebastianbergmann:
> Will the custom exporter have access to the defaul timplementation to defer some of the work to?

I think the best way would be for custom implementations of `PHPUnit\Util\Exporter` to create their own instance of `SebastianBergmann\Exporter\Exporter` when needed.

sebastianbergmann:
> where one can add multiple exporters

I just created `PHPUnit\Util\ExporterChain` as an implementation of `PHPUnit\Util\Exporter` to allow for multiple implementations to be used. The first one to return `true` for `PHPUnit\Util\ExporterChain::handles()` will be used. The default implementation of `PHPUnit\Util\Exporter` will be used if no other implementation handles the value to be exported.

sebastianbergmann:
I slept on it and now think that `SebastianBergmann\Exporter\Exporter` should be changed to allow for customizable (and chainable) handling of objects. I do not think that the exporting of scalar values needs to be customizable and the same goes for the recursive processing of arrays.

staabm:
Agree. I also wonder whether all callsites of `->export` should be configurable in the same way..? Do all call-sites work if the export shortens the information ? Or do some callsites exist which would break and always need the full export?

sebastianbergmann:
Here is a proof-of-concept change for `sebastian/exporter`: https://github.com/sebastianbergmann/exporter/pull/56

BladeMF:
> I do not think that the exporting of scalar values needs to be customizable and the same goes for the recursive processing of arrays.

Sounds reasonable.

PR #5774, opened 2024-03-28
Use `shortenedRecursiveExport()` for data-providers

staabm:
I just realized we already have a `shortenedRecursiveExport` version of the exporter, we just need to use

so instead of https://github.com/sebastianbergmann/exporter/pull/55 we could just do this single line change

tested this change on roave/BetterReflection:

phpunit 11.0.8
```
Time: 01:21.576, Memory: 3.79 GB

OK, but some tests were skipped!
Tests: 10383, Assertions: 65007, Skipped: 7.
```

this PR:
```
Time: 00:20.042, Memory: 1.04 GB

OK, but some tests were skipped!
Tests: 10383, Assertions: 65007, Skipped: 7.
```

---

maybe I did not yet understand that you actually try to make this type of exporter configurable via https://github.com/sebastianbergmann/phpunit/pull/5773 ?

### DIFF
--- a/src/Event/Value/Test/TestMethodBuilder.php
+++ b/src/Event/Value/Test/TestMethodBuilder.php
@@ -70,9 +70,11 @@ final readonly class TestMethodBuilder
                 $dataSetName = (int) $dataSetName;
             }
 
+            $providedData = $testCase->providedData();
+
             $testData[] = DataFromDataProvider::from(
                 $dataSetName,
-                $exporter->export($testCase->providedData()),
+                $exporter->shortenedRecursiveExport($providedData),
                 $testCase->dataSetAsStringWithData(),
             );
         }


staabm:
@sebastianbergmann does this PR depend on the exporter object PR or can it land independently for 11.x ?

sebastianbergmann:
Not completely exporting data provided by data providers is a BC break and should be opt-in. I am still not sure, though, what the right course of action is here "in the big picture". I hope to find time soon to think about this.

src/Event/Value/Test/TestMethodBuilder.php
```diff
@@ -70,9 +70,11 @@ private static function dataFor(TestCase $testCase): TestDataCollection
                 $dataSetName = (int) $dataSetName;
             }
 
+            $providedData = $testCase->providedData();
+
             $testData[] = DataFromDataProvider::from(
                 $dataSetName,
-                $exporter->export($testCase->providedData()),
+                $exporter->shortenedRecursiveExport($providedData),
                 $testCase->dataSetAsStringWithData(),
```
mvorisek, line 78: In our usecase, `$testCase->dataSetAsStringWithData()` consumes 99% of a runtime as big data provider arrays are dumper without limit. So I am not sure if this PR itself it really helpful...

The should be some test showing what data are problematic and how this helps.

staabm: Provide a reproducer for your problem, so we can an verify the fix or build a different one

mvorisek: Here is the slow provider testcode - https://github.com/mvorisek/sql-formatter/blob/de3d8d9808f130f7a8734f9e93ec3e216c77003f/tests/TokenizerTest.php#L90-L149.

The only dependency is this simple class - https://github.com/mvorisek/sql-formatter/blob/de3d8d9808f130f7a8734f9e93ec3e216c77003f/src/Token.php.

image

staabm: I tried reproducing what you described, but I can't see in profiles `dataSetAsStringWithData` taking a relevant share of resources.

please provide a full reproducer.

mvorisek: Here is a repro repo: https://github.com/mvorisek/sql-formatter/tree/repro_phpunit_5774_discussion

Repro command: composer update, run phpunit

The phpunit will output something like:
```
...
 Detected 2 tests where the duration exceeded the maximum duration.

1. ...
2. 3.249 (0.500) Doctrine\SqlFormatter\Tests\TokenizerTest::testTokenize#long concat

Time: 00:12.447, Memory: 397.35 MB

OK (283 tests, 294 assertions)
```

/wo dataprovider: https://github.com/doctrine/sql-formatter/actions/runs/9473730952/job/26101927778#step:7:25
/w dataprovider: https://github.com/mvorisek/sql-formatter/actions/runs/9536908812/job/26284385374#step:7:25

As the slowdown is non trivial, you should be able to measure/profile it easily, especially with `--filter 'testTokenize#long concat'` phpunit option.

staabm: so I can confirm, that this PR does not help for your case.

I wonder why you reported your problem here, as this PR solves a different problem (see PR description).
I think what you are reporting here is not related to this PR and should be reported as a separate issue...?

staabm:
btw: I am cooking up a alternative to this PR which contains of a fix in the exporter instead

staabm:
@sebastianbergmann I just saw you rebased this PR. do you have any new insights to share?

sebastianbergmann:
> @sebastianbergmann I just saw you rebased this PR. do you have any new insights to share?

I just noticed it's out-of-date and updated it.

staabm:
I see thanks. if we are still at a point in which we don't know whether this PR is acceptable, it might make sense to instead look more into things similar to https://github.com/sebastianbergmann/exporter/pull/55

(so looking into ways to shorten the export, when a `shortenedRecursiveExport` is requested

src/Event/Value/Test/TestMethodBuilder.php
staabm: fixed in https://github.com/sebastianbergmann/exporter/pull/59

staabm:
ondrej started upgrading PHPUnit to 11.x in PHPStan which leads to us facing this problem again.
the memory goes thru the roof and performance is considerably slower compared to PHPUnit 10.x/9.x (because of huge objects in dataproviders).

find details in https://github.com/phpstan/phpstan-src/pull/4057

Changing `vendor/phpunit/phpunit/src/Event/Value/Test/TestMethodBuilder.php` 

```diff
            $testData[] = DataFromDataProvider::from(
                $dataSetName,
-               Exporter::export($testCase->providedData()),
+               Exporter::shortenedRecursiveExport($testCase->providedData()),
                $testCase->dataSetAsStringWithData(),
            );
```

brings us back to PHPUnit 10 levels.

Is there anything I can do to bring this topic onto the agenda again?

src/Event/Value/Test/TestMethodBuilder.php
```diff
@@ -70,9 +70,11 @@ private static function dataFor(TestCase $testCase): TestDataCollection
                 $dataSetName = (int) $dataSetName;
             }
 
+            $providedData = $testCase->providedData();
+
             $testData[] = DataFromDataProvider::from(
                 $dataSetName,
-                $exporter->export($testCase->providedData()),
+                $exporter->shortenedRecursiveExport($providedData),
```
mvorisek, line 77: Does the exported correctly handle references/recursive arrays?

sebastianbergmann:
Addressed in 7ff096951d47b1e450a9b29daad84f273602d2c7 for PHPUnit 11.5 and b4a98841a44069c9ec9c4d2828fbf7390787ace8 for PHPUnit 12.2.

staabm:
Thank you 🙏

PR #5776, opened 2024-03-29
Narrow type of template parameter in Cloner

staabm:
fixes PHPStan errors, see https://phpstan.org/r/31bc15b5-0507-4739-a86d-9652e43bec9b

### DIFF
--- a/src/Util/Cloner.php
+++ b/src/Util/Cloner.php
@@ -17,7 +17,7 @@ use Throwable;
 final class Cloner
 {
     /**
-     * @psalm-template OriginalType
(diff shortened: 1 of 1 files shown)


src/Util/Cloner.php
```diff
@@ -17,7 +17,7 @@
 final class Cloner
 {
     /**
-     * @psalm-template OriginalType
+     * @psalm-template OriginalType of object
```
staabm, line 20: since the native type is already `object` this template narrowing is not a bc break, but just a bugfix to make PHPStan happier

PR #5777, opened 2024-03-29
Fix PHPStan reported issues

staabm:


src/TextUI/Output/Printer/DefaultPrinter.php
```diff
@@ -87,8 +89,6 @@ private function __construct(string $out)
             return;
         }
 
-        $this->isPhpStream = str_starts_with($out, 'php://');
```
staabm, line 90: > Class PHPUnit\TextUI\Output\DefaultPrinter has an uninitialized readonly property $isPhpStream. Assign it in the constructor.

src/Runner/PhptTestCase.php
```diff
@@ -186,7 +186,9 @@ public function run(): void
         $jobResult    = $this->phpUtil->runJob($code, $this->stringifyIni($settings));
         $this->output = $jobResult['stdout'] ?? '';
 
-        if (CodeCoverage::instance()->isActive() && ($coverage = $this->cleanupForCoverage())) {
```
staabm, line 189: > Right side of && is always true.

src/Runner/Baseline/Reader.php
```diff
@@ -85,7 +85,6 @@ public function read(string $baselineFile): Baseline
 
                     $description = $issueElement->textContent;
 
-                    assert(!empty($file));
```
staabm, line 88: > Variable $file in empty() always exists and is not falsy.

src/Framework/TestCase.php
```diff
@@ -1185,7 +1185,7 @@ protected function setLocale(mixed ...$arguments): void
             throw new Exception;
         }
 
-        $this->locale[$category] = setlocale($category, 0);
```
staabm, line 1188: > Parameter #2 $locale of function setlocale expects array|string|null, int given.

src/Framework/ExecutionOrderDependency.php
```diff
@@ -146,9 +149,6 @@ public function __construct(string $classOrCallableName, ?string $methodName = n
             $this->className  = $classOrCallableName;
             $this->methodName = !empty($methodName) ? $methodName : 'class';
         }
-
-        $this->deepClone    = $deepClone;
-        $this->shallowClone = $shallowClone;
```
staabm, line 151: > Class PHPUnit\Framework\ExecutionOrderDependency has an uninitialized readonly property $shallowClone. Assign it in the constructor.  
> Class PHPUnit\Framework\ExecutionOrderDependency has an uninitialized readonly property $deepClone. Assign it in the constructor.

src/TextUI/Help.php
```diff
@@ -47,7 +47,7 @@ public function __construct(?int $width = null, ?bool $withColor = null)
         foreach ($this->elements() as $options) {
             foreach ($options as $option) {
                 if (isset($option['arg'])) {
-                    $this->lengthOfLongestOptionName = max($this->lengthOfLongestOptionName, isset($option['arg']) ? strlen($option['arg']) : 0);
```
staabm, line 50: because of the outer `if (isset($option['arg']))` condition, this nested `isset()` can never be false

PR #5779, opened 2024-03-30
Test DefaultPrinter, ExecutionOrderDependency

staabm:
`DefaultPrinter` with a `socket://` url did not work before, because of the used readonly property

### DIFF
--- a/tests/unit/Framework/ExecutionOrderDependencyTest.php
+++ b/tests/unit/Framework/ExecutionOrderDependencyTest.php
@@ -118,5 +118,8 @@ class ExecutionOrderDependencyTest extends TestCase
         $empty = new ExecutionOrderDependency('');
         $this->assertFalse($empty->shallowClone());
         $this->assertFalse($empty->deepClone());
+        $this->assertFalse($empty->targetIsClass());
+        $this->assertSame('', $empty->getTargetClassName());
+
     }
 }
--- a/tests/unit/Framework/ExecutionOrderDependencyTest.php
+++ b/tests/unit/Framework/ExecutionOrderDependencyTest.php
@@ -112,4 +112,11 @@ class ExecutionOrderDependencyTest extends TestCase
             'Right side of merge could be empty',
         );
     }
+
+    public function testEmptyClassOrCallable(): void
+    {
+        $empty = new ExecutionOrderDependency('');
+        $this->assertFalse($empty->shallowClone());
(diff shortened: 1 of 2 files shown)


tests/unit/TextUI/Output/Default/DefaultPrinterTest.php
```diff
@@ -0,0 +1,44 @@
...
+#[CoversClass(DefaultPrinter::class)]
+#[Medium]
+final class DefaultPrinterTest extends TestCase
+{
+    public static function providePrinter(): array
+    {
+        return [
+            [DefaultPrinter::standardOutput()],
+            [DefaultPrinter::standardError()],
+            [DefaultPrinter::from('socket://hostname:port')],
```
staabm, line 28: this test failed before the fix in https://github.com/sebastianbergmann/phpunit/pull/5777#discussion_r1544726955

tests/unit/Framework/ExecutionOrderDependencyTest.php
```diff
@@ -112,4 +112,11 @@ public function testMergeHandlesEmptyDependencyLists(): void
             'Right side of merge could be empty',
         );
     }
+
+    public function testEmptyClassOrCallable(): void
+    {
+        $empty = new ExecutionOrderDependency('');
```
staabm, line 118: this test failed before the fix in https://github.com/sebastianbergmann/phpunit/pull/5777#discussion_r1544727534

tests/unit/TextUI/Output/Default/DefaultPrinterTest.php
staabm: to be more clear: I think the default printer would have created a fatal runtime error when used with a socket url before the recent readonly property fix.

I wonder whether its a feature used in the wild, as such a fatal runtime error would have been reported by someone - I guess..?

sebastianbergmann: I have never seen it used in the wild. I implemented it 10+ years ago and used it myself once or twice.

sebastianbergmann:
Just merged this, thanks.

However, I now get

```
$ ./phpunit
PHP Warning:  fsockopen(): Unable to connect to hostname:0 (Failed to parse address "hostname") in /usr/local/src/phpunit/src/TextUI/Output/Printer/DefaultPrinter.php on line 90
```

Can you have a look, please? Thanks!

PR #5781, opened 2024-03-30
Support static analysis in `PHPUnit\Framework\callback()`

staabm:


sebastianbergmann:
Please note that `src/Framework/Assert/Functions.php` is a generated file.

staabm:
ohh didn't realized it. where is the generator?

sebastianbergmann:
> ohh didn't realized it. where is the generator?

`build/scripts/generate-global-assert-wrappers.php` is invoked by `ant generate-global-assert-wrappers`.

PR #5782, opened 2024-03-30
More precise param types in `TestBuilder`

staabm:


src/Framework/TestBuilder.php
```diff
@@ -88,10 +88,9 @@ private function buildDataProviderTestSuite(string $methodName, string $classNam
         );
 
         foreach ($data as $_dataName => $_data) {
+            /** @psalm-suppress UnsafeInstantiation */
             $_test = new $className($methodName);
```
staabm, line 92: I wonder whether we should/need to declare `TestCase->construct()` as `final`?

sebastianbergmann: I wish we could, so far I was too "scared" to make this change.

staabm: We could try with a `@final` phpdoc for a softer migration

Issue #5775, opened 2024-03-29
Emit events for `E_*` and `E_USER_*` triggered in data provider methods


nikophil:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     |11.0.9
| PHP version         | 8.3.4
| Installation Method | Composer / PHAR

#### Summary

Hello,

I don't know if it is a known issue, I've looked for a related one and did not find it.

It seems that a deprecation triggered in a data provider will not be reported by phpunit.

#### Current behavior

If we call `@trigger_error('some deprecation', \E_USER_DEPRECATED)` in a data provider, no deprecation warning is displayed. Even with `ignoreSuppressionOfDeprecations="true"`

#### How to reproduce

call `trigger_error('some deprecation', \E_USER_DEPRECATED)` in a data provider

#### Expected behavior

a `D` should be displayed instead of a `.`

sebastianbergmann:
I do not consider this a bug, because the current behavior is intentional. I must think about whether I want to change this behavior.

sebastianbergmann:
I started working on this in #5784.

sebastianbergmann:
PHPUnit now captures `E_NOTICE`, `E_USER_NOTICE`, `E_WARNING`, and `E_USER_WARNING` triggered in data provider methods and reports them as events associated with the test methods that use the respective data provider.

Previously, only deprecations (`E_DEPRECATED`, `E_USER_DEPRECATED`) triggered in data providers were captured and reported. Notices and warnings were not handled, making it harder to identify problems in data provider code.

With this change, a data provider that triggers a warning or notice will cause the corresponding test to be marked accordingly (`N` for notice, `W` for warning) in the progress output, just as it would if the same error were triggered during the test method itself.

The previous attempt at solving this (#5784, March 2024) required introducing an entirely new set of data provider-specific event classes, subscriber interfaces, and emitter methods. That approach was necessary at the time because there was no mechanism to associate errors from data providers with the test methods that consumed them.

In August 2025, 83f64496c ("Relate deprecations from DataProvider with corresponding tests") added exactly that mechanism: a collect-and-replay approach that captures errors during data provider execution and replays them through the existing error handler during test execution, where the test is on the call stack and the existing `Test\Issue` events can be reused. This infrastructure was built for deprecations only. Extending it to also cover notices and warnings was now a one-line change to the error handler bitmask.

sebastianbergmann:
[...]

PR #5783, opened 2024-03-31
Fix "fsockopen(): Unable to connect to hostname:0" warning

staabm:
another try at fixing https://github.com/sebastianbergmann/phpunit/pull/5779#issuecomment-2027951918

after looking a bit into it, it seems that phpunit will only fail a test when a warning is triggered from the test itself, but not the dataprovider.. intentional?


sebastianbergmann:
> it seems that phpunit will only fail a test when a warning is triggered from the test itself

This is similar to #5775.

PR #5790, opened 2024-04-02
gitignore .github folder

clxmstaab:
having these files in the releases is unnecessary bloat.

(sneeeking in a commit with my coporate account, so you hopefully don't need to re-affirm my CI runs)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,9 @@
 # Composer
 /vendor
 
+# GitHub.com
+/.github
+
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it. `.gitignore` has nothing to do with which files are inside a release archive. That is handled by `.gitattributes`.

PR #5791, opened 2024-04-02
Type arrays in TestSuite

clxmstaab:
(sneeeking in a commit with my coporate account, so you hopefully don't need to re-affirm my CI runs)

### DIFF
--- a/src/Framework/TestSuite.php
+++ b/src/Framework/TestSuite.php
@@ -63,7 +63,8 @@ class TestSuite implements IteratorAggregate, Reorderable, SelfDescribing, Test
     /**
      * @psalm-var array<string,list<Test>>
      */
-    private array $groups         = [];
+    private array $groups = [];
+
     /**
      * @psalm-var ?list<ExecutionOrderDependency>
      */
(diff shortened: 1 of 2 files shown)


src/Framework/TestSuite.php
```diff
@@ -64,12 +64,18 @@ class TestSuite implements IteratorAggregate, Reorderable, SelfDescribing, Test
      * @psalm-var array<string,list<Test>>
      */
     private array $groups         = [];
+    /**
+     * @psalm-var list<ExecutionOrderDependency>|null
```
sebastianbergmann, line 68: I would prefer `?list<ExecutionOrderDependency>`.

src/Framework/TestSuite.php
```diff
@@ -64,12 +64,18 @@ class TestSuite implements IteratorAggregate, Reorderable, SelfDescribing, Test
...
+     */
     private ?array $requiredTests = null;
 
     /**
      * @psalm-var list<Test>
      */
     private array $tests             = [];
+    /**
+     * @psalm-var list<ExecutionOrderDependency>|null
```
sebastianbergmann, line 77: I would prefer `?list<ExecutionOrderDependency>`.

src/Framework/TestSuite.php
```diff
@@ -63,13 +63,21 @@ class TestSuite implements IteratorAggregate, Reorderable, SelfDescribing, Test
     /**
      * @psalm-var array<string,list<Test>>
      */
-    private array $groups         = [];
+    private array $groups = [];
```
clxmstaab, line 66: the indentation change was done by php-cs-fixer. not sure this is intended

PR #5797, opened 2024-04-05
Prevent creation of `socket:` folder when running infection

staabm:
Running the DefaultPrinter test with mutations lead to files/directories beeing created in unexpected locations. 

Since the test will have sideeffects when the DefaultPrinter is mutated, we should not run in with infection


staabm:
Alternatively we could move the test out of the "unit" group so it won't run in the infection job

Issue #5795, opened 2024-04-04
Using `@testWith` annotation may generate `PHP Warning:  Uninitialized string offset 0`


fredericgboutin-yapla:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 10.5.16 (also in 9.x)
| PHP version         | 8.2.14
| Installation Method | Composer

#### Summary

Using `@testWith` might generate `PHP Warning:  Uninitialized string offset: 0`

#### Current behavior

You launch phpunit and before you see the header you see 2 Warnings

#### How to reproduce

This is quite easy. You have to use `@testwith` in a way that it is followed with a blank line. For example,

```php
/**
 * @testWith		 [1]
 * 			 [2]
 * 			 [3]
 * 
 * @testdox This test should make phpunit spit a PHP Warning !
 */
public function testExample($arg): void
{
	$this->assertIsInt($arg);
}	
```

This is probably related to the `explode("\n", ...)` that includes empty lines in between annotations
 https://github.com/sebastianbergmann/phpunit/blob/cd49a71c9a0a2f3c3ee48f097c4d4130eb0c301a/src/Metadata/Api/DataProvider.php#L252

#### Expected behavior

No PHP warning.

#### Workaround

You can move the `@testWith` statement at the end of the comment section.

```php
/**
 * @testdox This test should NOT make phpunit spit a PHP Warning anymore...
 * 
 * @testWith		 [1]
 * 			 [2]
 * 			 [3]
 */
public function testExample($arg): void
{
	$this->assertIsInt($arg);
}
```

staabm:
will look into it

sebastianbergmann:
Please note that bugs in PHPUnit 9 are no longer fixed. And while I agree that we should fix this for newer version, please also note that annotations are deprecated and that you should use attributes instead sooner rather than later.

staabm:
@fredericgboutin-yapla I tried reproducing the issue but on my machine the code sample does not emit warnings. see https://github.com/sebastianbergmann/phpunit/pull/5802.

are you sure he warnings are emitted by a plain phpunit setup and not because of some other thing? do you have a backtrace of the warning?

fredericgboutin-yapla:
Hum, that's surprising. 

I'm running PHP from the official Docker image using [php.ini-development](https://github.com/php/php-src/blob/master/php.ini-development) configuration.

But sure, here is the backtrace JSON encoded for readability

<details>

```json
[
    {
[...]

PR #5802, opened 2024-04-06
Using `@testWith` annotation may generate `PHP Warning: Uninitialized string offset 0`

staabm:
Closes https://github.com/sebastianbergmann/phpunit/issues/5795

### DIFF
--- a/tests/end-to-end/regression/5795/Issue5795Test.php
+++ b/tests/end-to-end/regression/5795/Issue5795Test.php
@@ -1,0 +1,27 @@
+<?php declare(strict_types=1);
+/*
+ * This file is part of PHPUnit.
+ *
+ * (c) Sebastian Bergmann <sebastian@phpunit.de>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+namespace PHPUnit\TestFixture\Issue5795;
+
+use PHPUnit\Framework\TestCase;
+
+final class Issue5795Test extends TestCase
+{
+    /**
+     * @testWith		 [1]
+     * 			 [2]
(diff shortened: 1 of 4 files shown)


tests/end-to-end/regression/5795/Issue5795Test.php
```diff
@@ -0,0 +1,27 @@
...
+
+use PHPUnit\Framework\TestCase;
+
+final class Issue5795Test extends TestCase
+{
+    /**
+     * @testWith		 [1]
+     * 			 [2]
+     * 			 [3]
+     *
```
staabm, line 20: for the test to fail this line needs to contain a space after the `*`.

my local phpstorm automatically strips this space and therefore I was not able to reproduce

staabm:
btw: I never seen this feature used in the wild. I am not sure when to use it, if one case use a regular DataProvider.

writing data for tests in a phpdoc comment, which will be parsed and mangled by a lot of other tools sounds like a footgun.

sebastianbergmann:
> I never seen this feature used in the wild

Me neither.

> I am not sure when to use it, if one can use a regular DataProvider instead

This was proposed in #1151 and implemented in #1728. The argument was that "[data providers are] overcomplicated and require to[o] much code for most tasks".

I should have said "No!" at the time, but did not :-(

> writing data for tests in a phpdoc comment, which will be parsed and mangled by a lot of other tools sounds like a footgun

At least this footgun will be gone once support for annotations in comments is removed.

staabm:
> I should have said "No!" at the time, but did not :-(

Without trying we wouldn't know. So it was a small failed experiment.. just look forward to :-)

PR #5803, opened 2024-04-07
Preserve whitespace during configuration migration

ghostwriter:
This patch preserves formatting, new lines between attributes in the `phpunit` tag, during configuration migration, via the following changes:

 - Add `src/Util/Xml/Document.php` extends `DOMDocument` and overrides both `loadXML()` and `saveXML()` 
 - Update `src/Util/Xml/Loader.php`

### DIFF
--- a/tests/end-to-end/migration/migration-from-95.phpt
+++ b/tests/end-to-end/migration/migration-from-95.phpt
@@ -5,12 +5,19 @@ Configuration migration from PHPUnit 9.5 format works
 $_SERVER['argv'][] = '--do-not-cache-result';
 $_SERVER['argv'][] = '--migrate-configuration';
 
+$originalPHPUnitXMLFile = __DIR__ . '/_files/migration-from-95/phpunit-9.5.xml';
+$phpunitXMLFile = sys_get_temp_dir() . '/phpunit.xml';
+
 chdir(sys_get_temp_dir());
-copy(__DIR__ . '/_files/migration-from-95/phpunit-9.5.xml', 'phpunit.xml');
+copy($originalPHPUnitXMLFile, $phpunitXMLFile);
 
 require_once __DIR__ . '/../../bootstrap.php';
 
 (new PHPUnit\TextUI\Application)->run($_SERVER['argv']);
+
+assert(
+    file_get_contents($originalPHPUnitXMLFile) === file_get_contents($phpunitXMLFile)
+);
 --EXPECTF--
 PHPUnit %s by Sebastian Bergmann and contributors.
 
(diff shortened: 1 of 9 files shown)


staabm:
Could you cover it with a few tests?

ghostwriter:
> Could you cover it with a few tests?

No, because PHPUnit `exits` anytime a command is executed.

https://github.com/sebastianbergmann/phpunit/blob/51e342a0bc987e0ea8418105d0711f08ae116de3/src/TextUI/Application.php#L271-L280

staabm:
you could add a config file, let it process via configuration migration and assert the resulting file looks like your expectations?

ghostwriter:
The test I added should have failed but have not, because `exit` is called by PHPUnit.

eg. The `001.php` that is created when testing `001.phpt`, the testing command will `exit` via PHPUnit before reaching the code that asserts the resulting file looks like our expectations.

tests/end-to-end/migration/migration-from-95.phpt
```diff
@@ -5,12 +5,19 @@ Configuration migration from PHPUnit 9.5 format works
...
-copy(__DIR__ . '/_files/migration-from-95/phpunit-9.5.xml', 'phpunit.xml');
+copy($originalPHPUnitXMLFile, $phpunitXMLFile);
 
 require_once __DIR__ . '/../../bootstrap.php';
 
 (new PHPUnit\TextUI\Application)->run($_SERVER['argv']);
+
+assert(
+    file_get_contents($originalPHPUnitXMLFile) === file_get_contents($phpunitXMLFile)
+);
```
staabm, line 20: disclaimer: I am no official phpunit developer and sebastian might have a different idea how such a test should look like.

----

I think you could do something like

```php
register_shutdown_function(function() use ($originalPHPUnitXMLFile, $phpunitXMLFile) {
    assert(
        file_get_contents($originalPHPUnitXMLFile) === file_get_contents($phpunitXMLFile)
    );
});

(new PHPUnit\TextUI\Application)->run($_SERVER['argv']);
```

instead

sebastianbergmann:
I have to say that I am not really comfortable with such a huge change for "just" preserving whitespace during configuration file migration.

tests/end-to-end/migration/migration-from-95.phpt
ghostwriter: This works, thanks for your suggestions!

ghostwriter:
> I have to say that I am not really comfortable with such a huge change for "just" preserving whitespace during configuration file migration.

That is absolutely fine, we can close this.

----

Thank you both for your feedback.

I'll add a `DOMPrinter` to `ghostwriter/coding-standard` and sanitize/format migrated `phpunit.xml{.dist}` across all packages via command line.

----

image

PR #5804, opened 2024-04-07
Support doubling `readonly` classes

sebastianbergmann:
- [X] Refactor test double runtime to hold state in `readonly` properties
- [X] Allow `readonly` classes to be doubled
- [X] Investigate failing test for cloning of test doubles
- [x] Emit error when `clone` is used on the test double for a `readonly` class on PHP 8.2
- [x] Test doubling of `readonly` classes


sebastianbergmann:
I was able to implement support for doubling `readonly` classes.

However, this implementation breaks support for cloning test double objects on PHP 8.2 using `clone`, regardless of whether they were created based on an extendable class, an extendable `readonly` class, or an interface. With PHP >= 8.3, [cloning test double objects continues to work](https://wiki.php.net/rfc/readonly_amendments).

I cannot be sure, but I assume that cloning test double objects is not commonly done and that this breakage is acceptable. The current state of this PR reflects this.

Alternatively, we could delay support for doubling `readonly` classes until PHPUnit 12 which will require PHP 8.3. This way, PHPUnit 11 would not break existing behaviour for PHP 8.2. Of course, it would delay support for doubling `readonly` classes even more.

PR #5814, opened 2024-04-14
Added CoversClass annotations in unit tests

staabm:
created via [new rector rule](https://github.com/rectorphp/rector-phpunit/pull/319)

### DIFF
--- a/tests/unit/Framework/MockObject/MockObjectTest.php
+++ b/tests/unit/Framework/MockObject/MockObjectTest.php
@@ -10,6 +10,7 @@
 namespace PHPUnit\Framework\MockObject;
 
 use function call_user_func_array;
+use PHPUnit\Framework\Attributes\CoversClass;
 use PHPUnit\Framework\Attributes\DoesNotPerformAssertions;
 use PHPUnit\Framework\Attributes\Group;
 use PHPUnit\Framework\Attributes\IgnorePhpunitDeprecations;
@@ -27,6 +28,7 @@ use ReflectionProperty;
 #[Group('test-doubles/mock-object')]
 #[TestDox('Mock Object')]
 #[Medium]
+#[CoversClass(MockObject::class)]
 final class MockObjectTest extends TestDoubleTestCase
 {
     public function testExpectationThatMethodIsNeverCalledSucceedsWhenMethodIsNotCalled(): void
--- a/tests/unit/Framework/MockObject/MockObjectTest.php
+++ b/tests/unit/Framework/MockObject/MockObjectTest.php
@@ -10,7 +10,6 @@
 namespace PHPUnit\Framework\MockObject;
 
 use function call_user_func_array;
-use PHPUnit\Framework\Attributes\CoversClass;
 use PHPUnit\Framework\Attributes\DoesNotPerformAssertions;
 use PHPUnit\Framework\Attributes\Group;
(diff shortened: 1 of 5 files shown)


tests/unit/Framework/MockObject/MockObjectTest.php
```diff
@@ -27,6 +28,7 @@
 #[Group('test-doubles/mock-object')]
 #[TestDox('Mock Object')]
 #[Medium]
+#[CoversClass(MockObject::class)]
```
sebastianbergmann, line 31: `MockObject` is an interface and cannot/should not be targeted for coverage.

tests/unit/Framework/MockObject/StubTest.php
```diff
@@ -17,6 +18,7 @@
 #[Group('test-doubles/test-stub')]
 #[Medium]
 #[TestDox('Test Stub')]
+#[CoversClass(Stub::class)]
```
sebastianbergmann, line 21: `Stub` is an interface and cannot/should not be targeted for coverage.

tests/unit/Framework/MockObject/MockObjectTest.php
staabm: yeah, I also realized it.. I am in the process of fixing the rule in https://github.com/rectorphp/rector-phpunit/pull/325

staabm:
I wonder whether CoversClass is meant to be used with enums

sebastianbergmann:
> I wonder whether CoversClass is meant to be used with enums

Enumerations are just a special kind of (final) classes, IMO.

staabm:
Agree. Just want to make sure there is no oversight in the implementation

staabm:
should be good to go

tests/unit/Metadata/Parser/AttributeParserTestCase.php
```diff
@@ -53,6 +54,7 @@
 use PHPUnit\TestFixture\Metadata\Attribute\UsesTest;
 use PHPUnit\TestFixture\Metadata\Attribute\WithoutErrorHandlerTest;
 
+#[CoversClass(AttributeParser::class)]
```
sebastianbergmann, line 57: This has no effect as `AttributeParserTestCase` is an abstract class and only metadata on concrete test classes is used.

staabm: thanks for the feedback. is this something phpunit should error about?

sebastianbergmann: Good questions. Yes, it probably should. However, figuring out that there is something to warn about would require additional work at runtime for each test. I am not convinced that this is worth it.

staabm: agree it might be not worth it. fwiw we already do some heavy lifting using reflection in this case:

https://github.com/sebastianbergmann/phpunit/blob/f6c0e6d7c80e07b37886b05c6e738b5fd760368c/src/Metadata/Api/CodeCoverage.php#L315-L323

sebastianbergmann: That is code for classes targeted by `#[CoversClass]` and `#[UsesClass]`, not for test classes that use these attributes.

PR #5822, opened 2024-04-19
In callback there is no "file" in trace

DarkSide666:
In callback there is no "file" in trace.

For example, trace will look like
```
    [0] => Array
        (
            [file] => C:\wamp\www\atk4\validate\vendor\vlucas\valitron\src\Valitron\Validator.php
            [line] => 760
            [function] => strtotime
        )
    [1] => Array
        (
            [function] => validateDate
            [class] => Valitron\Validator
            [type] => ->
        )
    [2] => Array
        (
            [file] => C:\wamp\www\atk4\validate\vendor\vlucas\valitron\src\Valitron\Validator.php
            [line] => 1251
            [function] => call_user_func
        )
   ...
```
and `[1]` element has only function, class and type attributes, but no file and line.

### DIFF
--- a/src/Runner/ErrorHandler.php
+++ b/src/Runner/ErrorHandler.php
@@ -254,19 +254,18 @@ final class ErrorHandler
 
         $trace = $this->filteredStackTrace($filterTrigger);
 
-        assert(isset($trace[0]['file']));
-        assert(isset($trace[1]['file']));
-
         $triggeredInFirstPartyCode       = false;
         $triggerCalledFromFirstPartyCode = false;
 
-        if ($trace[0]['file'] === $test->file() ||
-            $this->sourceFilter->includes($this->source, $trace[0]['file'])) {
+        if (isset($trace[0]['file']) && (
+            $trace[0]['file'] === $test->file() ||
+            $this->sourceFilter->includes($this->source, $trace[0]['file']))) {
             $triggeredInFirstPartyCode = true;
         }
 
-        if ($trace[1]['file'] === $test->file() ||
-            $this->sourceFilter->includes($this->source, $trace[1]['file'])) {
+        if (isset($trace[0]['file']) && (
+            $trace[1]['file'] === $test->file() ||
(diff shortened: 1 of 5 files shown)


sebastianbergmann:
Thank you for your pull request.

Please include a **minimal**, **self-contained**, reproducing test case that shows the problem you are aiming to solve.

Also note that pull requests for bug fixes should not target `main` but the oldest release branch that still receives bug fixes where the problem exists.

DarkSide666:
I've rebased this PR to `11.1` branch. In versions `10.x` this is not relevant.

Speaking about tests I'm really not sure how to write proper one for phpunit.
This simple test class shows that in case `debug_backtrace` is called from callback method, then it do not contain `file` and `line` array keys. So asserting that `file` and `line` keys are always present in backtrace in phpunit ErrorHandler class is wrong and will throw error in such  cases.

```
class MyTest extends TestCase
{
    public function testDebugBacktrace()
    {
        $result = $this->callUserFuncExample();

        // Check if the debug_backtrace has been generated
        $this->assertNotNull($result);

        // Check if the first element of the debug_backtrace is missing file and line elements
        $this->assertArrayHasKey('function', $result[0]);
        $this->assertArrayNotHasKey('file', $result[0]);
        $this->assertArrayNotHasKey('line', $result[0]);
    }

    private function callUserFuncExample()
    {
        return call_user_func([$this, 'exampleCallback']);
    }

    private function exampleCallback()
    {
        return debug_backtrace(DEBUG_BACKTRACE_PROVIDE_OBJECT, 1);
    }
}```

DarkSide666:
Added testcase which after fix will no more fail, but before fix fails.

mvorisek:
@sebastianbergmann this PR should be retarget instead of closed

sebastianbergmann:
@DarkSide666 I did not intend to close this. Please send a new PR against the `11.2` branch. Thanks!

sebastianbergmann:
Merged manually into `11.2`.

PR #5832, opened 2024-05-02
Psalm baseline improvements - src/Logging/JUnit/JunitXmlLogger.php

mgleska:
As part of #4093, updated a few files to make the baseline shorter.


staabm:
Looks like this PR is obsolete (phpunit no longer uses psalm)

sebastianbergmann:
Correct.

sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, as @staabm already pointed out, we no longer use Psalm.

PR #5852, opened 2024-06-02
Output buffer - fix 5851

### LINKED ISSUE #5851: Output buffer manipulation in tests causes incorrect capture, hangs, and silent failures
| Q                   | A
| PHPUnit version     | 11.1.3
| PHP version         | 8.2.x, 8.3.x
| Installation Method | Composer / PHAR

#### Summary

1) when a test (or any function called from the test) flushes or cleans the output buffer, this incorrectly changes the output making tests pass that fail with expectOutputString and/or would cause a PHP notice when the code is actually executed
2) if a test ends but starts a new output buffer, no error is reported, since only the ob level is compared instead of comparing the output
3) when a test (or any function called from the test) register a non-removable ob callback, this leads to complete failure/unexpected behavior
4) non-removable ob callbacks cannot currently be tested
5) if stopping output buffering fails, it will try again creating the same notices twice

I will provide a PR.

kkmuffme:
Fix https://github.com/sebastianbergmann/phpunit/issues/5851

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -32,10 +32,13 @@ use function is_object;
 use function is_string;
 use function libxml_clear_errors;
 use function method_exists;
+use function ob_clean;
 use function ob_end_clean;
-use function ob_get_clean;
+use function ob_end_flush;
+use function ob_flush;
 use function ob_get_contents;
 use function ob_get_level;
+use function ob_get_status;
 use function ob_start;
 use function preg_match;
 use function restore_error_handler;
@@ -173,6 +176,7 @@ abstract class TestCase extends Assert implements Reorderable, SelfDescribing, T
     private ?string $outputExpectedString    = null;
     private bool $outputBufferingActive      = false;
     private int $outputBufferingLevel;
+    private ?int $outputBufferingFlushed      = null;
     private bool $outputRetrievedForAssertion = false;
     private bool $doesNotPerformAssertions    = false;
 
@@ -544,11 +548,13 @@ abstract class TestCase extends Assert implements Reorderable, SelfDescribing, T
         $outputBufferingStopped = false;
 
         if (!isset($e) &&
-            $this->hasExpectationOnOutput() &&
-            $this->stopOutputBuffering()) {
+            $this->hasExpectationOnOutput()) {
+            // if it fails now, we shouldn't try again later either
             $outputBufferingStopped = true;
 
-            $this->performAssertionsOnOutput();
+            if ($this->stopOutputBuffering()) {
+                $this->performAssertionsOnOutput();
+            }
         }
 
         if ($this->status->isSuccess()) {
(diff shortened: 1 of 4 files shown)


tests/end-to-end/regression/5342.phpt
```diff
@@ -17,21 +17,15 @@ F                                                                   1 / 1 (100%)
...
 
---
-
-There was 1 risky test:
-
-1) PHPUnit\TestFixture\Issue5342Test::testFailure
-Test code or tested code closed output buffers other than its own
-
-%sIssue5342Test.php:%i
+2) PHPUnit\TestFixture\Issue5342Test::testFailure
```
sebastianbergmann, line 27: Is this escalation from risky test to test failure intentional?

src/Framework/TestCase.php
```diff
@@ -1813,43 +1819,235 @@ private function markSkippedForMissingDependency(ExecutionOrderDependency $depen
...
     }
 
+    /**
+     * @throws Exception
+     * @throws NoPreviousThrowableException
+     */
     private function stopOutputBuffering(): bool
     {
-        $bufferingLevel = ob_get_level();
+        $bufferingLevel            = ob_get_level();
```
mvorisek, line 1862: At first, it is a bugfix, so it must target the lowest supported and problematic branch.

Also, it seems this PR fixes several ob issues. I would like to see a separate commit for each issue ("Fix ob is not cleanable", ...) so each can be reviewed individually.

sebastianbergmann:
Let me start by thanking you for taking the time to investigate this issue!

@theseer, @localheinz, @staabm, @Schrank, @sebastianheuer, @Tesla91, and I discussed this issue during the PHPUnit Code Sprint in Munich today.

We are not convinced we need to distinguish all the different cases. Here is our attempt to solve the two main issues of "do not close PHPUnit's buffer" and "do not leave your buffer open":

```patch
diff --git a/src/Framework/TestCase.php b/src/Framework/TestCase.php
index 2d749ea24..cabbee413 100644
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -10,6 +10,7 @@
 namespace PHPUnit\Framework;
 
 use const PHP_EOL;
+use const PHP_OUTPUT_HANDLER_REMOVABLE;
 use function array_keys;
 use function array_merge;
 use function array_reverse;
@@ -36,6 +37,7 @@
 use function ob_get_clean;
 use function ob_get_contents;
 use function ob_get_level;
+use function ob_get_status;
 use function ob_start;
 use function preg_match;
 use function restore_error_handler;
@@ -1434,9 +1436,18 @@ private function markSkippedForMissingDependency(ExecutionOrderDependency $depen
         $this->status = TestStatus::skipped($message);
     }
 
+    private function outputBufferCallback(): string
+    {
+        if (!$this->outputBufferingActive) {
+            return '';
+        }
+
+        throw new Exception('Test code or tested code flushed or cleaned global output buffers other than its own');
+    }
+
     private function startOutputBuffering(): void
     {
-        ob_start();
+        ob_start([$this, 'outputBufferCallback']);
 
         $this->outputBufferingActive = true;
         $this->outputBufferingLevel  = ob_get_level();
@@ -1449,27 +1460,28 @@ private function stopOutputBuffering(): bool
         if ($bufferingLevel !== $this->outputBufferingLevel) {
             if ($bufferingLevel > $this->outputBufferingLevel) {
                 $message = 'Test code or tested code did not close its own output buffers';
-            } else {
-                $message = 'Test code or tested code closed output buffers other than its own';
+
+                Event\Facade::emitter()->testConsideredRisky(
+                    $this->valueObjectForEvents(),
+                    $message,
+                );
+
+                $this->status = TestStatus::risky($message);
             }
 
             while (ob_get_level() >= $this->outputBufferingLevel) {
+                if (!(ob_get_status()['flags'] & PHP_OUTPUT_HANDLER_REMOVABLE)) {
+                    break;
+                }
+
                 ob_end_clean();
             }
 
-            Event\Facade::emitter()->testConsideredRisky(
-                $this->valueObjectForEvents(),
-                $message,
-            );
-
-            $this->status = TestStatus::risky($message);
-
             return false;
         }
 
-        $this->output = ob_get_clean();
-
         $this->outputBufferingActive = false;
+        $this->output                = ob_get_clean();
         $this->outputBufferingLevel  = ob_get_level();
 
         return true;
```

What do you think?

src/Framework/TestCase.php
sebastianbergmann: @mvorisek While I appreciate the contributions you have made to this project over the years, I do not like how you sometimes interact in discussions.

To somebody like @kkmuffme, whose first pull request this is, it must sound like you are a maintainer for this project when you write statements such as "it must target the lowest supported and problematic branch" or "I would like to see a separate commit for each issue" and that they have to do what you tell them to do. I can only imagine how frustrating it must be for @kkmuffme to read your comments and then do not get answers to the questions they asked in response.

If you look at the [Supported Versions of PHPUnit](https://phpunit.de/supported-versions.html) page then you will see that as of today, PHPUnit 10.5 is the oldest version of PHPUnit that still receives bug fixes. This page is linked from https://github.com/sebastianbergmann/phpunit/blob/main/.github/CONTRIBUTING.md#workflow-for-pull-requests.

As I wrote in https://github.com/sebastianbergmann/phpunit/pull/5852#issuecomment-2422906999, we spent quite some time yesterday trying to understand the problems @kkmuffme intends to solve as well as their proposed changes to solve them. The way I understand the proposed changes, they do not make much sense in isolation from each other. Therefore, I do not require separate commits to review.

I hope that, as outlined in https://github.com/sebastianbergmann/phpunit/pull/5852#issuecomment-2422906999, a simpler solution than the one proposed by @kkmuffme is possible. However, if the problems in questions can only be solved using the changes proposed here then I will gladly merge this pull request. However, given the complexity of the proposed changes I am reluctant to accept this changes into the `10.5` branch for PHPUnit 10.5. I would rather include them in the next minor version, which would be PHPUnit 11.5.

mvorisek: @sebastianbergmann thank you for your feedback. I have written my post mainly because I worked on this and this PR was/is too complex. It simply felt/feels very long to me - in the https://github.com/sebastianbergmann/phpunit/issues/5851 issue there are multiple problems mentioned, so to help me, you, and all others to understand the changes, I asked/ask to commit each problem with minimal fix&test separately.

Only then I would be willing to spend my time to review each LoC and probably would propose changes to simplify the PR as you did yesterday. Kudos to you.

I hope we can agree on this. Not only because of me, but also because the core code needs to be readable as much as possible to make it easy to understand while debugging any other ob issue.

I am rereading your post again and this is basically what you have written in your last paragraph. That you are unsure about the changes proposed here.

> it must sound like you are a maintainer for this project when you write statements such as "it must target the lowest supported and problematic branch" ... and that they have to do what you tell them to do

Does it matter who expresses the truth about the contribution guidelines? Or is the main issue you had with post not mentioning/linking https://github.com/sebastianbergmann/phpunit/blob/main/.github/CONTRIBUTING.md#workflow-for-pull-requests?

theseer:
> 
> Also there is a bug in the changed code still:
> 
> > $this->output                = ob_get_clean();
> 
> This will throw a notice if your buffer was closed and replaced by a non-removable one, since the ob level is the same. 
> This issue is fixed in my PR though.

I admit I didn't verify this by actually runnig a test for it. But I cannot see how this could be the case? In case *our* buffer is closed, the callback should have been called - which throws an exception. It should make the potentially problematic line unreachable.  Which *should* cover your case 2 just fine.

But I'll create a testcase for that and see :)

I'll try to create separate cases for each of your other issues mentioned and see if our simplified oversimplified things ;)

> 
> > I discussed this issue during the PHPUnit Code Sprint in Munich today.
> > We are not convinced we need to distinguish all the different cases. Here is our attempt to solve the two main issues of "do not close PHPUnit's buffer" and "do not leave your buffer open"
> 
> If you could provide a very short summary of the discussion/why these cases don't need to be differentiated, it would greatly appreciate it. (if those are in German that's alright too)

SirLouen:
@theseer any updates about this? Did you have the opportunity to see the last @kkmuffme comment?

sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it. This should be fixed by ea51aa4da5af03400bbab6e2265b618aeb13eae6.

Issue #5857, opened 2024-06-08
Mocked method called in constructor and a additional method throws an unexpected exception


sbuerk:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 11.2.0
| PHP version         | 8.2, 8.3
| Installation Method | Composer + phpunit repo directly

#### Summary

A class method is called in the constructor and in other
class methods. Using `onlyMethods()` to mock that method
does not regonized the calling (count) until now, but at
least subsequential calls could be tested and ensured.

This worked until 11.2.0 of phpunit and was detected in
the TYPO3 nightly pipeline automatically pulling in new
dependency versions, like phpunit.

Note: In manual tests this has been reduced to phpunit
and other dependency ruled out. Can be reproduced within
the phpunit source code directly.

#### Current behavior

Since 11.2.0 this no longer works and a exception like following
is thrown:

```
Error: Typed property MockObject_ClassCallingMethodInConstructor_b5ada611::$__phpunit_state
must not be accessed before initialization
```

#### How to reproduce

This can be reproduced with a simple class:

```php
<?php declare(strict_types=1);

namespace Vendor\Package;

class ClassCallingMethodInConstructor
{
    public function __construct()
    {
        $this->reset();
    }

    public function reset(): void
    {
    }

    public function second(): void
    {
        $this->reset();
    }
}
```

and having a test like following:

```php
use Vendor\Package\ClassCallingMethodInConstructor;

public function testOnlyMethodCalledInConstructorWorks(): void
{
  $testClassMock = $this->getMockBuilder(ClassCallingMethodInConstructor::class)
     ->onlyMethods(['reset'])
     ->getMock();

  $testClassMock->expects($this::once())->method('reset');
  $testClassMock->second();
}
```

#### Expected behavior

Test passes without error (green), which worked until 11.2.0.

sebastianbergmann:
PHPUnit 11.2 has added support for extendable classes that are declared readonly. This required extensive changes to the runtime internals of the test double functionality. These changes are the reason why tests like the one you have provided as an example no longer work. I am grateful to you for reporting this, as I was not aware of this problem before.

Side note: maybe we can talk about this at the TYPO3 code sprint this week, whether the TYPO3 project can also test in its nightly CI pipeline with the current development version of PHPUnit. This would allow us to uncover such problems earlier.

[...]

PR #5858, opened 2024-06-08
Provide a test to demonstrate a regresision

sbuerk:
This change adds a fixture class and a MockBuilder test
to showcase a regression introduced with version 11.2.0.

Given is a class, where one method is mocked using the
`onlyMethods()` option to check the method gets called.
The method is called in the class constructor and also
in other public method.

Providing a test mocking that method this breaks since
11.2.0 - albeit having the count in previous version
only working for calls to the mock method AFTER the
constructor call.

Following exception thrown is not expectd:

    Error: Typed property
    MockObject_ClassCallingMethodInConstructor_b5ada611::$__phpunit_state
    must not be accessed before initialization

Related: #5857


sebastianbergmann:
Cherry-picked into `11.2` and merged to `main` from there.

PR #5862, opened 2024-06-12
Do not cache test double declarations

sebastianbergmann:
While profiling the test double code generator today, I noticed that (counter-intuitively) memory consumption increases when test double code declarations are not cached. For instance, for PHPUnit's own test suite from 48 MB to 52 MB.

This is PR is not intended to be merged, but I would like to have a place where the investigation of this issue can be discussed.

### DIFF
--- a/src/Framework/MockObject/Generator/Generator.php
+++ b/src/Framework/MockObject/Generator/Generator.php
@@ -93,11 +93,6 @@ final class Generator
     ];
 
     /**
-     * @psalm-var array<non-empty-string, MockClass>
-     */
-    private static array $cache = [];
-
-    /**
      * Returns a test double for the specified class.
      *
      * @throws ClassIsEnumerationException
@@ -415,21 +410,17 @@ final class Generator
             serialize($callOriginalMethods),
         );
 
-        if (!isset(self::$cache[$key])) {
-            self::$cache[$key] = $this->generateCodeForTestDoubleClass(
-                $type,
-                $mockObject,
-                $markAsMockObject,
-                $methods,
-                $mockClassName,
-                $callOriginalClone,
-                $callAutoload,
-                $cloneArguments,
-                $callOriginalMethods,
-            );
-        }
-
-        return self::$cache[$key];
+        return $this->generateCodeForTestDoubleClass(
+            $type,
+            $mockObject,
+            $markAsMockObject,
+            $methods,
+            $mockClassName,
+            $callOriginalClone,
+            $callAutoload,
+            $cloneArguments,
+            $callOriginalMethods,
+        );
     }
 
     /**


sebastianbergmann:
This might be of interest to @staabm and @lolli42.

src/Framework/MockObject/Generator/Generator.php
```diff
@@ -415,21 +410,17 @@ public function generate(string $type, bool $mockObject, bool $markAsMockObject,
...
-                $mockClassName,
-                $callOriginalClone,
-                $callAutoload,
-                $cloneArguments,
-                $callOriginalMethods,
-            );
-        }
-
-        return self::$cache[$key];
+        return $this->generateCodeForTestDoubleClass(
```
staabm, line 413: > I noticed that (counter-intuitively) memory consumption increases when test double code declarations are not cached.

sounds like this method is somehow triggering logic which when invoked repeatedly builds up a buffer or uses a reasource which leaks memory.

staabm: when running blackfire on 778070d58, I don't see `generateCodeForTestDoubleClass` at all in the profile -> its using very little memory and therefore blackfire decides its not worth having information about this method in the profile.

when running blackfire on this PR, we get a bit more information. `generateCodeForTestDoubleClass` is still using very little memory, but at least it now shows up in the profile.

in this screen you can see how the memory used by `generateCodeForTestDoubleClass` spreads over calls triggered from this method:

grafik

it looks like the lions share of memory is consumed by autoloading and reflection, which makes sense to me.
both are apis which have side-effects.

if there is a way to reduce/remove calls to these 2 php-src apis you might be able to remove the caching.

----

I ran blackfire on the unit test-suite as a whole. in case we could find a command in which the test-double generator itself would take a bigger share of the overall memory we could get more precise numbers of the diff between main and this PR.

src/Framework/MockObject/Generator/Generator.php
```diff
@@ -92,11 +92,6 @@ final class Generator
         '__halt_compiler' => true,
     ];
 
-    /**
-     * @psalm-var array<non-empty-string, MockClass>
-     */
-    private static array $cache = [];
-
     /**
```
staabm, line 95: as I have the numbers at hand atm, maybe this is also interessting (even though not related to this PRs topic):

running the test-suite as a whole, these calls are the most memory consuming ones:

grafik

and here you can find the slowest calls:

grafik

---

recorded with

```
$ php -v
PHP 8.2.12 (cli) (built: Oct 24 2023 21:15:35) (NTS Visual C++ 2019 x64)
Copyright (c) The PHP Group
Zend Engine v4.2.12, Copyright (c) Zend Technologies
    with blackfire v1.92.16~win-x64-non_zts82, https://blackfire.io, by Blackfire
```

PR #5869, opened 2024-06-18
Add option to limit export of arrays

sebastianbergmann:


src/Util/Exporter.php
```diff
@@ -0,0 +1,49 @@
...
+final class Exporter
+{
+    private static ?OriginalExporter $exporter = null;
+
+    public static function export(mixed $value): string
+    {
+        return self::exporter()->export($value);
+    }
+
+    public static function shortenedRecursiveExport(array &$data): string
```
mvorisek, line 27:
```suggestion
    public static function shortenedRecursiveExport(array $data): string
```

sebastianbergmann: I would rather not have a different signature in PHPUnit's wrapper for the method from `Exporter`.

mvorisek: AFAIK in exporter the `&` is needed to prevent recursion, here it is not needed, thus IMO, `&` is useless and only confusing here.

PR #5870, opened 2024-06-18
Do not collect unnecessary information using `debug_backtrace()`

staabm:
see https://www.php.net/debug_backtrace

grafik


sebastianbergmann:
It should be safe for this to target `10.5`.

staabm:
Agree. I am afk for 2 hours

sebastianbergmann:
Cherry-picked into `11.2` and merged to `main` from there.

PR #5875, opened 2024-06-20
Also destruct `TestCase` objects early that use a data provider

talkinnl:
Refactor unsetting tests during TestSuite::run, faster destruction.

- Continuing on the destruction fix: The iterator also holds the array of tests. For tests with dataProviders, this results in not immediately destructing each Test, but only when the deeper TestSuite is done.
- Refactored cleanup tricks: Just do a first loop so we no longer need properties nor the iterator, and then array_shift as we go.
- This also reduces amount of code executed: One loop up front instead of the unset loop per test.
- Added test coverage.

PR #5877, opened 2024-06-20
Use `array_pop()` instead of `array_shift()` for processing `Test` objects in `TestSuite::run()` and optimize `TestSuite::isEmpty()`

mvorisek:
improve #5875 performance, `array_shift` is very slow with large arrays


mvorisek:
https://3v4l.org/4vGK2/rfc#vgit.master_jit - with even more items the performance difference is even much higher - one has O(N) complexity as the array is always reallocated, the other one has O(1)

sebastianbergmann:
Thank you!

PR #5887, opened 2024-07-03
Ignore suppression when generating baseline

soyuka:
If an event is suppressed, while generating the baseline, we only checked for `ignoreSuppressionOfPhpWarnings` although we should check every option.


soyuka:
some of these lines are hard to cover, lmk if I need to do more

nikophil:
Hi!  :100: for this PR

I tried it and works perfectly.

Before the PR: `vendor/bin/phpunit --generate-baseline tests/baseline.xml` creates an empty file, whereas some deprecations are actually triggered. Those deprecations are created with the function `trigger_deprecation()` in `symfony/deprecation-contracts`, which uses `@trigger_error()`.

After the PR, the baseline is fully created.

sebastianbergmann:
Please target the `10.5` branch. Thanks!

PR #5888, opened 2024-07-03
Configure global exception/error handlers checks

soyuka:
This makes https://github.com/sebastianbergmann/phpunit/pull/5619 configurable and potentially mitigates https://github.com/symfony/symfony/issues/53812#issuecomment-2204410941. 
On one hand I agree that not releasing error_handlers is probably a mistake, on the other it's cpu intensive to add/remove handlers and is something that you should be able to disable.

Let me know your thought, I'm not sure how this is tested on the phpunit code base and I didn't wanted to do too much in case this is a bad idea, thanks!

### DIFF
--- a/src/Framework/TestBuilder.php
+++ b/src/Framework/TestBuilder.php
@@ -138,6 +138,9 @@ final readonly class TestBuilder
         }
 
         $test->setBackupStaticPropertiesExcludeList($backupSettings['backupStaticPropertiesExcludeList']);
+
+        $test->setShouldBackupGlobalErrorHandlers(ConfigurationRegistry::get()->backupGlobalErrorHandlers());
+        $test->setShouldBackupGlobalExceptionHandlers(ConfigurationRegistry::get()->backupGlobalExceptionHandlers());
     }
 
     /**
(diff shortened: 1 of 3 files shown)


soyuka:
Do you think I should fix the tests and rebase this @sebastianbergmann or we should close?

staabm:
@soyuka could you describe what your motivation is for this change? what is the actual root issue?

https://github.com/symfony/symfony/issues/53812#issuecomment-2204410941 does not describe the problem but only possible solutions

> On one hand I agree that not releasing error_handlers is probably a mistake, on the other it's cpu intensive to add/remove handlers and is something that you should be able to disable.



this sounds like you are trying to fix a performance problem. if so, do you have numbers to proof that?

not restoring global state after a test might lead to random looking side-effects in followup tests

soyuka:
I thought it'd be a good idea to be able to configure this feature but if no one needs this it's probably useless. 

Feel free to close this.

staabm:
I am fine with making things configurable in case someone has a use-case for it.
if we don't have a concrete use-case I would close it, as it makes PHPUnit more complicated for no gain.

just my personal opinion though

sebastianbergmann:

PR #5890, opened 2024-07-03
Priority for hook methods

### LINKED ISSUE #5889: Add a priority for hooks
Hello,

would you accept a PR which would add a concept of "priority" between hooks like `Before`, `After`...?

This would allow to control the order of execution of these hooks, mainly useful when used along with traits:
```php
trait SomeTraitWithHook
{
	#[Before]
	protected function someBeforeHook(): void
    {
		// do something
    }
}

trait SomeOtherTraitWithHook
{
	#[Before(priority: 5)]
	protected function someOtherBeforeHook(): void
    {
		// will be executed before `SomeTraitWithHook::someBeforeHook`
    }
}
```

`priority` should be any integer (even negative ones), and it would default to `0`.

thanks for your answer on this.

nikophil:
fixes #5889

I've introduced two new classes:
- `PHPUnit\Metadata\Api\HookMethod`: which is a value object with method's name and its priority
- `PHPUnit\Metadata\Api\HookMethodsCollection` which holds a collection of `HookMethod`s, and will return methods as string with the right priority when iterating on it

`HookMethods::hookMethods()` will now return an array of `HookCollections`, and the handling of the priority will be delegated to this class.

BTW, I'm not really sure that `PHPUnit\Metadata\Api\` is the right namespace for those classes.

### DIFF
--- a/src/Metadata/Api/HookMethod.php
+++ b/src/Metadata/Api/HookMethod.php
@@ -1,0 +1,24 @@
+<?php
+
+declare(strict_types=1);
+/*
+ * This file is part of PHPUnit.
+ *
+ * (c) Sebastian Bergmann <sebastian@phpunit.de>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+namespace PHPUnit\Metadata\Api;
+
+/**
+ * @internal This class is not covered by the backward compatibility promise for PHPUnit
+ */
+final readonly class HookMethod
+{
+    public function __construct(
+        public string $methodName,
+        public int $priority = 0,
+    ) {
+    }
+}
--- a/src/Metadata/Api/HookMethods.php
+++ b/src/Metadata/Api/HookMethods.php
@@ -9,11 +9,16 @@
  */
 namespace PHPUnit\Metadata\Api;
 
-use function array_unshift;
 use function assert;
 use function class_exists;
 use PHPUnit\Framework\TestCase;
+use PHPUnit\Metadata\After;
+use PHPUnit\Metadata\AfterClass;
+use PHPUnit\Metadata\Before;
+use PHPUnit\Metadata\BeforeClass;
 use PHPUnit\Metadata\Parser\Registry;
+use PHPUnit\Metadata\PostCondition;
+use PHPUnit\Metadata\PreCondition;
 use PHPUnit\Util\Reflection;
 use ReflectionClass;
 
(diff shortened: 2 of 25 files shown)


src/Metadata/Api/HookMethod.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+declare(strict_types=1);
+
+namespace PHPUnit\Metadata\Api;
```
nikophil, line 5: not really sure this is the right namespace for this class :shrug:

src/Metadata/Api/HookMethods.php
```diff
@@ -53,55 +58,76 @@ public function hookMethods(string $className): array
 
             if ($method->isStatic()) {
                 if ($metadata->isBeforeClass()->isNotEmpty()) {
-                    array_unshift(
-                        self::$hookMethods[$className]['beforeClass'],
-                        $methodName,
+                    $beforeClass = $metadata->isBeforeClass()->asArray()[0];
+                    assert($beforeClass instanceof BeforeClass);
+
+                    self::$hookMethods[$className]['beforeClass']->prepend(
```
nikophil, line 64: because some of the hooks were added with `array_unshift`  and other ones were added at the end of the array, I introduced `prepend()` and `add()` methods, in order to keep the initial order when all hooks have default priority

src/Metadata/Parser/AttributeParser.php
nikophil, line 1: I only modified the attribute parser, since annotations are deprecated

src/Metadata/Api/HookMethodsCollection.php
```diff
@@ -0,0 +1,95 @@
...
+    public function add(HookMethod $hookMethod): static
+    {
+        if ($this->shouldPrepend) {
+            $this->hookMethods = [$hookMethod, ...$this->hookMethods];
+        } else {
+            $this->hookMethods[] = $hookMethod;
+        }
+
+        return $this;
+    }
```
nikophil, line 78: I made this to mimic the different behavior where certain hook were added with `array_unshift`, and other with array concatenation, in order to keep BC when methods are added without priority

sebastianbergmann: Thank you. I cannot tell at the moment when I will have the time to review this. I would like to note the following, though: the status quo is, AFAIR, that there is an implicit priority when multiple hook methods are configured using attributes: "before" hook methods from a parent class before "before" hook methods from a child class, "after" hook methods from a parent class after "after" hook methods from a child class, etc. There is also an implicit order within a class based on the line number of the method declaration (in other words: the order in which the Reflection API finds the methods). This implicit ordering must be preserved.

nikophil: This was more or less what I what trying to explain 😅

nikophil: Hi @sebastianbergmann 

I've added the test [`tests/end-to-end/metadata/hook-methods-order.phpt`](https://github.com/sebastianbergmann/phpunit/pull/5890/files#diff-d87d5bdbabdcb5f9b18e7b4ad9cb06638a6e39057d540dd97369ace116d98565) in order to ensure that this "natural order" is still respected

sebastianbergmann:
Thank you!

Issue #5883, opened 2024-06-30
Testdox no longer outputs in real-time


Bilge:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 10.5.24
| PHP version         | 8.2.15
| Installation Method | Composer

#### Summary

The sole reason I enabled testdox in PHPUnit 9 was for the immediate result output. For long-running test suites, displaying dots is not the level of immediate feedback we want from a test runner. Moreover, some CI runners (including GHA) do not emit lines until the newline character is reached, and some other CI services (like classic Travis CI) will fail a build if it doesn't receive a (complete) log line within a certain timeout. Testdox solved all these problems in PHPUnit 9 with immediate test result output.

#### Current behavior

In PHPUnit 10, testdox has no bearing on the test runner, other than printing a report at the end.

#### How to reproduce

`--testdox` or `testdox="true"`.

#### Expected behavior

Print results in real-time.

sebastianbergmann:
This works as intended.

The purpose of the TestDox output is documentation, not "real-time output" and/or being a replacement for the regular output.

Bilge:
That's awful, but it's your software. Normally, when I "upgrade", I expect my suite of options to expand, not decrease. We have now lost the only out-of-the-box way to have real-time test output.

warpr:
Is there any other way to see the names of the tests being run in real-time?  The dots don't provide any useful info while tests are running.

kAlvaro:
@warpr As far as I know there is only `--debug`, though it is a little bit more cluttered and certainly not as visual.

@sebastianbergmann May I humbly suggest you reconsider this as quality of life improvement, as long as it doesn't create too much maintenance burden? I find real-time texdox output useful in two contexts:

- While writing or fixing tests locally (or working on a feature and using tests to verify it works), when you typically need to run them many times in a row, it gives a useful progress highlight. Quite often, you don't need to run the whole set if you spot a common pattern in failures (&laquo;Oops, I made a typo in data provider argument name&raquo;).
- When running lengthy CI pipelines (e.g. GitLab), it allows to determine whether a given test (related to the feature you've just worked on) passes without having to wait for the whole job to finish. That allows to abort the job and get back to work (if fail) or grunt with satisfaction and switch to another task (if pass). In theory, tests that pass locally will also pass in CI. In practice, percentage is not 100% because infrastructure is complicated.

[...]

PR #5895, opened 2024-07-11
Test Anything Protocol (TAP)

sebastianbergmann:
> [TAP, the Test Anything Protocol](https://testanything.org/tap-version-14-specification.html), is a simple text-based interface between testing modules in a test harness. It decouples the reporting of errors from the presentation of the reports.

In other words: TAP is a plain text log file format for test results that is agnostic to the test framework or test runner used.

Earlier versions of PHPUnit had support for TAP, but this support was removed some time ago.

[I think that "output in CI pipelines" is a separate use case and as such should be addressed using a new/separate output printer](https://github.com/sebastianbergmann/phpunit/issues/5883#issuecomment-2219890403) and I think that the TAP format might be suitable for that purpose.

So far, this PR implements

- an XML configuration option for logging test results in TAP format to a file
- a `--tap` CLI option for replacing the default test runner output with output in TAP format
- a `--log-tap` CLI option for logging test results in TAP format to a file
- a minimal implementation of a TAP logger (which does not handle all possible outcomes and issues a test may have)

Here is an example of what the TAP output currently looks like:

https://github.com/sebastianbergmann/phpunit/blob/4cf45887d298eb07b2ee44a81aaa1819009a3cb4/tests/end-to-end/logging/tap.phpt#L14-L109

@TimWolla @edorian @Bilge What do you think?

I will not pursue this further if the consensus is that TAP does not solve the "output in CI pipelines" use case.

### DIFF
--- a/tests/end-to-end/logging/tap.phpt
+++ b/tests/end-to-end/logging/tap.phpt
@@ -14,26 +14,37 @@ require_once __DIR__ . '/../../bootstrap.php';
 TAP version 14
 
 # successfully prepared PHPUnit\TestFixture\OutcomesAndIssuesTest::testSuccessWithoutIssues
+ok 1 - PHPUnit\TestFixture\OutcomesAndIssuesTest::testSuccessWithoutIssues
 
 # successfully prepared PHPUnit\TestFixture\OutcomesAndIssuesTest::testSuccessWithRisky
+ok 2 - PHPUnit\TestFixture\OutcomesAndIssuesTest::testSuccessWithRisky
 
 # successfully prepared PHPUnit\TestFixture\OutcomesAndIssuesTest::testSuccessWithDeprecation
+ok 3 - PHPUnit\TestFixture\OutcomesAndIssuesTest::testSuccessWithDeprecation
 
 # successfully prepared PHPUnit\TestFixture\OutcomesAndIssuesTest::testSuccessWithNotice
+ok 4 - PHPUnit\TestFixture\OutcomesAndIssuesTest::testSuccessWithNotice
 
 # successfully prepared PHPUnit\TestFixture\OutcomesAndIssuesTest::testSuccessWithWarning
+ok 5 - PHPUnit\TestFixture\OutcomesAndIssuesTest::testSuccessWithWarning
 
 # successfully prepared PHPUnit\TestFixture\OutcomesAndIssuesTest::testFailWithDeprecation
+not ok 6 - PHPUnit\TestFixture\OutcomesAndIssuesTest::testFailWithDeprecation
 
 # successfully prepared PHPUnit\TestFixture\OutcomesAndIssuesTest::testFailWithNotice
+not ok 7 - PHPUnit\TestFixture\OutcomesAndIssuesTest::testFailWithNotice
 
 # successfully prepared PHPUnit\TestFixture\OutcomesAndIssuesTest::testFailWithWarning
+not ok 8 - PHPUnit\TestFixture\OutcomesAndIssuesTest::testFailWithWarning
 
 # successfully prepared PHPUnit\TestFixture\OutcomesAndIssuesTest::testErrorWithDeprecation
+not ok 9 - PHPUnit\TestFixture\OutcomesAndIssuesTest::testErrorWithDeprecation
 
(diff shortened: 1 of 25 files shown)


tests/end-to-end/logging/tap.phpt
```diff
@@ -0,0 +1,109 @@
...
+<?php declare(strict_types=1);
+$_SERVER['argv'][] = '--do-not-cache-result';
+$_SERVER['argv'][] = '--no-configuration';
+$_SERVER['argv'][] = '--tap';
+$_SERVER['argv'][] = __DIR__ . '/../_files/OutcomesAndIssuesTest.php';
+
+require_once __DIR__ . '/../../bootstrap.php';
+
+(new PHPUnit\TextUI\Application)->run($_SERVER['argv']);
+--EXPECTF--
```
TimWolla, line 13: FWIW: PHPT tests themselves are intended to be valid PHP files, so I would recommend adding the closing PHP tag.

TimWolla (commented):
Thank you. I was primarily looking for an indicator of which specific test is currently running and the “successfully prepared …” comments nicely solve this.

One note that may or may not be related to this being a WIP: It appears that the “ok …” output is not working for PHPT tests, the “not ok” works correctly.

edorian:
I've only had second hand problems with this as it was not possible for a colleague to see which test was hanging in CI. They were using `--testdox`, which incidentally solved this problem, and when that went away, the PHPUnit 11 update needed a solution.

A **very** simple PHPUnit event listener [1] was good enough to solve this problem for us. For me, this particular problem is a bit of a homemade edge case anyway.

Over the years I've had a few cases where a "--give-me-each-test-name-before-starting-and-print-its-runtime-afterwards" option would have been handy, but that may be beside the point :)

In this particular case, the `--tap' output would have done the job as well. I also forgot about `--debug`, which would also be fine, but a bit more tedious to read.

So all I can say is that I'm happy with what PHPUnit gives me here and I don't have a strong need for TAP output, but it would work for our current "issue".

[1]

<details>
<summary>Print current test</summary>
<pre><code>
class PrintCurrentTest implements \PHPUnit\Runner\Extension\Extension
{
    public function bootstrap(
        \PHPUnit\TextUI\Configuration\Configuration $configuration,
        \PHPUnit\Runner\Extension\Facade $facade,
        \PHPUnit\Runner\Extension\ParameterCollection $parameters,
    ): void {
        $facade->registerSubscriber(new class implements PHPUnit\Event\Test\PreparedSubscriber {
            public function notify(\PHPUnit\Event\Test\Prepared $event): void
            {
                $name = explode('/', $event->test()->name()); // PHPT test, so the test name is a file path
                fwrite(STDOUT, "\nRunning: " . $name[count($name) - 2]);
            }
        });
    }
}
</code></pre>
</details>

Bilge:
Let's be clear, this isn't just for CI. Having immediate output is useful in any environment, including locally. To that end, I'm not really sure why you're married to the classic PHPUnit output; to me it feels outdated and should be replaced, but I'm not here to convince you of that.

As far as TAP is concerned, of course it solves the immediate output problem, though I've never been a fan of its style.

Counter-point: [Pip](https://github.com/ScriptFUSION/PHPUnit-Immediate-Printer). I just resurrected this printer from the PHPUnit 5/6 days since the new event framework makes it much easier to write and maintain, and of course, because this is the first version without an immediate printer built-in. I don't think its classic format I cooked up seven years ago looks much better than TAP, but I will now begin designing a more modern theme, probably modelled on [Collision](https://github.com/nunomaduro/collision). I would be happy to submit it for inclusion in PHPUnit, either directly (e.g. imported into this repo or as a Composer dependency) or indirectly (e.g. just a link in the docs), if you like.

tests/end-to-end/logging/tap.phpt
sebastianbergmann: I know that https://qa.php.net/write-test.php has

> The FILE section is used as the body of the .php file, so don't forget to open and close your php tags.

but I do not consider a PHP file without a closing PHP tag to be invalid and have never run into problems related to not using closing PHP tags in PHPT tests (that I am aware of).

Bilge:
@sebastianbergmann Since you closed this, what _is_ your plan to support real-time test output going forward?

sebastianbergmann:
I do not plan to support real-time output, at least not at the moment.

PR #5905, opened 2024-07-30
Add support for external logger

### LINKED ISSUE #5903: --printer replacement in PHPUnit 10.x/11.x
With the removal of --printer, using PHPUnit with Eclipse has been broken. This isn’t your problem but I’ve been considering a solution. It seems that currently there isn’t a way to provide PHPUnit with an external printer/logger/event handler. Can you confirm that is the case, or am I missing something?

From what I have been able to glean from the code so far it seems like, if one could create a class with a constructor that takes a Printer and a Façade, this would be the start of returning the --printer functionality. Then Application::registerLogfileWriters() could check if a configuration exists (e.g. --eventListener), confirm that it refers to a class that implements an interface that defines that constructor, and if so then create an instance of the class. Having a constructor in an interface isn’t ideal but it seems it would be necessary in this case as otherwise you’d have no idea how to construct an instance.

Does this make any sense, or am I heading off in the wrong direction?

interface EventLoggerIntf
{
    public static function createInstance(Printer $printer, Facade $facade);
}

BlairCooper:
---
name: ⚙ Add support for external logger
labels: type/enhancement
about: Added support for both XML and CLI definition of an external logger. This --log-external option adds similar functionality that existed with the --printer option in 9.x and earlier.
---
resolve #5903 

Implementations of an external logger do need to implement the \PHPUnit\Logger\ExternalLogger interface and must use the PHPUnit 10.x and later event mechanism for capturing logging events.

### DIFF
--- a/schema/11.3.xsd
+++ b/schema/11.3.xsd
@@ -2,7 +2,7 @@
 <xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema">
     <xs:annotation>
         <xs:documentation source="https://phpunit.de/documentation.html">
-            This Schema file defines the rules by which the XML configuration file of PHPUnit 11.3 may be structured.
+            This Schema file defines the rules by which the XML configuration file of PHPUnit 11.2 may be structured.
         </xs:documentation>
         <xs:appinfo source="https://phpunit.de/documentation.html"/>
     </xs:annotation>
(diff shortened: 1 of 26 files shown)


sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it. Let me explain.

A [test runner extension](https://docs.phpunit.de/en/11.2/extending-phpunit.html#implementing-an-extension) can use the `replacesProgressOutput()`, `replacesResultOutput()`, and `replacesOutput()` methods in its `bootstrap()` method to communicate to PHPUnit's test runner that it replaces the default output, in whole or in part. Just loading such an extension replaces PHPUnit's output and therefore achieves the same effect that the old `--printer` CLI option had.

So far, extensions can only be configured in PHPUnit's XML configuration file. There is no CLI option to load/configure an extension.

I think that a better solution in general and to your use case in particular would be to introduce a CLI option `--extension <className>` that allows registering a test runner extension on the CLI. `<className>` would be the name of a class that implements the `PHPUnit\Runner\Extension\Extension` interface. I woll work on this in #5906.

Issue #4600, opened 2021-02-08
Remove support for old cache configuration


sebastianbergmann:
* Remove the  `cacheResultFile` attribute of the `<phpunit>` root element (as well as the `--cache-result-file <file>` CLI option)
* Remove the `cacheDirectory` attribute of the `<coverage>` element (as well as the `--coverage-cache <dir>` CLI option)

awslabspl:
@sebastianbergmann why would you want to remove old config? Someone may want to use it.....

PR #5911, opened 2024-08-05
Remove the `--warm-coverage-cache` option cause by issue #4600

kayw-geek:
#### Problem Description

This PR addresses the issue where using the `--warm-code-coverage-cache` option results in the following error message:
```
Cache for static analysis has not been configured.
```

#### Background

In issue [#4600](https://github.com/sebastianbergmann/phpunit/issues/4600), the `--coverage-cache <dir>` option was removed. However, the current implementation still checks for the configuration of `--coverage-cache <dir>` when the `--warm-code-coverage-cache` option is used, leading to the error message if the cache is not configured.

#### Solution

This PR resolves the issue by removing the outdated check for `--coverage-cache <dir>` configuration in the `--warm-code-coverage-cache` option handling. This will prevent the error message from appearing and ensure normal operation for future users.

#### References

- Related issue: [#4600](https://github.com/sebastianbergmann/phpunit/issues/4600)

---


sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it. The functionality in question still exists and its removal is not intended.

The CLI option for configuring the cache directory has been removed. It must be configured using the XML configuration file.

PR #5912, opened 2024-08-05
Constraint: improve valueToTypeStringFragment for objects

janedbal:
This improves default message of e.g. `self::assertInstanceOf(LogicException::class, $e);`. Previously, it gave no info about which object was actually asserted:

`Failed asserting that an `**`object`**` is an instance of class LogicException.`

### DIFF
--- a/src/Framework/Constraint/Constraint.php
+++ b/src/Framework/Constraint/Constraint.php
@@ -247,7 +247,7 @@ abstract class Constraint implements Countable, SelfDescribing
         }
 
         return match ($type) {
-            'array', 'integer', 'object' => 'an ' . $type . ' ',
+            'array', 'integer', 'object' => 'an ' . get_debug_type($type) . ' ',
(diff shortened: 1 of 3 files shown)


sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it as I think the approach implemented in 5e8b79e52717cf330ce1023ec4b255541812ba70 is even better.

janedbal:
@sebastianbergmann I do believe that `get_debug_type` is better than `::class` for anonymous classes: https://3v4l.org/bhQr4

sebastianbergmann:
Did not think about that (because I never use anonymous classes in production code). 42cfc7d5982eac708fab2a566021959b31a49aa5 should take care of this.

janedbal:
Do you really think we should have null bytes in the message?

https://3v4l.org/kMDQU

PR #5914, opened 2024-08-06
Add `TARGET` to Covers attributes

milwad-dev:
I'm writing some tests for a project and I migrated from `@cover` annotate to Covers attributes.
So when I want to use the `CoversMethod` attribute for the test function, I get an error because I need to add a target to that attribute.

```php
#[CoverMethod($className, $methodName)]
public function testRecurring()
{
    $user = User::factory()->create();
    $account = Account::factory()
        ->for($user)
        ->create();
        
     ........
}
```

### DIFF
--- a/src/Framework/Attributes/CoversFunction.php
+++ b/src/Framework/Attributes/CoversFunction.php
@@ -16,7 +16,7 @@ use Attribute;
  *
  * @no-named-arguments Parameter names are not covered by the backward compatibility promise for PHPUnit
  */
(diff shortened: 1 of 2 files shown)


sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it. By design, attributes for targetted code coverage data can only be used on the test case class level and not on the test method level.

milwad-dev:
I've used `@covers \App\Console\Tasks\Clear\ClearOldAccounts::clearOldAccounts`, but now I want to use Attributes.
How can I do it?

Issue #5913, opened 2024-08-05
Call `Extension::bootstrap()` before creating the test suite


nikophil:
Hello,

I'm wondering if it would be possible to call the `bootstrap()` methods of an extension before the test suite is created?

This would allow third-party libraries relying on PHPUnit to call some boostraping call way sooner, even sooner than calling the data providers.

I've naively tested to move the whole block where `bootstrapExtensions()` is called, before the call of `buildTestSuite()` in `\PHPUnit\TextUI\Application` and it worked in my current project. But I'm may not be aware of some PHPUnit internals which would be a good reason for things to happen in this! :sweat_smile:

nikophil:
Bu the way, I ran the whole test suite with this change, and only one test did fail:

```
1) /home/nikophil/works/github.com/sebastianbergmann/phpunit/tests/end-to-end/extension-xml/phar-extension.phpt
Failed asserting that string matches format description.
--- Expected
+++ Actual
@@ @@
 PHPUnit Started (PHPUnit 11.3.0-6-g220aee5a5 using PHP 8.3.9 (cli) on Linux)
 Test Runner Configured
-Test Suite Loaded (1 test)
 Extension Loaded from PHAR (phpunit/phpunit-test-extension 1.0.0)
 Extension Bootstrapped (PHPUnit\TestFixture\MyExtension\MyExtensionBootstrap)
+Test Suite Loaded (1 test)
 Event Facade Sealed
 Test Runner Started
 Test Suite Sorted
```

sebastianbergmann:
Feel free to send a PR with these change for `main`. Thanks!

sebastianbergmann:
Out of curiosity: what kind of bootstrapping do you need to perform "sooner than calling the data providers"?

nikophil:
I'm a maintainer of [Foundry](https://symfony.com/bundles/ZenstruckFoundryBundle/current/index.html) which is a data fixture manager for Symfony.

It's not unusual to need those fixtures as provided data, such as:
```php
#[DataProvider('provide')]
public function testSomething(MyObject $object): void
{
    // ... 
}

public static function provide(): iterable
{
    yield MyObjectFactory::createOne();
    yield MyObjectFactory::createOne(['some' => 'attribute']);
}
```

[...]

PR #5917, opened 2024-08-06
Seal event facade before loading the test suite

nikophil:
follow up of https://github.com/sebastianbergmann/phpunit/issues/5913#issuecomment-2270835127


sebastianbergmann:
Thank you!

PR #5920, opened 2024-08-08
fix: baseline should support aboslute file paths

nikophil:
Hi!

since `phpunit --generate-baseline` generates absolute file paths, the reader should support them

### DIFF
--- a/src/Runner/Baseline/Reader.php
+++ b/src/Runner/Baseline/Reader.php
@@ -13,9 +13,11 @@ use const DIRECTORY_SEPARATOR;
 use function assert;
 use function dirname;
 use function file_exists;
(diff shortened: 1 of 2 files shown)


soyuka:
Looks weird to be the baseline should not be using absolute paths.

nikophil:
:facepalm: yes you're absolutely right! the problem is not that absolute paths are badly handled, the problem is in certain cases, absolute paths are generated, although they should be relative:
```
$ vendor/bin/phpunit --generate-baseline deprecations-baseline.xml # relative paths are generated
$ vendor/bin/phpunit --generate-baseline tests/deprecations-baseline.xml # absolute paths are generated
```

PR #5923, opened 2024-08-09
Filter configured deprecation triggers when displaying deprecation details

### LINKED ISSUE #5921: Display deprecation stacktrace
Hello,

sometimes finding where a deprecation comes from could be challenging. I think a mechanism in PHPUnit which displays the stack trace of deprecations would be helpful.

In Symfony, they've solved it by the following: if the `SYMFONY_DEPRECATIONS_HELPER` is a regex (I guess, they detect it if the var starts and ends with a `/`, or something like this), they break PHPUnit execution when the first deprecation resolving this regex is found, and the stack trace is printed.
see https://symfony.com/doc/current/components/phpunit_bridge.html#display-the-full-stack-trace

I'm not sure that breaking the execution is the best solution, but at least, it prevents for giant outputs if the test suite is full of deprecations.

I would be willing to contribute for this, but I'm not sure of how this should work. Do you see any other more elegant solution for this? Maybe we can add a new option which would display the stack traces, or at least write them in a file in order to not pollute the output?

thanks!

nikophil:
fixes #5921

### DIFF
--- a/src/Runner/ErrorHandler.php
+++ b/src/Runner/ErrorHandler.php
@@ -155,11 +155,13 @@ final class ErrorHandler
                 break;
 
             case E_USER_DEPRECATED:
+                $deprecationFrame = $this->guessDeprecationFrame();
+
                 Event\Facade::emitter()->testTriggeredDeprecation(
                     $test,
                     $errorString,
-                    $errorFile,
-                    $errorLine,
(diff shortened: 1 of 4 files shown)


sebastianbergmann:
Thank you!

PR #5927, opened 2024-08-19
`#[RequiresPhpunitExtension]` attribute

### LINKED ISSUE #5926: New attribute `#[RequiresPhpUnitExtension(SomeExtension::class)]`
When developing a library which provides a PhpUnit extension, it would be useful to be able to skip some tests which do require this extension.

WDYT of a new attribute which would skip tests if the given extension is not enabled?

Maybe it can be nice to allow this attribute to be repeatable, or to accept multiple parameters, for the rare cases where two extensions are required for the given test?

nikophil:
fixes #5926

### DIFF
--- a/src/Framework/Attributes/RequiresPhpunitExtension.php
+++ b/src/Framework/Attributes/RequiresPhpunitExtension.php
@@ -1,0 +1,43 @@
+<?php declare(strict_types=1);
+/*
+ * This file is part of PHPUnit.
+ *
+ * (c) Sebastian Bergmann <sebastian@phpunit.de>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+namespace PHPUnit\Framework\Attributes;
+
+use Attribute;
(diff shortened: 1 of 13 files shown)


src/Framework/Attributes/RequiresPhpunitExtension.php
```diff
@@ -0,0 +1,43 @@
...
+
+use Attribute;
+use PHPUnit\Runner\Extension\Extension;
+
+/**
+ * @immutable
+ *
+ * @no-named-arguments Parameter names are not covered by the backward compatibility promise for PHPUnit
+ */
+#[Attribute(Attribute::TARGET_CLASS | Attribute::TARGET_METHOD)]
```
sebastianbergmann, line 20: To be consistent with existing attributes, please change this attribute to be repeatable (and only hold the name of a single extension bootstrapper class instead of an array of class names).

nikophil: yeah sorry, I have not seen that other requirements were repeatable. fixed.

PR #5934, opened 2024-09-02
When generating phpunit.xml, use phpunit.xsd from vendor folder if it exists

PrinsFrank:
When working in a lot of projects, it becomes quite cumbersome to download the xsd for every project the phpunit.xml file is edited in.

It is already possible to replace the absolute url with a relative one while editing the file manually (currently the only option when developing without a stable internet connection). I think this should be the default for when phpunit is actually used together with composer.

In #5140 there was already an attempt to implement this, but that was closed by the author for lack of support for multiple xsds in the spec.

Instead, I think it's fair to check for the existence of the xsd from the vendor folder and to use it if it exists, and not set a default otherwise.

### DIFF
--- a/src/TextUI/Command/Commands/GenerateConfigurationCommand.php
+++ b/src/TextUI/Command/Commands/GenerateConfigurationCommand.php
@@ -62,12 +62,17 @@ final readonly class GenerateConfigurationCommand implements Command
             $cacheDirectory = '.phpunit.cache';
         }
 
+        $vendorSchemaLocation = implode(DIRECTORY_SEPARATOR, ['vendor', 'bin', 'phpunit', 'phpunit', 'phpunit.xsd']);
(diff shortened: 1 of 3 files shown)


PrinsFrank:
@sebastianbergmann I've fixed the code style so that pipeline should now pass. 

What should I do about the test coverage? Should I implement a cli test as in `php-configuration.phpt` and make sure that the the file exists during the test and delete it afterwards? This might make the tests not be able to run in parallel with the existing configuration test.

sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it as I think the solution I just implemented in 0385463a576716506c442b3d8889686f5abb7bb4 is cleaner/simpler.

PrinsFrank:
@sebastianbergmann That looks good indeed! Thanks for the review!!

PR #5944, opened 2024-09-08
Fix variadic function as callback

### LINKED ISSUE #5891: `Callback` constraint does not handle variadic arguments correctly when used for mock object expectations
| Q                   | A
| PHPUnit version     | 9.6.13
| PHP version         | 8.2

#### Summary

I need to assert the number of elements being passed as an argument on a mocked method for a mocked service.
The argument being passed is an spread operator, and I experience a weird behaviour when unpacking the array. 
It always unpack the array as 1 element even the argument passed is an array with multiple elements, it just returns one.

#### Current behavior

Unpacking spread operator when trying to assert the expected arguments being passed to a mocked method returns always 1 element of the array being passed instead of the full array.

#### How to reproduce

//Class being mocked:
```php
interface EventBusInterface
{
    public function publish(DomainEvent ...$events): void;
}
```

//Class being used:
```php
$this->eventBus->publish(...$events);
```

//Test:
```php
$eventBusMock = $this->createMock(EventBusInterface::class);
$eventBusMock
    ->expects($this->once())
    ->method('publish')
    ->with(self::callback(function (...$receivedEvents) : bool {

        $this->assertCount($expectedNumber, $receivedEvents); //Received events is always 1 element...
        return true;
    }));
```

#### Expected behavior

Inside the self::callback function, when counting the number of elements inside of the array passed as spread operator should return the actual number of elements.

For example:

```php
$events = [$event1, $event2];
$this->eventBus->publish(...$events);
```

```php
$eventBusMock = $this->createMock(EventBusInterface::class);
$eventBusMock
    ->expects($this->once())
    ->method('publish')
    ->with(self::callback(function (...$receivedEvents) : bool {

        $this->assertCount(2, $receivedEvents); //Should return true, as $receivedEvents should contain 2 events.
        return true;
    }));
```

kubawerlos:
Fixes https://github.com/sebastianbergmann/phpunit/issues/5891

I'm really open to a better title for this PR.

### DIFF
--- a/src/Framework/Constraint/Callback.php
+++ b/src/Framework/Constraint/Callback.php
@@ -42,7 +42,7 @@ final class Callback extends Constraint
     /**
      * @psalm-suppress ArgumentTypeCoercion
      */
-    public function hasVariadicParam(): bool
+    public function isVariadic(): bool
     {
         foreach ((new ReflectionFunction($this->callback))->getParameters() as $parameter) {
             if ($parameter->isVariadic()) {
@@ -63,8 +63,10 @@ final class Callback extends Constraint
      */
     protected function matches(mixed $other): bool
     {
-        return $this->hasVariadicParam()
-            ? ($this->callback)(...$other)
-            : ($this->callback)($other);
+        if ($this->isVariadic()) {
+            return ($this->callback)(...$other);
+        }
+
+        return ($this->callback)($other);
     }
 }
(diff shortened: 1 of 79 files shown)


src/Framework/Constraint/Callback.php
```diff
@@ -37,14 +39,32 @@ public function toString(): string
...
      * Evaluates the constraint for parameter $value. Returns true if the
      * constraint is met, false otherwise.
      *
      * @psalm-param CallbackInput $other
+     *
+     * @psalm-suppress InvalidArgument
      */
     protected function matches(mixed $other): bool
     {
-        return ($this->callback)($other);
```
sebastianbergmann, line 48: Please avoid using the ternary operator.

PR #5948, opened 2024-09-11
Support for Property Hooks in Test Doubles

### LINKED ISSUE #5945: Doubling interfaces with property hooks
[With the introduction of property hooks, we'll likely have interfaces with properties defined](https://wiki.php.net/rfc/property-hooks#interfaces).  Let's say I have the following interface in PHP 8.4:

```php
interface Foo
{
    public string $bar { get; }
}
```

When creating a mock for `Foo`, I'll need to mock the get-value for `$bar`.  AFAIK, there is no way (or issue created for it) to be able to mock it, eg something like:

```php

class FooTest extends TestCase
{
    public function testBar(): void
    {
         $foo = $this->createMock(Foo::class);
         // This is a made up method, but gets what I'm trying to do across
         $foo->property('bar')
             ->willReturn('baz');

         // Do tests and assertions...
    }
}
```

Of course, I could just create an anonymous class that implements `Foo` and sets `$bar`, but I may need to mock other methods that exist in an interface in real test cases.  What's the best way to do this?  Or does additional functionality need to be added to PHPUnit to support this?

Thanks for all you do for the PHP community!!

sebastianbergmann:
PHP 8.4 introduces [property hooks](https://wiki.php.net/rfc/property-hooks).

PHPUnit's test double functionality should support doubling property hooks and allow the configuration of return values for get hooks and expectations for set hooks.

```php
interface I
{
    public string $property { get; set; }
}
```

Here is an example of what configuring test stubs and mock objects for the interface `I` shown above could look like:

```php
$stub = $this->createStub(I::class);
$stub->method(PropertyHook::get('property'))->willReturn('value');

$mock = $this->createMock(I::class);
$mock->expects($this->once())->method(PropertyHook::set('property'))->with('value');
```

- [X] Interfaces with property hooks
- [X] Extendable concrete classes with non-final concrete property hooks

Support for abstract classes with abstract property hooks will not be implemented as `getMockForAbstractClass()` is deprecated and will be removed in PHPUnit 12.


davidbyoung:
I know this still in draft, but we should also support mocking [abstract properties in abstract classes](https://wiki.php.net/rfc/property-hooks#abstract_properties) using similar functionality to the above.

sebastianbergmann:
@joseph-sentry Can you provide insight as to why we get `We are unable to process any of the uploaded JUnit XML files. Please ensure your files are in the right format.` in the comment made by Codecov? I assume (based on #5931) that you're with Sentry / Codecov, sorry in case I'm wrong.

davidbyoung:
FYI I'm playing around with you have so far (thank you for this!!), and found the following bug when mocking a nullable property getter:

```php

interface Foo
{
    public ?Bar $bar { get; }
}

interface Bar {}

class MyTest extends TestCase
{
    public function testNullablePropertyGetter(): void
    {
        $bar = $this->createMock(\Bar::class);
        $foo = $this->createMock(\Foo::class);
        $foo->method('$bar::get')
            ->willReturn($bar);
        // Dummy assertion...
        $this->assertTrue(true);
    }
}
````

Running this test will give you the following error:

```
Method $bar::get may not return value of type MockObject_Bar_878f2f68, its declared return type is "?Bar"
```

**Comment by sebastianbergmann** [comment]
> found the following bug when mocking a nullable property getter

Thank you, but that is expected. At least for me ;-) Jokes aside, yes, I know that not all bases are covered yet when it comes to type declarations.

**Comment by joseph-sentry** [comment]
The codecov JUnit error is due to a parser error on our end where I assumed that a `classname` attribute would always exist for `testcase` elements, I made a fix for it on the parser side. I will have to integrate the latest version of the parser with our backend before you see it as fixed however.

**Comment by sebastianbergmann** [comment]
> found the following bug when mocking a nullable property getter

Please try again with the current state of the branch.

**Comment by davidbyoung** [comment]
> > found the following bug when mocking a nullable property getter
> 
> Please try again with the current state of the branch.

Updated to the latest commit in this branch, but now I'm back to the same error I got before you implemented the feature, which is

```
PHP Fatal error:  Class MockObject_Foo_a2ab9d44 contains 1 abstract method and must therefore be declared abstract or implement the remaining methods (Foo::$bar::get)
```

Running `composer show phpunit/phpunit` generates the following, which indicates I'm using the latest commit:

```
name     : phpunit/phpunit
descrip. : The PHP Unit Testing framework.
keywords : phpunit, testing, xunit
versions : * dev-issue-5945/property-hooks
released : 2024-09-17, today
type     : library
license  : BSD 3-Clause "New" or "Revised" License (BSD-3-Clause) (OSI approved) https://spdx.org/licenses/BSD-3-Clause.html#licenseText
homepage : https://phpunit.de/
source   : [git] https://github.com/sebastianbergmann/phpunit.git e3ce86cd281b656541857cffab9d187ffab608f0
dist     : [zip] https://api.github.com/repos/sebastianbergmann/phpunit/zipball/e3ce86cd281b656541857cffab9d187ffab608f0 e3ce86cd281b656541857cffab9d187ffab608f0
path     : /home/dyoung/PHPStormProjects/aphiria/vendor/phpunit/phpunit
names    : phpunit/phpunit

support
issues : https://github.com/sebastianbergmann/phpunit/issues
security : https://github.com/sebastianbergmann/phpunit/security/policy
source : https://github.com/sebastianbergmann/phpunit/tree/issue-5945/property-hooks

autoload
classmap
src/
files

requires
ext-dom *
ext-json *
ext-libxml *
ext-mbstring *
ext-xml *
ext-xmlwriter *
myclabs/deep-copy ^1.12.0
phar-io/manifest ^2.0.4
phar-io/version ^3.2.1
php >=8.2
phpunit/php-code-coverage ^11.0.6
phpunit/php-file-iterator ^5.1.0
phpunit/php-invoker ^5.0.1
phpunit/php-text-template ^4.0.1
phpunit/php-timer ^7.0.1
sebastian/cli-parser ^3.0.2
sebastian/code-unit ^3.0.1
sebastian/comparator ^6.1.0
sebastian/diff ^6.0.2
sebastian/environment ^7.2.0
sebastian/exporter ^6.1.3
sebastian/global-state ^7.0.2
sebastian/object-enumerator ^6.0.1
sebastian/type ^5.1.0
sebastian/version ^5.0.1

suggests
ext-soap To be able to generate mocks based on WSDL files
```

sebastianbergmann:
> now I'm back to the same error I got before you implemented the feature

You will also need the current state of PHP 8.4 (https://github.com/php/php-src/commit/d75a289f6f9abe50934fe6f682666942ebbe6d7f and https://github.com/php/php-src/commit/2fce0bb8777ba0e6a016d6736a6063045b787295).

davidbyoung:
> > now I'm back to the same error I got before you implemented the feature
> 
> You need the current state of PHP 8.4 (with [php/php-src@d75a289](https://github.com/php/php-src/commit/d75a289f6f9abe50934fe6f682666942ebbe6d7f) and [php/php-src@2fce0bb](https://github.com/php/php-src/commit/2fce0bb8777ba0e6a016d6736a6063045b787295)).

Ah, good catch.  Compiled latest on `master`, and can confirm the issue I reported is fixed.  Thanks!

PR #5966, opened 2024-09-25
fix preg_match() warning for most filter usecases

verfriemelt-dot-org:
when a --filter option is provided and it starts with an invalid regex-delimiter, it will emit the following warning:

    Warning: preg_match(): Delimiter must not be alphanumeric, backslash, or NUL

this is really cumbersome when xdebug.scream=true is in use. also while stepdebugging with phpstorm for example, this will spam the console output of phpstorm.
we can and should prevent that most obvious case by just checking if the supplied filter starts with an alphanumeric character.

example:

image

### DIFF
--- a/src/Runner/Filter/NameFilterIterator.php
+++ b/src/Runner/Filter/NameFilterIterator.php
@@ -14,6 +14,7 @@ use function implode;
 use function preg_match;
 use function sprintf;
 use function str_replace;
+use function substr;
 use Exception;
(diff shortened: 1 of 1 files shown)


verfriemelt-dot-org:
i understand the reasoning about `xdebug.scream=true`, but please note, that the output from phpstorm is soley `xdebug.mode=debug` and _not_ with scream! :thinking:

verfriemelt-dot-org:
:) thank you!

Issue #5973, opened 2024-09-28
PHPT requirements before process creation


staabm:
I wonder whether there is a way when using PHPT to detect test requirements early and prevent starting a subprocess at all

Same as https://github.com/sebastianbergmann/phpunit/issues/4391  but for PHPT

staabm:
//cc @MauricioFauth since you worked recently on similar stuff

PR #5974, opened 2024-09-28
Inline SKIPIF evaluation without side-effects

staabm:
## Proof of concept

while looking into PHPT performance, I realized that the `--SKIPIF--` condition is evaluated in a separate subprocess.

this means a major overhead for a maybe pretty simple side-effect-less condition like:

```php
--SKIPIF--
<?php version_compare(PHP_VERSION, "8.0", ">=")
      or echo "skip because attributes are only available since PHP 8.0"; ?>

```

I [built a small class](https://github.com/staabm/side-effects-detector) in which you can pass a string of source-code and it returns whether the given code would have side-effects when evaluated.

In PHPUnit I utilize the detected side-effect-free SKIPIF conditions and evaluate such code in the main-process instead of creating a new one. In case we cannot say with 100% certainty the code does not have side-effects we play save and run it in the subprocess.

----

when running `php ./phpunit tests/end-to-end/event/assert-failure.phpt` I can see the following results:

macos:

after this PR: `00:00.070 - 00:00.077`
before this PR: `00:00.092 - 00:00.098`

-> which means ~20% faster on my mac m1pro on a single test-case.

on windows with

```
PHP 8.3.11 (cli) (built: Aug 27 2024 21:28:35) (NTS Visual C++ 2019 x64)
Copyright (c) The PHP Group
Zend Engine v4.3.11, Copyright (c) Zend Technologies
```

after this PR: `00:00.360 - 00:00.395`
before this PR: `00:00.406 - 00:00.443`

-> seems also roughly ~20% faster
(see how slow running the same test is on windows vs. macos)

----

Possible future idea
- Don't use [subprocess for `—CLEAN—`](https://github.com/sebastianbergmann/phpunit/blob/e8d66de7723a905ebb741996959b09a223af9780/src/Runner/PHPT/PhptTestCase.php#L453-L468) as long as the code cannot modify the parent process (e.g. file IO is fine within the parent process)

----

I am pretty sure this needs more work but wanted to have opinions first, whether this is acceptable.

refs https://github.com/sebastianbergmann/phpunit/issues/5973

### DIFF
--- a/src/Runner/PHPT/PhptTestCase.php
+++ b/src/Runner/PHPT/PhptTestCase.php
@@ -27,6 +27,8 @@ use function is_file;
 use function is_readable;
 use function is_string;
 use function ltrim;
+use function ob_get_clean;
+use function ob_start;
 use function preg_match;
 use function preg_replace;
 use function preg_split;
@@ -66,6 +68,8 @@ use SebastianBergmann\CodeCoverage\Test\TestStatus\TestStatus;
 use SebastianBergmann\CodeCoverage\TestIdMissingException;
 use SebastianBergmann\CodeCoverage\UnintentionallyCoveredCodeException;
 use SebastianBergmann\Template\Template;
+use staabm\SideEffectsDetector\SideEffect;
+use staabm\SideEffectsDetector\SideEffectsDetector;
 use Throwable;
 
 /**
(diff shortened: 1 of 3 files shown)


staabm:
Tested this PR on a real world project with https://github.com/antecedent/patchwork/pull/168 (and described the results in the PR)

src/Runner/PHPT/PhptTestCase.php
```diff
@@ -450,6 +461,48 @@ private function shouldTestBeSkipped(array $sections, array $settings): bool
...
+
+        $detector    = new SideEffectsDetector;
+        $sideEffects = $detector->getSideEffects($skipIfCode);
+
+        if ($sideEffects === []) {
+            return false; // no side-effects
+        }
+
+        foreach ($sideEffects as $sideEffect) {
+            if ($sideEffect === SideEffect::STANDARD_OUTPUT) {
```
staabm, line 482: We could even report a SkipIf condition to be broken, when it does not contain a standard-out side-effect

staabm:
todo
- test-case (events)
- build.xml
- target version 11.5

staabm:
retargeted in https://github.com/sebastianbergmann/phpunit/pull/5998

PR #5977, opened 2024-10-05
TestDox result collector does not correctly handle baseline-ignored `E_DEPRECATED` issues

pcf0:
When comparing versions [11.3.6...11.4.0](https://github.com/sebastianbergmann/phpunit/compare/11.3.6...11.4.0), I noticed the change from commit https://github.com/sebastianbergmann/phpunit/commit/f36e89b74a7d19726ea2ebd1626a8c035a9e752d.
I think in the method `testTriggeredPhpDeprecation` the wrong line was deleted?

### DIFF
--- a/src/Logging/TestDox/TestResult/TestResultCollector.php
+++ b/src/Logging/TestDox/TestResult/TestResultCollector.php
@@ -258,7 +258,7 @@ final class TestResultCollector
             return;
         }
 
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
Cherry-picked into `11.4` and merged to `main` from there (after adding test cases). Thank you!

Issue #5838, opened 2024-05-19
Allow enabling process isolation on suite/test base class


mondrake:
Process isolation can only be set via attributes in test classes (either overall on a class or on a single method) or on execution level (with attribute in XML or as a command line argument).

It would be good to enable process isolation on a middle ground, for instance for an entire suite or at a test base class level.

Drupal completed support of PHPUnit 10 in https://www.drupal.org/project/drupal/issues/3417066, and we needed there to find a workaround for this - there are 5 test suites in configuration, 3 of which have tests that need to run in process isolation, the others don't. 

Since for each test suite there's a base test class extending from TestCase, we resolved by extending the constructor and setting process isolation there:

```
  public function __construct(string $name) {
    parent::__construct($name);
    $this->setRunTestInSeparateProcess(TRUE);
  }

```

https://git.drupalcode.org/project/drupal/-/blame/11.x/core/tests/Drupal/KernelTests/KernelTestBase.php?ref_type=heads#L107-L113

Prior to PHPUnit 10, this was possible by overriding TestCase properties, but this is no longer possible. The workaround works, but extends a constructor marked `@internal` and we are exposed to potential changes.

longwave:
Would it be possible to look for class level attributes in parent classes as well? Then we could specify `[#RunTestsInSeparateProcesses]` on the base class and it would automatically apply to all subclasses.

sebastianbergmann:
I will change PHPUnit so that process isolation can be configured at four levels:

1. At the finest grain, adding `#[RunInSeparateProcess]` to a single test method will isolate just that method.
2. One step broader, adding `#[RunTestsInSeparateProcesses]` to a test class will isolate every test method in that class.
3. **I will make this class-level attribute inheritable:** when a base class carries `#[RunTestsInSeparateProcesses]`, every subclass at any depth will also run its tests in separate processes: if `A` has the attribute and `B extends A` and `C extends B`, tests in all three classes are isolated.
4. Finally, at the broadest level, setting `processIsolation="true"` on the `<phpunit>` element in `phpunit.xml` (or passing `--process-isolation` on the command line) will isolate every test in the entire suite.

PHP's `ReflectionClass::getAttributes()` deliberately does not return attributes from parent classes. This is by design in the language. PHPUnit follows this convention: its `AttributeParser::forClass()` faithfully reports only what is declared on the class itself.

Making all attributes inheritable would be wrong. Most class-level attributes are metadata that should not propagate, for example:

[...]

PR #5981, opened 2024-10-08
Merge metadata class level attributes with ancestor classes

mondrake:
This is to address #5838.

The idea, per https://github.com/sebastianbergmann/phpunit/issues/5838#issuecomment-2127634446, is to traverse the class hierarchy and merge class level attributes of the test class with those of its ancestors.


sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it. It is by design that metadata is not inherited.

PR #5989, opened 2024-10-13
Disable Xdebug in subprocesses when it is not used

staabm:
before this PR xdebug was only auto-disabled for PHPT.

as of this PR we disable xdebug for all sub-processes to make them faster - but only if no active debugger is attached.

this means subprocess isolated tests can still be debugged without the need for a additional cli-option or config setting changed just by attaching e.g. PHPStorm (so no change in DX).

---

I initially tried disabling xdebug for the whole sub-process by not loading the extension at all, but this would either require a new [composer/xdebug-handler](https://github.com/composer/xdebug-handler) dependency or a way more involved fix.

disabling xdebug already provides most perf benefits.

---

Benchmark (xdebug loaded, but no debugger attached): 

`/c/tools/php83/php phpunit tests/end-to-end/data-provider/log-junit-isolation.phpt` with phpunit@a82cc0548 takes 2.5 - 2.6 seconds (phpunit reported runtime)

After this PR: `/c/tools/php83/php phpunit tests/end-to-end/data-provider/log-junit-isolation.phpt` takes 0.5-0.6 seconds

tested with PHP 8.3.11 with Xdebug v3.4.0beta1 on windows11x64

php.ini settings:
```
zend_extension=xdebug
xdebug.mode=debug
xdebug.start_with_request=yes
```

### DIFF
--- a/src/Util/PHP/DefaultJobRunner.php
+++ b/src/Util/PHP/DefaultJobRunner.php
@@ -27,6 +27,7 @@ use function sys_get_temp_dir;
 use function tempnam;
 use function trim;
 use function unlink;
+use PHPUnit\Runner\CodeCoverage;
 use SebastianBergmann\Environment\Runtime;
 
 /**
@@ -184,6 +185,14 @@ final readonly class DefaultJobRunner implements JobRunner
                     array_keys($xdebugSettings),
                 ),
             );
+
+            // disable xdebug if not required to reduce xdebug performance overhead in subprocesses
+            if (
+                !CodeCoverage::instance()->isActive() &&
+                xdebug_is_debugger_active() === false
+            ) {
+                $phpSettings['xdebug.mode'] = 'xdebug.mode=off';
+            }
         }
 
         $command = array_merge($command, $this->settingsToParameters($phpSettings));
--- a/tests/end-to-end/event/_files/XdebugIsDisabled.php
+++ b/tests/end-to-end/event/_files/XdebugIsDisabled.php
@@ -1,0 +1,23 @@
+<?php declare(strict_types=1);
+/*
+ * This file is part of PHPUnit.
+ *
+ * (c) Sebastian Bergmann <sebastian@phpunit.de>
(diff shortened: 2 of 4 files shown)


tests/end-to-end/event/_files/XdebugIsDisabled.php
```diff
@@ -0,0 +1,23 @@
...
+use function extension_loaded;
+use function ini_get;
+use PHPUnit\Framework\TestCase;
+
+final class XdebugIsDisabled extends TestCase
+{
+    public function testOne(): void
+    {
+        $this->assertTrue(extension_loaded('xdebug'));
+        $this->assertSame('', (string) ini_get('xdebug.mode'));
```
staabm, line 21: when `xdebug.mode=off` is passed, `xdebug.mode` returns `''`

src/Util/PHP/DefaultJobRunner.php
```diff
@@ -174,16 +175,23 @@ private function buildCommand(Job $job, ?string $file): array
...
-                $phpSettings,
-                $runtime->getCurrentSettings(
-                    array_keys($xdebugSettings),
-                ),
-            );
+            if (
+                !CodeCoverage::instance()->isActive() &&
+                xdebug_is_debugger_active() === false
+            ) {
+                $phpSettings[] = 'xdebug.mode=off';
```
staabm, line 182: Hm maybe we should append all but the mode xdebug settings here

staabm:
I think this one is now good to go

tests/end-to-end/event/process-isolation-disable-xdebug.phpt
```diff
@@ -0,0 +1,38 @@
+--TEST--
```
staabm, line 1: should this test be moved in a different folder?

sebastianbergmann:
Cherry-picked into `11.4` and merged to `11.5` and `main` from there. Thanks!

PR #5992, opened 2024-10-15
Keep information which data provider provided current data set

AdamPDotty:
For better error messages and lookup capability the names of data providers are kept. First in the existing data set structure then in InvalidDataProviderException if thrown and in the TestCase itself.
When you have multiple data providers for a test function it is really helpful to know which one is invalid, and you do not always start off with named sets.
Because this PR is about error messages a lot of own tests are affected.


sebastianbergmann:
@staabm You might be interested in this.

PR #5997, opened 2024-10-18
Enhancement: Configure tokens for `no_extra_blank_lines` fixer

localheinz:
This pull request

- [x] configures tokens for `no_extra_blank_lines` fixer
- [x] runs `tools/php-cs-fixer fix`

💁‍♂️ For reference, see https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/blob/v3.64.0/doc/rules/whitespace/no_extra_blank_lines.rst


sebastianbergmann:
```
Detected deprecations in use:
- Option "tokens: use_trait" used in `no_extra_blank_lines` rule is deprecated, use the rule `class_attributes_separation` with `elements: trait_import` instead.
```

PR #5998, opened 2024-10-18
Do not run `SKIPIF` section of PHPT test in separate process when it is free of side effects

staabm:
while looking into PHPT performance, I realized that the `--SKIPIF--` condition is evaluated in a separate subprocess.

this means a major overhead for a maybe pretty simple side-effect-less condition like:

```php
--SKIPIF--
<?php version_compare(PHP_VERSION, "8.0", ">=")
      or echo "skip because attributes are only available since PHP 8.0"; ?>

```

I [built a small class](https://github.com/staabm/side-effects-detector) in which you can pass a string of source-code and it returns whether the given code would have side-effects when evaluated.

In PHPUnit I utilize the detected side-effect-free SKIPIF conditions and evaluate such code in the main-process instead of creating a new one. In case we cannot say with 100% certainty the code does not have side-effects we play save and run it in the subprocess.

----

when running `php ./phpunit tests/end-to-end/event/assert-failure.phpt` I can see the following results:

macos:

after this PR: `00:00.070 - 00:00.077`
before this PR: `00:00.092 - 00:00.098`

-> which means ~20% faster on my mac m1pro on a single test-case.

on windows with

```
PHP 8.3.11 (cli) (built: Aug 27 2024 21:28:35) (NTS Visual C++ 2019 x64)
Copyright (c) The PHP Group
Zend Engine v4.3.11, Copyright (c) Zend Technologies
```

after this PR: `00:00.360 - 00:00.395`
before this PR: `00:00.406 - 00:00.443`

-> seems also roughly ~20% faster
(see how slow running the same test is on windows vs. macos)

refs https://github.com/sebastianbergmann/phpunit/issues/5973

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -272,6 +272,13 @@
             </fileset>
         </copy>
 
+        <copy file="${basedir}/vendor/staabm/side-effects-detector/LICENSE" tofile="${basedir}/build/tmp/phar/staabm-side-effects-detector/LICENSE"/>
+        <copy todir="${basedir}/build/tmp/phar/staabm-side-effects-detector">
+            <fileset dir="${basedir}/vendor/staabm/side-effects-detector/lib">
+                <include name="**/*.php" />
+            </fileset>
+        </copy>
+
         <copy file="${basedir}/vendor/myclabs/deep-copy/LICENSE" tofile="${basedir}/build/tmp/phar/myclabs-deep-copy/LICENSE"/>
         <copy todir="${basedir}/build/tmp/phar/myclabs-deep-copy">
             <fileset dir="${basedir}/vendor/myclabs/deep-copy/src">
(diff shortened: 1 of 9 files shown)


build.xml
```diff
@@ -272,6 +272,13 @@
             </fileset>
         </copy>
 
+        <copy file="${basedir}/vendor/staabm/side-effects-detector/LICENSE" tofile="${basedir}/build/tmp/phar/staabm-side-effects-detector/LICENSE"/>
+        <copy todir="${basedir}/build/tmp/phar/staabm-side-effects-detector">
+            <fileset dir="${basedir}/vendor/staabm/side-effects-detector/lib">
+                <include name="**/*.php" />
+            </fileset>
+        </copy>
```
staabm, line 280: is this correct like that?

staabm:
Blogged about it: https://staabm.github.io/2024/10/19/phpunit-codesprint-munich.html

PR #5999, opened 2024-10-18
Do not run `CLEAN` section of PHPT test in separate process when it is free of side effects that modify the parent process

staabm:
Don't use [subprocess for `—CLEAN—`](https://github.com/sebastianbergmann/phpunit/blob/e8d66de7723a905ebb741996959b09a223af9780/src/Runner/PHPT/PhptTestCase.php#L453-L468) as long as the code cannot modify the parent process (e.g. file IO is fine within the parent process).

utilizes https://github.com/staabm/side-effects-detector

followup to https://github.com/sebastianbergmann/phpunit/pull/5998
 
`php phpunit tests/end-to-end/event/phpt-clean.phpt`
before this PR 109-110ms
after this PR: 85-87ms

tested on PHP 8.3.12 macos m1 pro

### DIFF
--- a/src/Runner/PHPT/PhptTestCase.php
+++ b/src/Runner/PHPT/PhptTestCase.php
@@ -469,6 +469,37 @@ final class PhptTestCase implements Reorderable, SelfDescribing, Test
     /**
      * @param array<non-empty-string, non-empty-string> $sections
      */
+    private function shouldRunCleanInSubprocess(array $sections, string $cleanCode): bool
+    {
+        if (isset($sections['INI'])) {
+            // to get per-test INI settings, we need a dedicated subprocess
+            return true;
+        }
+
+        $detector    = new SideEffectsDetector;
+        $sideEffects = $detector->getSideEffects($cleanCode);
+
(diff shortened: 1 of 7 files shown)


staabm:
todo: skip or fix cases, in which path exists which are relative to the test-file

```
--CLEAN--
<?php
unlink(__DIR__ . '/stream-url-stat.src.txt');
unlink(__DIR__ . '/stream-url-stat.dest.txt');
```

tests/end-to-end/event/phpt-clean.phpt
```diff
@@ -24,8 +24,6 @@ Test Prepared (%sphpt-clean.phpt)
 Child Process Started
 Child Process Finished
 Test Passed (%sphpt-clean.phpt)
-Child Process Started
-Child Process Finished
```
staabm, line 28: here we can see the new feature in action. a pre-existing test no longer needs a subprocess to cleanup

staabm:
Blogged about it: https://staabm.github.io/2024/10/19/phpunit-codesprint-munich.html

PR #6002, opened 2024-10-19
Enhancement: Configure `target_commitish` to target branch to show commits to branch since release

### LINKED ISSUE #5761: Configure GitHub Releases to show "N commits to X.Y since this release" instead of "N commits to main since this release"
@ghostwriter already provided input:

- https://phpc.social/deck/@ghostwriter/112109909998118813
- https://phpc.social/deck/@ghostwriter/112109993950954210

localheinz:
This pull request

- [x] indirectly configures the `target_commitish` field to show commits to branch since last release to that branch

Fixes https://github.com/sebastianbergmann/phpunit/issues/5761.

💁‍♂️ For reference, see

- https://github.com/ncipollo/release-action/blob/v1.14.0/src/Action.ts#L128-L140
- https://github.com/ncipollo/release-action/blob/v1.14.0/src/Releases.ts#L72-L97
- https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#create-a-release--parameters

Also see https://github.com/ergebnis/playground/releases/tag/0.14.0 for a demonstration:

CleanShot 2024-10-19 at 11 01 32@2x

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -41,6 +41,7 @@ jobs:
           tag: ${{ env.RELEASE_TAG }}
           name: PHPUnit ${{ env.RELEASE_TAG }}
           bodyFile: release-notes.md
+          commit: "8.5"
 
       - name: Announce release
         id: mastodon


.github/workflows/release.yaml
```diff
@@ -41,6 +41,7 @@ jobs:
           tag: ${{ env.RELEASE_TAG }}
           name: PHPUnit ${{ env.RELEASE_TAG }}
           bodyFile: release-notes.md
+          commit: "8.5"
```
localheinz, line 44: Unless we can find a way to obtain this from the GitHub context (see [`push`](https://docs.github.com/en/webhooks/webhook-events-and-payloads#push) event), we need to hard-code this value here (and update in other branches).

ghostwriter: > we need to hard-code this value here (and update in other branches).

...or

`BRANCH_NAME: ${{ github.ref_name }} `

| Property name |Description |
|--------|--------|
| `github.ref_name` | The short ref name of the branch or tag that triggered the workflow run. This value matches the branch or tag name shown on GitHub. For example, `feature-branch-1`.| 

https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#github-context

Issue #5993, opened 2024-10-16
PHPUnit process is blocked when there's a lot of output and a test with RunInSeparateProcess


MauricioFauth:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 10.5.36
| PHP version         | 8.3.12
| Installation Method | Composer
| SO                  | Debian testing with deb.sury.org repository

#### Summary

When there's a lot of errors triggered before running PHPUnit, for example when a installed package triggers a lot of deprecation messages when being auto loaded, and there's a test that runs in a separate process, PHPUnit will hang indefinitely.

#### Current behavior

The process blocks here when reading stdout:
```php
$stdout = stream_get_contents($pipes[1]);
```
https://github.com/sebastianbergmann/phpunit/blob/0c843b0dfe6781c1f62d2b958766674ab10a4a40/src/Util/PHP/DefaultPhpProcess.php#L115-L125

Replacing both stream_get_contents calls (stdout and stderr) with this old code (without the timeout check) fixed the issue (removed by ccb3b24112153f349cec5c5b78e7186228af7aed):
https://github.com/sebastianbergmann/phpunit/blob/dc7281e50005b5e01582a5f857765ecf99a347c6/src/Util/PHP/DefaultPhpProcess.php#L127-L182

#### How to reproduce

`IssueTest.php`:
```php
use PHPUnit\Framework\Attributes\RunInSeparateProcess;
use PHPUnit\Framework\TestCase;

final class IssueTest extends TestCase
{
    #[RunInSeparateProcess]
    public function testOne(): void
    {
        $this->assertTrue(true);
    }
}
```
`file_that_trigger_errors.php`:
```php
<?php
trigger_error("error 1");
trigger_error("error 2");
trigger_error("error 3");
// ...
trigger_error("error 9997");
trigger_error("error 9998");
trigger_error("error 9999");
```
`php.ini`:
```ini
error_reporting=-1
display_errors=1
display_startup_errors=1
memory_limit=-1
zend.assertions=1
assert.exception=1
```
`composer.json`:
```json
    "autoload-dev": { "files": [ "file_that_trigger_errors.php" ] }
```

staabm:
great report. 

adding back async streams would also help in [not blocking the phpunit main process](https://github.com/sebastianbergmann/phpunit/issues/5986) while workers do their thing

staabm:
I can reproduce using

### `IssueTest5993.php`

```
--TEST--
[...]

PR #6003, opened 2024-10-19
PHPUnit process is blocked when there's a lot of output and a test with separate process

staabm:
closes https://github.com/sebastianbergmann/phpunit/issues/5993

### DIFF
--- a/src/Util/PHP/DefaultPhpProcess.php
+++ b/src/Util/PHP/DefaultPhpProcess.php
@@ -12,12 +12,15 @@ namespace PHPUnit\Util\PHP;
 use function array_merge;
 use function fclose;
 use function file_put_contents;
+use function fread;
 use function fwrite;
 use function is_array;
 use function is_resource;
 use function proc_close;
 use function proc_open;
-use function stream_get_contents;
+use function proc_terminate;
+use function sprintf;
+use function stream_select;
 use function sys_get_temp_dir;
 use function tempnam;
 use function unlink;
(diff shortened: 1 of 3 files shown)


tests/end-to-end/regression/5993/Issue5993Test.php
```diff
@@ -0,0 +1,29 @@
...
+error_reporting=-1
+display_errors=1
+display_startup_errors=1
+memory_limit=-1
+zend.assertions=1
+assert.exception=1
+--SKIPIF--
+<?php
+for ($i = 0; $i < 390; $i++) {
+    \trigger_error("error {$i}");
```
mvorisek, line 13: Are all the inis needed? Isn't standard long fwrite to stderr/stdout enough?

tests/end-to-end/regression/5993.phpt
```diff
@@ -0,0 +1,14 @@
...
+$_SERVER['argv'][] = '--do-not-cache-result';
+$_SERVER['argv'][] = '--no-configuration';
+$_SERVER['argv'][] = '--debug';
+$_SERVER['argv'][] = __DIR__ . '/5993/Issue5993Test.php';
+
+require_once __DIR__ . '/../../bootstrap.php';
+
+(new PHPUnit\TextUI\Application)->run($_SERVER['argv']);
+--EXPECTF--
+%A
```
mvorisek, line 14: I would expect the exact exception asserted here.

src/Util/PHP/DefaultPhpProcess.php
```diff
@@ -112,16 +115,64 @@ protected function runProcess(string $job, array $settings): array
...
+
+                        unset($pipes[$pipeOffset]);
+                    } elseif ($pipeOffset === 1) {
+                        $stdout .= $line;
+                    } else {
+                        $stderr .= $line;
+                    }
+                }
+
+                if (empty($pipes)) {
```
mvorisek, line 172:
```suggestion
                if ($pipes === []) {
```

src/Util/PHP/DefaultPhpProcess.php
```diff
@@ -112,16 +115,64 @@ protected function runProcess(string $job, array $settings): array
...
-            fclose($pipes[2]);
+                    foreach ($pipes as $i => $origPipe) {
+                        if ($pipe === $origPipe) {
+                            $pipeOffset = $i;
+
+                            break;
+                        }
+                    }
+
+                    if (!$pipeOffset) {
```
mvorisek, line 155:
```suggestion
                    if ($pipeOffset === 0) {
```

src/Util/PHP/DefaultPhpProcess.php
```diff
@@ -112,16 +115,64 @@ protected function runProcess(string $job, array $settings): array
...
 
-        if (isset($pipes[1])) {
-            $stdout = stream_get_contents($pipes[1]);
+        unset($pipes[0]);
+        $timeout = 5;
 
-            fclose($pipes[1]);
-        }
+        while (true) {
+            $r = $pipes;
```
mvorisek, line 122:
```suggestion
            $readPipes = $pipes;
```

tests/end-to-end/regression/5993/Issue5993Test.php
staabm: if you want us to assist in fixing this problem, you could try to reduce the reproducer to a very minimum.

we suppose the problem is a underlying php-src bug and having a reproducer small enough to report to the php-src project would be valuable.

mvorisek: Is the main or the child blocked?

If the main (reading) process, then output to long stdout/stderr must be enough.

If the child process, then I do not understand the problem.

staabm: I think 3 processes are involved. Needs further investigation

sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it. This should be solved by 802790fab7a3ac85f8284b024b679390f08d316b.

PR #6004, opened 2024-10-19
Enhancement: Enable `bump-after-update` in `composer.json`

localheinz:
This pull request

- [x] enables nable `bump-after-update` in `composer.json`

💁‍♂️ For reference, see https://getcomposer.org/doc/06-config.md#bump-after-update

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -23,17 +23,11 @@
         </exec>
     </target>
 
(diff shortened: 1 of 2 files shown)


sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it as I prefer the explicitness of the current solution.

PR #6006, opened 2024-10-20
Removed unused PHPUnit\TextUI\Configuration\Builder class

staabm:
since this class is not `@internal` or similar, maybe we cannot delete it right away?
do we need some deprecation process?
or is this class meant for use outside PHPUnit or something I am not aware of?


sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it. While this class is not used by PHPUnit itself, it is [offered](https://docs.phpunit.de/en/11.4/extending-phpunit.html#wrapping-the-test-runner) for implementors of wrappers around PHPUnit.

PR #6008, opened 2024-10-20
Remove `*\ResultPrinter::flush()` which is only used from tests

staabm:
I guess this method mistakenly exists because we have a `Printer` interface which requires a `flush` method.
since the changed printers don't implement said interface, I think the methods are obsolete.


sebastianbergmann:
Cherry-picked into `10.5` and merged to `11.4`, `11.5`, and `main` from there. Thanks!

PR #6010, opened 2024-10-21
Improve parameter name for `assertGreaterThan()`, `assertGreaterThanOrEqual()`, `assertLessThan()`, and `assertLessThanOrEqual()`

### LINKED ISSUE #5849: Parameter name `$expected` does not make sense for `assertLessThan()` etc.
The parameter names $expected, $actual make sense when you're comparing two things that you expect to be the same.

But for assertLessThan() (and similar assertions), these do not make sense. $expected is not the value you expect to get, since you expect the value you are testing to be less than $expected.

semkoolennl:
Resolves #5849 

Improved the `$expected` parameter name for the following constraints:
- `assertGreaterThan()`
- `assertGreaterThanOrEqual()`
- `assertLessThan()`
- `assertLessThanOrEqual()`

I decided to use `minimum` and `maximum` as parameter names as these make the comparisons clear and intuitive.

### DIFF
--- a/src/Framework/Assert.php
+++ b/src/Framework/Assert.php
@@ -553,9 +553,9 @@ abstract class Assert
      *
      * @throws ExpectationFailedException
      */
-    final public static function assertGreaterThan(mixed $expected, mixed $actual, string $message = ''): void
+    final public static function assertGreaterThan(mixed $minimum, mixed $actual, string $message = ''): void
     {
-        self::assertThat($actual, self::greaterThan($expected), $message);
+        self::assertThat($actual, self::greaterThan($minimum), $message);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


sebastianbergmann:
Cherry-picked into `11.4` and merged to `11.5` and `main` from there. Thanks!

NanoSector:
Just wanted to note that this broke some of our tests with a dependency update, as we used named parameters with these functions.

Not sure if that can be seen as a backwards compatibility break in a patch version.

sebastianbergmann:
> Not sure if that can be seen as a backwards compatibility break in a patch version.

With PHPUnit, this is never a backwards compatibility break. Every source file of PHPUnit has this at the top:

```
/**
 * @no-named-arguments Parameter names are not covered by the backward compatibility promise for PHPUnit
 */
```

See https://github.com/sebastianbergmann/phpunit/blob/11.4.3/src/Framework/Assert.php#L70-L72, for example and the [Backward Compatibility](https://phpunit.de/backward-compatibility.html) page.

PR #6012, opened 2024-10-23
Remove empty lines between TeamCity events

smirok:
Hi, PhpStorm is here!

[TeamCity output](https://www.jetbrains.com/help/teamcity/service-messages.html) support was introduced on #1858. We have a workaround that allows us to cut such lines out. However, it's better for our internal events processing to avoid empty lines between teamcity events.


smirok:
Fixed CS in the last commit

sebastianbergmann:
Cherry-picked into `10.5` and merged to `11.4`, `11.5`, and `main` from there. Thank you!

PR #6014, opened 2024-10-26
strip closing PHP tags from code run in local sandbox

xabbuh:
According to https://qa.php.net/phpt_details.php#skipif_section the code in this section is enclosed by PHP tags. While it's perfectly valid to omit the closing tag we need to make sure that running the skipif section does not fail when a closing tag is used.

### DIFF
--- a/src/Runner/PHPT/PhptTestCase.php
+++ b/src/Runner/PHPT/PhptTestCase.php
@@ -528,7 +528,7 @@ final class PhptTestCase implements Reorderable, SelfDescribing, Test
 
     private function runCodeInLocalSandbox(string $code): string
     {
-        $code = preg_replace('/^<\?(php)?/', '', $code);
+        $code = preg_replace('/^<\?(?:php)?|\?>\s*+$/', '', $code);
         $code = preg_replace('/declare\S?\([^)]+\)\S?;/', '', $code);
 
         // wrap in immediately invoked function to isolate local-side-effects of $code from our own process
--- a/tests/end-to-end/_files/phpt-skipif-closing-php-tag.phpt
+++ b/tests/end-to-end/_files/phpt-skipif-closing-php-tag.phpt
@@ -1,0 +1,10 @@
+--TEST--
+PHPT skip condition which is terminated with a closing PHP tag
+--FILE--
+<?php declare(strict_types=1);
+print "Nothing to see here, move along";
+--SKIPIF--
+<?php declare(strict_types=1);
+print "skip: something terrible happened\n";
+?>
+--EXPECT--
--- a/tests/end-to-end/event/phpt-skipif-closing-php-tag.phpt
+++ b/tests/end-to-end/event/phpt-skipif-closing-php-tag.phpt
@@ -1,0 +1,30 @@
+--TEST--
+The right events are emitted in the right order for a skipped PHPT test
+--FILE--
+<?php declare(strict_types=1);
+$_SERVER['argv'][] = '--do-not-cache-result';
+$_SERVER['argv'][] = '--no-configuration';
+$_SERVER['argv'][] = '--debug';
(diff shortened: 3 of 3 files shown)


xabbuh:
see https://github.com/xabbuh/symfony/actions/runs/11522923171/job/32079815289#step:8:3350 for an example job where the Symfony test suite fails with the new code in PHPUnit 11.5 while it used to succeed before (the phpt file in question is https://github.com/symfony/symfony/blob/7.2/src/Symfony/Bridge/PhpUnit/Tests/DeprecationErrorHandler/php_deprecation_from_vendor_class.phpt#L4).

src/Runner/PHPT/PhptTestCase.php
```diff
@@ -529,6 +529,7 @@ private function shouldRunSkipIfInSubprocess(array $sections, string $skipIfCode
     private function runCodeInLocalSandbox(string $code): string
     {
         $code = preg_replace('/^<\?(php)?/', '', $code);
```
mvorisek, line 531:
```suggestion
        $code = preg_replace('/^<\?(php)?|\?>\s*/', '', $code);
```

src/Runner/PHPT/PhptTestCase.php
```diff
@@ -529,6 +529,7 @@ private function shouldRunSkipIfInSubprocess(array $sections, string $skipIfCode
...
+        $code = preg_replace('/\?>$/', '', rtrim($code));
```
mvorisek, line 532:
```suggestion
        $code = preg_replace('/^<\?(php)?|\?>\s*/', '', $code);
```

staabm (commented):
Lgtm - Thanks 🙏

staabm:
@xabbuh out of curiosity- do you have any experience to share regarding the new subprocess avoiding skip-if feature?

xabbuh:
> @xabbuh out of curiosity- do you have any experience to share regarding the new subprocess avoiding skip-if feature?

Apart from one failing test which resulted from the issue that is fixed by this PR the Symfony test suite succeeds with the latest 11.5 commit. Not sure if that fully answers your question though. :) Let me know if you are interested in something specific.

staabm:
Any perf differences? :-)

staabm:
Any perf differences? :-)

xabbuh:
I did a quick test with 11.3.6, 11.4.2 and 11.5 running the PHPUnitBridge test suite:

11.3.6: 5.9186s
11.4.2: 5.4252s (91.66% compared to 11.3)
11.5: 5.3022s (97.73% compared to 11.4, 89.59% compared to 11.3)

It's a bit faster. I will do some more investigation as the test suite does not exclusively consist of phpt tests and we also have other components that one the other hand also have phpt tests. I'll keep you posted.

staabm:
Related https://staabm.github.io/2024/10/19/phpunit-codesprint-munich.html

> ProTip™: In the future try to prevent use of die() or exit() in --SKIPIF-- sections of PHPT tests, and use echo or similar instead, so PHPUnit can run your --SKIPIF-- code in the main process. In case you are not sure whether PHPUnit is starting subprocesses, try running your test with --debug, and watch for the newly added Child Process Started, Child Process Finished events.

src/Runner/PHPT/PhptTestCase.php
```diff
@@ -528,7 +528,7 @@ private function shouldRunSkipIfInSubprocess(array $sections, string $skipIfCode
 
     private function runCodeInLocalSandbox(string $code): string
     {
-        $code = preg_replace('/^<\?(php)?/', '', $code);
+        $code = preg_replace('/^<\?(?:php)?|\?>\s*+$/', '', $code);
         $code = preg_replace('/declare\S?\([^)]+\)\S?;/', '', $code);
```
staabm, line 532: Since you introduced a non-capturing group above, you could here too

PR #6017, opened 2024-10-28
Deprecation Triggers are not registered when process isolation is used

ruudk:
When process isolation is used, the deprecation triggers are not registered on the ErrorHandler.

Related to https://github.com/sebastianbergmann/phpunit/pull/5923


ruudk:
@sebastianbergmann This one is ready. We first made sure the test was failing before applying our fix. Please have a look when you have time 🙏

sebastianbergmann:
Cherry-picked into `11.4` and merged to `11.5` and `main` from there.

PR #6022, opened 2024-11-01
Avoid subprocess when SKIPIF triggers IO

staabm:
I realized that since we already separate include/require from other IO functions, we can run tests which only do IO (like for filesystem cleanup stuff) in the main process (like we already do for `--CLEAN--`).

followup to https://github.com/sebastianbergmann/phpunit/pull/5998, https://github.com/sebastianbergmann/phpunit/pull/5999

### DIFF
--- a/tests/end-to-end/event/phpt-skipif-io.phpt
+++ b/tests/end-to-end/event/phpt-skipif-io.phpt
@@ -1,0 +1,31 @@
+--TEST--
+The right events are emitted in the right order for PHPT test using IO in skipif
+--FILE--
+<?php declare(strict_types=1);
+$_SERVER['argv'][] = '--do-not-cache-result';
+$_SERVER['argv'][] = '--no-configuration';
+$_SERVER['argv'][] = '--debug';
+$_SERVER['argv'][] = __DIR__ . '/../_files/phpt-skipif-io.phpt';
+
+require __DIR__ . '/../../bootstrap.php';
+
+(new PHPUnit\TextUI\Application)->run($_SERVER['argv']);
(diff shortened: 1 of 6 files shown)


tests/end-to-end/event/phpt-skipif-require.phpt
```diff
@@ -0,0 +1,33 @@
...
+Test Runner Started
+Test Suite Sorted
+Test Runner Execution Started (1 test)
+Test Suite Started (%s%ephpt-skipif-require.phpt, 1 test)
+Test Preparation Started (%s%ephpt-skipif-require.phpt)
+Test Prepared (%s%ephpt-skipif-require.phpt)
+Child Process Started
+Child Process Finished
+Child Process Started
+Child Process Finished
```
staabm, line 27: require + test-itself is running in subprocess

tests/end-to-end/event/phpt-skipif-io.phpt
```diff
@@ -0,0 +1,31 @@
...
+Event Facade Sealed
+Test Suite Loaded (1 test)
+Test Runner Started
+Test Suite Sorted
+Test Runner Execution Started (1 test)
+Test Suite Started (%s%ephpt-skipif-io.phpt, 1 test)
+Test Preparation Started (%s%ephpt-skipif-io.phpt)
+Test Prepared (%s%ephpt-skipif-io.phpt)
+Child Process Started
+Child Process Finished
```
staabm, line 25: only the test itself is running in subprocess. skipif subprocess avoided.

PR #6030, opened 2024-11-14
Remove unnecessary ext-json requirement

pereorga:
JSON extension is always available as of PHP 8.0+.


sebastianbergmann:
> JSON extension is always available as of PHP 8.0+.

That does not prevent some distributions to patch PHP to disable it.

PR #6033, opened 2024-11-16
Fix type errors in PhptTestCase

staabm:
fixes
```
------ -------------------------------------------------------------------------------------------------------------------------------------------- 
  Line   Runner/PHPT/PhptTestCase.php                                                                                                                
 ------ -------------------------------------------------------------------------------------------------------------------------------------------- 
  634    Parameter &$sections by-ref type of method PHPUnit\Runner\PhptTestCase::parseExternal() expects array<non-empty-string, non-empty-string>,  
         non-empty-array<non-empty-string, string|false> given.                                                                                      
         💡 You can change the parameter out type with @param-out PHPDoc tag.                                                                        
  767    Parameter &$job @param-out type of method PHPUnit\Runner\PhptTestCase::renderForCoverage() expects non-empty-string, string given.          
 ------ -------------------------------------------------------------------------------------------------------------------------------------------- 
```

on my way to ease the PHPStan 2.0 update


src/TextUI/Command/Commands/VersionCheckCommand.php
```diff
@@ -41,8 +41,6 @@ public function execute(): Result
 
         $latestCompatibleVersion = $this->downloader->download('https://phar.phpunit.de/latest-version-of/phpunit-' . $this->majorVersionNumber);
 
-        assert($latestCompatibleVersion !== false);
```
staabm, line 44: we explicitly handle the `false` case below, so no need to narrow it beforehand

sebastianbergmann:
Please send these changes to `11.4`. Thanks!

PR #6041, opened 2024-11-19
Fixed before and after class methods also called from primary process when test class or method is run in separated process.

akotulu:
Our project uses heavily `define` constants in multiple sub modules and integration tests are ran inside separated test class process using `RunClassInSeparateProcess` attribute. Every `TestCase` sets its own defines (constants) in `setUpBeforeClass`, but PHPUnit calls the function on every registered `TestCase` from primary process thus failing to initialize sub module resources after first call, and primary test process receives an exception (e.g. invalid path).

```
// Define application path
define('APPLICATION_PATH', realpath(PROJECT_APPLICATION_PATH . '/module/xxx'));
// Define assets' path
define('ASSETS_DIR', realpath(__DIR__ . '/../assets'));
// Define server name
define('SERVER_NAME', 'xxx.abc.ee');
// Boot
include __DIR__ . '/../../../boot.php';
```

Before and after class methods must not be called at all from primary process if the tests are ran inside separated process (sandbox).

Also there is a issue with the `src/Framework/TestRunner/templates/class.tpl`. Idk if the parameter `RunClassInSeparateProcess` isn't properly implemented, but the template is for sure invalid when set and `setUpBeforeClass` is called for every test defined in `TestCase`.


sebastianbergmann:
> Also there is a issue with the `src/Framework/TestRunner/templates/class.tpl`. Idk if the parameter `RunClassInSeparateProcess` isn't properly implemented, but the template is for sure invalid when set and `setUpBeforeClass` is called for every test defined in `TestCase`.

That would be #5230.

sebastianbergmann:
Please target the `10.5` branch as this is the oldest branch that still receives bug fixes. And please add at least one test case that fails without the proposed changes but passes with them. Thank you!

akotulu:
I've added a test and fixed after class being called twice. How do I target this branch? Checked out only master.

tests/end-to-end/sandbox/_files/BeforeAndAfterClassMethodCallCountTest.php
```diff
@@ -0,0 +1,43 @@
...
+use function file_get_contents;
+use function file_put_contents;
+use function intval;
+use PHPUnit\Framework\Attributes\RunClassInSeparateProcess;
+use PHPUnit\Framework\Attributes\RunTestsInSeparateProcesses;
+use PHPUnit\Framework\TestCase;
+
+//#[RunTestsInSeparateProcesses]
+#[RunClassInSeparateProcess]
+final class BeforeAndAfterClassMethodCallCountTest extends TestCase
```
mvorisek, line 21: Simillar test would be good for a class with some tests run in a separate process and some not.

akotulu:
I've fixed all the issues with separated process handling. Re-implemented separated process class template and added extra tests to validate all the before and after class methods are properly called.

src/Framework/TestRunner/templates/class.tpl
```diff
@@ -69,52 +72,67 @@ function __phpunit_run_isolated_test()
 
     ErrorHandler::instance()->useDeprecationTriggers($deprecationTriggers);
 
-    $test = new {className}('{name}');
+    ini_set('xdebug.scream', '0');
```
staabm, line 75: is this line fixing a separate problem and should be separated from this PR?

akotulu: Well problem began with PHPUnit calling before and after class methods from primary process when run in separated process. After fixing the issue I stumbled on another one where those methods are called for every test while `RunClassInSeparateProcess` is defined. So yes, this PR contains fixes to before and after class methods being falsely called while using `RunTestsInSeparateProcesses`, `RunClassInSeparateProcess` and `RunInSeparateProcess` attribute.

akotulu: This line was in the source before my redesign. Copied common functionality from previous version. 

```
$test->setData('{dataName}', unserialize('{data}'));
$test->setDependencyInput(unserialize('{dependencyInput}'));
```

These lines I don't known what they do. Are they for proxing primary process global state?
Our project now runs all the tests properly without any issues. We have suites without and with separated process. Also mixed tests.

sebastianbergmann: > ```
> $test->setData('{dataName}', unserialize('{data}'));
> $test->setDependencyInput(unserialize('{dependencyInput}'));
> ```
> 
> These lines I don't known what they do.

The pass data from data providers and depended-upon tests to the test.

sebastianbergmann:
Please fix the coding guidelines violations.

sebastianbergmann:
This pull request has grown quite a bit since it was first proposed, making it hard to review. For instance, the change from an array to an object structure seems to be unrelated to problems that you want to address. And "problems" is a problem: this pull request addressed multiple problems. It would be easier to review smaller, more focused pull requests.

akotulu:
Well this pull request can't be split cause changes are related to each other. Running tests in a separated process returns an array containing tests information. To distinguish it from method template returning test data as an array as, I changed returning array to object in both templates. Tests in separated process return array containing objects and method returns object.

sebastianbergmann:
Please have a look at the failing tests.

akotulu:
Added a check for skipped and incomplete tests. This 'improvement' is kinda as is and can't be done the right way at the moment. `TestBuilder` is loading all the tests to the `TestSuite` in all cases from primary process. Loading tests step should come after analyzing the test file for attributes so `TestRunner` can be executed for a specific test file in either same or separated process and these checks with the boolean values can be removed. This would simplify a lot of unnecessary code. Also `TestStatus` class is kinda of overhead, you can't check for same class instance without using string `::class` or static variable.

sebastianbergmann:
Thank you for looking into the tests. However, I am unable to review this pull request which seems to get bigger and bigger.

akotulu:
Bigger? 5 files have changed + tests.
Changes:
1. Rewrote `class.tpl` template to support running whole test file in separated process;
2. Changed `method.tpl` returned array to object;
3. Changed `src/Framework/TestRunner/SeparateProcessTestRunner.php` function `processChildResult` to support reading an array of results;
2. In `TestSuite` added two booleans, changed `invokeMethodsAfterLastTest` and `invokeMethodsAfterLastTest` to public (moved private mothods to bottom);
3. Fixed your `tests/end-to-end/regression/5884/tests/FooTest.php` not unlinking temp files when test fails by moving assert to end of test;
4. Added new test directory called sandbox with 3 new tests;

If this is counted a big change then I don't know what is a small change. For me this is tiny change.

src/Framework/TestRunner/SeparateProcessTestRunner.php
```diff
@@ -252,22 +253,28 @@ static function (int $errno, string $errstr, string $errfile, int $errline): nev
...
         if ($childResult !== false) {
-            if (!empty($childResult['output'])) {
-                $output = $childResult['output'];
+            if (!is_array($childResult)) {
+                $childResult = [$childResult];
             }
 
-            Facade::instance()->forward($childResult['events']);
-            PassedTests::instance()->import($childResult['passedTests']);
+            foreach ($childResult as $result) {
```
staabm, line 260: Could this array to object transformation be extracted into a separate refactoring PR, without the actual bugfix?

akotulu: Hmm, will try. Have to close this pull request as I cannot fork again.

staabm: You can create a new branch on your fork

PR #6049, opened 2024-11-29
Fixed inconsistent arg usage

davesavic:
There's an inconsistency with args provided.

Args are documented as:

`arg <value>`

For example:

`--colors <flag>`

When using said arg it actually requires it to be provided as:

`--colors=<flag>`

This PR rectifies that and makes the inconsistent arg format the same as the others.


src/TextUI/Configuration/Cli/Builder.php
```diff
@@ -1,4 +1,6 @@
-<?php declare(strict_types=1);
+<?php
+
+declare(strict_types=1);
```
davesavic, line 3:
```suggestion
<?php declare(strict_types=1);
```

src/TextUI/Configuration/Cli/Builder.php
```diff
@@ -7,6 +7,7 @@
  * For the full copyright and license information, please view the LICENSE
  * file that was distributed with this source code.
  */
+
```
davesavic, line 10:
```suggestion
```

sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it as it is wrong. This can also be seen in the test failures.

* `--colors`, `--colors=never`, `--colors=auto`, and `--colors=always` are valid
* `--coverage-text` and `--coverage-text=<file>` are valid
* `--stop-on-deprecation` and `--stop-on-deprecation="string"` are valid

davesavic:
> Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it as it is wrong. This can also be seen in the test failures.
> 
> * `--colors`, `--colors=never`, `--colors=auto`, and `--colors=always` are valid
> * `--coverage-text` and `--coverage-text=<file>` are valid
> * `--stop-on-deprecation` and `--stop-on-deprecation="string"` are valid

I see, I think the documentation may need to be updated to reflect that then.

For example, running `phpunit --help` shows this:

image

But when I run the command as shown in the help: `phpunit --colors 'auto'`, I get this:

image

On the contrary, if I run this command: `phpunit --columns 3`, it works perfectly fine.

Based on the documentation, I would assume that `phpunit --colors never` would be a valid command but it's not.

PR #6050, opened 2024-11-29
Remove duplicate `is_file()` check

staabm:
there is only a single place where [`new PhptTestCase` is used](https://github.com/sebastianbergmann/phpunit/blob/5960490e8fef8ba47bb234d453562e4de608abd3/src/Framework/TestSuite.php#L227).
at this call-site we already check with `is_file` for the file existance, so there is no value in doing the same IO check a second time


drupol:
Hi,

Wouldn't it be better to remove it from the call-site(s) instead?

staabm:
I did also think about that.

IMO at the call-site we are more flexible and better know when/how todo such checks (more efficient).

Doing side effectfull or expensive operations in __construct feel like a bad practice IMO

PR #6051, opened 2024-11-29
Use `is_file` over `file_exists`

staabm:
[`is_file` is faster](https://bugs.php.net/bug.php?id=78285)


sebastianbergmann:
`AbstractPhpProcess.php` does not exist in newer branches so the changes in that file were not merged to `11.4`, `11.5`, and `main`. Please check whether `11.4` needs these changes and send a new pull request. Thanks! I will take care of merging from `11.4` to `11.5` and `main` then.

PR #6062, opened 2024-11-30
Changed separated process runner templates to return object instead of array allowing class template to return array of objects when refactored.

akotulu:
Here is first pass for the separated process improvement/fix for the class template. I've changed the return type from array to object and corrected the 5884 regression test to remove temp file. It throws error when executed inside WSL.

### DIFF
--- a/tests/end-to-end/regression/5884/tests/FooTest.php
+++ b/tests/end-to-end/regression/5884/tests/FooTest.php
@@ -61,10 +61,11 @@ final class FooTest extends TestCase
 
         // Now verify the original file is unchanged.
         $contents = file_get_contents($filename);
-        $this->assertSame('foo', $contents);
 
         chmod($filename, 0o755);
         unlink($filename);
+
+        $this->assertSame('foo', $contents);
     }
 
     #[WithoutErrorHandler]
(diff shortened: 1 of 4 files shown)


tests/end-to-end/regression/5884/tests/FooTest.php
```diff
@@ -61,10 +61,11 @@ public function testStreamToNonWritableFileWithPHPUnitErrorHandler(): void
 
         // Now verify the original file is unchanged.
         $contents = file_get_contents($filename);
-        $this->assertSame('foo', $contents);
 
         chmod($filename, 0o755);
         unlink($filename);
+
+        $this->assertSame('foo', $contents);
```
staabm, line 68: whats the background of this change?

staabm:
looks like a great simple first step, thank you.

tests/end-to-end/regression/5884/tests/FooTest.php
akotulu: I'm atm  working on Windows. Running tests inside WSL container throws error when particular test is executed. After debugging why it's happening (WSL host file-system ignores permissions and file is truncated), corrected the test to `unlink` before failing. It generated new temp file for each failure.

sebastianbergmann:
> in case CI is green

I just disabled the "new merge experience" experimental feature ... and suddenly see the "workflow awaiting approval" / "Approve and run" UI again. Clicked that button just now :)

sebastianbergmann:
I merged this although I do not understand why an object should be used instead of an array, especially considering the fact that this object is created by casting an `array` to `object`.

Issue #6065, opened 2024-12-02
introduce `#[RequiresEnvVar]` attribute


nikophil:
Hello,

what do you think of adding a `#[RequiresEnvVar]` (or `RequiresEnv`?) attribute?

````php
// would run the test if the env var is "truthy"
#[RequiresEnvVar('SOME_ENV_VAR')]
#[Test]
public function some_test(){}

// would run the test if the env var is strictly equals to 'foo'
#[RequiresEnvVar('SOME_ENV_VAR', value: 'foo')]
#[Test]
public function some_other_test(){}
````

nikophil:
Hi @sebastianbergmann 

I saw you've added some labels. If you're ok with this proposal, I'm willing to work on this :blush:

sebastianbergmann:
Sorry, yes: feel free to work on this.

PR #6074, opened 2024-12-08
`#[RequiresEnvironmentVariable]`

nikophil:
Implements #6065

### DIFF
--- a/src/Framework/Attributes/RequiresEnvironmentVariable.php
+++ b/src/Framework/Attributes/RequiresEnvironmentVariable.php
@@ -1,0 +1,40 @@
+<?php declare(strict_types=1);
+/*
+ * This file is part of PHPUnit.
+ *
+ * (c) Sebastian Bergmann <sebastian@phpunit.de>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+namespace PHPUnit\Framework\Attributes;
+
+use Attribute;
+
+/**
+ * @immutable
+ *
+ * @no-named-arguments Parameter names are not covered by the backward compatibility promise for PHPUnit
+ */
+#[Attribute(Attribute::TARGET_CLASS | Attribute::TARGET_METHOD | Attribute::IS_REPEATABLE)]
+final readonly class RequiresEnvironmentVariable
+{
+    private string $environmentVariableName;
+    private null|bool|int|string $value;
+
+    public function __construct(string $environmentVariableName, null|bool|int|string $value = null)
+    {
+        $this->environmentVariableName = $environmentVariableName;
+        $this->value                   = $value;
+    }
+
+    public function environmentVariableName(): string
+    {
+        return $this->environmentVariableName;
+    }
+
+    public function value(): null|bool|int|string
+    {
+        return $this->value;
+    }
+}
--- a/src/Metadata/Api/Requirements.php
+++ b/src/Metadata/Api/Requirements.php
@@ -14,6 +14,7 @@ use const PHP_OS_FAMILY;
 use const PHP_VERSION;
 use function addcslashes;
 use function array_column;
+use function array_key_exists;
 use function assert;
 use function extension_loaded;
 use function function_exists;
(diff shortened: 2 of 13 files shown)


src/Framework/Attributes/RequiresEnvironmentVariable.php
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * @immutable
+ *
+ * @no-named-arguments Parameter names are not covered by the backward compatibility promise for PHPUnit
+ */
+#[Attribute(Attribute::TARGET_CLASS | Attribute::TARGET_METHOD | Attribute::IS_REPEATABLE)]
+final readonly class RequiresEnvironmentVariable
+{
+    private string $environmentVariableName;
+    private null|bool|int|string $value;
```
nikophil, line 23: I'm not sure about the typing of this :thinking:

sebastianbergmann: I do not think the type for an environment variable can be narrowed more than that.

mvorisek, line 23: Environment variable can be strictly only string.

What is the meaning of `null`, `bool`, `int`?

Can I test environment variable presence/absence /wo value?

mvorisek: Based on https://github.com/sebastianbergmann/phpunit/pull/6074/files#diff-8f0e8d461aea381bf08d49440fbc5da892d38eb782518cd44bfe1dddce7638f4R119 it seems `bool` and `int` should be removed and `null` is used for testing if the variable is present /wo specific value.

nikophil: > Can I test environment variable presence/absence /wo specific value?

you can only test environment variable presence or with a specific value

> it seems bool and int should be removed and as never possible 

this was more or less my concern [here](https://github.com/sebastianbergmann/phpunit/pull/6074#discussion_r1875030847). I'm also wondering if we should only check the env var in `$_ENV` , or use `getenv()` or check at least one of them

nikophil: any thoughts @sebastianbergmann? I can provide a patch to my PR

mvorisek: > this was more or less my concern [here](https://github.com/sebastianbergmann/phpunit/pull/6074#discussion_r1875030847). I'm also wondering if we should only check the env var in `$_ENV` , or use `getenv()` or check at least one of them

Please test the behaviour with E2E test and some tests in separate process. The separate process must see the same environment variables in both linux and Windows.

sebastianbergmann: AFAICS, an environment variable can only be `string`. And just checking whether an environment variable is set or not without checking for an expected value does not make much sense to me.

nikophil: ok, let me fix this

mvorisek: see https://github.com/php-lock/lock/blob/2.3.0/tests/mutex/MutexConcurrencyTest.php#L26 for legit usecases of "variable is set"

nikophil: yeah, actually in https://github.com/zenstruck/foundry we also have legit usecases for only "variable is set" :sweat_smile:

sebastianbergmann: Fair enough.

nikophil: do you think in the case of "no value", I check if the env var exist, or should I check it is ok-ish? empty string and `'0'` would not pass

mvorisek: In my opinion:

- If an environment variable is tested for a presence (/wo value defined in the Attribute), I would expect `'0'` to be consideted as set.

- For binary `'1'`/`'0'` detection I would expect user to write `#[RequiresEnvironmentVariable('foo', '1')]` explicitly.

- `''` should be considered as unset as in many operating systems `FOO=` unsets the variable. Also, in CI parametrization/matrix empty value might be set, but with "no variable" meaning.

nikophil: > '' should be considered as unset as in many operating systems FOO= unsets the variable

yeah, good point

PR #6079, opened 2024-12-09
fix: RequiresEnvironmentVariable should only allow strings

nikophil:
see https://github.com/sebastianbergmann/phpunit/pull/6074#pullrequestreview-2488488457

some questions towards the behavior of the feature:
- should I only test `$_ENV` or maybe `$_ENV ?? getenv()`?
- when no value is provided, should I at least test the value of the env var is okish?
- not sure what @mvorisek meant [here](https://github.com/sebastianbergmann/phpunit/pull/6074#discussion_r1875896248) by "in separate process"?

### DIFF
--- a/tests/end-to-end/_files/requires_environment_variable/SomeTest.php
+++ b/tests/end-to-end/_files/requires_environment_variable/SomeTest.php
@@ -1,0 +1,62 @@
+<?php
+
+declare(strict_types=1);
+/*
+ * This file is part of PHPUnit.
+ *
+ * (c) Sebastian Bergmann <sebastian@phpunit.de>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+namespace PHPUnit\TestFixture\requires_environment_variable;
+
+use PHPUnit\Framework\Attributes\RequiresEnvironmentVariable;
+use PHPUnit\Framework\Attributes\RunInSeparateProcess;
+use PHPUnit\Framework\TestCase;
+
+final class SomeTest extends TestCase
+{
+    #[RequiresEnvironmentVariable('FOO', 'bar')]
+    public function testShouldNotRunFOOHasWrongValue(): void
+    {
+        $this->fail();
+    }
+
+    #[RequiresEnvironmentVariable('BAR')]
+    public function testShouldNotRunBARIsEmpty(): void
+    {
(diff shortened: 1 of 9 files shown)


tests/end-to-end/_files/requires_environment_variable/SomeTest.php
```diff
@@ -0,0 +1,35 @@
...
+    {
+        self::assertTrue(true);
+    }
+
+    #[RequiresEnvironmentVariable('FOO', '1')]
+    public function testTwoShouldRun(): void
+    {
+        self::assertTrue(true);
+    }
+}
```
mvorisek, line 35: > not sure what @mvorisek meant https://github.com/sebastianbergmann/phpunit/pull/6074#discussion_r1875896248 by "in separate process"?

https://docs.phpunit.de/en/10.5/annotations.html#runinseparateprocess

nikophil: I've added on test with `#[RunInSeparateProcess]`, is it good enough?

sebastianbergmann:
> when no value is provided, should I at least test the value of the env var is okish?

You mean when `RequiresEnvironmentVariable` is used without an expected value and only the existence of an environment variable is required? In that: no, I do not think that we need to perform any check beyond whether or not an environment variable of the specified name exists.

sebastianbergmann:
> should I only test `$_ENV` or maybe `$_ENV ?? getenv()`?

I almost never work with environment variables, so bear with me please: is there a situation where an environment variable is not accessible through `$_ENV` but is accessible through `getenv()` or vice versa?

nikophil:
>  I do not think that we need to perform any check beyond whether or not an environment variable of the specified name exists

Yes, that's what I've done.
The only check, I've made is that the env var is not an empty string, as [suggested)(https://github.com/sebastianbergmann/phpunit/pull/6074#discussion_r1875987867) by @mvorisek 

> is there a situation where an environment variable is not accessible through $_ENV but is accessible through getenv() or vice versa?

I think when it is manually set, an env var can be added to one or both: if only added to $_ENV, it is not accessible with getenv() and vice versa. [`symfony/dotenv` does not even use `putenv()` by default](https://github.com/symfony/dotenv/blob/7.2/Dotenv.php#L57:68)

PR #6081, opened 2024-12-11
`DefaultResultCache::mergeWith()` for merging result cache instances

### LINKED ISSUE #6078: Make `DefaultResultCache` being able to merge another instance of itself
Hello, in ParaTest we need to be able to run subsets of the Test Suite and then merge the results in a single output.
`\PHPUnit\TestRunner\TestResult\TestResult` does this job flawlessly, but it only covers the tests' output.

The current `\PHPUnit\Runner\ResultCache\DefaultResultCache` doesn't allow ParaTest to run different jobs each with its own cache file and then merge them all together.

I'm here to ask you to implement a way to do that.

Few questions you may answer:

1. Q: What does the merge functionality should do if the `version` differs between the two instances?
    A: Don't care, I'm fine to raise an Exception
2. Q: What does the merge functionality should do if multiple `defects` or `times` exist for the same `$id`?
    A: Don't care, it won't happen in ParaTest, I'm fine to raise an Exception

Slamdunk:
Fix https://github.com/sebastianbergmann/phpunit/issues/6078

### DIFF
--- a/src/Runner/ResultCache/DefaultResultCache.php
+++ b/src/Runner/ResultCache/DefaultResultCache.php
@@ -87,6 +87,17 @@ final class DefaultResultCache implements ResultCache
         return $this->times[$id] ?? 0.0;
     }
 
+    public function mergeWith(self $other): void
+    {
+        foreach ($other->defects as $id => $defect) {
+            $this->defects[$id] = $defect;
+        }
+
+        foreach ($other->times as $id => $time) {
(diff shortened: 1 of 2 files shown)


sebastianbergmann:
Can you send this for `10.5`? It should also "just work" there and if it does, I'll merge it from there to `11.5` and `main`. Thanks!

Slamdunk:
Done :sunglasses:

PR #6084, opened 2024-12-14
Fix PHPStan level 7 errors

staabm:
fixing low hanging fruits. not all errors are resolved yet

Some more are fixed with https://github.com/phpstan/phpstan-src/pull/3731


src/Metadata/Api/Requirements.php
```diff
@@ -72,7 +72,7 @@ public function requirementsNotSatisfiedFor(string $className, string $methodNam
 
                 if (!extension_loaded($metadata->extension()) ||
                     ($metadata->hasVersionRequirement() &&
-                        !$metadata->versionRequirement()->isSatisfiedBy(phpversion($metadata->extension())))) {
+                        !$metadata->versionRequirement()->isSatisfiedBy(phpversion($metadata->extension()) ?: ''))) {
```
staabm, line 75: extensions might not provide version info and this line can return false

src/Metadata/Version/Requirement.php
```diff
@@ -23,7 +23,7 @@
  */
 abstract readonly class Requirement
 {
-    private const VERSION_COMPARISON = '/(?P<operator>[<>=!]{0,2})\s*(?P<version>[\d\.-]+(dev|(RC|alpha|beta)[\d\.])?)[ \t]*\r?$/m';
+    private const VERSION_COMPARISON = "/(?P<operator>!=|<|<=|<>|=|==|>|>=)?\s*(?P<version>[\d\.-]+(dev|(RC|alpha|beta)[\d\.])?)[ \t]*\r?$/m";
```
staabm, line 26: operators copied from ComparisonRequirement constructor

src/Framework/MockObject/Generator/Generator.php
```diff
@@ -89,7 +89,7 @@ final class Generator
      *
      * @param class-string            $type
      * @param ?list<non-empty-string> $methods
-     * @param list<mixed>             $arguments
+     * @param array<mixed>            $arguments
```
staabm, line 92: Named args cannot be a list

PR #6112, opened 2025-01-29
Improve performance of `SourceMapper`

dantleech:
Improve source map performance for directories with same path

This PR aggregates the directories to a map of paths to tuples containing prefixes and suffixes.

This will reduce the overhead of scanning paths when multiple directory elements are defined with the same path but differnet prefixes or suffixes.

For example:

```xml
<exclude>
    <directory suffix="Controller.php">src/</directory>
    <directory suffix="Factory.php">src/</directory>
    <directory suffix="Bus.php">src/</directory>
    <directory suffix="Car.php">src/</directory>
</exclude>
```

Would be resolved as:

```php
[
   'src/' => [
       [], // no prefixes
       ['Controller.php', 'Factory.php', 'Bus.php', 'Car.php'],
   ],
],
```

The File Iterator already accepts an array of suffixes or prefixes.

---

Imroves perforamance in the [example repository](https://github.com/dantleech/phpunit-6111) from 3.5 to 1.8 seconds.

### DIFF
--- a/src/TextUI/Configuration/SourceMapper.php
+++ b/src/TextUI/Configuration/SourceMapper.php
@@ -40,8 +40,10 @@ final class SourceMapper
 
         $files = [];
 
-        foreach ($source->includeDirectories() as $directory) {
-            foreach ((new FileIteratorFacade)->getFilesAsArray($directory->path(), $directory->suffix(), $directory->prefix()) as $file) {
+        $directories = $this->aggregateDirectories($source->includeDirectories());
+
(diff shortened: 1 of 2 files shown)


sebastianbergmann:
This could, and probably should, target the `11.5` branch instead of `main`.

dantleech:
I've rebased, squashed and retargeted at 11.5 - I also added a few more tests as we were missing ones with multiple prefixes/suffixes on the same path.

PR #6118, opened 2025-02-05
Add `expectErrorLog()` for expecting `error_log()` output

### LINKED ISSUE #2155: expectOutputString and error_log
I've been banging my head against tests for a class involving error_log for a bit, and I'm starting to wonder if I'm running into some sort of output buffer bug in PHPUnit.  PHPUnit appears not to be capturing error_log output as I would expect.  Here's a quick-and-dirty test I wrote.

ObTest.php:

``` php
<?php
namespace Some\Test;

class ObTest
{
    public function writeToErrorLog($message)
    {
        error_log($message);
    }
}
```

The corresponding test:

``` php
<?php

namespace Some\Test;

include "ObTest.php";

class ObTestTest extends \PHPUnit_Framework_TestCase
{
    public static $error_log_config;

    public static function setUpBeforeClass()
    {
        self::$error_log_config = ini_get("error_log");

        ini_set("error_log", "/dev/stdout");
    }

    public static function tearDownAfterClass()
    {
        ini_set("error_log", self::$error_log_config);
    }

    public function testWriteToErrorLogOutputs()
    {
        $ob_test = new ObTest();

        $this->expectOutputString("Test");

        $ob_test->writeToErrorLog("Test");
    }
}
```

And the output:

```
PHPUnit 5.4-g9b5b99b by Sebastian Bergmann and contributors.

[06-May-2016 01:05:51 UTC] Test

Failed asserting that two strings are equal.
Expected :'Test'
Actual   :''
```

I've also tried setting error_log to php://output thinking that the issue had something to do with stdout not getting captured in PHPUnit's output buffering at all; PHPUnit still returns the empty string as "Actual", which is ultimately what led me to believe this might be a bug.

I get the same result with PHPUnit 5.3.2, and in general I'm running PHP 5.6.20-1+deb.sury.org~trusty+1

staabm:
summary:

- when a test which invokes `error_log()` without a call to `expectsErrorLog()`, the output is directly rendered in-between the summary (as tested in https://github.com/sebastianbergmann/phpunit/pull/6127)
- when a test which invokes `error_log()` with a call to `expectsErrorLog()`, no more output is rendered in-between the test summary printing. when no `error_log()` is invoked the test is considered failed.

----

backstory:

while looking into https://github.com/sebastianbergmann/phpunit/issues/2155#issuecomment-2636149476 I realized PHPunit is already somehow capturing the `error_log()` output.

I think this means we don't need a workaround like https://github.com/sebastianbergmann/phpunit/issues/2155#issuecomment-584749044 in phpunit itself, but just a API which allows us to make assertions on the already captured output.

DX wise I think there is also some room for improvement since the `error_log`'ged string just shows up in the test-debug output without any indication where this string is coming from (the test-author needs to look into the code beeing tested to realize that this string originates from a `error_log`) or how to make assertions on it.

With this PR I am just sending a test so we can see the status quo and discuss where to move from here

closes https://github.com/sebastianbergmann/phpunit/issues/2155


staabm:
I tried debugging it a bit more to find out how phpunit knows what the unit-under-test passed to `error_log`.

```
PHPUnit Started (PHPUnit 12.0-gd39bf28f4 using PHP 8.3.16 (cli) on Darwin)
Test Runner Configured
Bootstrap Finished (/Users/m.staab/Documents/GitHub/phpunit/tests/bootstrap.php)
Event Facade Sealed
Test Suite Loaded (1 test)
Test Runner Started
Test Suite Sorted
Test Runner Execution Started (1 test)
Test Suite Started (PHPUnit\TestFixture\Issue2155\Issue2155Test, 1 test)
Test Preparation Started (PHPUnit\TestFixture\Issue2155\Issue2155Test::testOne)
Test Prepared (PHPUnit\TestFixture\Issue2155\Issue2155Test::testOne)
logged a side effect
Test Passed (PHPUnit\TestFixture\Issue2155\Issue2155Test::testOne)
Test Finished (PHPUnit\TestFixture\Issue2155\Issue2155Test::testOne)
Test Suite Finished (PHPUnit\TestFixture\Issue2155\Issue2155Test, 1 test)
Test Runner Execution Finished
Test Runner Finished
PHPUnit Finished (Shell Exit Code: 0)
```

I can see the string also be mentioned in the event log stream of `--debug`.
@localheinz could you give me a hint how I can find out where this `logged a side effect` string is coming from?

staabm:
ohh I think I understand the problem now. `error_log` is just writing into stderr. its not a event happening and beeing rendered by the event system

tests/end-to-end/regression/2155/Issue2155Test.php
```diff
@@ -0,0 +1,34 @@
...
+}
+
+final class Issue2155Test extends TestCase
+{
+    public function testOne(): void
+    {
+        $foo = new Foo;
+
+        $this->assertSame('', $foo->doFoo());
+        self::expectErrorOutputRegex('/logged a side effect/');
```
staabm, line 32: the string written into `error_log()` will get a date format prefix. therefore I have only implemented `expectErrorOutputRegex` as because of the ever changing date format, adding expectatings without a regex will not work.

tests/end-to-end/regression/2155/Issue2155Test.php
```diff
@@ -0,0 +1,34 @@
...
+namespace PHPUnit\TestFixture\Issue2155;
+
+use function error_log;
+use PHPUnit\Framework\TestCase;
+
+class Foo
+{
+    public function doFoo()
+    {
+        error_log('logged a side effect');
```
staabm, line 19: with the current implementation a test-case which itself fiddles with `ini_set('error_log', ...); ` would render the in this PR added expectation handling useless. I am not sure we can make this more robust, or whether thats good enough.

sebastianbergmann: I do not have time at the moment to look into this, especially as I consider this a rather niche use case. I will look into this, though. Thank you for your work so far!

staabm: Btw: the main case for such new api is, that testing code which calls error_log() currently is considered risky.

The main goal is not asserting  expectations on the error_log()'ed string.

Maybe something like `expectsErrorLog()` or similar would be a even simpler api

sebastianbergmann: Sounds reasonable.

staabm:
I think we are in better shape now

src/Framework/TestCase.php
```diff
@@ -1007,6 +1014,11 @@ final protected function expectOutputString(string $expectedString): void
         $this->outputExpectedString = $expectedString;
     }
 
+    final protected function expectsErrorLog(): void
```
sebastianbergmann, line 1017: This should be `expectErrorLog()` to be consistent with `expectOutputString`, for example.

tests/end-to-end/regression/2155-no-error_log.phpt
```diff
@@ -0,0 +1,31 @@
+--TEST--
```
sebastianbergmann, line 1: `expectErrorLog()` is a new feature. Therefore, we should not put its end-to-end tests into the `regression` directory. The `generic` would be acceptable for now.

staabm:
great catch, fixed it. thank you!

sebastianbergmann:
Merged manually, thanks.

src/Framework/TestCase.php
```diff
@@ -1104,8 +1116,20 @@ final protected function runTest(): mixed
     {
         $testArguments = array_merge($this->data, array_values($this->dependencyInput));
 
+        $capture          = tmpfile();
+        $errorLogPrevious = ini_set('error_log', stream_get_meta_data($capture)['uri']);
```
mvorisek, line 1120: I saw your php-src feature request to support in memory stream. I think stream wrapper would work here - https://www.php.net/manual/en/stream.streamwrapper.example-1.php

mvorisek: Also creating tmp. file for every testcase is far from ideal.

staabm: feel free to improve the implementation with a followup PR

mvorisek: > I think stream wrapper would work here - https://www.php.net/manual/en/stream.streamwrapper.example-1.php

sadly even this is currently not possible - https://github.com/php/php-src/issues/18530

PR #6126, opened 2025-02-09
feat: create attribute `#[WithEnvironmentVariable]`

### LINKED ISSUE #6116: Provide `#[WithEnvironmentVariable()]` attribute to set env variables
Hello,

would you accept a PR with a new attribute `#[WithEnv('name', 'value')]` or `#[WithEnvironmentVariable('name', 'value')]`, which will set the given env var and run the test with this env?

This would allow to fully test applications which rely on environment variables for feature flags, for example. Relying on environment variables is quite common these days...

thanks for your answer

nikophil:
fixes https://github.com/sebastianbergmann/phpunit/issues/6116

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -9,6 +9,7 @@
  */
 namespace PHPUnit\Framework;
 
+use PHPUnit\Runner\BackupEnvironmentVariable;
 use const PHP_EOL;
 use function array_keys;
 use function array_merge;
@@ -134,7 +135,7 @@ abstract class TestCase extends Assert implements Reorderable, SelfDescribing, T
     private null|int|string $expectedExceptionCode  = null;
 
     /**
-     * @var array<string, false|string>
+     * @var list<BackupEnvironmentVariable>
      */
     private array $backupEnvironmentVariables = [];
 
@@ -1768,7 +1769,7 @@ abstract class TestCase extends Assert implements Reorderable, SelfDescribing, T
         }
 
         foreach ($environmentVariables as $environmentVariableName => $environmentVariableValue) {
-            $this->backupEnvironmentVariables[$environmentVariableName] = $_ENV[$environmentVariableName] ?? getenv($environmentVariableName);
+            $this->backupEnvironmentVariables = [...$this->backupEnvironmentVariables, ...BackupEnvironmentVariable::create($environmentVariableName)];
 
             if ($environmentVariableValue === null) {
                 unset($_ENV[$environmentVariableName]);
@@ -1782,14 +1783,8 @@ abstract class TestCase extends Assert implements Reorderable, SelfDescribing, T
 
     private function restoreEnvironmentVariables(): void
     {
-        foreach ($this->backupEnvironmentVariables as $environmentVariableName => $value) {
-            if ($value === false) {
-                unset($_ENV[$environmentVariableName]);
-                putenv($environmentVariableName);
-            } else {
-                $_ENV[$environmentVariableName] = $value;
-                putenv("{$environmentVariableName}={$value}");
-            }
+        foreach ($this->backupEnvironmentVariables as $backupEnvironmentVariable) {
+            $backupEnvironmentVariable->restore();
         }
 
         $this->backupEnvironmentVariables = [];
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -23,6 +23,7 @@ use function defined;
 use function error_clear_last;
 use function explode;
 use function getcwd;
+use function getenv;
 use function implode;
 use function in_array;
 use function is_array;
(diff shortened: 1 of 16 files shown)


src/Framework/TestCase.php
```diff
@@ -130,6 +133,11 @@ abstract class TestCase extends Assert implements Reorderable, SelfDescribing, T
     private ?string $expectedExceptionMessageRegExp = null;
     private null|int|string $expectedExceptionCode  = null;
 
+    /**
+     * @var array<string, false|string>
```
nikophil, line 137: false means that the env var was not set

src/Framework/TestCase.php
```diff
@@ -1741,6 +1751,48 @@ private function compareGlobalStateSnapshotPart(array $before, array $after, str
...
+        foreach ($withEnvironmentVariables as $metadata) {
+            assert($metadata instanceof WithEnvironmentVariable);
+
+            $environmentVariables[$metadata->environmentVariableName()] = $metadata->value();
+        }
+
+        foreach ($environmentVariables as $environmentVariableName => $environmentVariableValue) {
+            assert($metadata instanceof WithEnvironmentVariable);
+
+            $this->backupEnvironmentVariables[$environmentVariableName] = $_ENV[$environmentVariableName] ?? getenv($environmentVariableName);
```
nikophil, line 1769: I was torn on this... in PHP, nothing prevents to have a different value for `$_ENV['VAR']` and `getenv('VAR')`.
Do you think I should store both representations, in order to have 100% the same environment afterand before the test?

tests/end-to-end/_files/with_environment_variable/SomeTest.php
```diff
@@ -0,0 +1,124 @@
...
+        $this->assertSame('foo', getenv('FOO'));
+
+        $this->assertSame('bar', $_ENV['BAR']);
+        $this->assertSame('bar', getenv('BAR'));
+    }
+
+    #[WithEnvironmentVariable('BAZ', '1')]
+    #[WithEnvironmentVariable('BAZ', '2')]
+    #[WithEnvironmentVariable('BAZ', '3')]
+    public function testMultipleAttributesKeepTheLastValue(): void
```
nikophil, line 111: maybe we should throw an exception for this case?

tests/end-to-end/_files/with_environment_variable/WithEnvironmentVariableHookTest.php
```diff
@@ -0,0 +1,66 @@
...
+
+    protected function tearDown(): void
+    {
+        $this->assertEnvironmentVariablesHaveCustomValues();
+    }
+
+    public static function tearDownAfterClass(): void
+    {
+        self::assertEnvironmentVariablesHaveDefaultValues();
+    }
```
nikophil, line 41: not dure of what would be the desired behavior here :thinking:

src/Framework/TestCase.php
sebastianbergmann: I think so, yes.

nikophil: any hints how I should store this?
my first guess would be to store this in an array:
```php
    /**
     * @var array{getenv: array<string, false|string>, global: array<string, false|string>}
     */
    private array $backupEnvironmentVariables = ['getenv' => [], 'global'  => []];
```

but I don't like it. Another idea would be to use a value object:
```php
final readonly class BackupEnvVar
{
    public static function asGetenv(): self;
    public static function asSuperGlobal(): self;    
}
```

nikophil: I've updated the code to use a value object. (some tests are failing, I'll check them after we agree on the implementation)

tests/end-to-end/_files/with_environment_variable/WithEnvironmentVariableTest.php
```diff
@@ -0,0 +1,124 @@
...
+        $this->assertSame('3', getenv('BAZ'));
+    }
+
+    #[WithEnvironmentVariable('BAZ', '1')]
+    #[RequiresEnvironmentVariable('BAZ', '1')]
+    public function testUsingAlongWithRequiresEnvironmentVariableAttribute(): void
+    {
+        $this->assertSame('1', $_ENV['BAZ']);
+        $this->assertSame('1', getenv('BAZ'));
+    }
```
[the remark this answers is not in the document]
sebastianbergmann: Ideally, PHPUnit would emit a warning when `WithEnvironmentVariable` and `RequiresEnvironmentVariable` are 1) used on the same test and 2) refer to the same environment variable. But this can be done in a future PR.

sebastianbergmann:
Merged manually, thanks.

nikophil:
hey @sebastianbergmann 

> Merged manually, thanks.

cool, thanks :)

out of curiosity: why do you sometimes close the MR and merge manually?

sebastianbergmann:
In this case I did it manually because I wanted to fix the build 😄

nikophil:
haha sorry for this, I was waiting for your approval about the implementation before fixing the tests :sweat_smile: 

anyway, I'm glad this was merged :+1:

PR #6127, opened 2025-02-09
Regression test for error_log()

staabm:
test the current behaviour when `error_log` is used in a test


tests/end-to-end/regression/2155-no-expects-log.phpt
```diff
@@ -0,0 +1,26 @@
...
+
+require_once __DIR__ . '/../../bootstrap.php';
+
+(new PHPUnit\TextUI\Application)->run($_SERVER['argv']);
+--EXPECTF--
+PHPUnit %s by Sebastian Bergmann and contributors.
+
+Runtime: %s
+
+logged a side effect
```
staabm, line 18: output of `error_log` gets rendered into the test summary

Issue #6132, opened 2025-02-17
`Stub::method` returns `InvocationStubber` interface, which does not have `with` method


medunrad:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 12.0.2
| PHP version         | 8.3.3
| Installation Method | Composer

#### Summary

`Stub::method` returns `InvocationStubber` interface, which does not have `with` method.

If I take code from official presentation: https://thephp.cc/presentations/testing-with-doubles-why-when-and-how.

```php
$writer = $this->createMock(EventWriter::class);
$writer->expects($this->once())
  ->method('write')
  ->with($event);
```

I am calling method `with`, which is not defined on `InvocationStubber`

#### Current behavior

If I use PHPStan on code above, I will get an error:  `Call to an undefined method PHPUnit\Framework\MockObject\Builder\InvocationStubber::with().` And I agree with it.

#### How to reproduce

```php
<?php

namespace Foo;

class Test {

	public function foo(int $id): int {
		return $id;
	}

}
```

```php
<?php

namespace Foo;

use PHPUnit\Framework\TestCase;

class FooTest extends TestCase {

	public function testFoo(): void {
		$mock = $this->createMock(Test::class);
		$mock->method('foo')->with(1);
	}

}
```

#### Expected behavior

No error.

sebastianbergmann:
I cannot reproduce what you are trying to report:

#### `src/Service.php`
```php
<?php declare(strict_types=1);
namespace PHPUnit\TestFixture\Issue6132;

interface Service
{
    public function doSomething(): void;
}
```
#### `tests/Issue6132Test.php`
```php
<?php declare(strict_types=1);
namespace PHPUnit\TestFixture\Issue6132;

use PHPUnit\Framework\Attributes\CoversNothing;
use PHPUnit\Framework\Attributes\Small;
use PHPUnit\Framework\TestCase;

#[CoversNothing]
#[Small]
final class Issue6132Test extends TestCase
{
    public function testOne(): void
    {
        $service = $this->createMock(Service::class);

        $service
            ->expects($this->once())
            ->method('doSomething');

        $service->doSomething();
    }
}
```
#### `phpunit.xml`
```xml
<?xml version="1.0" encoding="UTF-8"?>
<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:noNamespaceSchemaLocation="vendor/phpunit/phpunit/phpunit.xsd"
[...]

Issue #5416, opened 2023-06-20
Remove support for doubling interfaces (or classes) that have a method named `method`


sebastianbergmann:
Follows #5415.

PR #6137, opened 2025-02-21
Fix MockObject::method() return type

tscni:
https://github.com/sebastianbergmann/phpunit/issues/5416 / https://github.com/sebastianbergmann/phpunit/commit/98cc1116a3ddc8ca2bd9d3e503a2795dfb06009b removed the `MockObject::method()` pseudo method but didn't add an actual implemention on `MockObject`. A call to `MockObject::method()` thus only guarantees an `InvocationStubber` return as declared on `Stub::method()`.

With that, method mocking like this is technically not correct anymore:
```php
$this->createMock('SomeClass')
    ->method('call') // returns InvocationStubber
    ->with('foo') // only available on InvocationMocker, not InvocationStubber
    ->willReturn('bar');
```

and instead a `->expects($this->any())` is technically required in the chain to keep the existing behavior:
```php
$this->createMock('SomeClass')
    ->expects($this->any()) // returns InvocationMocker
    ->method('call') // InvocationMocker::method() returns InvocationMocker
    ->with('foo')
    ->willReturn('bar');
```

This seems like an unintended requirement(?), hence the PR.

(relates to the closed issue https://github.com/sebastianbergmann/phpunit/issues/6132)

### DIFF
--- a/src/Framework/MockObject/Runtime/Interface/MockObject.php
+++ b/src/Framework/MockObject/Runtime/Interface/MockObject.php
@@ -9,8 +9,10 @@
  */
 namespace PHPUnit\Framework\MockObject;
 
+use PHPUnit\Framework\Constraint\Constraint;
 use PHPUnit\Framework\MockObject\Builder\InvocationMocker;
 use PHPUnit\Framework\MockObject\Rule\InvocationOrder;
+use PHPUnit\Framework\MockObject\Runtime\PropertyHook;
 
 /**
  * @no-named-arguments Parameter names are not covered by the backward compatibility promise for PHPUnit
@@ -18,4 +20,6 @@ use PHPUnit\Framework\MockObject\Rule\InvocationOrder;
 interface MockObject extends Stub
 {
     public function expects(InvocationOrder $invocationRule): InvocationMocker;
+
+    public function method(Constraint|PropertyHook|string $constraint): InvocationMocker;
 }


sebastianbergmann:
Your code

```php
$this->createMock('SomeClass')
    ->method('call')
    ->with('foo')
    ->willReturn('bar');
```

does not make sense.

Consider this complete example

```php
namespace PHPUnit\TestFixture\Issue6137;

interface AnInterface
{
    public function doSomething(string $s): bool;
}

final class Issue6137Test extends TestCase
{
    {
        $o = $this
            ->createMock(AnInterface::class)
            ->method('doSomething')
            ->with('string')
            ->willReturn(true);

        $this->assertTrue($o->doSomething('string'));
    }
}
```

Running the code above will yield the output shown below:

```
❯ ./vendor/bin/phpunit tests
PHPUnit 12.0.4 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.4.4

E                                                                   1 / 1 (100%)

Time: 00:00.010, Memory: 8.00 MB

There was 1 error:

1) PHPUnit\TestFixture\Issue6137\Issue6137Test::testOne
Error: Call to undefined method PHPUnit\Framework\MockObject\Builder\InvocationMocker::doSomething()

/home/sb/issue-6137/tests/Issue6137Test.php:21

ERRORS!
Tests: 1, Assertions: 0, Errors: 1.
```

You must assign the result of test double creation methods such as `createStub()` or `createMock()` to a variable. This variable is your test double. You must start using the fluent API for configuration on this variable. The object returned by the fluent API's configuration methods such as `method()`, `with()`, or `willReturn()` are not your test double.

Changing the code to ...

```php
namespace PHPUnit\TestFixture\Issue6137;

interface AnInterface
{
    public function doSomething(string $s): bool;
}

final class Issue6137Test extends TestCase
{
    {
        $o = $this->createMock(AnInterface::class);

        $o
            ->method('doSomething')
            ->with('string')
            ->willReturn(true);

        $this->assertTrue($o->doSomething('string'));
    }
}
```

... makes it work, but not more meaningful or sensible as you create a mock object using `createMock()` but then do not configure any expectations.

Issue #6114, opened 2025-01-30
Reimplement `SourceFilter::includes()` to not depend on `SourceMapper` and to not perform filesystem traversal


sebastianbergmann:
Originally suggested by @dantleech in https://github.com/sebastianbergmann/phpunit/issues/6111#issuecomment-2620981394.

I think we should focus on `SourceFilter::includes()` first. Once we have a good solution for that we can look into extracting and reusing it for https://github.com/sebastianbergmann/php-code-coverage/issues/386.

dantleech:
Just as a heads up, I've started working on this and am currently mapping out the current behavior in tests. There _is_ a third-paty package that may work or provide a basis for an implementation: https://github.com/webmozarts/glob but not yet sure if the bejavior will be the same.

sebastianbergmann:
Superseded by #6493.

PR #6146, opened 2025-03-10
GH-6114: Static path matching

dantleech:
This (early stage) WIP PR introduces a static path matcher which intends to emulate the behavior of the PHPUnit `FileIterator` in order to prevent PHPUnit traversing the filesystem when a [deprecation is triggered](https://github.com/sebastianbergmann/phpunit/issues/6114).

The PHPUnit FileIterator uses `glob` to find directories and we therefore need to support the glob patterns - which can [vary](https://www.php.net/manual/en/function.glob.php) according to the platform. This PR uses https://man7.org/linux/man-pages/man7/glob.7.html as a reference in addition to testing the behavior locally to confirm assumptions.

The [webmozart/glob](https://github.com/webmozarts/glob/blob/4.8.x/src/Glob.php#L313) provides a similar feature however it's behavior is different as it supports curly braces, and `*` is restricted to a single directory level, while `*` in PHPUnit will return all descendants and I'm sure there are other differences - however I've used that as a starting point.

TODO:

- [x] Escaping unhandled regex characters, or escaping the regex before parsing the pattern.
- [x] Character classes `[:alnum:]` etc.
- [ ] Collating symbols
- [ ] Equivalence class expressions
- [x] Emulating `glob` behavior of unterminated `[` character groups.
- [ ] Seems to be a PHPUnit globstar bug whereby `/a**` will match `/b` and all other directories, where as `/ab*` will not match anything. We can either copy that behavior or "fix" it.

~and maybe writing the implementation from scratch if regex turns out to be a bad fit.~

Usages on Github:

- wildcard: 1.8k https://github.com/search?q=directory+%22*%22+path%3A*.xml+path%3A**%2Fphpunit.xml+language%3AXML&type=code
- globstar: 378: https://github.com/search?q=directory+%22**%22+path%3A*.xml+path%3A**%2Fphpunit.xml+language%3AXML&type=code
- Character range/class (`directory.*\[`): 0: https://github.com/search?q=%2Fdirectory.*%5C%5B%2F+path%3A*.xml+path%3A**%2Fphpunit.xml+language%3AXML&type=code
- `?` any char: 0 https://github.com/search?q=%2Fdirectory.*%5C%5B%2F+path%3A*.xml+path%3A**%2Fphpunit.xml+language%3AXML&type=code

### DIFF
--- a/src/TextUI/Configuration/SourceFilter.php
+++ b/src/TextUI/Configuration/SourceFilter.php
@@ -9,6 +9,7 @@
  */
 namespace PHPUnit\TextUI\Configuration;
 
+use PHPUnit\Util\FileMatcherPattern;
 use function array_map;
 use PHPUnit\Util\FileMatcher;
 use PHPUnit\Util\FileMatcherRegex;
@@ -52,11 +53,11 @@ final class SourceFilter
         $this->source                  = $source;
         $this->includeDirectoryRegexes = array_map(static function (FilterDirectory $directory)
         {
-            return FileMatcher::toRegEx($directory->path());
+            return FileMatcher::toRegEx(new FileMatcherPattern($directory->path()));
         }, $source->includeDirectories()->asArray());
         $this->excludeDirectoryRegexes = array_map(static function (FilterDirectory $directory)
         {
-            return FileMatcher::toRegEx($directory->path());
+            return FileMatcher::toRegEx(new FileMatcherPattern($directory->path()));
         }, $source->excludeDirectories()->asArray());
     }
 
--- a/src/TextUI/Configuration/SourceFilter.php
+++ b/src/TextUI/Configuration/SourceFilter.php
@@ -9,6 +9,10 @@
  */
 namespace PHPUnit\TextUI\Configuration;
 
+use PHPUnit\Util\FileMatcher;
+use PHPUnit\Util\FileMatcherRegex;
+
+
 /**
  * @no-named-arguments Parameter names are not covered by the backward compatibility promise for PHPUnit
  *
@@ -18,19 +22,23 @@ final class SourceFilter
 {
     private static ?self $instance = null;
 
+    private Source $source;
+
     /**
-     * @var array<non-empty-string, true>
+     * @var list<FileMatcherRegex>
      */
-    private readonly array $map;
+    private array $includeDirectoryRegexes;
+
+    /**
+     * @var list<FileMatcherRegex>
+     */
+    private array $excludeDirectoryRegexes;
 
     public static function instance(): self
     {
         if (self::$instance === null) {
-            self::$instance = new self(
-                (new SourceMapper)->map(
-                    Registry::get()->source(),
-                ),
-            );
+            $source = Registry::get()->source();
+            return new self($source);
         }
 
         return self::$instance;
@@ -39,13 +47,21 @@ final class SourceFilter
     /**
      * @param array<non-empty-string, true> $map
      */
-    public function __construct(array $map)
+    public function __construct(Source $source)
     {
-        $this->map = $map;
+        $this->source = $source;
+        $this->includeDirectoryRegexes = array_map(function (FilterDirectory $directory) {
+            return FileMatcher::toRegEx($directory->path());
+        }, $source->includeDirectories()->asArray());
+        $this->excludeDirectoryRegexes = array_map(function (FilterDirectory $directory) {
+            return FileMatcher::toRegEx($directory->path());
+        }, $source->excludeDirectories()->asArray());
     }
 
     public function includes(string $path): bool
     {
+        foreach ($this->source->includeDirectories() as $directory) {
+        }
         return isset($this->map[$path]);
     }
 }
--- a/src/TextUI/Configuration/SourceFilter.php
+++ b/src/TextUI/Configuration/SourceFilter.php
@@ -9,11 +9,13 @@
  */
 namespace PHPUnit\TextUI\Configuration;
 
+use function array_map;
 use PHPUnit\Util\FileMatcher;
 use PHPUnit\Util\FileMatcherRegex;
 
-
 /**
+ * TODO: Does not take into account suffixes and prefixes - and tests don't cover it.
+ *
  * @no-named-arguments Parameter names are not covered by the backward compatibility promise for PHPUnit
  *
  * @internal This class is not covered by the backward compatibility promise for PHPUnit
@@ -21,7 +23,6 @@ use PHPUnit\Util\FileMatcherRegex;
 final class SourceFilter
 {
     private static ?self $instance = null;
-
     private Source $source;
 
     /**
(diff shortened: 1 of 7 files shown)


src/Util/FileMatcherPattern.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+namespace PHPUnit\Util;
+
+class FileMatcherPattern
+{
+    public function __construct(public string $path)
```
dantleech, line 7: this will/could include the `$suffix`, `$prefix`  and `$exclude` also

dantleech:
Have refactored to tokenize the glob string, while less performant it's easier to reason about and we only need to compile the regex for each `<include` or `<exclude` - not for each path.

src/Util/FileMatcher.php
```diff
@@ -0,0 +1,260 @@
...
+                }
+
+                // otherwise it's a harmless literal
+                $resolved[] = [self::T_CHAR, ':' . $class];
+            }
+
+            if ($bracketOpen === true && $type === self::T_BRACKET_OPEN) {
+                // if bracket is already open, interpret everything as a
+                // literal char
+                $resolved[] = [self::T_CHAR, $char];
```
dantleech, line 230: rename T_CHAR => T_LITERAL?

src/Util/FileMatcherPattern.php
dantleech: or could simply remove this and handle the prefix/suffix independently of the "file matcher"

src/Util/FileMatcher.php
```diff
@@ -0,0 +1,260 @@
...
+            }
+
+            // normal globstar
+            if (
+                $type === self::T_SLASH &&
+                ($tokens[$offset + 1][0] ?? null) === self::T_ASTERIX && ($tokens[$offset + 2][0] ?? null) === self::T_ASTERIX && ($tokens[$offset + 3][0] ?? null) === self::T_SLASH
+            ) {
+                $resolved[] = [self::T_GLOBSTAR, '**'];
+
+                // we eat the two `*` in addition to the slash
```
dantleech, line 153: we eat two * AND the trailing slash

src/Util/FileMatcher.php
```diff
@@ -0,0 +1,297 @@
...
+                // TODO: this is not tested
+                $bracketOpen = false;
+
+                array_pop($brackets);
+                $resolved[] = [$type, $char];
+
+                continue;
+            }
+
+            $resolved[] = [$type, $char];
```
dantleech, line 287: this handles any "left over tokens" - including `T_ASTERISK` - maybe all tokens should be handled explicitly?

src/Util/FileMatcher.php
```diff
@@ -0,0 +1,297 @@
...
+    private static function tokenize(string $glob): array
+    {
+        $length = strlen($glob);
+
+        $tokens = [];
+
+        for ($i = 0; $i < $length; $i++) {
+            $c = $glob[$i];
+
+            $tokens[] = match ($c) {
```
dantleech, line 117: could inline `$c`

src/Util/FileMatcher.php
```diff
@@ -0,0 +1,293 @@
...
+
+                // match any segment up until the next directory separator
+                self::T_ASTERIX         => '[^/]*',
+                self::T_GREEDY_GLOBSTAR => '.*',
+                self::T_GLOBSTAR        => '/([^/]+/)*',
+                self::T_BRACKET_OPEN    => '[',
+                self::T_BRACKET_CLOSE   => ']',
+                self::T_HYPHEN          => '-',
+                self::T_COLON           => ':',
+                self::T_BACKSLASH       => '\\',
```
dantleech, line 92: COLON and BACKSLASH are not tested

dantleech:
@sebastianbergmann this PR is indicative of the approach but needs the final 20% of work:

- Refactor the source filter/mapper tests to create the paths they need in a temporary directory? this would make it easier to test edge cases without contaminating other tests and grant mode confidence.
- Implement the exclude/include logic (either by compiling that into the regex or as a separate concern).
- We can implement or forbid the use of `[[:character-ranges:]]` and other "esoteric" glob patterns.

But I'm keen for any thought you have on the approach so far - the biggest concern for me is providing confidence that it will be compatible with at least 99.99% of configuration files.

src/TextUI/Configuration/SourceFilter.php
```diff
@@ -9,43 +9,86 @@
...
+        $included = false;
+
+        foreach ($this->source->includeFiles() as $file) {
+            if ($file->path() === $path) {
+                $included = true;
+            }
+        }
+
+        foreach ($this->includeDirectoryRegexes as $directoryRegex) {
+            if ($directoryRegex->matches($path)) {
```
dantleech, line 75: this will need to change to account for include/exclude matching on the `basename` either in the regex or otherwise.

sebastianbergmann:
> @sebastianbergmann this PR is indicative of the approach but needs the final 20% of work:
> 
>     * Refactor the source filter/mapper tests to create the paths they need in a temporary directory? this would make it easier to test edge cases without contaminating other tests and grant mode confidence.
> 
>     * Implement the exclude/include logic (either by compiling that into the regex or as a separate concern).
> 
>     * We can implement or forbid the use of `[[:character-ranges:]]` and other "esoteric" glob patterns.

Thank you for your work so far. I am afraid that I might not be able to look into this before the PHPUnit Code Sprint in April in Berlin. There I will be able to look into this, for sure, and together with other developers. I hope I can look into this before then, but I cannot promise that.

> But I'm keen for any thought you have on the approach so far - the biggest concern for me is providing confidence that it will be compatible with at least 99.99% of configuration files.

We could delay this until PHPUnit 13 and break backward compatibility (for edge cases, "esoteric" glob patterns, etc.).

sebastianbergmann:
Is it expected that the tests are currently failing?

dantleech:
Yes - the prefix/suffix filtering hasn't been added yet and the tests fail due to the tests we added upstream. The PR is at an inflection point:

- Do we want to commit to supprting all the glob features? or could we simplify it.
- Do we want to refactor the source-finding/filtering tests to be more dynamic?[1]
- Are we happy with the _general_ approach in this PR?

And then I was also considering how to implement the suffix/prefix logic - as it probably makes sense to include them in the compiled regex - but then it intrudes on the "copy the behavior of glob" logic - so if we decide to simplify the globbing behavior it makes that decision easier too.

[1] currently all the source mapping/filtering tests share the same fixture which makes it hard to test edge cases without breaking other tests - I think it could generate the file tree instead.

sebastianbergmann:
I discussed this today with @edorian, @theseer, and @sebastianheuer at the PHPUnit Code Sprint. Even for more complex `<source>` configurations, we were unable to measure a runtime of more than 150 ms for the current implementation of `SourceMapper::map()`. We do not think that a potential reduction in runtime would justify a more complicated implementation.

Can you share how much time is spent in `SourceMapper::map()` in your environment and what the `<source>` section of your configuration looks like?

dantleech:
Hi @sebastianbergmann  - this is the [repoducer repository](https://github.com/dantleech/phpunit-6111) - on my local envrionment it pauses for around 3-4 seconds while loading the source map.

sebastianbergmann:
> Hi @sebastianbergmann - this is the [repoducer repository](https://github.com/dantleech/phpunit-6111) - on my local envrionment it pauses for around 3-4 seconds while loading the source map.

Thank you. How much does the time spent go down with your new implementation?

dantleech:
filtering without file traversal basically removes the performance issue. the only issue really is replicating the `glob` behavior - or deliberately breaking compatiblity and simplifying the behavior to some extent.

sebastianbergmann:
Thank you. I have to say, though, that I do not think that 3-4 seconds is too bad considering there are 120K files in the reproducer. I am now less convinced than I originally was that this is an issue worth fixing, sorry.

dantleech:
On the project for which I reported this issue it takes 1 minute and 22 seconds. We've had to remove the source mapping entintirely in order to be able to use PHPUnit. I will endeavour to update the reproduction repository to replicate that.

TomA-R:
I thought I'd share that this is also a problem for others who are running PHPUnit on relatively large codebases. We have ~9.2k files and executing [this loop](https://github.com/sebastianbergmann/phpunit/blob/e1a44ec6f048cb7b8c7ff288f5aef135275c2e81/src/TextUI/Configuration/SourceMapper.php#L69-L83) takes ~8 seconds on my machine the first time E_USER_DEPRECATED is emitted. It's not the biggest issue in the world for us and the workaround is to fix all deprecation warnings (easier said than done) but addressing this would really improve our quality of life. Thank you for your consideration!

dantleech:
@sebastianbergmann and I had a call and decided to simplify the approach in the PR to support the bare minumum glob syntax and that it would be part of the next major release and merged around December time.

sebastianbergmann:
Once we have the new implementation, we could introduce it into the current stable version in addition to the old implemention. When old and new implementation "disagree" we could then emit a deprecation warning.

sebastianbergmann:
@dantleech Did you have a chance to work on this yet? Thanks!

dantleech:
hi, no not yet, need to find some time

sebastianbergmann:
@dantleech FYI: `main` is not PHPUnit 13.0-dev (where these changes should land). This hopefully makes working on this easier/simpler/more convenient.

dantleech:
would you be option to using a third-party package, namely [webmozarts/glob](https://github.com/webmozarts/glob)? We could evaluate it with that implementation at least I guess.

sebastianbergmann:
Sure, we can evaluate @webmozart's package. I am usually reluctant to add a new dependency (that I did not create / do not maintain myself), but if this package provides what we need and is maintained on the PHP versions we need then I do not see a problem.

dantleech:
Replaced by #6398

PR #6148, opened 2025-03-14
SourceFilter test coverage

dantleech:
This MR aims to improve the test coverage for the SourceFilter in anticipation of #6146 

- Extracts an `AbstractSouceFilterTestCase` as both SourceMapper and SourceFilter have the same requirements.
- Refactor test to use factory methods to reduce boilerplate.
- Introduces tests for wildcards, globstars and question marks (i.e.single char wildcard)

The tests are _far_ from exhaustive as we are testing the behavior of `glob` in order to be able to avoid calling it in the future notably we are not testing character classes and all the variants therein - and it's an open question if PHPUnit should support that going forward? A quick scan of github doesn't reveal anybody making uses of character classes.

### DIFF
--- a/tests/unit/TextUI/SourceFilterTest.php
+++ b/tests/unit/TextUI/SourceFilterTest.php
@@ -9,256 +9,422 @@
  */
 namespace PHPUnit\TextUI\Configuration;
 
-use const DIRECTORY_SEPARATOR;
-use function realpath;
+use function json_encode;
+use function sprintf;
 use PHPUnit\Framework\Attributes\CoversClass;
 use PHPUnit\Framework\Attributes\DataProvider;
 use PHPUnit\Framework\Attributes\Small;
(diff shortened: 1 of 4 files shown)


tests/unit/TextUI/SourceFilterTest.php
```diff
@@ -9,256 +9,422 @@
...
                                 '.php',
                             ),
                         ],
                     ),
-                    FileCollection::fromArray([]),
-                    FilterDirectoryCollection::fromArray([]),
-                    FileCollection::fromArray(
+                ),
+            ],
+            'globstar with any more than a single char prefix does not include sibling files' => [
```
dantleech, line 332: this behavior is very strange. `/<any char>**` will be equivilent to `/*` but `<more than 1 char>**` will not

Issue #6162, opened 2025-03-25
PHPUnit 12 badly evaluates time for very first test


niconoe-:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 12.0.x
| PHP version         | 8.3.x
| Installation Method | Composer

#### Summary

When running PHPUnit 12, the very first test is reported to have a time-consumption really higher than with PHPUnit 11 on JUnit and cache-files reports. I bet this might be caused by an unwanted including of tests preparations before the very first test plays, but I can't find any location in source code so I wasn't able to provide a PR about it.

#### Current behavior

The "time" value on JUnit report is currently not the actual execution time for the very first test only.

On PHPUnit 12.0.10 :
```xml
<?xml version="1.0" encoding="UTF-8"?>
<testsuites>
  <testsuite name="/data/www/phpunit.xml" tests="3" assertions="3" errors="0" failures="0" skipped="0" time="0.452475">
    <testsuite name="myproject" tests="3" assertions="3" errors="0" failures="0" skipped="0" time="0.452475">
      <testsuite name="MyProject\Tests\FooBarTest" file="/data/www/src/Tests/FooBarTest.php" tests="3" assertions="3" errors="0" failures="0" skipped="0" time="0.452475">
        <testsuite name="MyProject\Tests\FooBarTest::testTrue" tests="3" assertions="3" errors="0" failures="0" skipped="0" time="0.452475">
          <testcase name="testTrue with data set &quot;test-case #1&quot;" file="/data/www/src/Tests/FooBarTest.php" line="17" class="MyProject\Tests\FooBarTest" classname="MyProject.Tests.FooBarTest" assertions="1" time="0.445116"/>
          <testcase name="testTrue with data set &quot;test-case #2&quot;" file="/data/www/src/Tests/FooBarTest.php" line="17" class="MyProject\Tests\FooBarTest" classname="MyProject.Tests.FooBarTest" assertions="1" time="0.003829"/>
          <testcase name="testTrue with data set &quot;test-case #3&quot;" file="/data/www/src/Tests/FooBarTest.php" line="17" class="MyProject\Tests\FooBarTest" classname="MyProject.Tests.FooBarTest" assertions="1" time="0.003530"/>
        </testsuite>
      </testsuite>
    </testsuite>
  </testsuite>
</testsuites>
```

On PHPUnit 11.5.15 :
```xml
<?xml version="1.0" encoding="UTF-8"?>
<testsuites>
[...]

PR #6167, opened 2025-03-30
Fix timing issue

marac19901990:
#6162


sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it as I do not believe this to be the right solution.

PR #6169, opened 2025-03-31
Bugfix/dependency fix

### LINKED ISSUE #6104: Test with dependencies and data provider fails
| Q                   | A
| PHPUnit version     | 11.5.2
| PHP version         | 8.2.18

#### Summary

This is a follow-up to #5827 which was closed for a specific case ("mixture of associative and non-associative arrays") but the originally reported issue wasn't answered.

#### Current behavior

Test cases having both a dependency and a data provider don't work anymore in PHPUnit 11 using the attribute syntax. The same specification worked fine in PHPUnit 10 using the annotation syntax.

#### How to reproduce

Here's a minimal example with two test functions. The second one depends on the first one and uses a data provider. Running the test yields this error:
```
1) dependencyTest::testTwo with data set "case 1" ('')
Error: Cannot use positional argument after named argument during unpacking
```

Code example:
```php
<?php

class dependencyTest extends \PHPUnit\Framework\TestCase
{
    {
        $this->assertEmpty([]);
    }

    #[\PHPUnit\Framework\Attributes\Depends('testOne')]
    #[\PHPUnit\Framework\Attributes\DataProvider('dataProvider')]
    public function testTwo(string $example): void
    {
        $this->assertEmpty($example);
    }

    public static function dataProvider(): array
    {
        return [
            'case 1' => [
                'example' => ''
            ]
        ];
    }
}
```

#### Expected behavior

I'd expect this to work because the data provider uses an associative array, i.e. the (single) argument *IS* a named one.
(Note: If I remove the dependency specification, the test runs successfully.)

marac19901990:
fixes #6104

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -1235,9 +1235,19 @@ abstract class TestCase extends Assert implements Reorderable, SelfDescribing, T
     final protected function runTest(): mixed
     {
         $testArguments = array_merge($this->data, array_values($this->dependencyInput));
+        $positionalArgs = [];
+        $namedArgs = [];
+
+        foreach ($testArguments as $key => $value) {
+            if (is_int($key)) {
+                $positionalArgs[] = $value;
+            } else {
+                $namedArgs[$key] = $value;
+            }
+        }
 
         try {
-            $testResult = $this->{$this->methodName}(...$testArguments);
+            $testResult = $this->{$this->methodName}(...$namedArgs, ...$positionalArgs);
         } catch (Throwable $exception) {
             if (!$this->shouldExceptionExpectationsBeVerified($exception)) {
                 throw $exception;
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -1234,9 +1234,9 @@ abstract class TestCase extends Assert implements Reorderable, SelfDescribing, T
      */
     final protected function runTest(): mixed
     {
-        $testArguments = array_merge($this->data, array_values($this->dependencyInput));
+        $testArguments  = array_merge($this->data, array_values($this->dependencyInput));
         $positionalArgs = [];
-        $namedArgs = [];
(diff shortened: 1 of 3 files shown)


sebastianbergmann:
Merged manually, thanks.

sebastianbergmann:
@seggewiss I think you might be right. I will revert this change and release new versions of PHPUnit 11.5 and PHPUnit 12.1 ASAP. Then we can look into a better fix for the problem @marac19901990 wanted to address. Thank you for letting us know!

marac19901990:
@seggewiss I will take a look also

seggewiss:
@marac19901990 Sure thing 👍 
The follwing code fails with the error that the argument $i is not provided anymore.

```php
public static function provideSlots(): array
{
    return [
        [
            0,
            'name' => 'left',
            'type' => 'manufacturer-logo',
            'config' => [
                'displayMode' => [
                    'source' => 'static',
                    'value' => 'cover',
                ],
            ],
            'position' => 0,
        ],
    ];
}

#[DataProvider('provideSlots')]
public function testSlotsFromXml(int $i, string $name, string $type, array $config, int $position): void {}
```

marac19901990:
```
    public static function provideSlots(): array
    {
        return [
            [
                0,
                'name' => 'left',
                'type' => 'manufacturer-logo',
                'config' => [
                    'displayMode' => [
                        'source' => 'static',
                        'value' => 'cover',
                    ],
                ],
                'position' => 0,
            ],
        ];
    }

    #[DataProvider('provideSlots')]
    public function testSlotsFromXml(string $name, string $type, array $config, int $position, int $i): void {}
```
This works because the index arguments are know after the named ones.
It is a breaking change unfortunatelly :(

seggewiss:
@marac19901990 That's exactly what I thought. This should also be documented if you do it as it's quiet the opposite of what I would expect.

Issue #5718, opened 2024-03-01
Bring back `--repeat` CLI option


sebastianbergmann:
IIRC, @epdenouden wants to look into this during next week's code sprint.

curkan:
Yes please return

epdenouden:
Returning does nobody any good when I'm still `void`

Nutickets:
Hell to the yeah

ERuban:
So then?)

Bilge:
Want to see this backported for v10 also.

PerryRylance:
We often use factories during testing, which in turn user faker and friends. I reached for this when running tests that deal with models that may or may not appear in a collection based on date, I understand we can restructure our tests (and this is what I was in the process of doing) - but it would be useful to have this back.

benascbr:
make phpunit great again!

jdmaguire:
This is a great library and I love it a lot.

At my company we were waiting for this CLI option could come back. We've tried to make our tests as non-flaky as possible but there are some old tests that seem inherently flaky (we can decrease the flakiness of these but not eliminate them).

We recently found https://github.com/liuggio/fastest. It has an -r / --rerun-failed option to rerun failed tests (it first runs all tests, then runs failed tests). We've not been using it for long but we've been liking it thus far. We have this in our composer script so this is as easy as running `composer fastest`:

```
"scripts": {
        "fastest": "find . -name '*Tests.php' -not -path './vendor/*' | ./vendor/liuggio/fastest/fastest -v -o -r -p 8 'vendor/phpunit/phpunit/phpunit {};'"
}
```

sebastianbergmann:
@theseer, @localheinz, @staabm, @Schrank, @sebastianheuer, @tesla91, and I discussed this issue during the PHPUnit Code Sprint in Munich today.

Way back when I originally implemented the old `--repeat` CLI option, my use case had nothing to do with uncovering flaky tests. It was purely about benchmarking. In hindsight, I consider it a mistake to have implemented this feature as benchmarking is out of scope for PHPUnit.

While working towards PHPUnit 10, there was a point in time when the `--repeat` CLI option was blocking the release. As the feature was not important to me, and I was not aware that others were relying on it for other use cases than the one I originally had in mind, I decided to remove it from PHPUnit 10.

[...]

PR #6182, opened 2025-04-13
`#[Retry]` attribute to support retrying flaky test

santysisi:
# ✨ Add Retry Attribute for Retrying Flaky Tests
This PR introduces a new Retry attribute that can be applied to individual test methods in PHPUnit. Its goal is to help deal with flaky or unstable tests that occasionally fail due to non-deterministic reasons (e.g. network issues, race conditions, external dependencies).

### 🔧 How it works
The Retry attribute allows a test to be automatically retried a specified number of times before being marked as failed. You can also control the delay between retries, and optionally restrict retries to specific exception types.

### 🧪 Usage
Here’s how you can use it:
```php
#[Retry(3)] // Retry up to 3 times on any exception
public function testFlakyApi(): void
{
    // test logic
}
```
You can also specify a delay (in seconds) between retries:
```php
#[Retry(2, delay: 2)] // Retry twice, with 2 seconds delay between attempts
```
And restrict retries to certain exception types only:
```php
#[Retry(4, retryOn: TimeoutException::class)]
```
All parameters:
* maxRetries (int) — how many times the test can be retried.
* delay (?int) — seconds to wait before retrying. Default: 0
* retryOn (?string) — exception type to retry on. If omitted, retries on any throwable.

### ✅ Example
```php
#[Retry(2)]
public function testRetriesUntilMaxAttemptsThenSucceeds(): void
{
    static $attempt = 0;

    if ($attempt++ < 2) {
        throw new RuntimeException('Transient failure');
    }

    $this->assertSame(3, $attempt);
}
```
### 📈 Benefits
* Stabilize CI pipelines by automatically retrying flaky tests
* Fine-grained control over retry behavior per test
* Non-intrusive: no need to change existing test logic
* Optional support for retrying only on specific exceptions
### 🔄 Prior art
Other popular testing frameworks already provide similar functionality:
* [NUnit: Retry Attribute](https://docs.nunit.org/articles/nunit/writing-tests/attributes/retry.html)
* [JUnit Pioneer: RetryingTest](https://javadoc.io/doc/org.junit-pioneer/junit-pioneer/1.5.0/org/junitpioneer/jupiter/RetryingTest.html)

### DIFF
--- a/src/Framework/Attributes/Retry.php
+++ b/src/Framework/Attributes/Retry.php
@@ -1,0 +1,57 @@
+<?php declare(strict_types=1);
+/*
+ * This file is part of PHPUnit.
+ *
+ * (c) Sebastian Bergmann <sebastian@phpunit.de>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+namespace PHPUnit\Framework\Attributes;
+
+use Attribute;
+
+/**
+ * @immutable
+ *
+ * @no-named-arguments Parameter names are not covered by the backward compatibility promise for PHPUnit
+ */
+#[Attribute(Attribute::TARGET_METHOD | Attribute::IS_REPEATABLE)]
+final readonly class Retry
+{
+    private int $maxRetries;
+    private int $delay;
+
+    /**
+     * @var ?non-empty-string
+     */
+    private ?string $retryOn;
+
+    /**
+     * @param ?non-empty-string $retryOn
+     */
+    public function __construct(int $maxRetries, ?int $delay = 0, ?string $retryOn = null)
+    {
+        $this->maxRetries = $maxRetries;
+        $this->delay      = $delay;
+        $this->retryOn    = $retryOn;
+    }
+
+    public function maxRetries(): int
+    {
+        return $this->maxRetries;
+    }
+
+    public function delay(): int
+    {
+        return $this->delay;
+    }
+
+    /**
+     * @return ?non-empty-string
+     */
+    public function retryOn(): ?string
+    {
+        return $this->retryOn;
+    }
+}
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -46,6 +46,7 @@ use function restore_error_handler;
 use function restore_exception_handler;
 use function set_error_handler;
 use function set_exception_handler;
+use function sleep;
 use function sprintf;
 use function str_contains;
 use function stream_get_contents;
@@ -79,6 +80,7 @@ use PHPUnit\Metadata\Api\Groups;
 use PHPUnit\Metadata\Api\HookMethods;
 use PHPUnit\Metadata\Api\Requirements;
 use PHPUnit\Metadata\Parser\Registry as MetadataRegistry;
+use PHPUnit\Metadata\Retry;
 use PHPUnit\Metadata\WithEnvironmentVariable;
 use PHPUnit\Runner\BackedUpEnvironmentVariable;
 use PHPUnit\Runner\DeprecationCollector\Facade as DeprecationCollector;
@@ -1248,7 +1250,7 @@ abstract class TestCase extends Assert implements Reorderable, SelfDescribing, T
      * @throws ExpectationFailedException
      * @throws Throwable
      */
-    private function runTest(): mixed
+    private function runTest(int $attempt = 0): mixed
     {
         $testArguments = array_merge($this->data, array_values($this->dependencyInput));
 
@@ -1276,6 +1278,12 @@ abstract class TestCase extends Assert implements Reorderable, SelfDescribing, T
             }
 
             if (!$this->shouldExceptionExpectationsBeVerified($exception)) {
+                $metadata = $this->getRetryMetadata($exception, $attempt);
+
+                if (null !== $metadata) {
+                    return $this->retryTest($metadata, $attempt);
+                }
+
                 throw $exception;
             }
 
@@ -2253,6 +2261,28 @@ abstract class TestCase extends Assert implements Reorderable, SelfDescribing, T
         if (isset($trace[0]['class']) && $trace[0]['class'] === InvokedCount::class) {
             $this->numberOfAssertionsPerformed++;
         }
+    }
+
+    private function getRetryMetadata(Throwable $th, int $attempt): ?Retry
+    {
+        foreach (MetadataRegistry::parser()->forMethod($this::class, $this->name())->isRetry() as $metadata) {
+            assert($metadata instanceof Retry);
+
+            if ($metadata->maxRetries() > $attempt && (null === $metadata->retryOn() || $th instanceof ($metadata->retryOn()))) {
+                return $metadata;
+            }
+        }
+
+        return null;
+    }
+
+    private function retryTest(Retry $metadata, int $attempt): mixed
+    {
+        if ($metadata->delay() > 0) {
+            sleep($metadata->delay());
+        }
+
+        return $this->runTest(++$attempt);
     }
 
     /**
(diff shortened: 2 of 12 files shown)


santysisi:
This PR could also offer a complementary approach to the discussion in [#5718: Bring back --repeat CLI option](https://github.com/sebastianbergmann/phpunit/issues/5718).

While `--repeat` focuses on re-running tests via the CLI, the #[Retry] attribute is designed for more fine-grained control, enabling retries at the test method level, optionally with delay or exception-based filtering. It might be a useful alternative for handling flaky tests directly in code, without requiring changes to CI configs or CLI commands.

Thanks again, looking forward to your feedback! 😊

src/Framework/TestCase.php
```diff
@@ -1276,6 +1278,12 @@ private function runTest(): mixed
             }
 
             if (!$this->shouldExceptionExpectationsBeVerified($exception)) {
+                $metadata = $this->getRetryMetadata($exception, $attempt);
```
sebastianbergmann, line 1281: Only had a very cursory look at the proposed changes yet, but this caught my eye: how are tests retried that use `expectException()`, for example?

src/Framework/TestCase.php
```diff
@@ -2255,6 +2263,34 @@ private function handleExceptionFromInvokedCountMockObjectRule(Throwable $t): vo
         }
     }
 
+    private function getRetryMetadata(Throwable $th, int $attempt): ?Retry
+    {
+        if (MetadataRegistry::parser()->forMethod($this::class, $this->name())->isRetry()->isEmpty()) {
+            return null;
+        }
+
+        $metadatas = MetadataRegistry::parser()->forMethod($this::class, $this->name())->isRetry()->getIterator();
```
sebastianbergmann, line 2272: The manual call to `getIterator()` is superfluous as `MetadataCollection` implements `IteratorAggregate`.

santysisi: Thanks again for your feedback! I think everything's in order now. If you need me to tweak anything else, please let me know.

src/Framework/TestCase.php
santysisi: That said, I'm not entirely sure if this approach fully aligns with your expectations or the overall goals of PHPUnit. If there's anything I'm not comfortable with or could be improved, I'd greatly appreciate your guidance. I'd be happy to review the logic and make adjustments to better fit PHPUnit's design and intent.

nikophil:
Hey @santysisi 

Thanks for this PR! Any chance that you add the CLI option along with the attribute?

sebastianbergmann:
> Any chance that you add the CLI option along with the attribute?

I do not see how a CLI option would be relevant in the context of what this PR is about.

sebastianbergmann:
I think that the functionality you propose here is useful.

However, the current implementation can "only" be considered an early proof of concept in my opinion. Retried tests are not run on a fresh instance of the test case object, the before-test, after-test, etc. template methods are not run for retried tests, etc. Furthermore, no events are emitted to record the fact that a test was retried. At least as far as I can see.

While I do think this would be a nice addition, I will not work on a "real" implementation myself. I am open to a more complete PR, of course.

santysisi:
Hi @sebastianbergmann , thanks for your comment :smile:.  I really appreciate it!

I'd like to give it another try, now taking the new requirements into account. Please let me know if I’m on the right track:
* **Fresh Test Case Instances for Retries**, allowing the proper execution of lifecycle/template methods (setUp, tearDown, etc.).
* **Emitting Events for Retried Tests**, so retries can be logged or tracked appropriately.

Does that align with what you had in mind?

Also, just to confirm, do you agree with the behavior described [here](https://github.com/sebastianbergmann/phpunit/pull/6182#discussion_r2042084712)

src/Framework/Attributes/Retry.php
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * @immutable
+ *
+ * @no-named-arguments Parameter names are not covered by the backward compatibility promise for PHPUnit
+ */
+#[Attribute(Attribute::TARGET_METHOD | Attribute::IS_REPEATABLE)]
+final readonly class Retry
+{
+    private int $maxRetries;
+    private int $delay;
```
mvorisek, line 23: I would expect this to be of type `float` to support delays like `0.5` (500 ms).

And usually this is very bad retry strategy. Exponential delay is much better. Maybe there should be no native delayed retry support as it can slowdown testing massively.

santysisi: You're right, but instead of passing a `float`, maybe it's a better idea to replace `sleep` with `usleep`. As for the performance issue, I think the solution is simply not to pass that parameter. Still, I believe the delay is a good idea since it's optional.

mvorisek: In SI units definitely, ie. seconds.

santysisi: I believe using milliseconds instead of seconds is a better and more flexible option, especially when it comes to controlling timeouts and delays more precisely. This approach is also more commonly adopted, for example, both [Symfony's HTTP client](https://symfony.com/blog/new-in-symfony-5-2-retryable-http-client) and [Guzzle](https://github.com/guzzle/guzzle/blob/7.9/src/Middleware.php#L174-L175) use milliseconds for time-related settings.

Additionally, we could consider adding a multiplier option (similar to Symfony) to allow increasing the delay between each retry. Personally, I prefer Guzzle's approach of using a callable for this logic. However, I understand that passing closures directly in PHP attributes isn't natively supported, 

Happy to hear your thoughts on this!

santysisi:
> Hi @sebastianbergmann , thanks for your comment 😄. I really appreciate it!
> 
> I'd like to give it another try, now taking the new requirements into account. Please let me know if I’m on the right track:
> 
> * **Fresh Test Case Instances for Retries**, allowing the proper execution of lifecycle/template methods (setUp, tearDown, etc.).
> * **Emitting Events for Retried Tests**, so retries can be logged or tracked appropriately.
> 
> Does that align with what you had in mind?
> 
> Also, just to confirm, do you agree with the behavior described [here](https://github.com/sebastianbergmann/phpunit/pull/6182#discussion_r2042084712)

Hi @sebastianbergmann, I hope you're doing well 🙂
Could you please confirm if this aligns with what you had in mind, or if I might be going in the wrong direction? 🙏
Thanks in advance! 😊

sebastianbergmann:
> Could you please confirm if this aligns with what you had in mind, or if I might be going in the wrong direction? 🙏

I am sorry that I have not gotten around to look into this yet.

tests/unit/Framework/Attributes/RetryTest.php
```diff
@@ -0,0 +1,108 @@
...
+ * file that was distributed with this source code.
+ */
+namespace PHPUnit\Framework\Attributes;
+
+use DateTime;
+use Exception;
+use LogicException;
+use PHPUnit\Framework\TestCase;
+
+final class RetryTest extends TestCase
```
mvorisek, line 17: Currently, it seems pre/post test hooks are not called. But that can be an issue.

I would like to have this tested by E2E tests to show the behaviour and also combined with other features like process isolation.

src/Framework/Attributes/Retry.php
mvorisek: Milliseconds are not base SI units.

Still, I am thinking if a delay is wanted to be supported natively.

I think there can be an elegant option - the "try number" can be accessible from `setUp` method and the use can wait/sleep the way he needs by any code.

sebastianbergmann:
Thank you for your contribution. I appreciate the time you spent preparing this pull request. However, I have decided not to merge it. I am planning to work on this myself and have already started researching the DX perspective and implementation approach, but unfortunately I do not have anything concrete to share at the moment. As soon as I am able, I will organise my thoughts and notes, and share my plan in a new issue.

PR #6191, opened 2025-04-26
Improve failed phpt test diff

sebastianfeldmann:
In case of multiline placeholder %a or %A we try to match multiple lines at the same time.

The tricky part is to keep the expected output and the real output in sync since they can be different in size and amount of lines.


sebastianbergmann:
I think this is a bug fix and as such should go into the `11.5` branch, sorry. But: thank you!

PR #6196, opened 2025-04-30
fix: PHPStan detecting type correctly after usage of Assert::assertContainsOnly*

mitelg:
Hey there :wave: 

I added more `phpstan-assert` annotation to the `Assert` class and to the standalone functions. Those cover the `assertContainsOnly**` methods.


mitelg:
thanks for merging :+1:

PR #6210, opened 2025-05-19
Set default Clover coverage project name

Staubiii:
Some Clover XML report processing tools require the project name to be set.

e.g. https://issues.jenkins.io/browse/JENKINS-75648

Could this be backported to phpunit 8.5?

### DIFF
--- a/src/Runner/CodeCoverage.php
+++ b/src/Runner/CodeCoverage.php
@@ -287,7 +287,7 @@ final class CodeCoverage
 
             try {
                 $writer = new CloverReport;
-                $writer->process($this->codeCoverage(), $configuration->coverageClover());
+                $writer->process($this->codeCoverage(), $configuration->coverageClover(), 'Clover Coverage');
 
                 $this->codeCoverageGenerationSucceeded($printer);
 


sebastianbergmann:
If I look at https://bitbucket.org/atlassian/clover/raw/master/etc/schema/clover.xsd and understand ...
```xml
<xs:element name="project">
    <xs:annotation>
        <xs:documentation>
            Project metrics relating to non-test source.
            @name - project name (optional)
            @timestamp - seconds since UTC
        </xs:documentation>
    </xs:annotation>
    <xs:complexType>
        <xs:sequence>
            <xs:element name="metrics" type="projectMetrics"/>
            <xs:element maxOccurs="unbounded" ref="package"/>
        </xs:sequence>
        <xs:attribute name="name"/>
        <xs:attribute name="timestamp" use="required" type="xs:integer"/>
    </xs:complexType>
</xs:element>
```
... correctly then the `name` attribute is optional whereas the `timestamp` attribute is not. Interesting that your parser seems to choke on the optional attribute that is missing but not on the non-optional attribute that is missing.

sebastianbergmann:
I totally missed in my first comment that the Clover XML writer in fact does generate the `timestamp` attribute, my bad.

sebastianbergmann:
I am open to accepting the change you propose. I am reluctant to change PHPUnit 8.5, PHPUnit 9.6, and PHPUnit 10.5 as these versions have reached the end of their bugfix support. The proposed change should go into `11.5`, `12.1`, and `main`.

It would help me if somebody could look into whether https://bitbucket.org/atlassian/clover/raw/master/etc/schema/clover.xsd is, in fact, the official/canonical schema for Clover XML and, if it is, look into whether there are any other changes PHPUnit (or [phpunit/php-code-coverage](https://github.com/sebastianbergmann/php-code-coverage), rather) needs to implement to comply with the schema.

sebastianbergmann:
I cherry-picked this into `11.5` and merged it to `12.1` and `main` from there.

PR #6213, opened 2025-05-20
Be strict about data providers having redundant data

### LINKED ISSUE #6211: Warn about impossible data provider
We had a hidden issue due to the dataprovider not being run and its data tested

    #[DataProvider('falseyValues')]
    public function testRenderSerializeWithFalseyValue(): void
    {

So effectively, the test only ran with the default value in there.

Would it be feasable for the testing harness to issue a warning (e.g. risky), if the data provider attached is "unused"?
As it would make no sense and is usually a dev or merge error of sorts.

kubawerlos:
Closes https://github.com/sebastianbergmann/phpunit/issues/6211

- 1st commit is the feature itself
- 2nd commit is the changes to align the codebase to follow the new rule

Should this be hidden behind a config option?
Should it make the test risky/with a warning rather than producing error?

### DIFF
--- a/src/Metadata/Api/DataProvider.php
+++ b/src/Metadata/Api/DataProvider.php
@@ -11,6 +11,7 @@ namespace PHPUnit\Metadata\Api;
 
 use function array_key_exists;
 use function assert;
+use function count;
 use function get_debug_type;
 use function is_array;
 use function is_int;
@@ -23,6 +24,7 @@ use PHPUnit\Metadata\MetadataCollection;
 use PHPUnit\Metadata\Parser\Registry as MetadataRegistry;
 use PHPUnit\Metadata\TestWith;
 use ReflectionClass;
+use ReflectionMethod;
 use Throwable;
 
 /**
@@ -61,6 +63,8 @@ final readonly class DataProvider
             );
         }
 
+        $testMethodNumberOfParameters = (new ReflectionMethod($className, $methodName))->getNumberOfParameters();
+
         foreach ($data as $key => $value) {
             if (!is_array($value)) {
                 throw new InvalidDataProviderException(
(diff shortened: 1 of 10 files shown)


sebastianbergmann:
> Should this be hidden behind a config option?

No.

sebastianbergmann:
Thank you!

sebastianbergmann:
I have "downgraded" this from an error to a warning in 7f1969ad314d922b2d47c6f023e0ab77d1281c13.

kubawerlos:
Cool. The warning will be on the test case level, not at the test method level, as the error was.

sebastianbergmann:
The warning can be seen here: https://github.com/sebastianbergmann/phpunit/blob/main/tests/end-to-end/data-provider/too-many-arguments.phpt#L23

dereuromark:
This might have a small regression for ...
```php
    /**
     * tests failure messages for assertions
     *
     * @param string $assertion Assertion method
     * @param string $message Expected failure message
     * @param string $url URL to test
     * @param mixed ...$rest
     */
    #[DataProvider('assertionFailureMessagesProvider')]
    public function testAssertionFailureMessages($assertion, $message, $url, ...$rest): void
    {
        $this->expectException(AssertionFailedError::class);
        $this->expectExceptionMessage($message);

        Security::setSalt($this->key);

        $this->get($url);

        call_user_func_array($this->$assertion(...), $rest);
    }
```
If there is a ...$var maybe the warning should be skipped as it is intentional?
```
1) Cake\Test\TestCase\TestSuite\IntegrationTestTraitTest::testAssertionFailureMessages
* Data set "assertCookie" has more arguments (5) than the test method accepts (4)
```

dereuromark:
Maybe a
```
$parameters = $method->getParameters();

$hasVariadic = false;

foreach ($parameters as $param) {
    if ($param->isVariadic()) {
        $hasVariadic = true;
        break;
    }
}
```
And if true we just skip?

sebastianbergmann:
@dereuromark Please open a new issue if you think you found a bug/regression. Thank you.

kubawerlos:
It is regression, I've raised the PR with the fix: https://github.com/sebastianbergmann/phpunit/pull/6228

PR #6218, opened 2025-05-20
Do not create `ReflectionClass` object only to call `getMethod` on it

kubawerlos:
Fewer objects created means better performance, right?


staabm (commented):
Lgtm

PR #6225, opened 2025-06-03
Checkout the proper branch on CI to replicate local behavior

soyuka:
Because `actions/checkout` uses a detached HEAD, the code at InfrastructureInformationProvider (`symbolic-ref --short HEAD`) doesn't get the correct information. This patch makes sure we checkout the correct branch to replicate our local usage.

I've tested this at https://github.com/soyuka/phpunit/actions/runs/15427391478/job/43417670031.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -169,6 +169,17 @@ jobs:
 
       - name: Checkout
         uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
We also need this change for the `code-coverage` job. Thank you for working on this!

soyuka:
interesting code coverage was green on my fork's CI. I added the patch on the coverage as well.

PR #6231, opened 2025-06-07
Fix saving information about passed tests

### LINKED ISSUE #6222: DataProvider seems to mess up Depends
| Q                   | A
| PHPUnit version     | 12.1.4
| PHP version         | 8.4.7

#### Summary

A test with a dependency to another test, which has a data provider, is skipped, although the dependency succeeded. This happens if there is any other test case failing before.

#### Current behavior

The third test case in the example is skipped.

#### How to reproduce

<details><summary>Run this test class as a whole</summary>

```php
<?php

declare(strict_types=1);

use PHPUnit\Framework\Attributes\Depends;

class PhpunitDependencyTest extends TestCase {
  public static function getTestCases(): array {
    return [
      ['FC4DF99B8C1F4156A68528E0B21A105B'],
    ];
  }

  public function testOne(): void {
    self::assertTrue(false);
  }

  #[DataProvider('getTestCases')]
  public function testTwo(): void {
    self::assertTrue(true);
  }

  #[Depends('testTwo')]
  public function testThree(): void {
    self::assertTrue(true);
  }
}

```

</details>

#### Expected behavior

The third test case in the example should run and not be skipped.

kubawerlos:
Fixes https://github.com/sebastianbergmann/phpunit/issues/6222

Current behaviour:
- when the test - let's call it `Foo` - with the data provider finishes, it calls [`testSuiteFinished`](https://github.com/sebastianbergmann/phpunit/blob/12.2.1/src/Runner/TestResult/Collector.php#L265) method
- [if](https://github.com/sebastianbergmann/phpunit/blob/12.2.1/src/Runner/TestResult/Collector.php#L267) there is a failed test case in the class (might be completely unrelated to the test `Foo`, like in https://github.com/sebastianbergmann/phpunit/issues/6222) the information that the `Foo` passed will not [be saved](https://github.com/sebastianbergmann/phpunit/blob/12.2.1/src/Runner/TestResult/Collector.php#L285).
- the information that `Foo` passed is not saved, so the test using `Depends` on `Foo` will be skipped because [`hasTestMethodPassed`](https://github.com/sebastianbergmann/phpunit/blob/12.2.1/src/Runner/TestResult/PassedTests.php#L92) will return `false`.

Proposed behaviour:
 - after the test suite, i.e. test with data provider finishes, we check if we have an event for the failed case of that test - if there is none, that means the test (all of its cases) passed.

### DIFF
--- a/src/Runner/TestResult/Collector.php
+++ b/src/Runner/TestResult/Collector.php
@@ -57,11 +57,10 @@ use PHPUnit\TestRunner\TestResult\Issues\Issue;
 final class Collector
 {
     private readonly IssueFilter $issueFilter;
-    private int $numberOfTests                       = 0;
-    private int $numberOfTestsRun                    = 0;
-    private int $numberOfAssertions                  = 0;
-    private bool $prepared                           = false;
-    private bool $currentTestSuiteForTestClassFailed = false;
+    private int $numberOfTests      = 0;
+    private int $numberOfTestsRun   = 0;
+    private int $numberOfAssertions = 0;
+    private bool $prepared          = false;
 
     /**
      * @var non-negative-int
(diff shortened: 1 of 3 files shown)


sebastianbergmann:
Thank you! I cherry-picked your commits into `11.5` and merged them to `12.2` and `main` from there.

PR #6237, opened 2025-06-14
Do not run tests when code coverage analysis is requested but code coverage data cannot be collected

### LINKED ISSUE #6233: Restart PHP process when code coverage data collection/processing is requested, Xdebug is loaded, but its mode is wrong
when I run `vendor/bin/phpunit --coverage-html coverage_report` phpunit is running and happily is working thru the test-suite, while after waiting for minutes for its completion it reports:

```
There was 1 PHPUnit test runner warning:

1) XDEBUG_MODE=coverage (environment variable) or xdebug.mode=coverage (PHP configuration setting) has to be set
```

when the xdebug extension is already loaded, but in the wrong mode, couldn't [phpunit restart itself with a proper `XDEBUG_MODE=coverage`](https://github.com/composer/xdebug-handler) set, so I don't need to wait for the whole testsuite to finish and re-run it manually?

this would be useful, as I usually have xdebug loaded in debugging mode

staabm:
doing a process restart and persisting all possible cli args etc, while not just re-using composer/xdebug-handler and re-implementing everything is not a useful way and requires lots of code we would need to maintain.

therefore I figured I would just make sure that phpunit exits early instead of running all tests, when the coverage driver cannot be activated

closes https://github.com/sebastianbergmann/phpunit/issues/6233

### DIFF
--- a/src/Runner/CodeCoverageInitializationStatus.php
+++ b/src/Runner/CodeCoverageInitializationStatus.php
@@ -1,0 +1,17 @@
+<?php declare(strict_types=1);
+/*
+ * This file is part of PHPUnit.
+ *
+ * (c) Sebastian Bergmann <sebastian@phpunit.de>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+namespace PHPUnit\Runner;
+
+enum CodeCoverageInitializationStatus
+{
+    case NOT_REQUESTED;
+    case SUCCESSFULLY;
+    case FAILED;
+}
--- a/tests/end-to-end/cli/coverage/coverage-no-tests-when-wrong-xdebug-mode.phpt
+++ b/tests/end-to-end/cli/coverage/coverage-no-tests-when-wrong-xdebug-mode.phpt
@@ -1,0 +1,30 @@
+--TEST--
+Don't run tests when wrong xdebug mode is set
+--SKIPIF--
+<?php declare(strict_types=1);
+if (!extension_loaded('xdebug')) {
+    print 'skip: Extension xdebug must be loaded.';
+}
+--ENV--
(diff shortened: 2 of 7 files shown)


tests/end-to-end/cli/coverage/coverage-no-tests-when-wrong-xdebug-mode.phpt
```diff
@@ -0,0 +1,30 @@
...
+PHPUnit %s by Sebastian Bergmann and contributors.
+
+Runtime: %s
+Configuration: %s
+
+There was 1 PHPUnit test runner warning:
+
+1) XDEBUG_MODE=coverage (environment variable) or xdebug.mode=coverage (PHP configuration setting) has to be set
+
+No tests executed!
```
staabm: here we see we get the warning while no more tests are run -> immediate feedback

src/Runner/CodeCoverageInitializationStatus.php
```diff
@@ -0,0 +1,17 @@
...
+ * This file is part of PHPUnit.
+ *
+ * (c) Sebastian Bergmann <sebastian@phpunit.de>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+namespace PHPUnit\Runner;
+
+enum CodeCoverageInitializationStatus
```
staabm, line 12:
```suggestion
/**
 * @internal This class is not covered by the backward compatibility promise for PHPUnit
 */
enum CodeCoverageInitializationStatus
```

?

staabm:
should be good to go

sebastianbergmann:
Merged manually, thanks.

staabm:
Thanks 🙏

PR #6238, opened 2025-06-15
Add missing `@no-named-arguments` annotations and ensure there won't be new missing ones anymore

kubawerlos:
This PR:
 - Updates "Run PHP-CS-Fixer" step in CI
 - Adds custom fixers to tools
 - Updates PHP CS Fixer config
 - Fix the codebase according to the new config (by adding missing `@no-named-arguments` annotations)


sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I am reluctant to merge it as I am already not happy that I have to use Composer to install PHPStan in order to use PHPStan plugins and this introduces a similar approach for PHP-CS-Fixer.

kubawerlos:
> I am already not happy that I have to use Composer to install PHPStan

Is there something wrong with it? I thought (based on how PHPStan is installed and used on `main` branch) that the goal is to have all code in `tools`, so Composer is only needed to update tools and never to use them.

How would you like to have the tools installed then? With phive?

kubawerlos:
Anyway, I've finished this PR. Feel free to cherry-pick some commits (if any) and close it.

Let me know how I can help to automate PHPStan and PHP CS Fixer the way you want them to be.

Or maybe if there is no such plan yet, we could go with this PR up to `main` branch (as there are new files with missing `@no-named-arguments`) for now.

sebastianbergmann:
I am subscribed to https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/pull/8906 and will update this project's configuration for PHP-CS-Fixer if/when `PhpdocTagNoNamedArgumentsFixer` is available in a stable release of PHP-CS-Fixer.

PR #6242, opened 2025-06-18
New attribute TestWithArray

sartor:
New attribute `TestWithArray` introduced.
More compact version of existing `TestWith` for many cases.

Example:
```php
    #[TestWithArray([
        'odd' => [1, true],
        'even' => [2, false],
    ])]
    public function testNumbers($number, $isOdd): void
    {
        $this->assertSame($isOdd, $number % 2 === 1);
    }
```


sebastianbergmann:

Instead of adding yet another mechanism for providing data to a test, I would rather deprecate (and later remove) `#[TestWith]` and `#[TestWithJson]`.

sartor:
Ok. But there will be no way of adHoc data provider, such convenient minimalist way without copy pasting.
No my tests looks like:
```php
#[DataProvider('userIsBlockedDataProvider')]
public function testUserIsBlocked(bool $filterValue, array $userRow, bool $expectedResult): void
...

#[DataProvider('ageDataProvider')]
public function testAge(array $filterValue, array $userRow, bool $expectedResult): void
...
    
#[DataProvider('sitesDataProvider')]
public function testSites(array $filterValue, array $userRow, bool $expectedResult): void
...
```

Is there other way to not repeat myself with same names in test method and dataProvider?

PR #6248, opened 2025-06-25
Verbose error if temp directory is not writable

dmitryuk:
If tmpfile returned `false` where is an error that not easy to understand:
```
There was 1 error:

1) Tests\Test::testExample
TypeError: stream_get_meta_data(): Argument #1 ($stream) must be of type resource, false given
``` 
The PR fixes it to:
```
There was 1 failure:

1) Tests\ExampleTest::testExample
Failed to create tmp file. Directory "/tmp" is not writable.
```

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -50,6 +50,7 @@ use function sprintf;
 use function str_contains;
 use function stream_get_contents;
 use function stream_get_meta_data;
+use function sys_get_temp_dir;
 use function tmpfile;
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
I am reluctant to merge this (or similar changes) because I do not think that a development tool should support restricted environments where the system's temporary directory is not writable.

dmitryuk:
ok

dmitryuk:
ok

PR #6250, opened 2025-06-26
Show more meaningful error when `testsuite` is missing `name` attribute

kubawerlos:


sebastianbergmann:
The schema for PHPUnit configuration files [requires](https://github.com/sebastianbergmann/phpunit/blob/11.5.24/phpunit.xsd#L253) that the `<testsuite>` element has the `name` attribute.

The XML configuration file loader [validates](https://github.com/sebastianbergmann/phpunit/blob/11.5.24/src/TextUI/Configuration/Xml/Loader.php#L118) the XML configuration file that is loaded against this schema.

However, the XML configuration file loader does not error out when the XML configuration file that is loaded has validation errors. This is because we want to load as much valid information from the XML configuration file as possible and inform about possible automated migrations etc. later.

I will use the regression test from this pull request, but I will implement a different solution.

Thank you!

PR #6251, opened 2025-06-28
Fix test with data provider and dependency which does not return value

kubawerlos:
Fixes https://github.com/sebastianbergmann/phpunit/issues/6104

The problem is that currently in `PHPUnit\Framework\TestCase` both - test dependency returning `null` and test dependency not returning value - are treated the same, as technically they both "return" `null`.

As always, I've split the PR into multiple commits if you prefer a different fix, Sebastian.


sebastianbergmann:
Thank you!

PR #6255, opened 2025-06-30
Use pcov as coverage driver again

staabm:
reverts https://github.com/sebastianbergmann/phpunit/commit/6eb1bc9dec8f982a2bd014f1cd54b9814ecd1b96

as https://github.com/krakjoe/pcov/issues/59 was fixed in pcov and many [more improvements are currently in the making](https://github.com/krakjoe/pcov/commits/develop/), lets try to switch back to pcov (or at least uncover issues, which might get fixed as pcov is currently beeing worked on)

---

[support for property hooks](https://github.com/krakjoe/pcov/pull/121) was also recently implemented. the changes are not yet released though.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -211,7 +211,7 @@ jobs:
         uses: shivammathur/setup-php@v2
         with:
           php-version: 8.4
-          coverage: xdebug
+          coverage: pcov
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
I would like to stick to Xdebug for every project of mine with one exception: [php-code-coverage](https://github.com/sebastianbergmann/php-code-coverage/actions/runs/15730378093). This is tested using both PCOV and Xdebug, because this is the library with code specific to code coverage drivers.

staabm:
I see, thanks for the info.

PR #6257, opened 2025-07-03
Allow named arguments with InvocationStubber::with()

### LINKED ISSUE #6163: `@no-named-arguments` leads to static analysis errors for variadic arguments
| Q                   | A
| PHPUnit version     | at least version 9

#### Summary

The `InvocationMocker` class is annotated with `@no-named-arguments` - as this is on a class-level, it also affects the `with()` function, that only has a single variadic argument.  This leads to static analysis issues.

#### Current behavior

When using PHPStan in a project where `with()` is used with an array (not a list), this may lead to a static analysis error (although the keys are ignored anyways):

     Method […]\InvocationMocker::willReturn() invoked with unpacked array with possibly string key, but it's not allowed because of @no-named-arguments.

#### How to reproduce

```php
$input = ['a' => 'foo', 'b' => 'bar'];

$mock->with(...$input); // invalid
$mock->with(...array_values($input)); // valid

$mock->willReturn(/* … */);
```

#### Expected behavior

Both cases are a valid usage, so I'd expect there would be no static analysis errors.

greg0ire:
`with()` uses a unique, variadic parameter, and there should be no static analysis issue when calling it with named arguments, because those will not refer to the parameters of that method, but to the parameters of the method specified when calling `method()`.

Fixes #6163

### DIFF
--- a/src/Framework/MockObject/Runtime/Interface/InvocationStubber.php
+++ b/src/Framework/MockObject/Runtime/Interface/InvocationStubber.php
@@ -14,12 +14,11 @@ use PHPUnit\Framework\MockObject\Runtime\PropertyHook;
 use PHPUnit\Framework\MockObject\Stub\Stub;
 use Throwable;
 
-/**
- * @no-named-arguments Parameter names are not covered by the backward compatibility promise for PHPUnit
- */
 interface InvocationStubber
 {
     /**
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
Thank you!

sebastianbergmann:
I backported this to PHPUnit 11.5.

staabm:
Would be great to have a test for this, so we don't regress

Issue #6254, opened 2025-06-29
`defects,random`configuration is supported by implementation, but it is not allowed by the XML configuration file schema


staabm:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 11.5.24
| PHP version         | 8.3.22
| Installation Method | Composer / PHAR

#### Summary

a execution order of "defects,random" seems to be a useful choice, as it combines a very fast feedback loop when re-running previously failed tests while it also allows random test order to harden tests against ordering problems

#### Current behavior

using a phpunit.xml like

```
<?xml version="1.0" encoding="UTF-8"?>
<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:noNamespaceSchemaLocation="./vendor/phpunit/phpunit/phpunit.xsd"
         bootstrap="./vendor/autoload.php"
         colors="true"
         executionOrder="defects,random"
         failOnWarning="true"
         failOnRisky="true"
         requireCoverageMetadata="true"
         displayDetailsOnPhpunitDeprecations="true"
         displayDetailsOnTestsThatTriggerDeprecations="true"
         displayDetailsOnTestsThatTriggerWarnings="true"
         cacheDirectory="./build/cache/phpunit"
>
```

works like expected, but xml validation fails, since `defects,random` is not a valid combiation per XSD:
see https://github.com/sebastianbergmann/phpunit/blob/6c7855fff66d6e324d613d6a6f605801fb7c694b/phpunit.xsd#L117-L138

PHPStorm screenshot:

<img width="1077" alt="Image" src="https://github.com/user-attachments/assets/0ae89e66-e1f8-43a0-a837-bbd890dea5b0" />

#### How to reproduce

change phpunit.xml to 
`executionOrder="defects,random"`

#### Expected behavior

it should be a valid option to use `"defects,random"`

staabm:
hmm looking at it again, maybe it even does not order properly in this case and this might be more of a feature request.

sebastianbergmann:
Probably. To be honest, I do not know the test sorting functionality as well as I should. Every time I would like to look into it, something more important comes up. I would appreciate it if you could look into this. You know what you are doing :-), and you seem to be interested in this feature.

staabm:
I had a look into it and I think what I see in my testing is, that the order of data-provider tests is not ordered by defects.
[...]

PR #6258, opened 2025-07-04
Test defects,random executionOrder

staabm:
Refs https://github.com/sebastianbergmann/phpunit/issues/6254

### DIFF
--- a/phpunit.xsd
+++ b/phpunit.xsd
@@ -118,6 +118,7 @@
         <xs:restriction base="xs:string">
             <xs:enumeration value="default"/>
             <xs:enumeration value="defects"/>
+            <xs:enumeration value="defects,random"/>
             <xs:enumeration value="depends"/>
             <xs:enumeration value="depends,defects"/>
             <xs:enumeration value="depends,duration"/>
--- a/tests/unit/Runner/TestSuiteSorterTest.php
+++ b/tests/unit/Runner/TestSuiteSorterTest.php
@@ -503,6 +503,42 @@ final class TestSuiteSorterTest extends TestCase
         $this->assertSame($expectedOrder, $sorter->getExecutionOrder());
     }
 
+    public function testCanSetRandomizationWithDefectsFirst(): void
+    {
+        $cache = new DefaultResultCache;
+
+        $runState = [
+            'testOne'   => ['state' => TestStatus::success(), 'time' => 1],
+            'testTwo'   => ['state' => TestStatus::success(), 'time' => 1],
+            'testThree' => ['state' => TestStatus::error(), 'time' => 1],
+            'testFour'  => ['state' => TestStatus::success(), 'time' => 1],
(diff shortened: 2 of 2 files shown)


staabm:
the errored static analysis build is related to a recent update of `DeepCopy` - not the PR

phpunit.xsd
```diff
@@ -118,6 +118,7 @@
         <xs:restriction base="xs:string">
             <xs:enumeration value="default"/>
             <xs:enumeration value="defects"/>
+            <xs:enumeration value="defects,random"/>
```
staabm, line 121: since this order seems to be working as expected (as long as we don't take data-providers into account), it seems there is no need to not call this order as "supported".

fixing the data provider ordering is a separate problem, which also affects `defect` alone, without random ordering afterwards

tests/unit/Runner/TestSuiteSorterTest.php
```diff
@@ -503,6 +503,42 @@ public function testCanSetRandomizationWithASeed(): void
...
+        mt_srand(54321);
+        $sorter->reorderTestsInSuite($suite, TestSuiteSorter::ORDER_RANDOMIZED, false, TestSuiteSorter::ORDER_DEFECTS_FIRST);
+
+        $expectedOrder = [
+            MultiDependencyTest::class . '::testFive',
+            MultiDependencyTest::class . '::testThree',
+            MultiDependencyTest::class . '::testTwo',
+            MultiDependencyTest::class . '::testFour',
+            MultiDependencyTest::class . '::testOne',
+        ];
```
staabm, line 537: we can see in this test the we correctly order the defect tests before the succeeding ones (as long as no dataprovider is used). this behaviour was not tested, therefore I have added the test

sebastianbergmann:
> the errored static analysis build is related to a recent update of `DeepCopy` - not the PR

The static analysis does not fail for `main` (or any other branch) at the moment using `myclabs/deep-copy` at the version locked in `composer.lock`. What am I missing?

staabm:
The tool folder seems not locked 

See the composer install step https://github.com/sebastianbergmann/phpunit/actions/runs/16076919539/job/45374197952?pr=6258

sebastianbergmann:
> The tool folder seems not locked

Good catch!, but this has nothing to do with the `tools` folder. Somehow we used `composer update` instead of `composer install` for installing the dependencies in the static analysis CI job.

sebastianbergmann:
Please rebase so that you have 5a943ab4c2508f1317b12f891eac1395ab3c81a7. Thanks!

staabm:
> Somehow we used composer update instead of composer install for installing the dependencies in the static analysis CI job.

When no composer.lock exists running `composer install` effectively is an `composer update`

sebastianbergmann:
Cherry-picked into `10.5` and merged to `11.5`, `12.2`, and `main` from there. Thank you!

PR #6261, opened 2025-07-05
Fix order of tests which use data from data providers is not affected by test sorting

staabm:
closes https://github.com/sebastianbergmann/phpunit/issues/6259

PR #6262, opened 2025-07-05
Refactor ResultCache interface to use dedicated ResultCacheId type

staabm:
groundwork refactoring to ease later bugfix

extracted from https://github.com/sebastianbergmann/phpunit/pull/6261


staabm:
should be good to go

PR #6264, opened 2025-07-06
Reference dataprovider label in more error messages

staabm:
Introduces `ProvidedData`, a container class which holds the raw dataprovider data.
As we no longer pass arround raw arrays but a `array<ProvidedData>` we can now stick the dataprovider label onto the `ProvidedData`-objects, which means we can later on build more precise error messages which note the dataproviders name.

closes https://github.com/sebastianbergmann/phpunit/pull/5992

### DIFF
--- a/src/Framework/Exception/InvalidDataProviderException.php
+++ b/src/Framework/Exception/InvalidDataProviderException.php
@@ -9,6 +9,8 @@
  */
 namespace PHPUnit\Framework;
 
+use Throwable;
+
 /**
  * @no-named-arguments Parameter names are not covered by the backward compatibility promise for PHPUnit
  *
@@ -16,4 +18,30 @@ namespace PHPUnit\Framework;
  */
 final class InvalidDataProviderException extends Exception
 {
+    private ?string $providerLabel = null;
+
+    public static function forException(Throwable $e, string $providerLabel): self
+    {
+        $exception = new self(
+            $e->getMessage(),
+            $e->getCode(),
+            $e,
+        );
+        $exception->providerLabel = $providerLabel;
+
+        return $exception;
+    }
+
+    public static function forProvider(string $message, string $providerLabel): self
+    {
+        $exception                = new self($message);
+        $exception->providerLabel = $providerLabel;
+
+        return $exception;
+    }
+
+    public function getProviderLabel(): ?string
+    {
+        return $this->providerLabel;
+    }
 }
--- a/src/Framework/Exception/InvalidDataProviderException.php
+++ b/src/Framework/Exception/InvalidDataProviderException.php
@@ -32,14 +32,6 @@ final class InvalidDataProviderException extends Exception
         return $exception;
     }
 
-    public static function forProvider(string $message, string $providerLabel): self
-    {
-        $exception                = new self($message);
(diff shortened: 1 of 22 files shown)


src/Framework/Exception/InvalidDataProviderException.php
```diff
@@ -9,11 +9,39 @@
...
 namespace PHPUnit\Framework;
 
+use Throwable;
+
 /**
  * @no-named-arguments Parameter names are not covered by the backward compatibility promise for PHPUnit
  *
  * @internal This class is not covered by the backward compatibility promise for PHPUnit
  */
 final class InvalidDataProviderException extends Exception
```
staabm, line 19: in https://github.com/sebastianbergmann/phpunit/blob/a73ec4950712a4ff809ff730abe244f460160d11/src/Metadata/Api/DataProvider.php#L60-L64 there is a single use-case in which we still use `InvalidDataProviderException` but do not reference a single dataprovider, therefore this method still has a `public __construct` inherited from the parent-class.

we could introduce another named-constructor for this use-case which allows us to turn the `InvalidDataProviderException::__construct()` private, in case you prefer.

src/Metadata/Api/DataProvider.php
```diff
@@ -135,32 +139,35 @@ private function dataProvidedByMethods(string $className, string $methodName, Me
...
                         ),
+                        $providerLabel,
                     );
                 }
 
                 if ($method->getNumberOfParameters() > 0) {
-                    throw new InvalidDataProviderException(
+                    throw InvalidDataProviderException::forProvider(
                         sprintf(
                             'Data Provider method %s::%s() expects an argument',
```
staabm: looking at the final result again, I think I might have been to close to the idea https://github.com/sebastianbergmann/phpunit/pull/5992 wanted to implement.

which dataprovider related to this exception cases, was already visible via exception messages before this PR.
🤔

did I miss the point?

staabm: reworked the initial commit and I think the result is now more useful.

in cases where we did not have a reference to the dataprovider before, we now have more precise error messages

tests/end-to-end/event/data-provider-duplicate-key.phpt
```diff
@@ -19,7 +19,7 @@ Data Provider Method Finished for PHPUnit\TestFixture\Event\DataProviderDuplicat
 - PHPUnit\TestFixture\Event\DataProviderDuplicateKeyTest::provider
 Test Triggered PHPUnit Error (PHPUnit\TestFixture\Event\DataProviderDuplicateKeyTest::testSomething)
 The data provider specified for PHPUnit\TestFixture\Event\DataProviderDuplicateKeyTest::testSomething is invalid
-The key "key" has already been defined by a previous data provider
+The key "key" has already been defined by provider PHPUnit\TestFixture\Event\DataProviderDuplicateKeyTest::provider
```
staabm, line 22: I went with a fully qualified method reference in the error messages, as these can easily be navigated to by e.g. PHPStorm.

staabm:
I think this PR is a good foundation so I can explore "how to reference a single data-set" in a error message in a followup PR

Issue #6236, opened 2025-06-13
PHPUnit warning cannot be ignored


nikophil:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 12.2.2 
| PHP version         | 8.3.20
| Installation Method | Composer

#### Summary

PHPUnit warnings are always failing the testsuite, diregarding of `failOnWarning` configuration.

This wasn't a problem until https://github.com/sebastianbergmann/phpunit/pull/6213 which started to trigger PHPUnit warnings when the number of arguments returned by the data provider does not match the number of arguments of the test method.

This warning seems totally legit, but in some case, it could be useful to allow those numbers of arguments to not match, example:

```php
final class SomeTest extends TestCase
{
    #[Before]
    public function initializationMethod(): void
    {
        // do some initialization using `$this->providedData()`
    }

    #[Test]
    #[DataProvider('provider')]
    public function testMethod() void
    {
        self::assertTrue(true);
    }

    public static function provider(): iterable
    {
        yield ['foo'];
    }
}
```

I know this is a little bit hacky, but in some complex scenario, it is pretty handy.

#### How to reproduce

```php
final class ReproducingTest extends TestCase
{
    #[Test]
    #[DataProvider('provider')]
    public function testMethod() void
    {
        self::assertTrue(true);
    }

    public static function provider(): iterable
    {
        yield ['foo'];
    }
}
```

#### Expected behavior

I'm not sure what's the best way to fix this... There could be some "global" solutions:
- make the config `failOnWarning` to act on PHPUnit warning as well
- add a new config `failOnPHPUnitWarning`

but both of them would hide other more legit warnings, I think they are not a really good solution.

Or maybe we could go for a more fine-grained solution:
- allow to ignore PHPUnit warning along with a baseline, as well as deprecations
- add a new parameter to `DataProvider` attribute? `#[DataProvider('providerMethodName', allowExtraArguments: true)]`
- introduce a new attribute `#[IgnoreWarnings]` / `#[IgnorePHPUnitWarnings]`

I'd be willing to provide a PR if we agree on a solution to implement.

thanks!

sebastianbergmann:
> PHPUnit warnings are always failing the testsuite, diregarding of failOnWarning configuration.

This is not a bug, this is by design. I am willing to discuss this design and the change the behaviour, of course.

nikophil:
[...]

PR #6268, opened 2025-07-07
`#[IgnorePHPUnitWarnings]` attribute for ignoring PHPUnit warnings

nikophil:
This is a follow up PR of [this discussion](https://github.com/sebastianbergmann/phpunit/issues/6236#issuecomment-2990025288) 

I hope the involved complexity is not too high. I was actually pleasantly surprised how easy it was

Please, could you tell me what's your feeling about the implementation, if you're okay to go, I'll add some missing unit tests.

thanks!

### DIFF
--- a/src/Event/Emitter/DispatchingEmitter.php
+++ b/src/Event/Emitter/DispatchingEmitter.php
@@ -29,7 +29,7 @@ use PHPUnit\Event\TestSuite\Sorted as TestSuiteSorted;
 use PHPUnit\Event\TestSuite\Started as TestSuiteStarted;
 use PHPUnit\Event\TestSuite\TestSuite;
 use PHPUnit\Framework\TestCase;
-use PHPUnit\Metadata\IgnorePHPUnitWarnings;
+use PHPUnit\Metadata\IgnorePhpunitWarnings;
 use PHPUnit\Metadata\Parser\Registry;
 use PHPUnit\TextUI\Configuration\Configuration;
 use SebastianBergmann\Comparator\Comparator;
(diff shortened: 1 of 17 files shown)


src/Framework/Attributes/IgnorePHPUnitWarnings.php
```diff
@@ -0,0 +1,40 @@
...
+namespace PHPUnit\Framework\Attributes;
+
+use Attribute;
+
+/**
+ * @immutable
+ *
+ * @no-named-arguments Parameter names are not covered by the backward compatibility promise for PHPUnit
+ */
+#[Attribute(Attribute::TARGET_METHOD)]
```
nikophil: I'm wondering if we should allow the attribute on class level :thinking:

sebastianbergmann: I do not think so.

sebastianbergmann:
> Please, could you tell me what's your feeling about the implementation, if you're okay to go, I'll add some missing unit tests.

Looks good, thank you. I will merge once you're ready.

src/Event/Emitter/DispatchingEmitter.php
```diff
@@ -1008,6 +1011,17 @@ public function testTriggeredPhpunitError(Code\Test $test, string $message): voi
      */
     public function testTriggeredPhpunitWarning(Code\Test $test, string $message): void
     {
+        if (
+            $test->isTestMethod() &&
+            class_exists($test->className()) &&
```
sebastianbergmann, line 1016: Is the call to `class_exists()` really needed? I do not see how we could have a `Code\Test $test` object where `$test->className()` returns a string that is not the name of a class that exists.

nikophil: I've added it because the test `DispatchingEmitterTest::testTestTriggeredPhpunitWarningEmitsTestPhpunitWarningTriggeredEvent` uses a classname that does not exist

src/Framework/Attributes/IgnorePHPUnitWarnings.php
```diff
@@ -0,0 +1,40 @@
...
+final readonly class IgnorePHPUnitWarnings
```
nikophil: For the sake of coherence, I'll change the case to `IgnorePhpunitWarnings`

nikophil: done

src/Event/Emitter/DispatchingEmitter.php
nikophil: I think that otherwise, we should use existing classes/methods [here](https://github.com/sebastianbergmann/phpunit/blob/main/tests/unit/Event/Emitter/DispatchingEmitterTest.php#L3504-L3528)

nikophil:
@sebastianbergmann ready! :blush:

sebastianbergmann:
@nikophil I have changed the implementation in 4db30b16506b6e603821783d63995982868004a2 to be consistent with similar functionality.

PR #6269, opened 2025-07-08
Fix handling of regex containing `/` in IgnorePhpunitWarnings

staabm:
followup to https://github.com/sebastianbergmann/phpunit/pull/6268

### DIFF
--- a/src/Metadata/IgnorePhpunitWarnings.php
+++ b/src/Metadata/IgnorePhpunitWarnings.php
@@ -40,6 +40,6 @@ final readonly class IgnorePhpunitWarnings extends Metadata
     public function shouldIgnore(string $message): bool
     {
         return $this->messagePattern === null ||
-            (bool) preg_match("/{$this->messagePattern}/", $message);
+            (bool) preg_match('{' . $this->messagePattern . '}', $message);
     }
 }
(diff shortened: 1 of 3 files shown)


src/Metadata/IgnorePhpunitWarnings.php
```diff
@@ -40,6 +40,6 @@ public function isIgnorePhpunitWarnings(): true
     public function shouldIgnore(string $message): bool
     {
         return $this->messagePattern === null ||
-            (bool) preg_match("/{$this->messagePattern}/", $message);
```
staabm, line 43: I guess its more of a theoretical problem, as we likely don't have error messages containing `/`.

using `{}` as pattern delimiters allows us to concat a dynamic pattern which cannot error on meta characters

staabm:
//cc @nikophil

src/Metadata/IgnorePhpunitWarnings.php
nikophil: Hey @staabm 
Thanks to notifying me on this
Not sure to understand: are you using `{}` because they are less likely to be used in a regex pattern, or does it have another purpose?
Thanks!

staabm: Using `{}` cannot collide with other meta chars. 

You can nest `{}` into it, what you can't do with `/`.

`/warn/err/` is a syntax error (requires escaping of `/` as `\\/`)
While
`{warn{}err}` is not

nikophil: Good to know, thanks!

PR #6272, opened 2025-07-12
Use `@<data-set-name>` format (compatible with `--filter` CLI option) in defect messages

### LINKED ISSUE #6260: DX: filtering tests by "data provider name" vs. index
depending on how a dataprovider is keyed, the `--filter` parameter needs to be provided in a different format:

namely `@` vs. `#` formatted

---

for a dataprovider with named keys:
```php
function provideData(): iterable {
  yield['my name' => 1];
}
```

`--filter "testSuiteSorterDefectsWithDataProviderTest@my name"`

the corresponding test error is

`2) PHPUnit\TestFixture\FaillingDataProviderTest::testWithProvider with data set "my name" (1)`

=> I think it would be useful if the error message would contain `@my name` instead of just `my name`, to ease building the filter param

---

for a dataprovider with indexed keys:
```php
function provideData(): iterable {
  yield[1];
}
```

`--filter "testSuiteSorterDefectsWithDataProviderTest#1"`

the corresponding test error is

`PHPUnit\TestFixture\FaillingDataProviderTest::testWithProvider with data set #1 (1)`

=> here its easy to remember we can run the test with `#1` because the error message tells me so

staabm:
closes https://github.com/sebastianbergmann/phpunit/issues/6260

After this PR the default phpunit result printer will format errors the same way the `--filter` CLI option expects them.

that way the user can copy/paste the tool output and re-use it as the CLI filter option value:

---

phpunit output for numbered datasets:
```
There was 1 failure:

1) testSuiteSorterDefectsWithDataProviderTest::testOne#1 with data (false)
Failed asserting that false is true.
```

CLI option for numbered datasets:
`--filter "testSuiteSorterDefectsWithDataProviderTest::testOne#1"`

---

phpunit output for named datasets:
```
There was 1 failure:

1) testSuiteSorterDefectsWithDataProviderTest::testOne@data-set with data (false)
Failed asserting that false is true.
```

CLI option for named datasets:
`--filter "testSuiteSorterDefectsWithDataProviderTest::testOne@data-set"`


src/Framework/TestCase.php
```diff
@@ -843,17 +843,15 @@ final public function dataName(): int|string
      */
     final public function dataSetAsString(): string
```
staabm, line 844: since TestDox formatter does not use the `className::methodName <dataset>` format, I did not change the format in this method itself but created a new `dataSetAsFilterString` for use only in the default-printer

staabm:
commit 1&2 are pure refactoring. the actual change is in commit 3

PR #6273, opened 2025-07-13
Warn when `#[DataProvider*]` attributes are mixed with `#[TestWith*]` attributes

### LINKED ISSUE #6265: DX: error on mix of different dataprovider types
when mixing `#[TestWith(...)]` and `#[DataProvider(...)]` style of data-providers, [we prefer `#[DataProvider(...)]` and ignore `#[TestWith(...)]`](https://github.com/sebastianbergmann/phpunit/blob/a73ec4950712a4ff809ff730abe244f460160d11/src/Metadata/Api/DataProvider.php#L54-L58).

I think we should emit a PHPUnit warning in case a `#[TestWith(...)]` is ignored.
(I think its not worth supporting mixing these different dataprovider types on a single method)

```php
/*
 *
 *
 */
namespace PHPUnit\TestFixture;

use PHPUnit\Framework\Attributes\TestWith;

final class TestWithAttributeAndDataProviderTest extends TestCase
{
    public static function provider(): iterable
    {
        yield 'foo' => ['bar', 'baz'];
    }

    #[TestWith(['a', 'b'], 'foo')]
    #[DataProvider('provider')]
    public function testWithDuplicateName($one, $two): void
    {
        $this->assertTrue(true);
    }
}
```

staabm:
closes https://github.com/sebastianbergmann/phpunit/issues/6265

### DIFF
--- a/src/Metadata/Api/DataProvider.php
+++ b/src/Metadata/Api/DataProvider.php
@@ -52,6 +52,25 @@ final readonly class DataProvider
         }
 
         if ($dataProvider->isNotEmpty()) {
+            if ($testWith->isNotEmpty()) {
+                $method = new ReflectionMethod($className, $methodName);
+
+                Event\Facade::emitter()->testTriggeredPhpunitWarning(
+                    new TestMethod(
+                        $className,
+                        $methodName,
+                        $method->getFileName(),
+                        $method->getStartLine(),
+                        Event\Code\TestDoxBuilder::fromClassNameAndMethodName(
+                            $className,
+                            $methodName,
+                        ),
+                        MetadataCollection::fromArray([]),
+                        Event\TestData\TestDataCollection::fromArray([]),
+                    ),
+                    'Mixing #[DataProvider] and #[TestWith] attributes is not supported, only the data provided by #[DataProvider] will be used',
+                );
+            }
             $data = $this->dataProvidedByMethods($className, $methodName, $dataProvider);
         } else {
             $data = $this->dataProvidedByMetadata($testWith);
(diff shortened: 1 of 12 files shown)


src/Metadata/Api/DataProvider.php
```diff
@@ -52,6 +52,25 @@ public function providedData(string $className, string $methodName): ?array
         }
 
         if ($dataProvider->isNotEmpty()) {
+            if ($testWith->isNotEmpty()) {
+                $method = new ReflectionMethod($className, $methodName);
+
+                Event\Facade::emitter()->testTriggeredPhpunitWarning(
```
staabm, line 58: is this the right type of warning?

src/Metadata/Api/DataProvider.php
```diff
@@ -52,6 +52,25 @@ public function providedData(string $className, string $methodName): ?array
...
+                        $method->getFileName(),
+                        $method->getStartLine(),
+                        Event\Code\TestDoxBuilder::fromClassNameAndMethodName(
+                            $className,
+                            $methodName,
+                        ),
+                        MetadataCollection::fromArray([]),
+                        Event\TestData\TestDataCollection::fromArray([]),
+                    ),
+                    'Mixing #[DataProvider] and #[TestWith] attributes is not supported, only the data provided by #[DataProvider] will be used',
```
stof: this will be confusing when using `DataProviderExternal` or `TestWithJson`. The message should probably deal better with such cases

staabm: good point, didn't think about all these other types of providers. will adjust the wording.

src/Metadata/Api/DataProvider.php
sebastianbergmann: Yes.

Issue #6234, opened 2025-06-10
`expectProcessExit()`


sebastianbergmann:
When we have this code ...

#### `ProcessExit.php`
```php
<?php declare(strict_types=1);
final class ProcessExit
{
    public function doSomething(): never
    {
        exit(1);
    }
}
```

#### `ProcessExitTest.php`
```php
<?php declare(strict_types=1);
use PHPUnit\Framework\Attributes\CoversClass;
use PHPUnit\Framework\Attributes\RunInSeparateProcess;
use PHPUnit\Framework\TestCase;

#[CoversClass(ProcessExit::class)]
final class ProcessExitTest extends TestCase
{
    #[RunInSeparateProcess]
    public function testOne(): void
    {
        (new ProcessExit)->doSomething();
    }
}
```

and run PHPUnit then we currently get

```
PHPUnit 12.2.1 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.4.8
Configuration: /tmp/example/phpunit.xml

E                                                                   1 / 1 (100%)

Time: 00:00.156, Memory: 25.68 MB

There was 1 error:

1) ProcessExitTest::testOne
Test was run in child process and ended unexpectedly

ERRORS!
Tests: 1, Assertions: 0, Errors: 1.
```

Due to `#[RunInSeparateProcess]`, the test method is run in a separate process and the `exit` statement does not terminate the PHP process in which PHPUnit's test runner is running.

The child process that is used to run the test ends unexpectedly and this is interpreted by PHPUnit as an error.

There are situations, though, when the child process using `exit` is expected. However, this cannot be expected currently.

The following should be made possible:

```php
<?php declare(strict_types=1);
use PHPUnit\Framework\Attributes\CoversClass;
use PHPUnit\Framework\Attributes\RunInSeparateProcess;
use PHPUnit\Framework\TestCase;

#[CoversClass(ProcessExit::class)]
final class ProcessExitTest extends TestCase
{
    #[RunInSeparateProcess]
    public function testOne(): void
    {
        $this->expectProcessExit(1);

        (new ProcessExit)->doSomething();
    }
}
```

staabm:
Just had a look into implementing it.

The biggest complication is `$this->expectProcessExit(1);` beeing run in the child process, while result of the `exit(1)` code is available in the parent process.

[...]

PR #6275, opened 2025-07-15
Implement `TestCase->expectProcessExit($exitCode)`

staabm:
POC for https://github.com/sebastianbergmann/phpunit/issues/6234

### DIFF
--- a/ProcessExitTest.php
+++ b/ProcessExitTest.php
@@ -1,0 +1,33 @@
+<?php
+
+declare(strict_types=1);
+
+use PHPUnit\Framework\Attributes\CoversClass;
+use PHPUnit\Framework\Attributes\DataProvider;
+use PHPUnit\Framework\Attributes\RunInSeparateProcess;
+use PHPUnit\Framework\TestCase;
+
+final class ProcessExitTest extends TestCase
+{
+    #[RunInSeparateProcess]
+    #[DataProvider("provideExitCodes")]
+    public function testOne(?int $expectedExit, int $actualExitCode): void
+    {
+        if ($expectedExit !== null) {
+            $this->expectProcessExit($expectedExit);
+        }
+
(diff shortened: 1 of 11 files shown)


ProcessExitTest.php
```diff
@@ -0,0 +1,33 @@
...
+use PHPUnit\Framework\TestCase;
+
+final class ProcessExitTest extends TestCase
+{
+    #[RunInSeparateProcess]
+    #[DataProvider("provideExitCodes")]
+    public function testOne(?int $expectedExit, int $actualExitCode): void
+    {
+        if ($expectedExit !== null) {
+            $this->expectProcessExit($expectedExit);
```
staabm, line 17: we should add a PHPUnit error/warning when `expectProcessExit` is called in a test, which does not use process-isolation

src/Framework/TestRunner/ChildProcessResultProcessor.php
```diff
@@ -74,6 +74,16 @@ public function process(Test $test, string $serializedProcessResult, string $std
...
 
+        if ($childResult->expectedProcessExit !== null && $childResult->testCalledExit === true) {
+            assert($test instanceof TestCase);
+
+            $test->assertSame($childResult->expectedProcessExit, $exitCode, 'Process exit-code expectation failed');
+        } elseif ($childResult->expectedProcessExit !== null && $childResult->testCalledExit === false) {
+            $test->fail('Process expected exit() to be called but test did not call it');
+        } elseif ($childResult->expectedProcessExit === null && $childResult->testCalledExit === false) {
+            $test->fail('Process called exit() but the test did not expect it');
+        }
```
staabm: we need something here which allows us to combine parent-process level assertions with the child process results into a overall result

sebastianbergmann:

PR #6281, opened 2025-07-26
Exceptions raised in after-test method are not reported for skipped tests

LukeTowers:
Currently, if a test calls markTestSkipped() (throwing a SkippedWithMessageException) and tearDown() then throws its own exception (e.g. due to uninitialized state), PHPUnit swallows the tearDown() exception and only reports the skip and any resulting risky‑test warnings.

This change treats SkippedWithMessageException differently so that exceptions raised in tearDown() are still surfaced, making it easier to diagnose teardown issues even in skipped tests.

Related: https://github.com/laravel/framework/issues/49502

## Issue Fixed:
If a test is marked as skipped and your `tearDown()` throws an exception before Laravel’s cleanup runs, PHPUnit will swallow that teardown exception and instead show the vague “Test code or tested code did not remove its own exception handlers” message.

### Why:
`markTestSkipped()` throws a `SkippedWithMessageException`.
PHPUnit’s `runBare()` always calls `tearDown()` afterward, but if `tearDown()` throws, the exception is ignored because PHPUnit assumes the earlier skip exception is the “real” outcome.
Since Laravel’s cleanup never ran, global handlers remain, triggering the risky-test warning.

### Example:

```php
public function setUp(): void
{
    parent::setUp();

    if (!File::exists(base_path('node_modules'))) {
        $this->markTestSkipped('This test requires node_modules to be installed');
    }

    // Never reached if skipped:
    $this->fixturePath = base_path('modules/system/tests');
}

public function tearDown(): void
{
    // Uses $this->fixturePath but it’s null if skipped early
    if (File::isDirectory($this->fixturePath . '/node_modules')) {
        File::deleteDirectory($this->fixturePath . '/node_modules');
    }

    parent::tearDown();
}
```

### Result:
Instead of the teardown exception, PHPUnit only reports:

> Test code or tested code did not remove its own exception handlers.

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -620,7 +620,7 @@ abstract class TestCase extends Assert implements Reorderable, SelfDescribing, T
                 Event\Code\ComparisonFailureBuilder::from($e),
             );
         } catch (Throwable $exceptionRaisedDuringTearDown) {
-            if (!isset($e)) {
+            if (!isset($e) || $e instanceof SkippedWithMessageException) {
                 $this->status = TestStatus::error($exceptionRaisedDuringTearDown->getMessage());
                 $e            = $exceptionRaisedDuringTearDown;
 


sebastianbergmann:
Cherry-picked into `11.5` and merged to `12.3` and `main` from there.

raveren:
It's an unending ball of yarn, but I am pretty sure this change caused several issues on my code base, lets see if I'm the only one:

```
ArgumentCountError: Too few arguments to function PHPUnit\Runner\ErrorHandler::enable(), 0 passed in /var/www/vendor/laravel/framework/src/Illuminate/Foundation/Bootstrap/HandleExceptions.php on line 358 and exactly 1 expected
```

image

the line 77 in the output is just `parent::tearDown()` :

image

and then there's this 
image

I am running Laravel 11.45 and the latest PHPUnit 12.3.4

raveren:
Yeah, I am getting 
```
Test code or tested code removed error handlers other than its own
```
OR
```
ArgumentCountError: Too few arguments to function PHPUnit\Runner\ErrorHandler::enable(), 0 passed in /var/www/vendor/laravel/framework/src/Illuminate/Foundation/Bootstrap/HandleExceptions.php on line 358 and exactly 1 expected
```
for EVERY test if I update PHPUnit from 12.2.9 to 12.3.4.

Also, with this, my teamcity pipeline timeouts and nothing can kill it besides a server restart :D

Also I smell smoke brb (just kidding about this one!)

sebastianbergmann:
`PHPUnit\Runner\ErrorHandler` is `@internal` and not covered by the backward compatibility promise for PHPUnit. Its API was changed and it looks like `Illuminate/Foundation/Bootstrap/HandleExceptions.php` needs to be adapted.

LukeTowers:
@raveren it was this commit that's causing the issue: https://github.com/sebastianbergmann/phpunit/commit/83f64496c6366e4d0f2c8a620ed5e3ca8b3dbb5c

PR #6282, opened 2025-07-31
Add `assertSameDictionaryKeysValues` to compare keys and values without regard to key order

### LINKED ISSUE #6088: assertSame ignoring key order
Key order when we're talking about associative arrays is quite often irrelevant, especially when checking stuff that will end up being json serialized where dictionaries do not guarantee the key order.

Right now I'm stuck between using assertEqualsCanonicalizing and risking datatype errors, e.g.:

```
        $this->assertEqualsCanonicalizing(['foo' => 1], ['foo' => '1']); // true
```

Or use assertSame but then key order tends to make tests extra brittle. Having an option for assertSame or a new assertDictionarySame or something would be great.

samlitowitz:
Resolves #6088 

@sebastianbergmann assuming you are ok with the concept, I would love your input on the implementation. Notably a large portion of the constraint work is take from the [`ArrayComparator`](https://github.com/sebastianbergmann/comparator/blob/main/src/ArrayComparator.php) and the [`ObjectComparator`](https://github.com/sebastianbergmann/comparator/blob/main/src/ObjectComparator.php) making me wonder if a new comparator was the right approach. If it's ok with you,  I wanted to have the discussion here before opening issues or making PR's against the [`comparator` repository](https://github.com/sebastianbergmann/comparator).

### DIFF
--- a/src/Framework/Assert.php
+++ b/src/Framework/Assert.php
@@ -25,6 +25,7 @@ use PHPUnit\Framework\Constraint\ArrayHasKey;
 use PHPUnit\Framework\Constraint\Callback;
 use PHPUnit\Framework\Constraint\Constraint;
 use PHPUnit\Framework\Constraint\Count;
+use PHPUnit\Framework\Constraint\Dictionary\IsIdenticalKeysValues;
 use PHPUnit\Framework\Constraint\DirectoryExists;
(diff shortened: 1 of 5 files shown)


samlitowitz:
Closed because resolution is part of a larger effort, #6423

PR #6283, opened 2025-08-01
Fix `expectUserDeprecationMessage*()` failure when test is run in separate process

### LINKED ISSUE #6102: `expectUserDeprecationMessage*()` fails when test is run in separate process
| Q                   | A
| PHPUnit version     | 11.5.2
| PHP version         | 8.3.9

#### Summary

`expectUserDeprecationMessage*()` fails when test is run in separate process.

#### Current behavior

A test like this

```
<?php

declare(strict_types=1);

class ExpectDeprecationTest extends TestCase {

  /**
   * Tests expectDeprecation.
   */
  public function testExpectDeprecation(): void {
    $this->expectUserDeprecationMessage('Test deprecation');
    @trigger_error('Test deprecation', E_USER_DEPRECATED);
  }

  /**
   * Tests expectDeprecation in isolated test.
   */
  #[RunInSeparateProcess]
  public function testExpectDeprecationInIsolation(): void {
    $this->expectUserDeprecationMessage('Test isolated deprecation');
    @trigger_error('Test isolated deprecation', E_USER_DEPRECATED);
  }

}

```

leads to an `EventFacadeIsSealedException` error reported by the runner:

```
PHPUnit 11.5.2 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.3.9
Configuration: /Users/xxx/Dev/drupal/core/phpunit.xml.dist

DE                                                                  2 / 2 (100%)

Time: 00:01.879, Memory: 14.00 MB

There was 1 error:

1) ExpectDeprecationTest::testExpectDeprecationInIsolation
PHPUnit\Event\EventFacadeIsSealedException:

--

2 tests triggered 2 deprecations:

1) /Users/xxx/Dev/drupal/core/tests/Drupal/Tests/ExpectDeprecationTest.php:15
Test deprecation

2) /Users/xxx/Dev/drupal/core/tests/Drupal/Tests/ExpectDeprecationTest.php:24
Test isolated deprecation

ERRORS!
Tests: 2, Assertions: 2, Errors: 1, Deprecations: 2.
```

#### How to reproduce

See above

#### Expected behavior

No exception reported, only the deprecations.

mondrake:
Fixes #6102.


tests/end-to-end/regression/6102-2.phpt
```diff
@@ -0,0 +1,38 @@
+--TEST--
```
staabm, line 1: could we have a better test name then 6102-1 and 6102-2. something which reflects the difference of the 2 tests

mondrake: Would `6102-cli-argument` and `6102-class-attribute` be more relevant?

staabm: I would go with something like
`6102-process-isolated-arg`
`6102-separate-processes-attribute`

sebastianbergmann: How about `6102-process-isolation-via-attribute.phpt` and `6102-process-isolation-via-cli-option.phpt`?

mondrake: afk at the moment, will look into it when possible if noone beats me at it

sebastianbergmann:
Cherry-picked into `11.5`, cleaned up, and merged to `12.3` and `main` from there.

PR #6287, opened 2025-08-02
feat(baseline): can ignore deprecations with regex

nikophil:
I think it could be nice to be able to ignore deprecations using regex: see for instance [doctrine/deprecations](https://github.com/doctrine/deprecations/blob/1.1.x/src/Deprecation.php#L198-L206): they do insert the line number in their deprecation messages. Meaning that if they change some code in the file, if PHPUnit is not able to ignore deprecations based on regex, the code would break

### DIFF
--- a/src/Runner/Baseline/Issue.php
+++ b/src/Runner/Baseline/Issue.php
@@ -13,6 +13,7 @@ use const FILE_IGNORE_NEW_LINES;
 use function assert;
 use function file;
 use function is_file;
+use function preg_match;
 use function sha1;
 use PHPUnit\Runner\FileDoesNotExistException;
 
@@ -112,7 +113,7 @@ final readonly class Issue
         return $this->file() === $other->file() &&
                $this->line() === $other->line() &&
                $this->hash() === $other->hash() &&
-               $this->description() === $other->description();
+               ($this->description() === $other->description() || preg_match('{' . $this->description() . '}', $other->description()) > 0);
     }
 
     /**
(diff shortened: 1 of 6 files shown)


src/Runner/Baseline/Issue.php
```diff
@@ -112,7 +113,7 @@ public function equals(self $other): bool
         return $this->file() === $other->file() &&
                $this->line() === $other->line() &&
                $this->hash() === $other->hash() &&
-               $this->description() === $other->description();
+               ($this->description() === $other->description() || preg_match('{' . $this->description() . '}', $other->description) !== false);
```
nikophil: I'm actually wondering if we should use regex, or a simple `str_contains()`, WDYT? maybe both?

sebastianbergmann:
This pull request proposes a new feature and should therefore target the `main` branch.

nikophil:
@sebastianbergmann done

sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it as I fail to see the benefit.

The baseline file is not supposed to be written or edited manually, it must only be generated using the `--generate-baseline` CLI option.

nikophil:
@sebastianbergmann damn you're right 

I think I did mix up the two concepts which Symfony's PHPunit bridge offers: 
- [the baseline](https://symfony.com/doc/current/components/phpunit_bridge.html#baseline-deprecations), which is basically the same thing that PHPunit provides
- [the ignore file](https://symfony.com/doc/current/components/phpunit_bridge.html#ignoring-deprecations) which allows to rely on regex to globally ignore deprecations

sorry for wasting your time :grimacing:

PR #6290, opened 2025-08-04
Test `#[DataProvider]` with process isolation

staabm:
this combination of features seems not tested atm

### DIFF
--- a/tests/end-to-end/_files/TestProcessIsolationWithDataProvider.php
+++ b/tests/end-to-end/_files/TestProcessIsolationWithDataProvider.php
@@ -1,0 +1,30 @@
+<?php declare(strict_types=1);
+
+/*
+ * This file is part of PHPUnit.
+ *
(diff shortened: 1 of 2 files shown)


sebastianbergmann:
`tests/end-to-end/phar/tests/standard/GreeterTest.php` is not a regular test, but specific to end-to-end tests for the PHPUnit PHAR.

End-to-End tests for data providers should go into `tests/end-to-end/data-provider` and use separate fixture (test case class(es) that combine data provider and process isolation functionality in this case).

Also note that missing tests like this should be added to oldest branch that is still supported (currently `11.5`). In this case, I think that we should add the test(s) also to `10.5`.

staabm:
moved the logic into a new test

sebastianbergmann:
Thank you!

I cherry-picked this into `10.5` and merged to `11.5`, `12.3`, and `main` from there.

PR #6292, opened 2025-08-06
Added more detailed assertions in Issue5592Test

staabm:
Adding more assertions arround error and exception handler, so one can more easily see that the result of the test depends on PHPUnit inner workings (and the current state of the inner workings).

PHPUnit itself does only register a error-handler, but not a exception handler.
therefore only the error-handler case is considered risky in https://github.com/sebastianbergmann/phpunit/blob/558dfd63e8e823a26453011e50caaef0f6fa8380/tests/end-to-end/regression/5592-process-isolation.phpt#L47-L50

---

as we get different behaviour when the test runs in isolation or not, I have copied the test and adjusted assertions accordingly.

no functional changes - just reflecting current behaviour in assertions.

### DIFF
--- a/tests/end-to-end/regression/5592/Issue5592Test.php
+++ b/tests/end-to-end/regression/5592/Issue5592Test.php
@@ -14,19 +14,25 @@ use function restore_exception_handler;
 use function set_error_handler;
 use function set_exception_handler;
 use PHPUnit\Framework\TestCase;
+use PHPUnit\Runner\ErrorHandler;
+use Throwable;
 
 class Issue5592Test extends TestCase
 {
     public function testAddedAndRemovedErrorHandler(): void
     {
-        set_error_handler(static fn () => false);
+        $previous = set_error_handler([$this, 'addedAndRemovedErrorHandler']);
         restore_error_handler();
+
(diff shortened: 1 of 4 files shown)


tests/end-to-end/regression/5592/Issue5592Test.php
```diff
@@ -14,19 +14,25 @@
...
 use PHPUnit\Framework\TestCase;
+use PHPUnit\Runner\ErrorHandler;
+use Throwable;
 
 class Issue5592Test extends TestCase
 {
     public function testAddedAndRemovedErrorHandler(): void
     {
-        set_error_handler(static fn () => false);
+        $previous = set_error_handler([$this, 'addedAndRemovedErrorHandler']);
```
staabm, line 24: using a named error handler, so its easier to point out which error handler violates assertions.

before this PR we only got `Closure` as a error handler, which is hard to use for debugging, as we have multiple sites registering a `Closure`.

this will also help us identifiy which handler is leaking into what in case of isolation bugs

sebastianbergmann:
Cherry-picked into `11.5` and merged to `12.3` and `main` from there.

theofidry:
👏

PR #6165, opened 2025-03-28
Collect deprecations triggered by autoloading while loading/building the test suite

soyuka:
When one triggers a deprecation in a class directly like this:

```php
<?php
trigger_error('Depreciated interface', E_USER_DEPRECATED);

interface Foo {}
```

This technique is notably used in many Symfony components, for example:

https://github.com/symfony/symfony/blob/82cf90aefb7e9117ebd981fec83d7546e310e52b/src/Symfony/Component/PropertyInfo/Type.php#L12-L15

The deprecation will be triggered when autoloading the class, which happens in:

https://github.com/sebastianbergmann/phpunit/blob/4b6a4ee654e5e0c5e1f17e2f83c0f4c91dee1f9c/src/TextUI/Application.php#L181

At that point, the ErrorHandler is not registered and we loose the deprecation. 

This pull request gives an idea on how it may be fixed, indeed the `ErrorHandler` needs a `TestCase` therefore I'm collecting these deprecations and triggering them once we call `ErrorHandler::enable`. I'm open to suggestions on how to fix this differently. Obviously I'll add tests if/once a solution is found that can be merged.

Thanks!

PR #6293, opened 2025-08-06
Relate deprecations from DataProvider with corresponding tests

### LINKED ISSUE #6279: Deprecation triggered in data provider method affects all test methods using that data provider method
| Q                   | A
| PHPUnit version     | 12.2
| PHP version         | any

Since https://github.com/sebastianbergmann/phpunit/pull/6165, when a deprecation is triggered in a data provider, it affects **all** tests:

This test:

```php
class TriggersDeprecationInDataProviderTest extends TestCase
{
    #[Test]
    public function method1(): void
    {
        self::assertTrue(true);
    }

    #[Test]
    #[DataProvider('dataProvider')]
    public function method2(bool $value): void
    {
        self::assertTrue($value);
    }

    public static function dataProvider(): iterable
    {
        @trigger_error('some deprecation', \E_USER_DEPRECATED);

        yield [true];
    }

    #[Test]
    public function method3(): void
    {
        self::assertTrue(true);
    }
}
```

would display this output:
```
DDD                                                                 3 / 3 (100%)

Time: 00:00.001, Memory: 16.00 MB

3 tests triggered 1 deprecation:

1) /home/nicolas/works/github.com/sebastianbergmann/phpunit/tests/end-to-end/deprecation-trigger/_files/deprecation-in-data-provider/tests/TriggersDeprecationInDataProviderTest.php:26
some deprecation

Triggered by:

* PHPUnit\TestFixture\DeprecationInDataProvider\TriggersDeprecationInDataProviderTest::method1
  /home/nicolas/works/github.com/sebastianbergmann/phpunit/tests/end-to-end/deprecation-trigger/_files/deprecation-in-data-provider/tests/TriggersDeprecationInDataProviderTest.php:12

* PHPUnit\TestFixture\DeprecationInDataProvider\TriggersDeprecationInDataProviderTest::method2#0
  /home/nicolas/works/github.com/sebastianbergmann/phpunit/tests/end-to-end/deprecation-trigger/_files/deprecation-in-data-provider/tests/TriggersDeprecationInDataProviderTest.php:19

* PHPUnit\TestFixture\DeprecationInDataProvider\TriggersDeprecationInDataProviderTest::method3
  /home/nicolas/works/github.com/sebastianbergmann/phpunit/tests/end-to-end/deprecation-trigger/_files/deprecation-in-data-provider/tests/TriggersDeprecationInDataProviderTest.php:32

OK, but there were issues!
Tests: 3, Assertions: 3, Deprecations: 1.
```

I think the problem is because `ErrorHandler::$globalDeprecations` is not emptied in `ErrorHandler::triggerGlobalDeprecations()`.

A straightforward fix would be:
```diff
    private function triggerGlobalDeprecations(): void
    {
        foreach ($this->globalDeprecations ?? [] as $d) {
            $this->__invoke(...$d);
        }

+        $this->globalDeprecations = [];
    }
```

But this fix seems brittle:
- all the "data provider deprecations" would be affected to the first test, although they are not related to this specific test
- if this test has `#[IgnoreDeprecation]` attribute, all those "data provider deprecations" would suddenly disappear

One cool thing to do would be to manage to _connect_ the "data provider deprecations" to their related test method.
So that the above test would output:
```
.D.                                                                 3 / 3 (100%)

Time: 00:00.001, Memory: 16.00 MB

1 test triggered 1 deprecation:
```

Maybe in `ErrorHandler::deprecationHandler()` we could guess from which test we're coming from (something similar to https://github.com/sebastianbergmann/phpunit/blob/main/src/Runner/ErrorHandler.php#L101 or by leveraging `DataProviderMethodCalled`/`DataProviderMethodFinished` events), and defer the trigger of the deprecation only when we reach the test(s) connected to the data provider?

WDYT?

/cc @soyuka

staabm:
fixes https://github.com/sebastianbergmann/phpunit/issues/6279

idea is, that we relate deprecations when triggered while building a dataprovider to the corresponding test

---

we cannot use `$test = Event\Code\TestMethodBuilder::fromCallStack();` to infer the test-contex, as the dataprovider is built before we enter the actual test

### DIFF
--- a/tests/end-to-end/regression/6279/TriggersDeprecationInDataProviderTest.php
+++ b/tests/end-to-end/regression/6279/TriggersDeprecationInDataProviderTest.php
@@ -12,8 +12,10 @@ declare(strict_types=1);
 namespace PHPUnit\TestFixture\Issue6279;
 
 use const E_USER_DEPRECATED;
+use const E_USER_WARNING;
 use function trigger_error;
 use PHPUnit\Framework\Attributes\DataProvider;
+use PHPUnit\Framework\Attributes\DataProviderExternal;
 use PHPUnit\Framework\Attributes\Test;
 use PHPUnit\Framework\TestCase;
 
@@ -22,6 +24,15 @@ class TriggersDeprecationInDataProviderTest extends TestCase
     public static function dataProvider(): iterable
     {
         @trigger_error('some deprecation', E_USER_DEPRECATED);
+
+        yield [true];
+    }
+
+    public static function dataWith2Deprecations(): iterable
+    {
+        @trigger_error('first', E_USER_DEPRECATED);
+        @trigger_error('second', E_USER_DEPRECATED);
+        @trigger_error('warning', E_USER_WARNING);
 
         yield [true];
     }
@@ -43,5 +54,19 @@ class TriggersDeprecationInDataProviderTest extends TestCase
     public function method3(): void
     {
         $this->assertTrue(true);
+    }
+
+    #[Test]
+    #[DataProvider('dataProvider')]
+    public function method4(bool $value): void
+    {
+        $this->assertTrue($value);
+    }
+
+    #[Test]
+    #[DataProviderExternal(self::class, 'dataWith2Deprecations')]
+    public function method5(bool $value): void
+    {
+        $this->assertTrue($value);
     }
 }
--- a/tests/end-to-end/regression/6279/TriggersDeprecationInDataProviderTest.php
+++ b/tests/end-to-end/regression/6279/TriggersDeprecationInDataProviderTest.php
@@ -1,0 +1,47 @@
+<?php
+
+declare(strict_types=1);
+/*
+ * This file is part of PHPUnit.
+ *
+ * (c) Sebastian Bergmann <sebastian@phpunit.de>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+namespace PHPUnit\TestFixture\Issue6279;
+
+use const E_USER_DEPRECATED;
+use function trigger_error;
+use PHPUnit\Framework\Attributes\DataProvider;
+use PHPUnit\Framework\Attributes\Test;
+use PHPUnit\Framework\TestCase;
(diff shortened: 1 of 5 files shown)


tests/end-to-end/regression/6279/TriggersDeprecationInDataProviderTest.php
```diff
@@ -0,0 +1,47 @@
...
+ */
+namespace PHPUnit\TestFixture\Issue6279;
+
+use const E_USER_DEPRECATED;
+use function trigger_error;
+use PHPUnit\Framework\Attributes\DataProvider;
+use PHPUnit\Framework\Attributes\Test;
+use PHPUnit\Framework\TestCase;
+
+class TriggersDeprecationInDataProviderTest extends TestCase
```
nikophil, line 22: maybe we can add a case where two tests relate to the same data provider?

staabm: added more tests

nikophil:
@staabm shouldn't this PR target v12 ?

in my understanding it fixes a problem introduced by https://github.com/sebastianbergmann/phpunit/pull/6165 in 12.2

staabm:
You might be right. I will let this decision to Sebastian

xabbuh:
As commented in https://github.com/sebastianbergmann/phpunit/issues/6279#issuecomment-3173483558 this change would unblock Symfony to use PHPUnit > 12.1.

I think it could be worse to expand the tests with a test that uses the `#[IgnoreDeprecations]` attribute to prove that this is sufficient to silence deprecations.

I was also wondering how PHPUnit behaves if you use the same data provider with both tests ignoring deprecations as well as not ignoring them. The result looks expected to me but it may also be worth adding a test for that (maybe also doing two distinct cases to prove that the order doesn't matter, i.e. the behaviour is the same no matter if the ignoring deprecations test comes before or after the not ignoring test inside the test class).

staabm:
@xabbuh please suggest/PR concrete test examples you have in mind

sebastianbergmann:

PR #6295, opened 2025-08-08
Simplify TestCase

staabm:


src/Framework/TestCase.php
```diff
@@ -1295,7 +1295,8 @@ private function runTest(): mixed
     {
         $testArguments = array_merge($this->data, array_values($this->dependencyInput));
 
-        $capture          = tmpfile();
+        $capture = tmpfile();
+        assert($capture !== false);
         $errorLogPrevious = ini_set('error_log', stream_get_meta_data($capture)['uri']);
```
staabm, line 1300: before this PR, this line would already trigger a fatal error when `$capture` is `false`.
therefore make the error explicit so we can cleanup the `finally` block.

sebastianbergmann:

PR #6296, opened 2025-08-08
Prevent fatal errors not reported

### LINKED ISSUE #6294: Silent failure of PHP fatal errors
| Q                   | A
| PHPUnit version     | 12.3.0
| PHP version         | 8.4.3

#### Summary
Certain types of errors do no longer show up int he output

#### Current behavior

The script silently fails, without any error in the output. See cli output:

```
PHPUnit 12.3.0 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.4.3
```

#### How to reproduce

See the following test setup:

#### `A.php`
```php
class A
{
  public function someFunction(): void{
    echo 'A::someFunction' . PHP_EOL;
  }
}
```

#### `B.php`
```php
class B extends A
{
  // incorrect access level
  protected function someFunction(): void
  {
    parent::someFunction();

    echo 'B::someFunction' . PHP_EOL;
  }  
}
```

#### `Test.php`
```php
use \PHPUnit\Framework\TestCase;

final class ExampleTest extends TestCase
{
  {
    require_once('A.php');
    require_once('B.php');    

    $this->assertSame(1, 1);
  }
}
```

#### Expected behavior

Some form of error in the output, see CLI output:
```
PHPUnit 12.0.10 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.4.3

PHP Fatal error:  Access level to B::someFunction() must be public (as in class A) in ...
```

NOTE: it seems 12.0.10 was the latest version where the errors were shown

staabm:
happens when php.ini declares `display_errors=0`

closes https://github.com/sebastianbergmann/phpunit/issues/6294

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -26,6 +26,7 @@ use function fclose;
 use function getcwd;
 use function implode;
 use function in_array;
+use function ini_get;
 use function ini_set;
 use function is_array;
 use function is_callable;
@@ -83,6 +84,7 @@ use PHPUnit\Metadata\WithEnvironmentVariable;
 use PHPUnit\Runner\BackedUpEnvironmentVariable;
 use PHPUnit\Runner\DeprecationCollector\Facade as DeprecationCollector;
 use PHPUnit\Runner\HookMethodCollection;
+use PHPUnit\Runner\ShutdownHandler;
 use PHPUnit\TestRunner\TestResult\PassedTests;
 use PHPUnit\TextUI\Configuration\Registry as ConfigurationRegistry;
 use PHPUnit\Util\Exporter;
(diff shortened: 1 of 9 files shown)


sebastianbergmann:
I am not convinced (yet) that using `ini_set()` to "fix" the configuration is the right course of action here.

In the meantime, I have updated the documentation to also suggest `display_errors=On`: https://github.com/sebastianbergmann/phpunit-documentation-english/commit/b3b159cbe9bd7eb5656dd381fc6f028549601dce

staabm:
alternative fix could be to suggest 

```
<php>
  <ini name="display_errors" value="On"/>
</php>
```

in the configuration file when auto-generating.

but my gut feeling is, that people might miss a lot of obvious errors when not using the correct ini setting (which would mean lots of support tickets on phpunit).

---

we could also try to render a phpunit warning when display_errors=0, but atm phpunit crashes at the fatal error (so we would need a prepared error message in a shutdown function or similar)

staabm:
pushed an alternative and less invasive fix, which does not change the config with `ini_set()`, but instead prepares a error message which gets printed on premature fatal errors

sebastianbergmann:
This should target the `12.3` branch, right?

staabm:
yes

sebastianbergmann:

PR #6299, opened 2025-08-08
Emit warning when OPCache is enabled while collecting coverage

### LINKED ISSUE #1022: Since v10 some lines are ignores
| Q                         | A
| --------------------------| ---------------
| php-code-coverage version | 10.1.9
| PHP version               | 8.2.13
| Driver                    | Xdebug
| PCOV version (if used)    | n/a
| Xdebug version (if used)  | 3.2.2
| Installation Method       | Composer
| Usage Method              | PHPUnit / other
| PHPUnit version (if used) | 10.4.2

source: https://github.com/atk4/ui/blob/5.0.0/tests/JsTest.php#L38

image

Started to happen when PHPUnit was to v10 and php-code-coverage from v9.2.29 to v10.1.9. See the screenshot, l38 went from uncovered to ignored.

staabm:
because of byte-code optimizations results of coverage collection can be wrong, as the source beeing executed might no longer match the actual php script sources 

closes https://github.com/sebastianbergmann/php-code-coverage/issues/1022


.github/workflows/ci.yaml
```diff
@@ -204,14 +204,21 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 5
 
+    strategy:
+      fail-fast: false
+      matrix:
+        php-version:
+          - "8.4"
+          - "8.5"
```
staabm: was not sure how to configure CI so it runs the new test with opcache+coverage-driver.

I don't think it makes sense to run most other tests in this setup, as no other until now seem to need this combination.

---

alternatively we could use the phar tests which seem to also have this setup

staabm:
@sebastianbergmann any tips on how to run this test properly in the phar tests?

locally it works for me, but CI errors

src/Runner/CodeCoverage.php
```diff
@@ -124,6 +126,16 @@ public function init(Configuration $configuration, CodeCoverageFilterRegistry $c
             $this->codeCoverage()->excludeUncoveredFiles();
         }
 
+        if (
+            // opcache is built-in since 8.5
+            function_exists('opcache_compile_file') &&
+            ini_get('opcache.enable_cli') === '1'
+        ) {
```
sebastianbergmann: Just because `opcache` is always (unless a vendor patches PHP to change that) available should not have an effect how its availability is queried. For example, `extension_loaded('Zend OPcache')` still works for me with PHP 8.5-dev.

That being said, I prefer not to have code like this in PHPUnit itself. We have `Runtime::isOpcacheActive()` (https://github.com/sebastianbergmann/environment/blob/8.0.2/src/Runtime.php#L280C22-L295), `Runtime::performsJustInTimeCompilation()` (https://github.com/sebastianbergmann/environment/blob/8.0.2/src/Runtime.php#L93-L114), etc. for querying such information.

So, please use `Runtime` instead. If it does not yet have a query for the information we need here, we can add it.

sebastianbergmann:
> any tips on how to run this test properly in the phar tests?

Short answer: don't.

I do not see anything here that would be specific to the PHAR. A regular PHPT test should be all that we need here. Does that help?

staabm:
> I do not see anything here that would be specific to the PHAR. A regular PHPT test should be all that we need here. Does that help?

we don't currently run regular tests with opcache + coverage-driver, therefore I cannot make the new test in CI fail.
(or we need to add a new CI job)

src/Runner/CodeCoverage.php
staabm: good catch. I think [`Runtime::isOpcacheActive()`](https://github.com/sebastianbergmann/environment/pull/75) is already enough as even without JIT opcache might optimize code differently, right?

staabm:
should be ready to land/another review

sebastianbergmann:
I am not sure whether we still need/want this considering https://github.com/sebastianbergmann/php-code-coverage/issues/1022#issuecomment-3207033742.

PR #6308, opened 2025-08-11
Improve output of `--check-php-configuration`

### LINKED ISSUE #6307: `--check-php-configuration`: show current (incorrect) value
```
./vendor/bin/phpunit --check-php-configuration        
PHPUnit 12.3.2 by Sebastian Bergmann and contributors.

Checking whether PHP is configured according to https://docs.phpunit.de/en/12.3/installation.html#configuring-php-for-development

display_errors = On         ... ok
display_startup_errors = On ... ok
error_reporting = -1        ... not ok
zend.assertions = 1         ... ok
assert.exception = 1        ... ok
memory_limit = -1           ... not ok
```

This is my result. I now know that I should probably change my `error_reporting` and `memory_limit` for the CLI, BUT it wasn't my initial thought: at first I was thinking there is a but in PHPUnit since it's showing my that "my memory limit is -1 and it's not ok" xD

Well, actually it's showing the _desired_ value and `ok`/`not ok` – not the _current_ value ;-)

Suggested improvement: if the value is `not ok` -> then show the current value -> it will be more obvious what's going and and easier to fix.

MacDada:
Showing the actual configuration value if the value is not as expected.

Resolves #6307

### DIFF
--- a/src/TextUI/Command/Commands/CheckPhpConfigurationCommand.php
+++ b/src/TextUI/Command/Commands/CheckPhpConfigurationCommand.php
@@ -87,10 +87,12 @@ final class CheckPhpConfigurationCommand implements Command
                 }
             }
 
-            if (ini_get($name) === $setting['expectedValue']) {
+            $actualValue = ini_get($name);
+
+            if ($actualValue === $setting['expectedValue']) {
                 $check = $this->ok();
             } else {
-                $check         = $this->notOk();
+                $check         = $this->notOk($actualValue);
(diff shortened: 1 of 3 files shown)


sebastianbergmann:
Do not worry about the Psalm error(s), I will take care of that.

sebastianbergmann:
Cherry-picked into `10.5` and merged to `11.5`, `12.3`, and `main` from there.

PR #6309, opened 2025-08-11
fix(deprecations): testCaseContext should be computed with class name

nikophil:
This fixes a small bug in https://github.com/sebastianbergmann/phpunit/pull/6293

the so-called "test case context" in `ErrorHandler` was only relying on the test method's name, so two methods with the same name in different test classes would be related to the same deprecation from the data provider, although if one of them does not trigger a deprec

ping @staabm


sebastianbergmann:
This pull request has conflicts that must be resolved.

nikophil:
@sebastianbergmann done

staabm:
Thanks

PR #6310, opened 2025-08-11
test(deprecations): test deprecs in data provider can be ignored

nikophil:
add tests that validates `#[IgnoreDeprecations]`  can be used to ignore deprecs triggered by a data provider

as [suggested](https://github.com/sebastianbergmann/phpunit/pull/6293#issuecomment-3173495032) by @xabbuh 

ping @staabm


xabbuh:
That's what I had in mind. Thank you for doing this @nikophil. 👍

Issue #5863, opened 2024-06-12
TestDox printer does not show previous exception


donquixote:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 11.2.1
| PHP version         | 8.2.15
| Installation Method | Composer

#### Summary

`phpunit --testdox` does not show previous exception.

#### Current behavior

(good) Output for `phpunit ExampleTest.php` without `--testdox` shows outer and inner exception.
(bad) Output for `phpunit --testdox ExampleTest.php` shows outer exception, but not the inner (previous) exception.

```
~/projects/phplib/phpunit-test $ ./vendor/bin/phpunit --testdox ExampleTest.php 
PHPUnit 11.2.1 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.2.15

E                                                                   1 / 1 (100%)

Time: 00:00.008, Memory: 4.00 MB

Example
 ✘ Foo
   │
   │ Exception: Outer
   │
   │ /home/lemonhead/projects/phplib/phpunit-test/ExampleTest.php:10
   │

ERRORS!
Tests: 1, Assertions: 0, Errors: 1.
```
#### How to reproduce

```sh
mkdir testdir
cd testdir
composer require phpunit/phpunit
nano ExampleTest.php
```
Put the following php code:
```php
<?php

declare(strict_types=1);

use PHPUnit\Framework\TestCase;

class ExampleTest extends TestCase {

  public function testFoo() {
    throw new \Exception('Outer', previous: new \Exception('Inner'));
  }

}
```
Run these commands:
```sh
./vendor/bin/phpunit ExampleTest.php
./vendor/bin/phpunit --testdox ExampleTest.php
```

#### Expected behavior
Output for both commands contains "Exception: Outer" and "Exception: Inner".

#### Related issues
There are other issues and PRs that deal with showing previous exception in different use cases.
But it seems this was not done yet for testdox.

#### About my use case
I personally would not mind to run the test without `--testdox`.
But I am currently using a tool that adds `--testdox` parameter by default. See https://github.com/ddev/ddev-drupal-contrib/blob/903026a2fef564b1b7b1fb54213ebb669267995b/commands/web/phpunit#L21
For the steps above I isolated the problem to a minimal project with the latest phpunit I get with composer.

staabm:
Fixed with https://github.com/sebastianbergmann/phpunit/pull/6316

PR #6315, opened 2025-08-13
Print previous exception in Testdox

staabm:
closes https://github.com/sebastianbergmann/phpunit/issues/5863

color rendered result:

grafik


sebastianbergmann:
This should target `11.5`.

PR #6319, opened 2025-08-14
Detect premature end of PHPUnit's main PHP process

staabm:
when a test calls `exit()` phpunit main process halted and printed nothing.
since [PHPUnit 12.3 we have a `ShutdownHandler`](https://github.com/sebastianbergmann/phpunit/pull/6296) which we utilize with this PR to handle this situation and print a more usefull error message.

before this PR:
```
➜  phpunit git:(11.5) ✗ php phpunit Test.php
PHPUnit 11.5.32-19-g48f14e340 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.3.24
```

after this PR:
```
➜  phpunit git:(11.5) ✗ php phpunit Test.php
PHPUnit 11.5.32-19-g48f14e340 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.3.24

Fatal error: Premature end of PHP process when running PHPUnit\TestFixture\Issue6234Test::testExitWithoutProcessIsolation.
```

similar to https://github.com/sebastianbergmann/phpunit/issues/6234
-> [in a separate PR I will add expectation handling for subprocesses](https://github.com/sebastianbergmann/phpunit/pull/6275).

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -353,7 +353,7 @@ abstract class TestCase extends Assert implements Reorderable, SelfDescribing, T
 
         if (!$this->shouldRunInSeparateProcess() || $this->requirementsNotSatisfied()) {
             try {
-                ShutdownHandler::setMessage('Fatal error: Premature end of PHP process.');
+                ShutdownHandler::setMessage(sprintf('Fatal error: Premature end of PHP process in %s.', $this->toString()));
                 (new TestRunner)->run($this);
             } finally {
                 ShutdownHandler::resetMessage();
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -353,7 +353,7 @@ abstract class TestCase extends Assert implements Reorderable, SelfDescribing, T
 
         if (!$this->shouldRunInSeparateProcess() || $this->requirementsNotSatisfied()) {
             try {
-                ShutdownHandler::setMessage(sprintf('Fatal error: Premature end of PHP process in %s.', $this->toString()));
+                ShutdownHandler::setMessage(sprintf('Fatal error: Premature end of PHP process when running %s.', $this->toString()));
                 (new TestRunner)->run($this);
             } finally {
                 ShutdownHandler::resetMessage();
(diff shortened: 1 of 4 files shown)


src/Framework/TestCase.php
```diff
@@ -352,7 +352,12 @@ final public function run(): void
         }
 
         if (!$this->shouldRunInSeparateProcess() || $this->requirementsNotSatisfied()) {
-            (new TestRunner)->run($this);
+            try {
+                ShutdownHandler::setMessage('Fatal error: Premature end of PHP process.');
```
staabm: not sure whether the error is precise enough?

staabm: added the test-name

src/Runner/ShutdownHandler.php
```diff
@@ -39,6 +39,7 @@ private static function register(): void
             return;
         }
 
+        self::$registered = true;
```
staabm, line 42: fixes failling end-to-end test:
https://github.com/sebastianbergmann/phpunit/actions/runs/16960222681/job/48071048701?pr=6319

staabm: PHPStan should have detected this problem: https://github.com/phpstan/phpstan/issues/13384

tests/end-to-end/regression/6234/Issue6234Test.php
```diff
@@ -0,0 +1,23 @@
...
+use PHPUnit\Framework\TestCase;
+
+final class Issue6234Test extends TestCase
+{
+    public function testExitWithoutProcessIsolation(): void
+    {
+        $this->assertTrue(true);
+        $this->assertTrue(true);
+
+        exit(1);
```
staabm, line 21: just realized there is a even more critical case, namely a test calling `exit(0)`, from non-isolated tests.
this would make PHPUnit return a 0 exit-code and noone would notice (besides the newly added error message), that tests did not run successfully.

stuff for a separate issue though. maybe even stuff for phpstan-phpunit as I can't think of how we can make phpunit return a non-zero exit code, when a test itself calls already `exit(0)`.

src/Framework/TestCase.php
staabm: hmm maybe even `PHPUnit process` instead of `PHP process`?

PR #6320, opened 2025-08-14
Support error_reporting=E_ALL in CheckPhpConfigurationCommand

simPod:
`E_ALL` is same as `-1`

### DIFF
--- a/tests/end-to-end/check-php-configuration/failure-without-xdebug.phpt
+++ b/tests/end-to-end/check-php-configuration/failure-without-xdebug.phpt
@@ -8,7 +8,7 @@ if (extension_loaded('xdebug')) {
 --INI--
 display_errors=0
 display_startup_errors=1
-error_reporting=-1
+error_reporting=E_ALL
 zend.assertions=1
 assert.exception=1
 memory_limit=-1
--- a/src/TextUI/Command/Commands/CheckPhpConfigurationCommand.php
+++ b/src/TextUI/Command/Commands/CheckPhpConfigurationCommand.php
@@ -32,47 +32,49 @@ final readonly class CheckPhpConfigurationCommand implements Command
     /**
      * @var non-empty-array<non-empty-string, array{expectedValue: list<int|non-empty-string>|non-empty-string, valueForConfiguration: non-empty-string, requiredExtensions: list<non-empty-string>}>
      */
-    private const array SETTINGS = [
-        'display_errors' => [
-            'expectedValue'         => '1',
-            'valueForConfiguration' => 'On',
-            'requiredExtensions'    => [],
-        ],
-        'display_startup_errors' => [
-            'expectedValue'         => '1',
-            'valueForConfiguration' => 'On',
-            'requiredExtensions'    => [],
-        ],
-        'error_reporting' => [
-            'expectedValue'         => ['-1', E_ALL],
-            'valueForConfiguration' => '-1',
-            'requiredExtensions'    => [],
-        ],
-        'xdebug.show_exception_trace' => [
-            'expectedValue'         => '0',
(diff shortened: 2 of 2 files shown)


tests/end-to-end/check-php-configuration/success-with-xdebug.phpt
```diff
@@ -8,7 +8,7 @@ if (!extension_loaded('xdebug')) {
 --INI--
 display_errors=1
 display_startup_errors=1
-error_reporting=-1
+error_reporting=-E_ALL
```
staabm: `-E_ALL` (prefixed `-`) ?

sebastianbergmann:
`E_ALL` will not be equal to `-1` when another `E_*` constant is added in the future and the bitmask of `error_reporting` is expanded. `-1` will always mean "everything".

simPod:
Yup, E_ALL is not equal to `-1` value but the behaviour is the same and will always mean "everything" as well.

from docs: 

> Passing in the value -1 will show every possible error, even when new levels and constants are added in future PHP versions. The behavior is equivalent to passing [E_ALL](https://www.php.net/manual/en/errorfunc.constants.php#constant.e-all) constant.

tests/end-to-end/check-php-configuration/success-with-xdebug.phpt
simPod: hm, I guess that test is not testing it. Will investigate.

tests/end-to-end/check-php-configuration/failure-without-xdebug.phpt
```diff
@@ -8,7 +8,7 @@ if (extension_loaded('xdebug')) {
 --INI--
 display_errors=0
 display_startup_errors=1
-error_reporting=-1
```
staabm, line 11: should this test be duplicated so we have coverage for both supported values?

tests/end-to-end/check-php-configuration/success-with-xdebug.phpt
simPod: I have [validated](https://github.com/simPod/phpunit/actions/runs/16963908459/job/48083050651) that `without-xdebug` tests consider error_reporting. For some reason `with-xdebug` tests ignore it but did not dig into it.

tests/end-to-end/check-php-configuration/failure-without-xdebug.phpt
simPod: We have error_reporting=-1 already in 3 other tests (1 without xdebug)

image

simPod:
Tests pass

https://github.com/simPod/phpunit/actions/runs/16964704428?pr=1

I had to move settings to constructor as E_ALL needs to be cast to string because `ini_get()` returns string value.

sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it, as I am not happy with your implementation or the current implementation as a whole.

It would take me longer to explain what I want to do than to do it myself. I currently regret adding the `--check-php-configuration` CLI option in the first place. It was intended as a quick win, making it easy to check whether your PHP is configured sensibly. However, I have now realised that it is much harder to implement correctly than I originally thought.

simPod:
Sure, np. 🤞🏾

sebastianbergmann:
I should have looked more closely at your changes before I dismissed them, sorry. The only thing I was not happy with was setting the property in the constructor. I wanted to cherry-pick your changes into `10.5`, but that resulted in a conflict.

simPod:
Should I rebase? Lemme know.

sebastianbergmann:
It's taken care of.

Issue #6329, opened 2025-08-23
IgnorePhpunitDeprecations does not ignore "Metadata found in doc-comment.." in PHPUnit 11.x


staabm:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     |  11.5.34
| PHP version         |  8.3.2 NTS
| Installation Method | Composer 

#### Summary

inconsistent behaviour with IgnorePhpunitDeprecations

#### How to reproduce

running 

```php
<?php // test.php

class Test extends \PHPUnit\Framework\TestCase {
    /**
     * @dataProvider provideData
     */
    public function testFoo(int $foo): void {
        $this->assertTrue(true);
    }

    static public function provideData(): array {
        return [
            [1],
            [2],
            [3],
        ];
    }
}
```

results in

```
php phpunit test.php
PHPUnit 11.5.34 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.3.24
Configuration: /Users/staabm/workspace/phpunit/phpunit.xml

...                                                                 3 / 3 (100%)

Time: 00:00.004, Memory: 10.00 MB

There was 1 PHPUnit test runner deprecation:

1) Metadata found in doc-comment for method Test::testFoo(). Metadata in doc-comments is deprecated and will no longer be supported in PHPUnit 12. Update your test code to use attributes instead.

OK, but there were issues!
Tests: 3, Assertions: 3, PHPUnit Deprecations: 1.
```

----

adding `#[\PHPUnit\Framework\Attributes\IgnorePhpunitDeprecations]`

```php
<?php

class Test extends \PHPUnit\Framework\TestCase {
    /**
     * @dataProvider provideData
     */
    #[\PHPUnit\Framework\Attributes\IgnorePhpunitDeprecations]
    public function testFoo(int $foo): void {
        $this->assertTrue(true);
    }

    static public function provideData(): array {
        return [
            [1],
            [2],
            [3],
        ];
    }
}
```

... turns it into

```
php phpunit test.php
PHPUnit 11.5.34 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.3.24
Configuration: /Users/staabm/workspace/phpunit/phpunit.xml

E                                                                   1 / 1 (100%)

Time: 00:00.003, Memory: 10.00 MB

There was 1 error:

1) Test::testFoo
[...]

Issue #6304, opened 2025-08-09
Command hangs after update to PHPUnit 11.5.29


jcandan:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 11.5.29
| PHP version         | 8.3.23
| Installation Method | Composer

#### Summary

Since updating from `11.5.28` to `11.5.29`, the `phpunit` command hangs and never indicates tests are being performed after showing the Runtime.

#### Current behavior

The command shows the version and runtime output, then hangs with no further output:

```sh
$ ddev phpunit 
PHPUnit 11.5.29 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.3.23
```

This is also true of Drupal GitLab CI: https://git.drupalcode.org/project/require_on_publish/-/jobs/6159400.

The same tests complete successfully without delay on `11.5.28`.

#### How to reproduce

Unsure what specific details cause working tests to hang on `11.5.29`.

#### Expected behavior

Tests that complete successfully on a previous patch version should complete successfully on a patch update.

sebastianbergmann:
Thank you for your report.

Please provide a **minimal**, **self-contained**, reproducing test case that shows the problem you are reporting.

Without such a minimal, self-contained, reproducing test case I will not be able to investigate this issue.

sebastianbergmann:
Using the test runner's `--debug` CLI option should help finding where it hangs.

mably:
> Using the test runner's `--debug` CLI option should help finding where it hangs.

Debug does not seem to display much information: https://www.drupal.org/project/gitlab_templates/issues/3540591#comment-16222698

fjgarlin:
It's not complete isolation but we have so tests in here: https://www.drupal.org/project/gitlab_templates/issues/3540591

If we use the `--debug` option here https://git.drupalcode.org/project/page_to_pdf/-/jobs/6162154 it just seems to hang on 'Child process started'

On the same project, locking the version to 11.5.28 https://git.drupalcode.org/project/page_to_pdf/-/merge_requests/30/diffs?commit_id=1f61289de256f556304b775a958584f89422ae40, things work again.

fjgarlin:
The same seems to be happening with 11.5.30: https://git.drupalcode.org/project/keycdn/-/jobs/6163515

sebastianbergmann:
> it just seems to hang on 'Child process started'

That would indicate a problem with process isolation.

[...]

PR #6331, opened 2025-08-23
Take #[IgnorePhpunitDeprecations] attribute into account for test runner deprecations

staabm:
similar to https://github.com/sebastianbergmann/phpunit/commit/b3be029cd69c276031342ad13769b5dc0486a035 but with a fix to prevent the endless loop we saw in https://github.com/sebastianbergmann/phpunit/issues/6304#issuecomment-3173303380

closes https://github.com/sebastianbergmann/phpunit/issues/6329

### DIFF
--- a/tests/end-to-end/regression/6304.phpt
+++ b/tests/end-to-end/regression/6304.phpt
@@ -1,0 +1,29 @@
+--TEST--
+https://github.com/sebastianbergmann/phpunit/issues/6304
+--FILE--
+<?php declare(strict_types=1);
+$_SERVER['argv'][] = '--do-not-cache-result';
+$_SERVER['argv'][] = '--no-configuration';
+$_SERVER['argv'][] = '--display-phpunit-deprecations';
+$_SERVER['argv'][] = __DIR__ . '/6304/Issue6304Test.php';
+
+require __DIR__ . '/../../bootstrap.php';
+
+(new PHPUnit\TextUI\Application)->run($_SERVER['argv']);
+--EXPECTF--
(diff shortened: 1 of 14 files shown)


staabm:
@sebastianbergmann your initial commit was not accompanied by a test-case. do you know what was the motivation for the initial commit so we can properly test it?

tests/end-to-end/regression/6304.phpt
```diff
@@ -0,0 +1,29 @@
+--TEST--
```
staabm: added this test from https://github.com/sebastianbergmann/phpunit/issues/6304#issuecomment-3173303380 which seems to not find its way into the codebase?

staabm:
@jcandan @mably @fjgarlin @cmlara @jrfnl this PR re-adds a reworked commit which caused problems for you in https://github.com/sebastianbergmann/phpunit/issues/6304

could you please double check the new version of the initial fix will not lead to the same problems in your projects?

I tried to cover it with tests as good as possible, based on the information given at issue 6304.

jrfnl:
@staabm Thanks for the ping.

Without an opinion on the functionality in the PR.... tested locally on Windows 10 and I don't see the "hanging" anymore. Tried to test in CI on Linux too, but am running into rate limits. Will report back once I get CI running again.

sebastianbergmann:
This looks good to me, thank you for working on this. However, for the `12.3` and `main` branches we will need different tests. I hope to be able to look into this soon.

sebastianbergmann:
Merged manually, thanks.

Issue #6305, opened 2025-08-10
Nightly Build for all branches


sebastianbergmann:
We currently only have a nightly build for `main`.  [This nightly build started failing with the most recent snapshot of PHP 8.5.](https://github.com/sebastianbergmann/phpunit/actions/runs/16855438877/job/47747701586)

However, older branches such as `8.5` have different/more problems with that need to be resolved for PHP 8.5. But we do not see this in our nightly build.

mbeccati:
@sebastianbergmann do you need any help with this?

sebastianbergmann:
Yes, a pull request would help. Thank you!

mbeccati:
@sebastianbergmann I've checked again and all branches are now passing w/ latest PHP master on my CI too: https://github.com/mbeccati/php-latest-builds/actions/runs/17005556973

What kind of problems were you referring to?

sebastianbergmann:
We currently test all branches (`8.5`, `9.6`, `10.5`, `11.5`, `12.3`, `main`) for every push. But we only test `main` every night (even when there was no push). I would like to have the same for the other branches as well.

PR #6332, opened 2025-08-23
Fix test that was passing only when run sequentially

mbeccati:
Failure found here: https://github.com/mbeccati/phpunit-tmp/actions/runs/17174306040/job/48728153064

Refs #6305

### DIFF
--- a/tests/unit/Framework/MockObject/GeneratorTest.php
+++ b/tests/unit/Framework/MockObject/GeneratorTest.php
@@ -98,7 +98,7 @@ final class GeneratorTest extends TestCase
     {
         $this->expectException(RuntimeException::class);
 
-        $this->generator->getMock(stdClass::class, [], [], RuntimeException::class);
+        $this->generator->getMock(stdClass::class, [], [], \InvalidArgumentException::class);
     }
 
     public function testGetMockCanCreateNonExistingFunctions(): void


tests/unit/Framework/MockObject/GeneratorTest.php
```diff
@@ -98,7 +98,7 @@ public function testGetMockThrowsExceptionWithExistingClassAsMockName(): void
     {
         $this->expectException(RuntimeException::class);
 
-        $this->generator->getMock(stdClass::class, [], [], RuntimeException::class);
+        $this->generator->getMock(stdClass::class, [], [], 'RuntimeException');
```
staabm: why does this change fix the test?

staabm:
related error from the logs (which will be non-existant in 30 days) is :

```
1) PHPUnit\Framework\MockObject\GeneratorTest::testGetMockThrowsExceptionWithExistingClassAsMockName
Failed asserting that exception of type "ParseError" matches expected exception "PHPUnit\Framework\MockObject\RuntimeException". Message was: "syntax error, unexpected namespaced name "PHPUnit\Framework\MockObject\R...", expecting identifier" at
D:\a\phpunit-tmp\phpunit-tmp\src\Framework\MockObject\Generator.php:677
D:\a\phpunit-tmp\phpunit-tmp\src\Framework\MockObject\Generator.php:266
D:\a\phpunit-tmp\phpunit-tmp\tests\unit\Framework\MockObject\GeneratorTest.php:101
D:\a\phpunit-tmp\phpunit-tmp\src\Framework\TestCase.php:1504
D:\a\phpunit-tmp\phpunit-tmp\src\Framework\TestCase.php:1124
D:\a\phpunit-tmp\phpunit-tmp\src\Framework\TestResult.php:705
D:\a\phpunit-tmp\phpunit-tmp\src\Framework\TestCase.php:850
D:\a\phpunit-tmp\phpunit-tmp\src\Framework\TestSuite.php:626
D:\a\phpunit-tmp\phpunit-tmp\src\Framework\TestSuite.php:626
D:\a\phpunit-tmp\phpunit-tmp\src\Framework\TestSuite.php:626
D:\a\phpunit-tmp\phpunit-tmp\src\TextUI\TestRunner.php:645
D:\a\phpunit-tmp\phpunit-tmp\src\TextUI\Command.php:236
D:\a\phpunit-tmp\phpunit-tmp\src\TextUI\Command.php:195
```

tests/unit/Framework/MockObject/GeneratorTest.php
staabm: this string is likely beeing re-written by a codestyle tool to the previous `::class` line again.
would adding a `use RuntimeException` at the top of the file also fix the problem?

mbeccati: The previous code references the class in the local namespace, which is not loaded yet if the test is ran alone (or with a random order). Calling `new RuntimeException` or using any defined class (e.g. `\RuntimeException::class`) makes the test pass. The ParseError is caused by the template trying to define a class with the full namespace as class name, so I thought that the parameter was supposed to be the actual class name without namespace.

staabm: sounds like the test expects https://github.com/sebastianbergmann/phpunit/blob/8.5/src/Framework/MockObject/Exception/RuntimeException.php to be loaded.

if this is the case, maybe we should call `class_exists(RuntimeException::class)` to setup the test (instead of relying on other tests doing so)

mbeccati: Any defined class will do the job in terms of what is being tested...

mbeccati: @staabm I've changed both the class name being used and the notation, to make the intent clearer: it is not necessary that the `$mockClassName` parameter is the same as the exception being thrown.

Issue #6335, opened 2025-08-25
Trigger a test runner warning when both attributes and annotations are used


sebastianbergmann:
Follows https://github.com/sebastianbergmann/phpunit/issues/6329#issuecomment-3216561365.

PR #6336, opened 2025-08-26
Trigger a test runner warning when both attributes and annotations are used

staabm:
closes https://github.com/sebastianbergmann/phpunit/issues/6335

### DIFF
--- a/src/Metadata/Parser/ParserChain.php
+++ b/src/Metadata/Parser/ParserChain.php
@@ -12,6 +12,7 @@ namespace PHPUnit\Metadata\Parser;
 use function assert;
 use function class_exists;
 use function method_exists;
+use PHPUnit\Event\Facade as EventFacade;
 use PHPUnit\Metadata\MetadataCollection;
 
 /**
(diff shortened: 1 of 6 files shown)


src/Metadata/Parser/ParserChain.php
```diff
@@ -37,13 +38,20 @@ public function forClass(string $className): MetadataCollection
...
-        $metadata = $this->attributeReader->forClass($className);
+        $attributes  = $this->attributeReader->forClass($className);
+        $annotations = $this->annotationReader->forClass($className);
 
-        if (!$metadata->isEmpty()) {
-            return $metadata;
+        if (!$attributes->isEmpty()) {
+            if (!$annotations->isEmpty()) {
+                EventFacade::emitter()->testRunnerTriggeredPhpunitWarning(
+                    'PHPDoc annotations will be ignored when attributes are declared',
```
staabm, line 47: maybe we need to be more precise here, that we talk only about PHPUnit related annotations.. ?

wording suggestions welcome.

staabm:
when projects support a wide range of php versions, they likely need to run their test-suite across multiple phpunit versions (therefore they might configure their tests with phpdoc and attributes in tandem)

see e.g. https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/pull/8965#issuecomment-3216542707

the warning might get annoying in such cases

src/Metadata/Parser/ParserChain.php
```diff
@@ -37,13 +38,20 @@ public function forClass(string $className): MetadataCollection
...
         assert(class_exists($className));
 
-        $metadata = $this->attributeReader->forClass($className);
+        $attributes  = $this->attributeReader->forClass($className);
+        $annotations = $this->annotationReader->forClass($className);
 
-        if (!$metadata->isEmpty()) {
-            return $metadata;
+        if (!$attributes->isEmpty()) {
+            if (!$annotations->isEmpty()) {
```
mvorisek, line 45: In projects with old PHP versions support, multiple PHPUnit versions needs to be supported.

In order to be able to run multiple PHPUnit versions, annotations might be needed to be supported together with attributes.

The warning should be only if the annotations and attributes does not equal.

sebastianbergmann: > The warning should be only if the annotations and attributes does not equal.

This is not possible without parsing the annotations and PHPUnit >= 12 does not have the code for that anymore.

> multiple PHPUnit versions needs to be supported

I disagree.

staabm: Ok, so seems this is ready to land

sebastianbergmann:

PR #6338, opened 2025-08-27
Rewrite of tests for test reordering

sebastianbergmann:
Motivated by https://github.com/sebastianbergmann/phpunit/pull/6270#issuecomment-3174755852, @staabm and I discussed the current implementation of `PHPUnit\Runner\TestSuiteSorter` today.

We identified code in `PHPUnit\Runner\TestSuiteSorter` that is only used in the tests for this class. This code is called twice and is potentially slow. Markus performed a quick benchmark ...

5e3b1f7f9c874d67

... and we can see a miniscule, yet measurable, performance improvement after removing this code when running PHPUnit's own test suite.

We both do not understand the existing tests for test reordering and decided to delete them.

I have started to write end-to-end, event-based tests for common use cases of test reordering. In doing so, I already discovered behaviour that surprises me. I think of these tests as characterization tests that document the status quo.

Once the characterization tests are sufficient, we can decide on which behaviour to keep as-is, which behaviour that is currently "broken" (or is not as we assumed it to be) we want to change, and which behaviour to deprecate and later remove.

This pull request targets the `10.5` branch as I would like to have the same tests for PHPUnit >= 10. Apart from removing the aforementioned code in `PHPUnit\Runner\TestSuiteSorter` that is only used by tests, I do not intend to make any production code changes for PHPUnit 10.5. Pending our evaluation after we have a sufficient amount of characterization tests, we will fix bugs in PHPUnit >= 11.


sebastianbergmann:
The new end-to-end tests for test suite sorting are now at a point where only running them results in 100% Line Coverage and 100% Branch Coverage for the `TestSuiteSorter` class.

I am sure, however, that we are still missing more tests for combinations of sorting strategies. This is what I will look into next.

PR #6339, opened 2025-08-27
Prevent unnecessary duplicate work in `NamePrettifier->prettifyTestCase()`

staabm:
shows up in blackfire perf profiles

grafik


staabm:
the last commit would also apply to 11.x while the first/second ones only appliy to 12.x

PR #6341, opened 2025-08-28
Support for regular expressions with `#[IgnoreDeprecations]` attribute

nikophil:
This will allow to only ignore specific deprecations, instead of all of them

### DIFF
--- a/src/Metadata/IgnoreDeprecations.php
+++ b/src/Metadata/IgnoreDeprecations.php
@@ -16,8 +16,30 @@ namespace PHPUnit\Metadata;
  */
 final readonly class IgnoreDeprecations extends Metadata
 {
+    /** @var null|non-empty-string */
+    private ?string $messagePattern;
+
+    /**
+     * @param int<0, 1>             $level
+     * @param null|non-empty-string $messagePattern
+     */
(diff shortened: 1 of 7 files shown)


src/Metadata/IgnoreDeprecations.php
```diff
@@ -9,15 +9,37 @@
...
+
+        $this->messagePattern = $messagePattern;
+    }
+
     public function isIgnoreDeprecations(): true
     {
         return true;
     }
+
+    public function shouldIgnore(string $message): bool
```
sebastianbergmann: I am not comfortable with the fact that a value object, which is what a `Metadata` object is supposed to be, makes a decision like that.

This decision should be moved to `ErrorHandler::deprecationIgnoredByTest()`, for example.

nikophil: done.

I must mention that I did the same thing in `IgnorePhpunitWarnings`: https://github.com/sebastianbergmann/phpunit/blob/main/src/Metadata/IgnorePhpunitWarnings.php#L40

do you want that I change that too?

sebastianbergmann: Yes, please. This should be a separate PR to the `12.3` branch. Thanks!

PR #6352, opened 2025-09-05
Add a trailing \n in ShutdownHandler message

lyrixx:
The first execution is without my patch
The second one is with my patch.

As you can see, the output is cleaner with it

image

### DIFF
--- a/src/Runner/ShutdownHandler.php
+++ b/src/Runner/ShutdownHandler.php
@@ -10,6 +10,7 @@
 namespace PHPUnit\Runner;
 
 use function register_shutdown_function;
+use function rtrim;
 
 /**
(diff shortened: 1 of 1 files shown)


staabm:
needs a description why you want to change it and at best a test

lyrixx:
thanks, I added a description to the PR

sebastianbergmann:
> thanks, I added a description to the PR

In the future, [please do not post screenshots of code or output](https://github.com/sebastianbergmann/phpunit/blob/main/.github/CONTRIBUTING.md#write-bug-reports-with-detail-background-and-sample-code). Use proper markup instead. Thank you.

sebastianbergmann:

stof:
shouldn't the trailing newline be added only when the message is not empty ?

lyrixx:
Good idea. I opened a PR

PR #6353, opened 2025-09-06
Disable performance optimization for tests requiring on Xdebug

staabm:
closes https://github.com/sebastianbergmann/phpunit/issues/6105

PR #6359, opened 2025-09-11
Properly skip new test

staabm:
refs https://github.com/sebastianbergmann/phpunit/pull/6353#issuecomment-3279152242

a followup could be to run the test also with xdebug loaded

### DIFF
--- a/tests/end-to-end/regression/6105.phpt
+++ b/tests/end-to-end/regression/6105.phpt
@@ -4,6 +4,21 @@ https://github.com/sebastianbergmann/phpunit/issues/6105
 <?php if(!extension_loaded('xdebug')) {
     print 'skip: xdebug is not loaded';
 }
+
+if (ini_get('xdebug.start_with_request') === "1") {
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it as I have already implemented a similar solution. Sorry :(

tests/end-to-end/regression/6105.phpt
```diff
@@ -4,6 +4,21 @@ https://github.com/sebastianbergmann/phpunit/issues/6105
 <?php if(!extension_loaded('xdebug')) {
     print 'skip: xdebug is not loaded';
 }
+
+if (ini_get('xdebug.start_with_request') === "1") {
+    print 'skip: xdebug emits a warning when xdebug.start_with_request=1 which breaks output expectations';
+}
```
staabm, line 10: would be great to have this condition also in the test, as I am using it in my default settings locally

PR #6360, opened 2025-09-11
 Improve 2448-not-existing-test.phpt 

staabm:
this test failed when the phpunit project root folder contains a "test.php" file - which I regularily use to test things ;)


sebastianbergmann:
Good catch, thanks. Maybe sqash into single commit?

staabm:
here we go (sent the initial PR in a hurry, sorry)

PR #6364, opened 2025-09-12
`--filter` format used by PhpStorm stopped working

schlndh:
It was broken by PR #6272.  As far as I can tell the goal was to change the test name printed by the default printer to make it easier to copy-paste it into --filter.

Example from PHPUnit's own tests:

```
1) PHPUnit\TestFixture\DataProviderFilterTest::testFalse@other false test with data (false)
```

Can be copy-pasted into `--filter` like this:

```
--filter 'testFalse@other false test'
```

However, IMO the PR changed the test name more broadly than necessary. This broke the --filter pattern used by PHPStorm (i.e. `No tests executed!`). E.g:

```
--filter '/PHPUnit\\TestFixture\\DataProviderFilterTest::testFalse with data set \"other false test\"$/'
```

It would have have an extra `@` like this:

```
--filter '/PHPUnit\\TestFixture\\DataProviderFilterTest::testFalse with data set \"@other false test\"$/'
```

But the `@` is not present in the teamcity output that PHPStorm uses (they seem to use the `locationHint`). So PHPStorm (and similar tools) would have to hack the test name to add it (which would presumably make the whole thing even more tied to PHPUnit's internal behavior).

IMO the `--filter` used by PHPStorm is pretty reasonable, so I'm attempting to restore support for it, while maintaining the new name used by the default printer for easy copy-pasting.


sebastianbergmann:
CC @staabm

staabm:
@schlndh thanks for reporting. 

Could you describe how to reproduce the problem jn PHPStorm?

staabm:
If phpstorm is using data from a formatter and feeds it back into the filter cli option we  should have a test for this e2e scenario

schlndh:
@staabm I do it like this:
- Run a test with a bunch of cases from a data provider.
- Click on one of the test cases and run just that one (especially useful for debugging failing cases one-by-one).

Here is a screen recording to make it even easier to understand:

12.3 branch:

https://github.com/user-attachments/assets/9d9ee5d6-7946-4a3b-a2b4-df63f01cd23c

My branch:


https://github.com/user-attachments/assets/3ef469d3-4e82-458d-a8c2-9e31b34b4621

src/Runner/Filter/NameFilterIterator.php
```diff
@@ -109,7 +109,7 @@ private function prepareFilter(string $filter): array
             //  * testDetermineJsonError@JSON.*
             elseif (preg_match('/^(.*?)@(.+)$/', $filter, $matches)) {
                 $filter = sprintf(
-                    '%s.*with data set "@%s"$',
+                    '%s.*with data set "%s"$',
```
staabm, line 112: just to re-iterate: this reverts my change which caused the bug/BC break.

tests/end-to-end/cli/filter/filter-dataprovider-by-phpstorm-regexp.phpt
```diff
@@ -0,0 +1,22 @@
+--TEST--
```
staabm: we need another test with a numbered dataprovider. this one here covers only named dataproviders.

and I think we should land the new tests on PHPUnit 11+ to make sure all PHPUnit versions are on the same page

tests/end-to-end/cli/filter/filter-dataprovider-by-phpstorm-regexp.phpt
```diff
@@ -0,0 +1,22 @@
...
+phpunit --filter '/PHPUnit\\TestFixture\\DataProviderFilterTest::testFalse with data set \"other false test\"$/' ../../_files/DataProviderFilterTest.php
```
staabm: usually we put a url to the underlying issue/pr in `--TEST--`.

in this case it would be
```
--TEST--
https://github.com/sebastianbergmann/phpunit/pull/6364
```

staabm (commented):
I think this looks mostly good. thanks for the in-detail repro and the time you invested in making the screen recordings.

tests/end-to-end/cli/filter/filter-dataprovider-by-phpstorm-regexp.phpt
sebastianbergmann: Maybe even for PHPUnit >= 10.

staabm:
And as mentioned above, I think we should have a separate e2e test which feeds the output of the teamcity formatter back into the --filter option, so we mimick the phpstorm use-case as close as possible

src/Runner/Filter/NameFilterIterator.php
schlndh: The real bug/BC break was in `dataSetAsString`. This just has to match it, so that the short `test@data` filter syntax works. The PHPStorm's filter doesn't execute this code at all, because it's a full regex pattern (i.e. starting with `/`).

schlndh:
> And as mentioned above, I think we should have a separate e2e test which feeds the output of the teamcity formatter back into the --filter option, so we mimick the phpstorm use-case as close as possible

Do you have any suggestion about the best way to do it? I tried first running the whole test class with `--log-teamcity`, parsing the `locationHint` and running the test again with `--filter`. Unfortunately, it seems that `(new PHPUnit\TextUI\Application)->run($_SERVER['argv']);` cannot be called twice from one process, because of all the static properties it initializes.

I suppose that a weaker alternative might be hard-coding the `--filter` and running the test with `--teamcity` to make sure that the `locationHint` matches the contents of the filter. But I guess it would be easy to overlook this intentional coupling if the test breaks in the future and it could be fixed incorrectly.

staabm:
> Do you have any suggestion about the best way to do it? I tried first running the whole test class with `--log-teamcity`, parsing the `locationHint` and running the test again with `--filter`. Unfortunately, it seems that `(new PHPUnit\TextUI\Application)->run($_SERVER['argv']);` cannot be called twice from one process, because of all the static properties it initializes.

I think you could either
- use `exec` or similar to let phpunit ran in a subprocess and work on the resulting xml (feels closer to PHPStorm)
- or use `TeamCityLogger` directly to let it build the xml

maybe @sebastianbergmann has a better idea

tests/end-to-end/cli/filter/filter-dataprovider-by-phpstorm-regexp.phpt
schlndh: I opened a separate [PR](https://github.com/sebastianbergmann/phpunit/pull/6367) that adds E2E tests both for named and numbered data providers. If you want, I can also add a simpler version of the test like this to it. Otherwise, I'd just delete the test from this PR.

schlndh:
@sebastianbergmann Since the PhpStorm team hasn't responded to [your question](https://github.com/sebastianbergmann/phpunit/pull/6367#issuecomment-3288676389) on the E2E test PR, can we proceed with this PR regardless? IMO the fix makes sense in general, not just due to PhpStorm. I'd just add a simple test for the numbered dataprovider here for completeness.

sebastianbergmann:
I have merged the tests from #6367 after they were approved by @smirok on behalf of the PhpStorm team.

On the `12.4`, `12.5`, and `main` branches, the `filter-dataprovider-by-phpstorm-regex-string-key.phpt` is marked as "expected to fail". This PR (or any other PR aiming to address this issue) shoud remove the `XFAIL` section from this test.

sebastianbergmann:
@staabm What do you think? Thanks!

staabm:
So I justed tested this against the goals of https://github.com/sebastianbergmann/phpunit/pull/6272:

after applying this diff (to make tests fail)
```diff
diff --git a/tests/_files/DataProviderFilterTest.php b/tests/_files/DataProviderFilterTest.php
index 1c05d2eca..f26af6d9f 100644
--- a/tests/_files/DataProviderFilterTest.php
+++ b/tests/_files/DataProviderFilterTest.php
@@ -14,11 +14,12 @@
 
 final class DataProviderFilterTest extends TestCase
 {
+
     public static function truthProvider(): array
     {
         return [
             [true],
-            [true],
+            [false],
             [true],
             [true],
         ];
@@ -28,7 +29,7 @@ public static function falseProvider(): array
     {
         return [
             'false test'        => [false],
-            'false test 2'      => [false],
+            'false test 2'      => [true],
             'other false test'  => [false],
             'other false test2' => [false],
         ];
```

I can run

```
➜  phpunit git:(12.4) ✗ php phpunit tests/_files/DataProviderFilterTest.php --filter 'testTrue#1'
PHPUnit 12.4.0-22-g80e38f394 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.3.26
Configuration: /Users/m.staab/dvl/phpunit/phpunit.xml

Time: 00:00.003, Memory: 18.00 MB

There was 1 failure:

1) PHPUnit\TestFixture\DataProviderFilterTest::testTrue#1 with data (false)
Failed asserting that false is true.

/Users/m.staab/dvl/phpunit/src/Framework/Constraint/Constraint.php:120
/Users/m.staab/dvl/phpunit/src/Framework/Constraint/Constraint.php:73
/Users/m.staab/dvl/phpunit/src/Framework/Assert.php:2596
/Users/m.staab/dvl/phpunit/src/Framework/Assert.php:1599
/Users/m.staab/dvl/phpunit/tests/_files/DataProviderFilterTest.php:41
/Users/m.staab/dvl/phpunit/src/Framework/TestCase.php:1317
/Users/m.staab/dvl/phpunit/src/Framework/TestCase.php:517
/Users/m.staab/dvl/phpunit/src/Framework/TestRunner/TestRunner.php:99
/Users/m.staab/dvl/phpunit/src/Framework/TestCase.php:357
/Users/m.staab/dvl/phpunit/src/Framework/TestSuite.php:374
/Users/m.staab/dvl/phpunit/src/Framework/TestSuite.php:374
/Users/m.staab/dvl/phpunit/src/TextUI/TestRunner.php:64
/Users/m.staab/dvl/phpunit/src/TextUI/Application.php:229

FAILURES!
➜  phpunit git:(pr/6364) ✗ php phpunit tests/_files/DataProviderFilterTest.php --filter 'DataProviderFilterTest::testFalse@false test 2'
PHPUnit 12.4.0-22-g80e38f394 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.3.26
Configuration: /Users/m.staab/dvl/phpunit/phpunit.xml

Time: 00:00.003, Memory: 18.00 MB

There was 1 failure:

1) PHPUnit\TestFixture\DataProviderFilterTest::testFalse@false test 2 with data (true)
Failed asserting that true is false.

/Users/m.staab/dvl/phpunit/src/Framework/Constraint/Constraint.php:120
/Users/m.staab/dvl/phpunit/src/Framework/Constraint/Constraint.php:73
/Users/m.staab/dvl/phpunit/src/Framework/Assert.php:2596
/Users/m.staab/dvl/phpunit/src/Framework/Assert.php:1623
/Users/m.staab/dvl/phpunit/tests/_files/DataProviderFilterTest.php:47
/Users/m.staab/dvl/phpunit/src/Framework/TestCase.php:1317
/Users/m.staab/dvl/phpunit/src/Framework/TestCase.php:517
/Users/m.staab/dvl/phpunit/src/Framework/TestRunner/TestRunner.php:99
/Users/m.staab/dvl/phpunit/src/Framework/TestCase.php:357
/Users/m.staab/dvl/phpunit/src/Framework/TestSuite.php:374
/Users/m.staab/dvl/phpunit/src/Framework/TestSuite.php:374
/Users/m.staab/dvl/phpunit/src/TextUI/TestRunner.php:64
/Users/m.staab/dvl/phpunit/src/TextUI/Application.php:229

FAILURES!
➜  phpunit git:(pr/6364) ✗ php phpunit tests/_files/DataProviderFilterTest.php --filter 'PHPUnit\\TestFixture\\DataProviderFilterTest::testFalse@false test 2'
PHPUnit 12.4.0-22-g80e38f394 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.3.26
Configuration: /Users/m.staab/dvl/phpunit/phpunit.xml

Time: 00:00.003, Memory: 18.00 MB

There was 1 failure:

1) PHPUnit\TestFixture\DataProviderFilterTest::testFalse@false test 2 with data (true)
Failed asserting that true is false.

/Users/m.staab/dvl/phpunit/src/Framework/Constraint/Constraint.php:120
/Users/m.staab/dvl/phpunit/src/Framework/Constraint/Constraint.php:73
/Users/m.staab/dvl/phpunit/src/Framework/Assert.php:2596
/Users/m.staab/dvl/phpunit/src/Framework/Assert.php:1623
/Users/m.staab/dvl/phpunit/tests/_files/DataProviderFilterTest.php:47
/Users/m.staab/dvl/phpunit/src/Framework/TestCase.php:1317
/Users/m.staab/dvl/phpunit/src/Framework/TestCase.php:517
/Users/m.staab/dvl/phpunit/src/Framework/TestRunner/TestRunner.php:99
/Users/m.staab/dvl/phpunit/src/Framework/TestCase.php:357
/Users/m.staab/dvl/phpunit/src/Framework/TestSuite.php:374
/Users/m.staab/dvl/phpunit/src/Framework/TestSuite.php:374
/Users/m.staab/dvl/phpunit/src/TextUI/TestRunner.php:64
/Users/m.staab/dvl/phpunit/src/TextUI/Application.php:229

FAILURES!
```

which is what the initial PR wanted to implement.

therefore I think this PR is ready to land.

thanks @schlndh

PR #6367, opened 2025-09-13
End-to-End test(s) simulating how PhpStorm invokes PHPUnit

schlndh:
I'm adding an E2E test simulating how (as far as I can tell) PHPStorm runs individual tests, as requested [here](https://github.com/sebastianbergmann/phpunit/pull/6364#issuecomment-3287706580).


sebastianbergmann:
@pronskiy Maybe someone from the PhpStorm team can have a look at #6364 and this?

smirok:
Hi @sebastianbergmann, PhpStorm team is here! Thanks for the fix, I approve it.

tests/end-to-end/cli/filter/filter-dataprovider-by-phpstorm-regex-number-key.phpt
```diff
@@ -0,0 +1,72 @@
...
+    return 0;
+}
+
+if (preg_match('#php_qn://(?:[A-Z]:)?[^:]*::\\\\(.*)#', $matches[1], $locationHintMatches) !== 1) {
+    echo "Failed to parse locationHint.\n";
+    echo $matches[1];
+
+    return 0;
+}
+
```
sebastianbergmann, line 43: Should this not be two separate PHPT files? It looks wrong to me to invoke PHPUnit twice in a single PHPT file, once using `proc_open()` and once using `run()`.

schlndh: The test makes sure that I can use the `locationHint` to construct `--filter`. Therefore, it seems best to obtain `locationHint` dynamically. I presume that I can't do that with two independent test files.

Or:

> I suppose that a weaker alternative might be hard-coding the --filter and running the test with --teamcity to make sure that the locationHint matches the contents of the filter. But I guess it would be easy to overlook this intentional coupling if the test breaks in the future and it could be fixed incorrectly.

sebastianbergmann: Thank you.

PR #6380, opened 2025-10-07
Allow `Throwable` in `expectExceptionObject()`

VincentLanglet:
- expectException works with a Throwable class and is typed `class-string<Throwable>`
- Throwable has both Message and Code

So this method could works on Throwable.

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -1083,7 +1083,7 @@ abstract class TestCase extends Assert implements Reorderable, SelfDescribing, T
      * Information for expected exception class, expected exception message, and
      * expected exception code are retrieved from a given Exception object.
      */
-    final protected function expectExceptionObject(\Throwable $exception): void
+    final protected function expectExceptionObject(Throwable $exception): void
     {
         $this->expectException($exception::class);
         $this->expectExceptionMessage($exception->getMessage());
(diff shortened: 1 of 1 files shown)


staabm:
Do you have a real world use-case in mind?

VincentLanglet:
> Do you have a real world use-case in mind?

Sure, multiple.

Let's say I have a test with something like
```
$this->someSubService
            ->expects($this->once())
            ->method('execute')
            ->willThrowException($thrownException);

if (null !== $expectedException) {
     $this->expectExceptionObject($expectedException);
}
$this->myHandler->__invoke($message);
```
I'm restricted to the typehint
```
public function testThrowException(\Throwable $thrownException, ?\Exception $expectedException): void
```
because even if the Throwable is accepted by willThrowException it is not by expectExceptionObject.

For the same reason, I cannot do expectation on Error, like TypeError, ValueError, and so on while I can configure my mock to throw one.

I can still use expectException which supports Throwable, and combine it with expectExceptionMessage.
But I don't see a reason to not support it in expectExceptionObject too then.

sebastianbergmann:
> I cannot do expectation on Error, like TypeError, ValueError, and so on while I can configure my mock to throw one.

To me, that is not a valid argument, because expecting PHP errors such as `TypeError` does not make sense, at least not to me. I do see the asymmetry between `willThrowException()` and `expectExceptionObject()`, though.

PR #6386, opened 2025-10-10
Fix PHPStan missingType.generics ignore

staabm:


staabm:
I was not sure about the target branch for this fix. in case you prefer 11.x use this one instead: https://github.com/sebastianbergmann/phpunit/pull/6387

PR #6387, opened 2025-10-10
Fix PHPStan missingType.generics ignore

staabm:


staabm:
I was not sure about the target branch for this fix. in case you prefer 12.x use this one instead: https://github.com/sebastianbergmann/phpunit/pull/6386

src/Framework/MockObject/Generator/Generator.php
```diff
@@ -1110,11 +1110,15 @@ private function instantiateProxyTarget(?object $proxyTarget, object $object, st
...
+     * @template T of object
+     *
+     * @param class-string<T> $className
      *
      * @throws ReflectionException
      *
-     * @phpstan-ignore missingType.generics, throws.unusedType
+     * @return ReflectionClass<T>
+     *
+     * @phpstan-ignore throws.unusedType
```
staabm, line 1121: PHPStan "knows" that `new ReflectionClass($className)` cannot throw, in case you pass it a `class-string`.
thats why we need to ignore it manually, as I think we cannot guarantee that `$className` is 100% sure a valid class-name which is autoloadable

see https://github.com/phpstan/phpstan/issues/13667

sebastianbergmann:
PHPStan errors?

staabm:
ahh interessting. we don't have this errors on 12.x, because there we have more precise phpdocs arround this place.

maybe its easier to merge it in 12.x + and call it a day

PR #6398, opened 2025-10-25
Static Source Path Matching

### LINKED ISSUE #6114: Reimplement `SourceFilter::includes()` to not depend on `SourceMapper` and to not perform filesystem traversal

dantleech:
This PR is a proof-of-concept for statically matching paths without traversing the filesystem during the test run. It replaces #6146 and uses the [webmozarts/glob](https://github.com/webmozarts/glob) package.

Would fix #6114 

---

The primary goal is to ensure that the source mapper needn't iterate over all included directories recursively whenever the source map is required to validate if a file belongs to the mapped files or not - for example when a deprecation is encountered PHPUnit needs to know if the deprecation was issued from source code within the project's responsiblity - i.e. source that is mapped.

We can determine if a file is within the included source by converting the glob-patterns in the `<directory>` element to **regexes**. Currently the `<directory>` element in `<include>` and `<exclude>` has the attributes `prefix` and `suffix` and we also have `<file>` which specifies a single file.

This is more complicated than it could be:

- Current matching/traversal logic depends on PHP's `glob` function - the implementation   of which is not consistent across platforms and which has a number of rarely-used operators which while not common, would present a B/C break if they were removed.

We can expect the following breaks in behavior:

- Differences in the more esoteric syntax between the internal `glob` function and the
  `webmozarts` library. Note that the behavior of `glob` is also dependent on
  the platform making it hard-to-impossible to lock down the behavior
  completely (for reference [linux man page for glob](https://man7.org/linux/man-pages/man7/glob.7.html)).
- Consequently the behavior of `phpunit/php-file-iterator` will not exactly match the results of the static matching logic.

Things to consider:

- Making this an experiemental "feature flag"
- Updating the `phpunit/php-file-iterator` to use the same package so as to be consistent.
- Implementing the logic within PHPUnit rather than using an external package.
- Documenting the glob syntax.


composer.json
```diff
@@ -114,5 +115,8 @@
         "branch-alias": {
             "dev-main": "12.5-dev"
         }
+    },
+    "require-dev": {
+        "symfony/var-dumper": "^7.3"
```
dantleech: remove this

src/TextUI/Configuration/README.md
```diff
@@ -0,0 +1,128 @@
+SourceMapper Implementation Notes
```
dantleech: remove this

tests/unit/TextUI/SourceFilterTest.php
```diff
@@ -167,18 +179,18 @@ public static function provider(): array
...
+            //        self::fixturePath('a/PrefixSuffix.php') => false,
+            //    ],
+            //    self::createSource(
+            //        includeDirectories: FilterDirectoryCollection::fromArray(
+            //            [
+            //                new FilterDirectory(self::fixturePath(), 'a/*', '.php'),
+            //            ],
+            //        ),
+            //    ),
+            //],
```
dantleech: should this be commented ?

tests/unit/TextUI/SourceFilterTest.php
```diff
@@ -311,24 +323,25 @@ public static function provider(): array
...
+            //            [
+            //                new FilterDirectory(
+            //                    self::fixturePath('a/c/Z**'),
+            //                    '',
+            //                    '.php',
+            //                ),
+            //            ],
+            //        ),
+            //    ),
+            //],
```
dantleech: this just seems like very strange behavior (passes on main, fails here). (will remove commented code)

tests/unit/TextUI/SourceFilterTest.php
```diff
@@ -311,24 +321,6 @@ public static function provider(): array
...
-                        [
-                            new FilterDirectory(
-                                self::fixturePath('a/c/Z**'),
-                                '',
-                                '.php',
-                            ),
-                        ],
-                    ),
-                ),
-            ],
```
dantleech: this test, that passes on `main`, seems more like a bug, e.g. the directory path `a/c/Z**` wil include `a/c/PrefixSuffix.php`

sebastianbergmann:
What is the status of this? Thanks.

dantleech:
Hey, basically this is a proposal - and it should work but would need more time invested in testing, exploring the behavior and perhaps adding more test cases and as above, to consider:

- Should this be a"feature flag"?
- Do we want/need to update the phpunit/php-file-iterator to use the same package so as to be consistent.
- Do we want to use the external package or "vendor" the code within PHPUnit.
- Do we want to document the (new) glob syntax, and I guess, the difference between the old and new.

sebastianbergmann:
I started working on a new library, [file-filter](https://github.com/sebastianbergmann/file-filter), for this. This allows to work on this outside of PHPUnit and makes it easy to use the the functionality in both PHPUnit and php-code-coverage.

Help on this new library would be much appreciated. The most obvious thing to do right now would be to add more test cases.

dantleech:
> I started working on a new library, [file-filter](https://github.com/sebastianbergmann/file-filter)

that looks great!

>  Help on this new library would be much appreciated. 

I have to regretfully back away a bit, I've already got a number of unfinshed open-source plates spinning and I'm not sure I can add more value then what you've already provided there. But feel free to reach out with any specifics I can help with.

sebastianbergmann:
Superseded by #6493.

Issue #6401, opened 2025-11-05
PHPUnit hangs when using process isolation on FreeBSD


Rafagd:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 8.4.?
| PHP version         | 12.4.?
| Installation Method | Both Composer and Packaged

#### Summary

When using process isolation to run tests in separate processes, phpunit hangs right after it forks the child process and hangs there indefinitely.

I'm not going to disclose too much info about our configuration, as I haven't been formally allowed to do so, but it's running PHP 8.4, PHPUnit 12.4 and FreeBSD (which version I don't remember, but should be relatively new).

I have tried to reproduce the error using my machine (Debian Forky) and my work machine (WSL2 running Debian Trixie) without success, so I'm assuming it's either a FreeBSD quirk or it's some arcane combination of settings in our server's php.ini. 

I have investigated the error, it seems like when you generate the php file that will execute the test in the new process, there's a series of functions that are called with error suppression turned on:

```php

    ini_set('xdebug.scream', '0'); // >>>>> [0]

    // Not every STDOUT target stream is rewindable
    @rewind(STDOUT); // >>>> [1]

    if ($stdout = @stream_get_contents(STDOUT)) { // >>>> [2]
        $output         = $stdout . $output;
        $streamMetaData = stream_get_meta_data(STDOUT);

        if (!empty($streamMetaData['stream_type']) && 'STDIO' === $streamMetaData['stream_type']) {
            @ftruncate(STDOUT, 0);
            @rewind(STDOUT);
        }
    }
```
STDOUT in this context seems to pretty much always be actual `php://stdout` for my use-case.
When [1] executes, it will generate a warning like "Can't rewind stream" (suppressed by both [0] and the @ operator). This is ok, no side-effects happen, but as soon as we hit [2], it looks like in Linux-land it will give up and bail with a warning (also suppressed), but in FreeBSD-land it looks like it actually hangs there forever waiting for new data to come over. 

This is a link for my tentative fix, that I would like to PR: https://github.com/Rafagd/phpunit/commit/f4ae1134646c92af4f82bd39cd51074306cc4ec3

It basically checks the return value of `rewind` (true/false depending on success) and does not try to read STDOUT if we know rewind failed. I seems like it fixes the problem in my use-case.

#### Current behavior

When the process fork to execute the first test, it hangs forever until it's manually killed. 
(FreeBSD only?)

#### How to reproduce

[...]

PR #6402, opened 2025-11-05
Avoid reading from `STDOUT` when `rewind()` fails

Rafagd:
If rewind silently fails and the return is not checked, it's possible `stream_get_contents` will hang forever waiting for new input that will never come.

I have tested this on Linux and Linux on WSL and I couldn't reproduce the error, but it happens in our FreeBSD server.
I'm not sure if it's a FreeBSD quirk or something weird with our configs, but it seems like it's a sensible change to avoid reading the stream if it couldn't be rewound.

This is related to the issue: https://github.com/sebastianbergmann/phpunit/issues/6401

### DIFF
--- a/src/Framework/TestRunner/templates/class.tpl
+++ b/src/Framework/TestRunner/templates/class.tpl
@@ -87,9 +87,9 @@ function __phpunit_run_isolated_test()
     ini_set('xdebug.scream', '0');
 
     // Not every STDOUT target stream is rewindable
-    @rewind(STDOUT);
+    $hasRewound = @rewind(STDOUT);
 
-    if ($stdout = @stream_get_contents(STDOUT)) {
+    if ($hasRewound && $stdout = @stream_get_contents(STDOUT)) {
         $output         = $stdout . $output;
         $streamMetaData = stream_get_meta_data(STDOUT);
 
--- a/src/Framework/TestRunner/templates/method.tpl
+++ b/src/Framework/TestRunner/templates/method.tpl
@@ -87,9 +87,9 @@ function __phpunit_run_isolated_test()
     ini_set('xdebug.scream', '0');
 
     // Not every STDOUT target stream is rewindable
-    @rewind(STDOUT);
+    $hasRewound = @rewind(STDOUT);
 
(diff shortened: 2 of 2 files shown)


devnexen:
I was able to reproduce and indeed the php_stream_read part hangs. The fix looks right to me.

sebastianbergmann:
Thank you, David, for taking the time to look into this!

Rafagd:
Thanks for having a look at it @devnexen . Was I right that it's a FreeBSD thing?

Also thanks @sebastianbergmann for taking some time for this one too.

devnexen:
> Thanks for having a look at it @devnexen . Was I right that it's a FreeBSD thing?
> 

It might or it might not. There are other platforms out there such as netbsd/openbsd ... solaris/illumos.

staabm:
@Rafagd I am currently running into a very similar problem with PHPUnit 10.x
(on macos 26.1)

could you verify this problem still exists on 10.5 and we might need to backport this fix?

sebastianbergmann:
[PHPUnit 10.5 no longer receives bug fixes.](https://phpunit.de/supported-versions.html)

staabm:
hmm this means no more bugfixes for any PHP 8.1 related project?

staabm:
hmm ok.. doesn't matter.. the endless-loop in my case also happens on 11.5.44

sebastianbergmann:
This has nothing to do with the PHP version. PHPUnit >= 8.5 will be adapted to new PHP versions. AFAIU, this was about a bug in PHPUnit and this bug was fixed for the versions of PHPUnit that still receive bug fixes.

If you tell me that this change was made necessary by a change in PHP, then, of course, PHPUnit >= 8.5 should be adapted to the respective change in PHP.

Considering the fact that this change is rather small, I am open to backporting it to PHPUnit 10.5. Does this change solve the issue you experience with PHPUnit 10.5 on macOS 26.1?

staabm:
I need to play a bit more and see how to get a reproducer. since 11.5.44 does still show the endless-loop on my end, I guess there is another bug with PHPUnit/PHP on macos, which only has a similar problem of running into a endless loop with `stream_get_contents(STDOUT)`

Issue #6408, opened 2025-11-12
Exception in a data provider method leads to internal PHPUnit error


staabm:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     |  11.5.43-11-gaf583d4bb
| PHP version         | PHP 8.4.14
| Installation Method | Composer / PHAR

#### Summary

When calling `self::fail()` from a method which is called from a data-provider leads to a internal error:

> An error occurred inside PHPUnit.

#### Current behavior

```
➜  phpunit git:(11.5) ✗ php phpunit TestGenerator.php     

An error occurred inside PHPUnit.

Message:  error
Location: /Users/staabm/workspace/phpunit/src/Framework/Assert.php:3105

#0 /Users/staabm/workspace/phpunit/TestGenerator.php(18): PHPUnit\Framework\Assert::fail('error')
#1 /Users/staabm/workspace/phpunit/TestGenerator.php(13): TestGenerator::gatherAssertTypes()
#2 /Users/staabm/workspace/phpunit/src/Metadata/Api/DataProvider.php(183): TestGenerator::provideData()
#3 /Users/staabm/workspace/phpunit/src/Metadata/Api/DataProvider.php(69): PHPUnit\Metadata\Api\DataProvider->dataProvidedByMethods('TestGenerator', 'provideData', Object(PHPUnit\Metadata\MetadataCollection))
#4 /Users/staabm/workspace/phpunit/src/Framework/TestBuilder.php(47): PHPUnit\Metadata\Api\DataProvider->providedData('TestGenerator', 'testFoo')
#5 /Users/staabm/workspace/phpunit/src/Framework/TestSuite.php(508): PHPUnit\Framework\TestBuilder->build(Object(ReflectionClass), 'testFoo', Array)
#6 /Users/staabm/workspace/phpunit/src/Framework/TestSuite.php(109): PHPUnit\Framework\TestSuite->addTestMethod(Object(ReflectionClass), Object(ReflectionMethod), Array)
#7 /Users/staabm/workspace/phpunit/src/TextUI/Configuration/TestSuiteBuilder.php(121): PHPUnit\Framework\TestSuite::fromClassReflector(Object(ReflectionClass))
#8 /Users/staabm/workspace/phpunit/src/TextUI/Configuration/TestSuiteBuilder.php(58): PHPUnit\TextUI\Configuration\TestSuiteBuilder->testSuiteFromPath('/Users/staabm/w...', Array)
[...]

PR #6409, opened 2025-11-12
Report invalid data provider on exceptions

staabm:
closes https://github.com/sebastianbergmann/phpunit/issues/6408

diff at best reviewed, with whitespaces ignored:
https://github.com/sebastianbergmann/phpunit/pull/6409/files?w=1
(as I only moved a few lines code arround - from after a try/catch, now into said try/catch)

### DIFF
--- a/src/Metadata/Api/DataProvider.php
+++ b/src/Metadata/Api/DataProvider.php
@@ -167,6 +167,40 @@ final readonly class DataProvider
                 $className  = $_dataProvider->className();
                 $methodName = $_dataProvider->methodName();
                 $data       = $className::$methodName();
+
+                foreach ($data as $key => $value) {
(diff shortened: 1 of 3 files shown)


staabm:
Memo to me: I need to double check whether this fix currently swallows the stacktrace of the underlying exception

sebastianbergmann:
> diff at best reviewed, with whitespaces ignored

Thanks, I usually review changes locally because then I can use difftastic 😏

sebastianbergmann:
I merged your changes into `11.5` and reorganized the test. I tried merging the changes from `11.5` to `12.4` and ran into a merge conflict that I cannot deal with right now. I will look into this as soon as I can.

staabm:
thanks!

PR #6411, opened 2025-11-12
Stop shutdown handler outputting in child processes

### LINKED ISSUE #6410: Test runner's shutdown handler is called in system-under-test's child process
| Q                   | A
| PHPUnit version     | 12.4.2
| PHP version         | 8.3.27

#### Summary

If the SUT forks a child process and waits for it to exit, the test will output "Fatal error: Premature end of PHP process".

#### Current behavior

The test passes but outputs an error message. Interestingly `beStrictAboutOutputDuringTests="true"` has no effect here.

#### How to reproduce

```php
final readonly class Forker
{
    public function fork(): bool
    {
        $child = pcntl_fork();

        if ($child === -1) {
            // couldn't fork
            return false;
        } elseif ($child) {
            // we are the parent process
            pcntl_wait($child);
            return true;
        } else {
            // we are the child process
            exit(0);
        }
    }
}

final class ForkerTest extends TestCase
{
    public function testFork(): void
    {
        $actual = (new Forker())->fork();
        self::assertTrue($actual);
    }
}
```

Output:

```
✗ vendor/bin/phpunit test/ForkerTest.php 
PHPUnit 12.4.2 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.3.27
Configuration: /Users/xxx/yyy/phpunit.xml

Fatal error: Premature end of PHP process when running ForkerTest::testFork.

Time: 00:00.013, Memory: 20.00 MB

OK (1 test, 1 assertion)
```

(In the real world I'm testing Swoole stuff, which is a headache enough 😁 )

#### Expected behavior

I'm not sure if anything can be done with `ShutdownHandler` that would detect if it were in a child process. If not, it would be good to have a way to suppress the message. Right now I'm calling `ShutdownHandler::resetMessage()`, but that method is `@internal`.

kynx:
This fixes https://github.com/sebastianbergmann/phpunit/issues/6410

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -48,6 +48,9 @@
         "sebastian/version": "^6.0.0",
         "staabm/side-effects-detector": "^1.0.5"
     },
+    "require-dev": {
+        "ext-pcntl": "*"
+    },
     "config": {
         "platform": {
             "php": "8.3.0"
--- a/composer.json
+++ b/composer.json
@@ -48,9 +48,6 @@
         "sebastian/version": "^6.0.0",
         "staabm/side-effects-detector": "^1.0.5"
     },
-    "require-dev": {
-        "ext-pcntl": "*"
-    },
     "config": {
         "platform": {
             "php": "8.3.0"
--- a/tests/end-to-end/cli/child-process-output.phpt
+++ b/tests/end-to-end/cli/child-process-output.phpt
@@ -1,5 +1,10 @@
 --TEST--
 ShutdownHandler does not output when child process exits
+--SKIPIF--
+<?php declare(strict_types=1);
+if (!extension_loaded('pcntl')) {
+    print 'skip: Extension pcntl must be loaded.';
+}
 --FILE--
 <?php declare(strict_types=1);
 $_SERVER['argv'][] = '--do-not-cache-result';
--- a/tests/end-to-end/cli/child-process-output.phpt
+++ b/tests/end-to-end/cli/child-process-output.phpt
@@ -9,8 +9,6 @@ if (!extension_loaded('pcntl')) {
 <?php declare(strict_types=1);
 $_SERVER['argv'][] = '--do-not-cache-result';
 $_SERVER['argv'][] = '--no-configuration';
-$_SERVER['argv'][] = '--filter';
-$_SERVER['argv'][] = 'testChildProcessOutput';
 $_SERVER['argv'][] = __DIR__ . '/../_files/ChildProcessTest.php';
 
 require __DIR__ . '/../../bootstrap.php';
--- a/tests/end-to-end/cli/child-process-output.phpt
+++ b/tests/end-to-end/cli/child-process-output.phpt
@@ -1,0 +1,23 @@
+--TEST--
+ShutdownHandler does not output when child process exits
+--FILE--
+<?php declare(strict_types=1);
+$_SERVER['argv'][] = '--do-not-cache-result';
+$_SERVER['argv'][] = '--no-configuration';
+$_SERVER['argv'][] = '--filter';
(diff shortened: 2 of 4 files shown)


sebastianbergmann:
@staabm Can you have a look at this? Thanks!

tests/end-to-end/cli/child-process-output.phpt
```diff
@@ -0,0 +1,23 @@
+--TEST--
+ShutdownHandler does not output when child process exits
+--FILE--
```
staabm, line 8: needs a SKIPIF whether pcntl extension is available

composer.json
```diff
@@ -48,6 +48,9 @@
         "sebastian/version": "^6.0.0",
         "staabm/side-effects-detector": "^1.0.5"
     },
+    "require-dev": {
+        "ext-pcntl": "*"
+    },
```
staabm: I don't think we need this change, because we need pcntl only for this single test but not for general development of phpunit.

[`getmypid()`](https://www.php.net/getmypid) seems available even without pcntl.

we already have other tests depending on pcntl without this setting

sebastianbergmann: `composer.lock` needs to be updated after `composer.json` is changed. This is the reason why the CI pipeline failed.

tests/end-to-end/cli/child-process-output.phpt
sebastianbergmann, line 8: This needs a `--SKIPIF--` section to skip the test when the `pcntl` extension is not available.

sebastianbergmann: Also note that we might not set up PHP with the `pcntl` extension in our CI pipeline yet.

tests/end-to-end/cli/child-process-output.phpt
```diff
@@ -0,0 +1,23 @@
...
+<?php declare(strict_types=1);
+$_SERVER['argv'][] = '--do-not-cache-result';
+$_SERVER['argv'][] = '--no-configuration';
+$_SERVER['argv'][] = '--filter';
+$_SERVER['argv'][] = 'testChildProcessOutput';
```
staabm: we can drop the `--filter testChildProcessOutput` - it's unnecessary

tests/end-to-end/cli/child-process-output.phpt
staabm: > Also note that we might not set up PHP with the `pcntl` extension in our CI pipeline yet.

we need this change in earlier release branches. therefore I am doing it here: https://github.com/sebastianbergmann/phpunit/pull/6412

composer.json
kynx: I've removed it as suggested in https://github.com/sebastianbergmann/phpunit/pull/6411#discussion_r2521600662

staabm (commented):
lgtm - thanks

tests/end-to-end/cli/child-process-output.phpt
kynx: I've added the `--SKIPIF--`

composer.json
kynx: OK, done. Yes, `getmypid()` is part of the core.

tests/end-to-end/cli/child-process-output.phpt
kynx: Done

sebastianbergmann:
@kynx Thank you for working on this. In the future, please follow the [contribution guidelines](https://github.com/sebastianbergmann/phpunit/blob/main/.github/CONTRIBUTING.md#workflow-for-pull-requests) and configure your real name for your Git commits.

@staabm Thank you for reviewing!

Issue #6407, opened 2025-11-11
Expect calls to the same method of a mock object but with different arguments


sebastianbergmann:
It should be possible to expect calls to the same method of a mock object but with different arguments.

Consider this code:

#### `src/Event.php`
```php
<?php declare(strict_types=1);
namespace PHPUnit\TestFixture\Issue6406;

interface Event
{
}
```
#### `src/AnEvent.php`
```php
<?php declare(strict_types=1);
namespace PHPUnit\TestFixture\Issue6406;

final readonly class AnEvent implements Event
{
}
```
#### `src/AnotherEvent.php`
```php
<?php declare(strict_types=1);
namespace PHPUnit\TestFixture\Issue6406;

final readonly class AnotherEvent implements Event
{
}
```
#### `src/Dispatcher.php`
```php
<?php declare(strict_types=1);
namespace PHPUnit\TestFixture\Issue6406;

interface Dispatcher
{
    public function dispatch(Event $event): void;
}
```
#### `src/Service.php`
```php
final readonly class Service
{
    private Dispatcher $dispatcher;

    public function __construct(Dispatcher $dispatcher)
    {
        $this->dispatcher = $dispatcher;
    }

    public function doSomething(): void
    {
        $this->dispatcher->dispatch(new AnEvent);
        $this->dispatcher->dispatch(new AnotherEvent);
    }
}
```

In a test for `Service` we currently cannot (conveniently) configure that we expect two calls of the `Dispatcher::dispatch()` method, once with an instance of `AnEvent` and once with an instance of `AnotherEvent`:

#### `tests/ServiceTest.php`
```php
<?php declare(strict_types=1);
namespace PHPUnit\TestFixture\Issue6406;

use PHPUnit\Framework\Assert;
use PHPUnit\Framework\TestCase;

interface Event
{
}

final readonly class AnEvent implements Event
{
}

final readonly class AnotherEvent implements Event
{
}

interface Dispatcher
{
    public function dispatch(Event $event): void;
}

final readonly class Service
{
    private Dispatcher $dispatcher;

    public function __construct(Dispatcher $dispatcher)
    {
        $this->dispatcher = $dispatcher;
    }

    public function doSomething(): void
    {
        $this->dispatcher->dispatch(new AnEvent);
        $this->dispatcher->dispatch(new AnotherEvent);
    }
}

final class ServiceTest extends TestCase
{
    public function testDoesNotWork(): void
    {
[...]

PR #6416, opened 2025-11-17
Introduce UnorderedParameterSets and ConsecutiveParameterSets rules for improved invocation parameter matching in MockObject #6407

laloona:
Introduce two new parameter matching rules for the MockObject framework:

- UnorderedParameterSets: validates parameters across multiple invocations without requiring a specific order.
- ConsecutiveParameterSets: validates parameters for consecutive method invocations in a defined sequence.

These rules extend the flexibility of argument verification and integrate with the existing Rule API and invocation matching logic.


laloona:
Superseded by #6455 (same change from the correct feature branch; the original PR was opened from my fork’s main by mistake)

PR #6421, opened 2025-11-25
Report memory-usage while coverage report generation

### LINKED ISSUE #6419: Report memory usage when generating coverage report
PHPUnit atm reports

> Generating code coverage report in PHPUnit XML format ... done [00:31.317]

which means it only tells about the time it took to generate the report.

I think it would be useful to also see how much memory this report generation needed

staabm:
closes https://github.com/sebastianbergmann/phpunit/issues/6419

before this PR:
```
XDEBUG_MODE=coverage php phpunit tests/unit --coverage-html tmp/html
PHPUnit 13.0-g39f926a92 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.4.15 with Xdebug 3.4.7
.............................................................   61 / 3834 (  1%)
.............................................................  122 / 3834 (  3%)
...
............................................................. 3782 / 3834 ( 98%)
....................................................          3834 / 3834 (100%)

Time: 00:08.935, Memory: 102.50 MB

OK (3834 tests, 11391 assertions)

Generating code coverage report in HTML format ... done [00:01.557]
```

after this PR:
```
XDEBUG_MODE=coverage php phpunit tests/unit --coverage-html tmp/html
PHPUnit 13.0-g39f926a92 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.4.15 with Xdebug 3.4.7
.............................................................   61 / 3834 (  1%)
.............................................................  122 / 3834 (  3%)
...
............................................................. 3782 / 3834 ( 98%)
....................................................          3834 / 3834 (100%)

Time: 00:08.935, Memory: 102.50 MB

OK (3834 tests, 11391 assertions)

Generating code coverage report in HTML format ... done [Time: 00:01.468, Memory: 36.98 MB]
```

see the difference in the last line of the output

### DIFF
--- a/src/Runner/CodeCoverage.php
+++ b/src/Runner/CodeCoverage.php
@@ -14,6 +14,9 @@ use function file_put_contents;
 use function sprintf;
 use function sys_get_temp_dir;
 use PHPUnit\Event\Facade as EventFacade;
+use PHPUnit\Event\Telemetry\MemoryMeter;
+use PHPUnit\Event\Telemetry\MemoryUsage;
+use PHPUnit\Event\Telemetry\SystemMemoryMeter;
 use PHPUnit\Framework\TestCase;
 use PHPUnit\TextUI\Configuration\CodeCoverageFilterRegistry;
 use PHPUnit\TextUI\Configuration\Configuration;
@@ -52,16 +55,25 @@ use SebastianBergmann\Timer\Timer;
  */
 final class CodeCoverage
 {
+    /**
+     * @var array<string,int>
+     */
+    private const array SIZES = [
+        'GB' => 1073741824,
+        'MB' => 1048576,
+        'KB' => 1024,
+    ];
     private static ?self $instance                                      = null;
     private ?\SebastianBergmann\CodeCoverage\CodeCoverage $codeCoverage = null;
 
     /**
      * @phpstan-ignore property.internalClass
      */
-    private ?Driver $driver  = null;
-    private bool $collecting = false;
-    private ?TestCase $test  = null;
-    private ?Timer $timer    = null;
+    private ?Driver $driver                = null;
+    private bool $collecting               = false;
+    private ?TestCase $test                = null;
+    private ?Timer $timer                  = null;
+    private ?MemoryUsage $memoryStartUsage = null;
 
     public static function instance(): self
     {
@@ -448,6 +460,7 @@ final class CodeCoverage
         );
 
         $this->timer()->start();
+        $this->memoryStartUsage = $this->memoryMeter()->memoryUsage();
     }
 
     /**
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
That is not really the amount of memory spent for the report generation, but `memory_get_peak_usage()` for the entire run.

staabm:
hmm you are right. I had it more precise before but then fell in love with `ResourceUsageFormatter`, without realizing the results changed ;)

adjusted accordingly

sebastianbergmann:
Would this change in `phpunit/php-timer` help?

```
❯ git diff
src/ResourceUsageFormatter.php --- PHP
24            'MB' => 1048576,
25            'KB' => 1024,
26        ];
27    
28        public function resourceUsage(Duration $duration): string
32                $duration->asString(),
33                $this->bytesToString(memory_get_peak_usage(true)),
   28     public function resourceUsage(Result $result): string
   32             $result->duration()->asString(),
   33             $this->bytesToString($result->peakMemoryUsage()),
   34         );
   35     }
   36 
   37     /**

src/Timer.php --- 1/2 --- PHP
10    namespace SebastianBergmann\Timer;
11    
12    use function array_pop;
13    use function hrtime;
   14 use function memory_get_peak_usage;
   15 
   16 final class Timer
   17 {
   18     /**

src/Timer.php --- 2/2 --- PHP
18         * @var list<float>
19         */
20        private array $startTimes = [];
30        public function stop(): Duration
38            return Duration::fromNanoseconds((float) hrtime(true) - array_pop($this->startTimes));
   23     /**
   24      * @var list<int>
   25      */
   26     private array $peakMemoryUsages = [];
   31         $this->peakMemoryUsages[] = memory_get_peak_usage(true);
   37     public function stop(): Result
   45         return Result::from(
   46             Duration::fromNanoseconds((float) hrtime(true) - array_pop($this->startTimes)),
   47             array_pop($this->peakMemoryUsages),
   48         );
   49     }
   50 }
```

This would be a BC break as-is, but that would be okay if we wait with this until PHPUnit 13. Basically I am brainstorming future API ideas here. `Result` would be a new immutable value object that combines time and memory.

staabm:
I think the question needs to be: will the user of phpunit/php-timer always need time and memory togehter.

or do use cases exist, where you either only need time, or only need memory but not both.
(from a output format point of view)

sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it as I do not think this can reliably work.

Consider the case that multiple reports are created, for example the HTML report and an XML one. The reported memory usage will be different when only the HTML report is generated, when only the XML report is generated, and when both are generated. This would be confusing.

PR #6427, opened 2025-12-01
Add test for Assert::markTestSkipped static helper

miladev95:
Adds `testStaticMarksTestAsSkipped` to `tests/unit/Framework/Assert/markTestSkippedTest.php.`
The test calls `Assert::markTestSkipped('static message')` and asserts a SkippedWithMessageException with the expected message.


sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it as the method in question is already sufficiently tested.

PR #6428, opened 2025-12-01
deps: remove ext-json

alexislefebvre:
This extension is always active since PHP 8.0: https://www.php.net/manual/en/json.installation.php

I _stole_ the idea from: 

- https://github.com/VincentLanglet/Twig-CS-Fixer/issues/416

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,6 @@
     "require": {
         "php": ">=8.4.1",
         "ext-dom": "*",
-        "ext-json": "*",
         "ext-libxml": "*",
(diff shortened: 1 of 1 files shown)


sebastianbergmann:

The fact that an extension cannot be disabled without modifying the PHP source code does not mean it is always available, as binary vendors have previously modified the PHP source code to disable extensions that should not be disabled. I prefer to err on the side of caution and explicitly state that an extension is required, even if this information is not necessary in the best-case scenario.

PR #6429, opened 2025-12-02
Test .phar file is scoped

staabm:


sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it as I think 8eb184789975073952447cdbfe0a8dba8fc46518 is a better approach.

PR #6431, opened 2025-12-02
Add unit test for PHPUnit\Metadata\Parser\Registry singleton

miladev95:
Adds a focused unit test that verifies the singleton behavior of `PHPUnit\Metadata\Parser\Registry::parser()` and that the registry can be recreated after resetting its private static instance.

`testParserReturnsParserInstanceAndIsSingleton` — asserts `Registry::parser()` returns an instance implementing `PHPUnit\Metadata\Parser\Parser` and that repeated calls return the same instance (singleton).

`testParserRecreatedAfterResettingInstanceWithReflection` — uses reflection to reset the private static instance, asserts a new Parser is created and is not the same object as before (tests recreation path).

### DIFF
--- a/tests/unit/Metadata/Parser/RegistryTest.php
+++ b/tests/unit/Metadata/Parser/RegistryTest.php
@@ -1,0 +1,50 @@
+<?php declare(strict_types=1);
+/*
+ * This file is part of PHPUnit.
+ *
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it as the method in question is already sufficiently tested.

PR #6436, opened 2025-12-06
Cache help elements in TextUI\Help

miladev95:
Cache the array returned by `Help::elements()` in a readonly property and use it throughout the class instead of rebuilding the array on each call.

Files changed: `src/TextUI/Help.php`

Why: small, safe micro-optimization to reduce repeated allocations when generating help text.


sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it as this code is only ever called once.

PR #6444, opened 2025-12-12
Fix incorrect docblocks for assertArrayHasKey and assertArrayNotHasKey

techmahedy:
This PR updates the docblocks for `assertArrayHasKey()` and `assertArrayNotHasKey()` to accurately reflect their actual behavior and parameter requirements.

This makes the documentation clearer, more precise, and aligned with PHPUnit’s actual behavior.

### DIFF
--- a/src/Framework/Assert.php
+++ b/src/Framework/Assert.php
@@ -163,8 +163,10 @@ abstract class Assert
 
     /**
      * Asserts that an array has a specified key.
-     *
+     * 
+     * @param mixed $key
      * @param array<mixed>|ArrayAccess<array-key, mixed> $array
+     * @param string $message
      *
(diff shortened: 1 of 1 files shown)


staabm:
The added phpdoc types are redundant, since they are the same as already existing native types

techmahedy:
@staabm Yes, you are right. However, my VS Code editor is still reporting a parameter-mismatch issue.

sebastianbergmann:

techmahedy:
@sebastianbergmann okay, thank you for your time

staabm:
> However, my VS Code editor is still reporting a parameter-mismatch issue.

sounds like a bug in the editor which should be reported on VS Code

PR #6454, opened 2025-12-27
Introduce attribute `#[RequiresPackageVersion]`

nikophil:
Hello :wave: 

this new attribute is the missing piece along with other `RequiresXXX` attributes.

It can be helpful to adapt the test towards the version of the dependencies, in a test matrix for example.

The CI is failing because now `Requirements` triggers PHUnit warnings, but I did not found a way to mute them. Any idea?

thanks.

### DIFF
--- a/src/Framework/Attributes/RequiresPackageVersion.php
+++ b/src/Framework/Attributes/RequiresPackageVersion.php
@@ -1,0 +1,57 @@
+<?php declare(strict_types=1);
+/*
+ * This file is part of PHPUnit.
+ *
+ * (c) Sebastian Bergmann <sebastian@phpunit.de>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+namespace PHPUnit\Framework\Attributes;
+
+use Attribute;
+
+/**
+ * @immutable
+ *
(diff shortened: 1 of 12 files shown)


staabm:
I don't think its a good idea to wire up a test-runner with a api of the most used package manager. 

People can easily implement this in their own base classes (or use feature detects or other means).

Is this feature really useful for the majority of php projects?

sebastianbergmann:
I agree with what @staabm wrote and am rather reluctant to merge this.

nikophil:
> I don't think its a good idea to wire up a test-runner with a api of the most used package manager.

not sure to understand why? All the code of PHPUnit is not dependent of composer, only this tiny bit of code. Maybe the attribute can be called `#[RequiresComposerPackage]` or something like that

> People can easily implement this in their own base classes (or use feature detects or other means).

this argument could be applied to the majority of the other "RequiresXxx" attributes

> Is this feature really useful for the majority of php projects?

absolutely not :sweat_smile: it is a totally niche feature, but could be helpful in some rare cases. I think it is the same for a bunch of attributes of the "RequiresXxx" series attribute which may not be widely used either

--- 

That being said, I totally understand if this PR is not accepted. BTW, I just discovered `RequiresMethod` that I think I can leverage for my usecase

sebastianbergmann:

PR #6455, opened 2025-12-31
Introduce UnorderedParameterSets and ConsecutiveParameterSets rules for improved invocation parameter matching in MockObject #6407

laloona:
Introduce two new parameter matching rules for the MockObject framework:

- UnorderedParameterSets: validates parameters across multiple invocations without requiring a specific order.
- ConsecutiveParameterSets: validates parameters for consecutive method invocations in a defined sequence.

These rules extend the flexibility of argument verification and integrate with the existing Rule API and invocation matching logic.


sebastianbergmann:
I accidentally pushed the wrong button. I only wanted to rebase the pull request. I wanted to do this against the current state of  `main`. I wanted to do this before I start my review. Looks like will look at `main` now for my review 😄

laloona:
Oh ok, so please let me know if I could / should do something

sebastianbergmann:
Thank you for working on this, Christina, and for making #6407 a reality. I performed some minor cleanup after the merge, mostly to remove the IMO redundant end-to-end tests. This feature will ship with PHPUnit 13.0 in a couple of weeks.

PR #6474, opened 2026-01-20
Optimize prettifyTestMethodName

kamil-tekiela:
This should optimise the performance of this method by 10x. Regex is much more suitable for a task like this than a loop which reconstructs the string character by character. `ucfirst` is twice as fast as `strtoupper` with character substitution. Since this is hot code, the optimisation has a significant impact on the runtime. 

<s>One behavioural change is that the second regex also skips the digits at the start of the string. Since the return value is trimmed, it makes sense not to add a space at the beginning of the string.</s>

### DIFF
--- a/src/Logging/TestDox/NamePrettifier.php
+++ b/src/Logging/TestDox/NamePrettifier.php
@@ -36,9 +36,9 @@ use function str_replace;
 use function str_starts_with;
 use function strlen;
 use function strtolower;
-use function strtoupper;
 use function substr;
 use function trim;
+use function ucfirst;
 use PHPUnit\Event\Code\TestMethodBuilder;
 use PHPUnit\Event\Facade as EventFacade;
 use PHPUnit\Framework\TestCase;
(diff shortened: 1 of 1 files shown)


staabm:
Please share how you benchmark it

kamil-tekiela:
> Please share how you benchmark it

Using a very simple handwritten benchmark. https://3v4l.org/iH2iu/rfc#vgit.master

I have also verified that there is an impact by executing the change with phpMyAdmin's test suite. 

image
image

I am sure you can verify the changes much more scientifically.

staabm:
I don't need a scientific one.

How much does it speedup end-to-end on phpMyAdmin?

Thank you

kamil-tekiela:
> I don't need a scientific one.
> 
> How much does it speedup end-to-end on phpMyAdmin?
> 
> Thank you

As measured by xDebug, it's about 2s. Without it, it's difficult to measure because of the unpredictability of the tests due to I/O and HTTP requests. I estimate it's in the range of 100-200 ms for a suite that takes around 35s to execute.

staabm:
I see, thanks for the indication.

the initial PR description sounded more dramatic, so its nice to have a real number on a real project.

as the implementation also will be simplified by the change, I would welcome the change.

thank you.

sebastianbergmann:
Cherry-picked into `12.5` and merged to `main` from there. Thank you!

PR #6475, opened 2026-01-20
Optimize colorize

kamil-tekiela:
This is micro-optimisation for the hot path. For the average case, it is 1.5x-2x faster. For the common case, this is 3.6x-5x faster. In the worst case, it is 1.3x faster.

### DIFF
--- a/src/Util/Color.php
+++ b/src/Util/Color.php
@@ -86,11 +86,14 @@ final class Color
             return $buffer;
         }
 
-        $styles = [];
+        if (array_key_exists($color, self::ANSI_CODES)) {
+            return self::optimizeColor(sprintf("\x1b[%sm", self::ANSI_CODES[$color]) . $buffer . "\x1b[0m");
+        }
 
+        $styles = [];
         foreach (explode(',', $color) as $code) {
             $code = trim($code);
-            if (isset(self::ANSI_CODES[$code])) {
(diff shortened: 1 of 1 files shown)


kamil-tekiela:
I do not know why Coding Guidelines failed. I executed the cs-fixer as suggested by the guidelines and it didn't find anything on my machine. I might be doing it wrong. Is it because `array_key_exists` is not imported?

staabm:
locally I get

```diff

   1) /Users/m.staab/dvl/phpunit/src/Util/Color.php
      ---------- begin diff ----------
--- /Users/m.staab/dvl/phpunit/src/Util/Color.php
+++ /Users/m.staab/dvl/phpunit/src/Util/Color.php
@@ -11,6 +11,7 @@
 
 use const DIRECTORY_SEPARATOR;
 use const PHP_EOL;
+use function array_key_exists;
 use function array_map;
 use function array_walk;
 use function count;
@@ -91,8 +92,10 @@
         }
 
         $styles = [];
+
         foreach (explode(',', $color) as $code) {
             $code = trim($code);
+
             if (array_key_exists($code, self::ANSI_CODES)) {
                 $styles[] = self::ANSI_CODES[$code];
             }

      ----------- end diff -----------
```

for `tools/php-cs-fixer fix --diff`

Issue #6423, opened 2025-11-28
Introduce new array assertions


theofidry:
Related to #6088.

There is a bit of frustration when it comes to comparing arrays in PHPUnit with `assertEquals()`, `assertEquals(canonicalizing:true)` and `assertSame()`.

If we step back, what we may want to check with arrays is:

- the values are equal or strictly equal
- the keys matter
- the order of the keys matter

From this we can derive the following permutations:

| desired comparison | proposed API (1) | proposed API (2) | existing API |
|--|--|--|--|
| ordered strict comparison with keys (keys + order matter) | assertArraysIdentical | assertOrderedHashMapIdentical | assertSame |
| ordered loose comparison with keys (keys + order matter) | assertArraysEqual | assertOrderedHashMapEqual | |
| unordered strict comparison with keys (keys + order doesn't matter) | assertArraysIdenticalUnordered | assertUnorderedHashMapIdentical  | |
| unordered loose comparison with keys (keys + order doesn't matter) | assertArraysEqualUnordered | assertUnorderedHashMapEqual | assertEquals |
| ordered strict comparison without keys (order matter) | assertArraysIdenticalValues | assertListIdentical | |
| ordered loose comparison without keys (order matter) | assertArraysEqualValues | assertListEqual | |
| unordered strict comparison without keys (order doesn't matter) | assertArraysIdenticalUnorderedValues | assertSetIdentical | |
| unordered loose comparison without keys (order doesn't matter) | assertArraysEqualUnorderedValues | assertSetEqual | assertEquals(canonicalizing: true) |

Rationale:

- I choose `Identical` and `Equal` for respectively the strict equality `===` and loose equality `==` as I think it matches the current PHPUnit naming convention.
- I assumed ordered by default, just because when we do `assertSame()`, we assume "exactly the same" and the order is part of that. However, we could also make it more explicit and add an `Ordered` suffix.

Note that I proposed two APIs. I think the API 1 is the most natural and PHP like, as in, it's likely the expectation of a PHP developer that never really worked with proper data structures from other languages.

API 2 is another take where we convey whether the order or keys matters by using the "proper" corresponding data structure name. I provided it for completeness, but I don't think it would be a good idea for the following reasons:

- There is no universally agreed term for those data structures. Each language has their own although some uses the same term. E.g. `HashMap` is used in Java and Rust but Python will use `Dictionary` and Go & JavaScript a `Map`.
[...]

PR #6477, opened 2026-01-22
New array assertions

sebastianbergmann:
This implements the `assertArraysAreIdentical()`, `assertArraysAreIdenticalIgnoringOrder()`, `assertArraysHaveIdenticalValues()`, `assertArraysHaveIdenticalValuesIgnoringOrder()`, `assertArraysAreEqual()`, `assertArraysAreEqualIgnoringOrder()`, `assertArraysHaveEqualValues()`, and `assertArraysHaveEqualValuesIgnoringOrder()` assertions as requested in #6423.

The implementation follows the specification in https://github.com/sebastianbergmann/phpunit/issues/6423#issuecomment-3758745111:

| Method                                           | Comparison | Keys matter | Order matters |
|--------------------------------------------------|------------|-----------------------------------|-----------------------|
| `assertArraysAreIdentical()`                     | Strict     | ✅                                 | ✅                     |
| `assertArraysAreIdenticalIgnoringOrder()`        | Strict     | ✅                                 | ❌                     | 
| `assertArraysHaveIdenticalValues()`              | Strict     | ❌                                 | ✅                     |  
| `assertArraysHaveIdenticalValuesIgnoringOrder()` | Strict     | ❌                                 | ❌                     |  
| `assertArraysAreEqual()`                         | Loose      | ✅                                 | ✅                     | 
| `assertArraysAreEqualIgnoringOrder()`            | Loose      | ✅                                 | ❌                     | 
| `assertArraysHaveEqualValues()`                  | Loose      | ❌                                 | ✅                     |  
| `assertArraysHaveEqualValuesIgnoringOrder()`     | Loose      | ❌                                 | ❌                     |  

It is worth noting that the implementation uses the `==` operator for the new equality-related assertions and does not delegate the comparison to a chain of `Comparator` objects, as is the case with `assertEquals()`, for example.

### DIFF
--- a/src/Framework/Assert.php
+++ b/src/Framework/Assert.php
@@ -20,6 +20,8 @@ use ArrayAccess;
 use Countable;
 use Generator;
 use PHPUnit\Framework\Constraint\ArrayHasKey;
+use PHPUnit\Framework\Constraint\ArraysAreEqual;
+use PHPUnit\Framework\Constraint\ArraysAreIdentical;
 use PHPUnit\Framework\Constraint\Callback;
 use PHPUnit\Framework\Constraint\Constraint;
 use PHPUnit\Framework\Constraint\Count;
@@ -203,6 +205,154 @@ abstract class Assert
         self::assertThat(
             $array,
             new IsList,
+            $message,
+        );
+    }
+
+    /**
+     * Assert that two arrays are identical.
+     *
+     * The (key, value) relationship matters, the order of the (key, value) pairs in the array matters, and keys as well as values are compared strictly.
+     *
+     * @param array<mixed> $expected
+     * @param array<mixed> $actual
+     *
+     * @throws ExpectationFailedException
+     */
+    final public static function assertArraysAreIdentical(array $expected, array $actual, string $message = ''): void
+    {
+        self::assertThat(
+            $actual,
+            new ArraysAreIdentical($expected, true, true),
+            $message,
+        );
+    }
+
+    /**
+     * Assert that two arrays are identical while ignoring the order of their values.
+     *
+     * The (key, value) relationship matters, the order of the (key, value) pairs in the array does not matter, and keys as well as values are compared strictly.
+     *
+     * @param array<mixed> $expected
+     * @param array<mixed> $actual
+     *
+     * @throws ExpectationFailedException
+     */
+    final public static function assertArraysAreIdenticalIgnoringOrder(array $expected, array $actual, string $message = ''): void
+    {
+        self::assertThat(
+            $actual,
+            new ArraysAreIdentical($expected, true, false),
+            $message,
+        );
+    }
+
+    /**
+     * Assert that two arrays have identical values.
+     *
+     * The (key, value) relationship does not matter, the order of the (key, value) pairs in the array matters, and values are compared strictly.
+     *
+     * @param array<mixed> $expected
+     * @param array<mixed> $actual
+     *
+     * @throws ExpectationFailedException
+     */
+    final public static function assertArraysHaveIdenticalValues(array $expected, array $actual, string $message = ''): void
+    {
+        self::assertThat(
+            $actual,
+            new ArraysAreIdentical($expected, false, true),
+            $message,
+        );
+    }
+
+    /**
+     * Assert that two arrays have identical values while ignoring the order of these values.
+     *
+     * The (key, value) relationship does not matter, the order of the (key, value) pairs in the array does not matter, and values are compared strictly.
+     *
+     * @param array<mixed> $expected
+     * @param array<mixed> $actual
+     *
+     * @throws ExpectationFailedException
+     */
+    final public static function assertArraysHaveIdenticalValuesIgnoringOrder(array $expected, array $actual, string $message = ''): void
+    {
+        self::assertThat(
+            $actual,
+            new ArraysAreIdentical($expected, false, false),
+            $message,
+        );
+    }
+
+    /**
+     * Assert that two arrays are equal.
+     *
+     * The (key, value) relationship matters, the order of the (key, value) pairs in the array matters, and keys as well as values are compared loosely.
+     *
+     * @param array<mixed> $expected
+     * @param array<mixed> $actual
+     */
+    final public static function assertArraysAreEqual(array $expected, array $actual, string $message = ''): void
+    {
+        self::assertThat(
+            $actual,
+            new ArraysAreEqual($expected, true, true),
+            $message,
+        );
+    }
+
+    /**
+     * Assert that two arrays are equal while ignoring the order of their values.
+     *
+     * The (key, value) relationship matters, the order of the (key, value) pairs in the array does not matter, and keys as well as values are compared loosely.
+     *
+     * @param array<mixed> $expected
+     * @param array<mixed> $actual
+     *
+     * @throws ExpectationFailedException
+     */
+    final public static function assertArraysAreEqualIgnoringOrder(array $expected, array $actual, string $message = ''): void
+    {
+        self::assertThat(
+            $actual,
+            new ArraysAreEqual($expected, true, false),
+            $message,
+        );
+    }
+
+    /**
+     * Assert that two arrays have equal values.
+     *
+     * The (key, value) relationship does not matter, the order of the (key, value) pairs in the array matters, and values are compared loosely.
+     *
+     * @param array<mixed> $expected
+     * @param array<mixed> $actual
+     */
+    final public static function assertArraysHaveEqualValues(array $expected, array $actual, string $message = ''): void
+    {
+        self::assertThat(
+            $actual,
+            new ArraysAreEqual($expected, false, true),
+            $message,
+        );
+    }
+
+    /**
+     * Assert that two arrays have equal values while ignoring the order of these values.
+     *
+     * The (key, value) relationship does not matter, the order of the (key, value) pairs in the array does not matter, and values are compared loosely.
+     *
+     * @param array<mixed> $expected
+     * @param array<mixed> $actual
+     *
+     * @throws ExpectationFailedException
+     */
+    final public static function assertArraysHaveEqualValuesIgnoringOrder(array $expected, array $actual, string $message = ''): void
+    {
+        self::assertThat(
+            $actual,
+            new ArraysAreEqual($expected, false, false),
             $message,
         );
     }
--- a/src/Framework/Constraint/Array/ArrayComparison.php
+++ b/src/Framework/Constraint/Array/ArrayComparison.php
@@ -1,0 +1,142 @@
+<?php declare(strict_types=1);
+/*
+ * This file is part of PHPUnit.
+ *
+ * (c) Sebastian Bergmann <sebastian@phpunit.de>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+namespace PHPUnit\Framework\Constraint;
+
+use function array_values;
+use function is_array;
+use function ksort;
+use function sort;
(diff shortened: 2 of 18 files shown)


sebastianbergmann:
@Seldaek @theofidry I would greatly appreciate your feedback on this as soon as possible. For the purposes of this review, I suggest focusing on the test cases rather than the implementation.

@staabm Maybe this is of interest to you as well.

tests/unit/Framework/Assert/assertArraysAreIdenticalTest.php
```diff
@@ -0,0 +1,401 @@
...
+                [1 => 'a', 2 => 'b', 4 => 'c'],
+                [1 => 'a', 2 => 'b', 4 => 'c'],
+            ],
+
+            'explicit negative integer keys' => [
+                [-1 => 'a', -2 => 'b'],
+                [-1 => 'a', -2 => 'b'],
+            ],
+
+            'explicit numeric keys (literal strings, auto-cast to int)' => [
```
theofidry, line 54: I would add an additional test case for the sake of completion:


```php
            'different explicit numeric keys (literal strings, auto-cast to int)' => [
                [1 => 'a', 2 => 'b', 4 => 'c'],
                ['1' => 'a', '2' => 'b', '4' => 'c'],
            ],
```

Because whilst the written input is different, it might not be obvious to a user that from PHP's perspective (hence this assertion) those are identical.

tests/unit/Framework/Assert/assertArraysAreEqualTest.php
```diff
@@ -0,0 +1,322 @@
...
+                [true],
+                [1],
+            ],
+
+            'false vs 0 (loose comparison)' => [
+                [false],
+                [0],
+            ],
+
+            'different object instances (loose comparison)' => [
```
theofidry, line 122: wouldn't it be worth adding two more cases here?

- two loosely equal objects (i.e. the values of the properties are equal when loosely compared)
- same properties but in a different order (I do not expect the other of properties of an object to matter)

tests/unit/Framework/Assert/assertArraysAreEqualTest.php
```diff
@@ -0,0 +1,322 @@
...
+-        2 => 3,
++        2 => 1,
+     ],
+ ]
+
+EOT,
+                ['a' => [1, 2, 3]],
+                ['a' => [3, 2, 1]],
+            ],
+        ];
```
theofidry, line 396: no failing case for an object comparison?

tests/unit/Framework/Assert/assertArraysAreIdenticalTest.php
theofidry: I think this is also relevant for all the other tests except `assertArraysHaveEqualValues()`

tests/unit/Framework/Assert/assertArraysAreEqualTest.php
theofidry: IMO it would also be worth to have in the other tests too.

src/Framework/Constraint/Array/ArrayComparison.php
```diff
@@ -0,0 +1,142 @@
...
+ * @no-named-arguments Parameter names are not covered by the backward compatibility promise for PHPUnit
+ */
+abstract class ArrayComparison extends Constraint
+{
+    /**
+     * @var array<mixed>
+     */
+    protected readonly array $expected;
+    protected readonly bool $keysMatter;
+    protected readonly bool $orderMatters;
```
theofidry, line 31: side question, is there a reason to not use promoted properties here?

src/Framework/Constraint/Array/ArrayComparison.php
```diff
@@ -0,0 +1,142 @@
...
+    {
+        if (!is_array($other)) {
+            return false;
+        }
+
+        $expected = $this->expected;
+        $actual   = $other;
+
+        if ($this->keysMatter && !$this->orderMatters) {
+            ksort($expected);
```
theofidry: one thing I'm wondering is if there is a case where the flag could have an importance (same for `sort`), i.e. that two values which would for example be equal with `ksort()` but would not be with `ksort(flags: [SORT_LOCALE_STRING](https://www.php.net/manual/en/array.constants.php#constant.sort-locale-string)`.

But arguably probably out of scope of the PR, and maybe just a weird question in general.

theofidry (commented):
One question mark for me looking at the tests is what you expect the behaviour to be for objects with the same structure but of a different class.

I would expect them to be different, but it is not obvious this is what is expected when looking at the tests.

theofidry:
Looking forward to it 😄

src/Framework/Constraint/Array/ArrayComparison.php
sebastianbergmann: I do not use them on principle.

sebastianbergmann:
> One question mark for me looking at the tests is what you expect the behaviour to be for objects with the same structure but of a different class.
> 
> I would expect them to be different, but it is not obvious this is what is expected when looking at the tests.

I would expect that, too. As does PHP:

```php
final class A
{
    private string $foo = 'bar';
}

final class B
{
    private string $foo = 'bar';
}

$a = [new A];
$b = [new B];

var_dump($a == $b);
```
```
bool(false)
```

tests/unit/Framework/Assert/assertArraysAreEqualTest.php
sebastianbergmann: Did not think of that.

src/Framework/Constraint/Array/ArrayComparison.php
sebastianbergmann: I agree that this is out of scope and weird.

sebastianbergmann:
@theofidry I have merged #6478 and added placeholders to the expected comparison failures. All new tests pass now, except for this one:

```
1) PHPUnit\Framework\assertArraysAreEqualIgnoringOrderTest::testSucceedsWhenConstraintEvaluatesToTrue@different object instances with strictly equal properties, different order with data ([PHPUnit\TestFixture\ObjectComparison\ValueObjectA Object (...), PHPUnit\TestFixture\ObjectComparison\ValueObjectA Object (...)], [PHPUnit\TestFixture\ObjectComparison\ValueObjectA Object (...), PHPUnit\TestFixture\ObjectComparison\ValueObjectA Object (...)])
Failed asserting that two arrays are equal while ignoring order.
--- Expected
+++ Actual
@@ @@
 Array &0 [
-    0 => PHPUnit\TestFixture\ObjectComparison\ValueObjectA Object #5896 (
+    0 => PHPUnit\TestFixture\ObjectComparison\ValueObjectA Object #5853 (
+        'value' => 'v2',
+    ),
+    1 => PHPUnit\TestFixture\ObjectComparison\ValueObjectA Object #5995 (
         'value' => 'v1',
-    ),
-    1 => PHPUnit\TestFixture\ObjectComparison\ValueObjectA Object #5854 (
-        'value' => 'v2',
     ),
 ]
```

sebastianbergmann:
@theofidry The failing test you added now passes.

src/Framework/Constraint/Array/ArrayComparison.php
```diff
@@ -0,0 +1,153 @@
...
+        }
+
+        $expected = $this->expected;
+        $actual   = $other;
+
+        if ($this->keysMatter && !$this->orderMatters) {
+            $expectedKeys = array_keys($expected);
+            $actualKeys   = array_keys($actual);
+            sort($expectedKeys);
+            sort($actualKeys);
```
theofidry, line 68: I'm mentioning this here because `ArrayComparison` is introduced in this PR, but I think the problem should be dealt with separately.

One downside of sorting the keys as done here is that it can confuse the user about the order when displaying the diff, as the order was changed compared to the original one. It is, from the comparison view-point, correct as we say the order does not matter _for the comparison_. But from a user perspective analysing/debugging the issue, it can create confusion.

Do you prefer me to open an issue about it despite this not being merged yet (if you think this should be addressed)? Or should we wait for this to be merged first?

sebastianbergmann: Early on in the `evaluate()` method we have

```php
$expected = $this->expected;
$actual   = $other;
```

The `$expected` and `$actual` variables are used for the comparison. These variables are sorted as and when required.

When the comparison fails, the `ComparisonFailure` object is created using the `$this->expected` property and the `$other` parameter.

Unless I am misunderstanding something, the problem you describe does not exist.

theofidry: I'll check it out and add a test if I encounter the issue. Sorry I should have done that in the first place

sebastianbergmann: No worries!

Seldaek: Sorry I am also not sure here and cannot check on my phone rn but this makes me wonder.. For array functions where the order doesn't matter, I hope the diff if the arrays don't match doesnt show keys added/removed if they're only in a different order. If you diff using the unsorted array sounds like this might happen..

Seldaek: Sorry I am also not sure here and cannot check on my phone rn but this makes me wonder.. For array functions where the order doesn't matter, I hope the diff if the arrays don't match doesnt show keys added/removed if they're only in a different order. If you diff using the unsorted array sounds like this might happen..

sebastianbergmann: That should not happen, I agree, and I do not think it does. This sounds to me like we are missing more tests for the failure case(s) that show the textual representation of the comparison failure.

sebastianbergmann: I added some more tests that (hopefully) show what comparison failures currently look like.

src/Framework/Assert.php
```diff
@@ -207,6 +209,154 @@ final public static function assertIsList(mixed $array, string $message = ''): v
...
+     * Assert that two arrays are identical.
+     *
+     * The (key, value) relationship matters, the order of the (key, value) pairs in the array matters, and keys as well as values are compared strictly.
+     *
+     * @param array<mixed> $expected
+     * @param array<mixed> $actual
+     *
+     * @throws ExpectationFailedException
+     */
+    final public static function assertArraysAreIdentical(array $expected, array $actual, string $message = ''): void
```
VincentLanglet, line 222: It's the same than assertSame then ? (Might be worth to mention this in the phpdoc)

Seldaek: I'd say it's the same if both are arrays. But it asserts that they are arrays as well. So like assertArray + assertSame combined.

tests/unit/Framework/Assert/assertArraysAreEqualIgnoringOrderTest.php
```diff
@@ -0,0 +1,456 @@
...
+
+--- Expected
++++ Actual
+@@ @@
+ Array &0 [
++    'x' => 1,
++    'y' => 99,
+     'z' => 3,
+-    'y' => 2,
+-    'x' => 1,
```
Seldaek, line 389: This is kinda what I meant earlier. If you compare without caring about the order, this diff is way harder to read than it should be:

```suggestion
-    'y' => 2,
+    'y' => 99,
```

In theory that's all I care about in this case and I'd think it probably would be more helpful to just show that

theofidry: Agreed, but not a blocker for the feature though (IMO).

sebastianbergmann:
Thank you all for your feedback!

Issue #6476, opened 2026-01-21
Count on IteratorAggregate hangs


designermonkey:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 11.5.48
| PHP version         | 8.5.2
| Installation Method | Composer

#### Summary

When I have a class that implements IteratorAggregate and not Countable, if a test runs `assertCount` on an instance of that class, it hangs.

#### Current behavior

See Summary above.

#### How to reproduce

```php
<?php declare(strict_types=1);

namespace Something;

use Ds\Set;
use IteratorAggregate;
use PHPUnit\Framework\TestCase;
use Traversable;

class MyTestClass implements IteratorAggregate
{
    private Set $myset;

    public function __construct()
    {
        $this->myset = new Set();
    }

    public function getIterator(): Traversable
    {
        return $this->myset->getIterator();
    }
}

class ExampleTest extends TestCase
{
    public function testThatIteratorAggregateHangs(): void
    {
        $subject = new MyTestClass();

        $this->assertCount(2, $subject);
    }
}
```

#### Expected behavior

At least some kind of error.

sebastianbergmann:
Thank you for your report.

Please provide a **minimal**, **self-contained**, reproducing test case that shows the problem you are reporting.

Without such a minimal, self-contained, reproducing test case I will not be able to investigate this issue.

designermonkey:
Already done :)

If I try `return $this->myset;` instead, I get the same result.

If I try `return new IteratorIterator$this->myset);` the it works ok.

sebastianbergmann:
> Already done :)

No, you did not, sorry. Please provide an example that does not use `Ds\Set` and is self-contained.

designermonkey:
I don't know what else I can do. My example is how the error has manifested which is what you ask for.

If I provide an example any other way, it may not happen.

PR #6480, opened 2026-01-24
Infinite recursion in `Count::getCountOf()` for unusal implementations of `Iterator` or `IteratorAggregate`

kubawerlos:
Fixes https://github.com/sebastianbergmann/phpunit/issues/6476

The problem is that `PHPUnit\Framework\Constraint\Count::getCountOf` [calls `getIterator` recursively](https://github.com/sebastianbergmann/phpunit/blob/11.5.48/src/Framework/Constraint/Cardinality/Count.php#L71), which can fall into infinite recursion - which is happening for the case reported in https://github.com/sebastianbergmann/phpunit/issues/6476.

- 1st commit is the test case where "test was aborted after 4 seconds".
  - CI randomly fails here: running `php ./phpunit --testsuite end-to-end --order-by depends,random --display-all-issues tests/end-to-end/regression/2085.phpt tests/end-to-end/regression/6476.php` non-deterministically fails (I have restarted them until all passed in [my fork](https://github.com/6b7562617765726c6f73/phpunit/actions/runs/21311535008) - I will investigate it and raise a separate issue/PR 
- 2nd commit is the idea for the solution, which works for both -  `tests/end-to-end/regression/6476/Issue6476Test.php` and the case from https://github.com/sebastianbergmann/phpunit/issues/6476 with `Ds\Set`.

### DIFF
--- a/src/Framework/Constraint/Cardinality/Count.php
+++ b/src/Framework/Constraint/Cardinality/Count.php
@@ -12,6 +12,7 @@ namespace PHPUnit\Framework\Constraint;
 use function count;
 use function is_countable;
 use function iterator_count;
+use function spl_object_id;
 use function sprintf;
 use EmptyIterator;
 use Generator;
@@ -66,7 +67,16 @@ class Count extends Constraint
         }
 
         if ($other instanceof Traversable) {
+            $traversableSeen = [];
+
             while ($other instanceof IteratorAggregate) {
+                $id = spl_object_id($other);
+
+                if (isset($traversableSeen[$id])) {
+                    throw new Exception('IteratorAggregate::getIterator() returned an object that was already seen');
+                }
+                $traversableSeen[$id] = true;
+
                 try {
                     $other = $other->getIterator();
                 } catch (\Exception $e) {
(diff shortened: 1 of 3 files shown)


src/Framework/Constraint/Cardinality/Count.php
```diff
@@ -66,7 +67,16 @@ protected function getCountOf(mixed $other): ?int
         }
 
         if ($other instanceof Traversable) {
+            $traversableSeen = [];
+
             while ($other instanceof IteratorAggregate) {
+                $id = spl_object_id($other);
```
staabm, line 73: we are doing a similar thing in PHPStan. maybe you are interessted in refactoring the recursion-guard into a separate class:

https://github.com/phpstan/phpstan-src/blob/2.1.x/src/Type/RecursionGuard.php

sebastianbergmann: I created https://github.com/sebastianbergmann/recursion-context long ago and use that in PHPUnit. This fix should probably refactored to use that.

PR #6481, opened 2026-01-24
Debug non-deterministic tests failure

kubawerlos:
Running test at all 18 PHP 8.4 tags - to have 18 different runs.

Time limits - `2085.phpt` has a time limit of 1 second, `2085a.phpt` has 1 or 2 seconds
Timeout source - where the timeout happens - inside of PHPUnit or in a test case.
 
| Order | Time limits | Timeout source | non-deterministic tests failure?
| :--: | :--: | :--: | :--: |
| 2085, 2085a | different | PHPUnit | yes
| 2085, 2085a | the same  | PHPUnit | yes
| 2085, 2085a | the same  | test case | no
| 2085, 2085a | different | test case | no
| 2085a, 2085 | different | PHPUnit | yes
| 2085a, 2085 | the same | PHPUnit | yes

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -243,14 +243,31 @@ jobs:
       matrix:
         os:
           - ubuntu-latest
-          - windows-latest
+#          - windows-latest
 
         php-version:
-          - "8.2"
-          - "8.3"
-          - "8.4"
-          - "8.5"
-          - "8.6"
+#          - "8.2"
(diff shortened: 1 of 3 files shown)


kubawerlos:
So, it seems that it does not matter if time limits are the same or different.
The same is for the order of test - only 2085a was failing in PHPUnit, 2085 was failing in the test case.

Any hint what could be the reason for those non-deterministic failures?

Issue #6434, opened 2025-12-04
Issue reporting deprecations when calls being made from 3rd party code


mpdude:
I am trying to collect deprecation notices while running the PHPUnit test suite in a Symfony application.

In general, I‘d like to `ignoreIndirectDeprecations="true"`, since there are a lot of them between 3rd party packages I don‘t want to care about right now. But that currently also suppresses deprecations when my code is being called by framework code.

For example, there might be a Controller class that lives in my `src` and that triggers a deprecation notice. This controller, however, is not called directly by my test or other own source code, but during functional tests. 

In the case of a controller, it‘s probably Symfony‘s `HttpKernel` invoking the controller method. In other cases (like Symfony Console commands), it may be other Symfony components calling my code.

From looking at `\PHPUnit\Runner\ErrorHandler::trigger`, it seems to me that since calls are _not_ being made directly from my own (first party) code, this is always considered the indirect deprecation case which I‘d like to suppress.

I can try to work around this by adding the HttpKernel and other relevant parts from `vendor` as source files in the PHPUnit config. However, that does not exactly fit all cases since I don‘t want to see the cases where the Kernel is calling 3rd party code (which would be a direct deprecation if the Kernel is considered first party code).

[Edit: Moved second group of issues to #6435, since I think the cause and potential solutions are different.]

#### Additional information

PHPUnit 12.4.5, PHP 8.4.15

phpunit.xml.dist:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<phpunit
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
    xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/12.0/phpunit.xsd"
    bootstrap="tests/bootstrap.php"
>
    <source ignoreSuppressionOfDeprecations="true" ignoreIndirectDeprecations="true">
        <deprecationTrigger>
            <function>trigger_deprecation</function>
        </deprecationTrigger>
        <include>
            <directory>src</directory>
        </include>
    </source>
    <testsuites>
        <testsuite name="Project Test Suite">
            <directory>tests</directory>
        </testsuite>
    </testsuites>
</phpunit>

```

test/bootstrap.php:

```php
<?php

use Symfony\Component\ErrorHandler\DebugClassLoader;
use Symfony\Component\ErrorHandler\ErrorHandler;

require_once __DIR__.'/../vendor/autoload.php';

DebugClassLoader::enable();
```

[...]

Issue #5689, opened 2024-02-02
Distinguish between self, direct and indirect deprecations


DominicLuidold:
## Context

We've recently upgraded PHPUnit to version 10.5, up from 9.6, and also made the decision to remove the `symfony/phpunit-bridge` dev dependency. This choice was influenced by the insights from the "PHPUnit 10 for Symfony Developers" talk presented at SymfonyCon Brussels/SymfonyOnline, which indicated that the bridge is no longer necessary for reporting deprecations.

Our development involves a closed-source Symfony project for a client, and we've traditionally relied on the `symfony/phpunit-bridge` for deprecation reports. Specifically, we focus on identifying self and direct deprecations, such as instances where we call a method in our `src` directory from a third-party library in the `vendor` directory that will be removed in the next major version. This proactive approach helps us stay ahead during upgrades and the planning of our development tasks.

## Previous approach using `symfony/phpunit-bridge`

With `symfony/phpunit-bridge`, distinguishing between `self`, `direct`, and `indirect` deprecations was possible, see [here](https://symfony.com/doc/current/components/phpunit_bridge.html#making-tests-fail) and [here](https://symfony.com/doc/current/components/phpunit_bridge.html#direct-and-indirect-deprecations). We utilized this to make tests fail on `self` and especially `direct` deprecations while "ignoring" `indirect` deprecations, as they cannot directly be addressed by us.

## Current possibilites

With PHPUnit 10.5, it is already possible to get reports on deprecations using the following configuration (inspired by [slide \#66 from the conference's talk](https://thephp.cc/presentations/phpunit-10-for-symfony-developers)).

```xml
<?xml version="1.0" encoding="UTF-8"?>
<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/10.5/phpunit.xsd"
         bootstrap="tests/bootstrap.php"
         displayDetailsOnTestsThatTriggerDeprecations="true"
         failOnDeprecation="true">
    <!-- ... -->

    <source restrictDeprecations="true"
            ignoreSuppressionOfDeprecations="true">
        <include>
            <directory>src</directory>
            <file>vendor/symfony/deprecation-contracts/function.php</file>
        </include>
    </source>
</phpunit>
```

This will produce the following output when running tests that eventually trigger both direct and also indirect deprecations:

```
[...]

PR #6489, opened 2026-02-03
Classification of self/direct/indirect deprecation triggers is not aligned with Symfony's bridge for PHPUnit

sebastianbergmann:
This refactors the `IssueTrigger` classification to track both the callee (where the issue is triggered) and the caller (where it was called from) and aligns the self/direct/indirect classification with Symfony's phpunit-bridge semantics as discussed in [#6434](https://github.com/sebastianbergmann/phpunit/issues/6434#issuecomment-3613501430).


sebastianbergmann:
@stof @nicolas-grekas @xabbuh @mpdude This should align PHPUnit's issue trigger classification with Symfony's phpunit-bridge semantics. Please review! Thank you.

stof:
Looking at the PR, I have the impression that there is no difference between the handling of `E_DEPRECATED` and `E_USER_DEPRECATED`, unlike what we discussed on Slack. This distinction is important due to the stack trace having different meaning for them:
- For `E_USER_DEPRECATED`, the stack trace has a `trigger_error` call which is located in the code deciding to trigger the error (the callee) with the previous frame being the code using the deprecated API (the caller).
- For `E_DEPRECATED`, the deprecation is triggered in native code (which should be a "third-party" callee, as php-src is third-party), with a location pointing to the code using the deprecated API (which should be considered the caller, but is exactly in the same position in the stack trace than the callee of `E_USER_DEPRECATED`).

For anyone reading this without the context of our Slack discussion, this is a *simplified* description of the `E_DEPRECATED`. Full support would require handling the case of `#[Deprecated]` where PHP would trigger it on behalf of userland code (which should ideally change the callee to be the place defining the attribute instead of being php-src). However, `symfony/phpunit-bridge` does not implement that either (because that attribute did not exist at that time though).

An extra feature we have in `symfony/phpunit-bridge` is [special integration with the `DebugClassLoader`](https://github.com/symfony/phpunit-bridge/blob/f25c63002d4dc7ff3658fbf9840cc5db0e795bb9/DeprecationErrorHandler/Deprecation.php#L95-L109) of `symfony/error-handler`. I'm not sure how phpunit could support an equivalent (it would probably require exposing an API for extensions to implement such behavior instead of hardcoding it in PHPUnit as being not generic enough, allowing `symfony/phpunit-bridge` to provide this feature with symfony-specific knowledge). This special integration does 2 things:
1. any `E_USER_DEPRECATED` triggered (directly) inside the DebugClassLoader gets a virtual frame injected so that it looks like it is triggered inside the file being loaded by the autoloader (as DebugClassLoader reports deprecation happening due to the class definition itself)
2. for some specific messages where the triggered deprecattion is related to inheritance, it injects a *second* virtual frame, with the effect that the deprecation now appears triggered from the parent class (as callee), with the first virtual frame becoming the caller.
This special integration allows classifying the autoloader-triggered deprecations properly based on the class definitions using some *declarative* way to define things (`@final` for instance).

Thinking about building this based on extension point makes me think about _other_ classification cases that are not solved today in `symfony/phpunit-bridge` but are a bit similar: cases where a config file loader triggers deprecations due to the file content. For such cases, proper classification would benefit from considering the file being processed as being the caller of the processing logic. If building a feature to support the DebugClassLoader use case in a more generic way, maybe it would allow solving such cases as well.

sebastianbergmann:
Thank you for your feedback, @stof.

> I have the impression that there is no difference between the handling of `E_DEPRECATED` and `E_USER_DEPRECATED`

Your impression is correct. In the context of this pull request, the distinction between `E_DEPRECATED` and `E_USER_DEPRECATED` is irrelevant. This PR is about preserving the information of caller and callee and encapsulating the distinction of self, direct, indirect in the `IssueTrigger` object.

> For `E_USER_DEPRECATED`, the stack trace has a `trigger_error()` call which is located in the code deciding to trigger the error (the callee) with the previous frame being the code using the deprecated API (the caller).

And using the [`<deprecationTrigger>`](https://docs.phpunit.de/en/13.0/configuration.html#the-deprecationtrigger-element) element in PHPUnit's XML configuration file you can configure "[deprecation triggers](https://github.com/sebastianbergmann/phpunit/issues/5689#issuecomment-2006819389)" such as Symfony's API for triggering deprecations.

> For `E_DEPRECATED`, the deprecation is triggered in native code (which should be a "third-party" callee, as php-src is third-party)

Interesting, never thought about it this way.

> For anyone reading this without the context of our Slack discussion

~~That includes me, I'm afraid. I remember having a discussion with you on Slack, but I can no longer see it.~~

> Full support would require handling the case of `#[Deprecated]` where PHP would trigger it on behalf of userland code (which should ideally change the callee to be the place defining the attribute instead of being php-src).

Any special handling of `#[Deprecated]` is out of scope, at least for now.

> I'm not sure how phpunit could support an equivalent (it would probably require exposing an API for extensions to implement such behavior instead of hardcoding it in PHPUnit as being not generic enough, allowing symfony/phpunit-bridge to provide this feature with symfony-specific knowledge).

This is out of scope for this PR. If and when you have concrete ideas for what extension point you need from PHPUnit to implement what you need: let me know.

stof:
> This PR is about preserving the information of caller and callee and encapsulating the distinction of self, direct, indirect in the `IssueTrigger` object.

this *is* relevant. For `E_DEPRECATED`, the callee is php-src (not present in the PHP stack trace) and its caller is the first frame.

sebastianbergmann:
I have implemented a preparatory refactoring to introduce a type distinction between PHP-triggered and userland-triggered issues, allowing PHPUnit to classify them differently in the future.

The single concrete `IssueTrigger` class is being split into an abstract base class with two concrete subclasses, distinguishing between `E_DEPRECATED` issues and `E_USER_DEPRECATED` issues.

The new `PhpIssueTrigger` (for `E_DEPRECATED`) and `UserlandIssueTrigger` (for `E_USER_DEPRECATED`) classes currently have the same implementation, but thisdistinction enables future divergent behavior.

`ErrorHandler` has been changed to use `PhpIssueTrigger` or `UserlandIssueTrigger` depending on whether `E_DEPRECATED` or `E_USER_DEPRECATED` is handled.

The next step should be to change the implementation of `PhpIssueTrigger` to accommodate https://github.com/sebastianbergmann/phpunit/pull/6489#issuecomment-3883026906.

stof:
> The single concrete `IssueTrigger` class is being split into an abstract base class with two concrete subclasses, distinguishing between `E_DEPRECATED` issues and `E_USER_DEPRECATED` issues.
> 
> The new `PhpIssueTrigger` (for `E_DEPRECATED`) and `UserlandIssueTrigger` (for `E_USER_DEPRECATED`) classes currently have the same implementation, but thisdistinction enables future divergent behavior.

I'm not convinced about that split. What needs to change between both cases is the identification of the caller and callee to classify them differently. And the IssueTrigger is not the one responsible for that logic.

sebastianbergmann:
> What needs to change between both cases is the identification of the caller and callee to classify them differently. And the IssueTrigger is not the one responsible for that logic.

You are correct, and I have worked on this PR a bit more.

Previously, the `trigger()` method in `ErrorHandler` used the same stack-trace-based logic for both `E_DEPRECATED` and `E_USER_DEPRECATED` issues. But these two issue types have fundamentally different stack trace semantics, as you already explained.

I split the single `trigger()` method into two specialized methods:
1. `triggerForPhpDeprecation($class, $test, $errorFile)` handles `E_DEPRECATED` and uses the `$errorFile` argument to determine who the caller was:
    - If `$errorFile` matches the test file → classified as test
    - If `$errorFile` is in the source filter → caller is first-party
    - Otherwise → caller is third-party
    - The callee is always third-party as it's the PHP runtime
2. `triggerForUserlandDeprecation($class, $test, $trace)` handles `E_USER_DEPRECATED` and uses the existing stack-trace logic

sebastianbergmann:
I folded `PhpIssueTrigger` and `UserlandIssueTrigger` back into `IssueTrigger` and introduced `Code::PHP` to separate third-party code from PHP runtime. One final review from @stof and this should be good to be merged.

src/Event/Value/Test/Issue/IssueTrigger.php
```diff
@@ -9,89 +9,90 @@
...
     /**
-     * Your own code triggers an issue in third-party code.
-     *
-     * @phpstan-assert-if-true DirectTrigger $this
+     * First-party code triggers an issue in third-party code.
      */
     public function isDirect(): bool
     {
-        return false;
+        return $this->caller == Code::FirstParty && $this->callee == Code::ThirdParty;
```
stof: this should also cover the case where the callee is `Code::PHP`, as the PHP runtime is a special kind of third-party code.

src/Event/Value/Test/Issue/IssueTrigger.php
```diff
@@ -9,89 +9,90 @@
...
     /**
-     * Third-party code triggers an issue either in your own code or in third-party code.
-     *
-     * @phpstan-assert-if-true IndirectTrigger $this
+     * Third-party code triggers an issue.
      */
     public function isIndirect(): bool
     {
-        return false;
+        return $this->caller == Code::ThirdParty && $this->callee == Code::ThirdParty;
```
stof: same here

src/Event/Value/Test/Issue/IssueTrigger.php
stof: And what about the caller being test code ? I think we might need to also treat test code as first party for that matter. Maybe you should introduce `Code::isFirstParty()` and `Code::isThirdParty()` (naming to be discussed to avoid confusion between case names and the methods) to centralize the handling of enum cases that are split only to provide more granular messages.

src/Runner/ErrorHandler.php
```diff
@@ -268,14 +270,49 @@ private function ignoredByBaseline(string $file, int $line, string $description)
...
+        return $this->triggerForUserlandDeprecation($test, $trace);
+    }
+
+    /**
+     * @param non-empty-string $errorFile
+     */
+    private function triggerForPhpDeprecation(TestMethod $test, string $errorFile): IssueTrigger
+    {
+        if ($errorFile === $test->file()) {
+            return IssueTrigger::test();
```
stof: this looks weird to me. This IssueTrigger will set the *callee* as `CodeTest`, while this is a case where test code is the caller of deprecated PHP APIs.

src/Runner/ErrorHandler.php
stof: shouldn't we distinguish test code calling into first party code vs third party code ?

sebastianbergmann:
Thank you for your feedback, @stof. I will review/incorporate it ASAP.

sebastianbergmann:
@stof I have implemented your latest feedback and simplified the code a bit along the way.

I refactored both `triggerForPhpDeprecation()` and `triggerForUserlandDeprecation()` to not return early in favour of a simpler "default then override" pattern.
                                                                                                                                                                                                                                                                                                      
The key semantic change: when code is triggered from the test file itself, it is no longer short-circuited to a generic "test" trigger. Instead, it is classified with the full `from($callee, $caller)` path, which preserves what the test code was calling into (PHP runtime vs. first-party vs. third-party).

src/Event/Value/Test/Issue/IssueTrigger.php
```diff
@@ -9,89 +9,87 @@
...
 
-    abstract public function asString(): string;
+    public function asString(): string
+    {
+        if ($this->isUnknown()) {
+            return 'unknown if issue was triggered in first-party code or third-party code';
+        }
+
+        if ($this->isTest()) {
+            return 'issue triggered by test code';
```
stof: shouldn't this be removed now that deprecations in test code are classifying their callee as well ?

src/Runner/ErrorHandler.php
```diff
@@ -268,52 +270,75 @@ private function ignoredByBaseline(string $file, int $line, string $description)
...
-            $triggerCalledFromFirstPartyCode = true;
-        }
-
-        if ($triggerCalledFromFirstPartyCode) {
-            if ($triggeredInFirstPartyCode) {
-                return IssueTrigger::self();
-            }
-
-            return IssueTrigger::direct();
+            $caller = Code::FirstParty;
```
stof: shouldn't this also handle the case of a caller being test code ?

sebastianbergmann: You are correct, thank you. 2ba2aab4b0fcd5b9ecd7593742ed61b61e45246d should address this.

src/Event/Value/Test/Issue/IssueTrigger.php
sebastianbergmann: I have removed this special case from `asString()` in affc648afdcad043118c5193032bc37de4c38656.

sebastianbergmann:
In 469f117bba93ec451d2b000d8c37dadf225d5351, I introduced `Code::PHPUnit` to separate third-party code from PHPUnit.

src/Event/Value/Test/Issue/Code.php
```diff
@@ -0,0 +1,28 @@
...
+enum Code: string
+{
+    public function isFirstPartyOrTest(): bool
+    {
+        return $this === self::FirstParty || $this === self::Test;
+    }
+
+    public function isThirdPartyOrPhp(): bool
+    {
+        return $this === self::ThirdParty || $this === self::PHP;
```
stof: PHPunit should also be third-party AFAICT

stof: maybe it would make sense to have a test ensuring that all cases of the enum return true for one of the methods, to make sure that they are all getting their high-level qualification.

src/Event/Value/Test/Issue/IssueTrigger.php
```diff
@@ -9,89 +9,83 @@
...
     /**
-     * Third-party code triggers an issue either in your own code or in third-party code.
-     *
-     * @phpstan-assert-if-true IndirectTrigger $this
+     * Third-party code triggers an issue.
      */
     public function isIndirect(): bool
     {
-        return false;
+        return $this->caller !== null && $this->caller == Code::ThirdParty &&
```
stof: shouldn't this use `isThirdPartyOrPhp()` as well ?

src/Event/Value/Test/Issue/IssueTrigger.php
```diff
@@ -9,89 +9,83 @@
...
     /**
-     * Your test code triggers an issue.
-     *
-     * @phpstan-assert-if-true TestTrigger $this
+     * An issue is triggered in test code.
      */
     public function isTest(): bool
     {
-        return false;
+        return $this->callee == Code::Test;
```
stof: shouln't this check the caller instead ? If the callee is test, it would mean that the test is defining a deprecated API that is being called, which looks like a weird case to me.

src/Runner/ErrorHandler.php
```diff
@@ -268,52 +272,79 @@ private function ignoredByBaseline(string $file, int $line, string $description)
...
+    {
+        $caller = Code::ThirdParty;
 
-        if (isset($trace[1]['file']) &&
-            ($trace[1]['file'] === $test->file() ||
-            SourceFilter::instance()->includes($trace[1]['file']))) {
-            $triggerCalledFromFirstPartyCode = true;
+        if ($errorFile === $test->file()) {
+            $caller = Code::Test;
+        } elseif (SourceFilter::instance()->includes($errorFile)) {
```
stof: this should also have support for `Code::PHPUnit` IMO

src/Runner/ErrorHandler.php
```diff
@@ -268,52 +272,79 @@ private function ignoredByBaseline(string $file, int $line, string $description)
...
+    private function triggerForUserlandDeprecation(TestMethod $test, array $trace): IssueTrigger
+    {
+        $callee = Code::ThirdParty;
+        $caller = Code::ThirdParty;
+
+        if (isset($trace[0]['file'])) {
+            if ($trace[0]['file'] === $test->file()) {
+                $callee = Code::Test;
+            } elseif (SourceFilter::instance()->includes($trace[0]['file'])) {
+                $callee = Code::FirstParty;
```
stof: shouldn't this also have support for `Code::PHPUnit`, to handle the case where PHPUnit files would define a deprecated API using this reporting ?

stof: To ensure that all cases are handled for all places, maybe a private method `categorizeFile(string $file, TestMethod $test): Code` should be introduced to use it in the 3 places without duplication.

sebastianbergmann: > shouldn't this also have support for `Code::PHPUnit`, to handle the case where PHPUnit files would define a deprecated API using this reporting ?

PHPUnit does not use `E_USER_DEPRECATED` to report usage of deprecated PHPUnit APIs.

src/Event/Value/Test/Issue/Code.php
sebastianbergmann: > PHPunit should also be third-party AFAICT

Of course.

src/Runner/ErrorHandler.php
sebastianbergmann: > To ensure that all cases are handled for all places, maybe a private method `categorizeFile(string $file, TestMethod $test): Code` should be introduced to use it in the 3 places without duplication.

Done.

src/Event/Value/Test/Issue/IssueTrigger.php
sebastianbergmann: Done.

src/Runner/ErrorHandler.php
sebastianbergmann: Taken care of by the extracted `categorizeFile()` method.

src/Event/Value/Test/Issue/Code.php
sebastianbergmann: > maybe it would make sense to have a test ensuring that all cases of the enum return true for one of the methods, to make sure that they are all getting their high-level qualification.

Done.

tests/unit/Event/Value/Test/CodeTest.php
```diff
@@ -0,0 +1,59 @@
...
+use PHPUnit\Framework\Attributes\Small;
+use PHPUnit\Framework\TestCase;
+
+#[CoversClass(Code::class)]
+#[Small]
+final class CodeTest extends TestCase
+{
+    public function testFirstPartyIsFirstPartyOrTest(): void
+    {
+        $this->assertTrue(Code::FirstParty->isFirstPartyOrTest());
```
stof: I would suggest having a tests looking like that:

```php
#[DataProvider('provideCase')]
public function testCodeIsCategorizedAtHighLevel(Code $code): void
{
    $this->assertTrue($code->isFirstPartyOrTest() || $code->isThirdPartyOrPhpunitOrPhp(), 'Code is categorized in one of the high-level categories.')
    $this->assertFalse($code->isFirstPartyOrTest() && $code->isThirdPartyOrPhpunitOrPhp(), 'Code is categorized in only one of the high-level categories.')
}

public static function provideCase(): iterable
{
    foreach (Code::cases() as $code) {
        yield $code->value => [$code];
    }
}
```

Such test would prevent forgetting to add support for new enum cases in those methods when adding new cases (which is something that happened in this PR *each* time a new case has been added to split a case into more granular cases).

The test could be written using a single `assertTrue(... xor ...)` assertion but this does not distinguish anymore whether a case was forgotten or classified badly as being in both categories.

src/Event/Value/Test/Issue/IssueTrigger.php
stof: @sebastianbergmann what about this comment ?

sebastianbergmann: I have deleted the `isTest()` method now as it was no longer used anywherre.

sebastianbergmann:
When I work on this part of PHPUnit's code, after a while I always reach a point where I can no longer think clearly. Now I am at this point again.

@stof It would help me if we could look at these two questions separately:

1. Does `ErrorHandler` create `IssueTrigger` objects with the correct `Code` values for caller and callee now?
2. Are the `isSelf()`, `isDirect()`, and `isIndirect()` methods of `IssueTrigger` correct now?

Can you help me answer these questions? Thank you!

src/Event/Value/Test/Issue/IssueTrigger.php
```diff
@@ -9,89 +9,75 @@
...
 
-    public static function direct(): DirectTrigger
+    public static function unknown(): self
     {
-        return new DirectTrigger;
+        return new self(null, null);
     }
 
-    public static function indirect(): IndirectTrigger
+    public static function from(Code $callee, ?Code $caller = null): self
```
stof: I don't think there is any place passing `null` as caller when using this `from` method.

src/Runner/ErrorHandler.php
```diff
@@ -268,52 +276,73 @@ private function ignoredByBaseline(string $file, int $line, string $description)
...
 
-        $triggeredInFirstPartyCode       = false;
-        $triggerCalledFromFirstPartyCode = false;
+    /**
+     * @param list<array{file: string, line: int, class?: string, function?: string, type: string}> $trace
+     */
+    private function triggerForUserlandDeprecation(TestMethod $test, array $trace): IssueTrigger
+    {
+        $callee = Code::ThirdParty;
+        $caller = Code::ThirdParty;
```
stof: I'm wondering whether `Code::ThirdParty` is the appropriate fallback when we don't have a trace ti identify the callee and the caller. Indirect deprecations are the ones that are most commonly ignored in CI (as they are the ones that are not directly actionable), and this would be the classification for a third-party caller.

src/Runner/ErrorHandler.php
```diff
@@ -268,52 +276,73 @@ private function ignoredByBaseline(string $file, int $line, string $description)
...
-            $triggerCalledFromFirstPartyCode = true;
+        return IssueTrigger::from($callee, $caller);
+    }
+
+    /**
+     * @param non-empty-string $file
+     */
+    private function categorizeFile(string $file, TestMethod $test): Code
+    {
+        if ($file === $test->file()) {
```
stof, line 323: What about files defining reusable test utilities in the project `tests` directly ? They would not be classified as `Code::Test` by this logic, but would also likely not match the source filter (as this source filter is used to configure the code coverage report, and you generally don't include the testsuite in that report). Such utilities are still first-party as the high-level qualification as they are part of the project repositories, not of dependencies.

stof (commented):
I confirm that the implementation of `IssueTrigger::is*` methods is correct according to the semantic.
Regarding the ErrorHandler, the implementation looks correct to me, with 2 edge cases for which I added dedicated comments.

src/Runner/ErrorHandler.php
sebastianbergmann: What a coincidence: you wrote this comment while I wrote #6517.

src/Event/Value/Test/Issue/IssueTrigger.php
sebastianbergmann: Not anymore at least, good catch.

src/Runner/ErrorHandler.php
stof: this issue is proposing a nice solution indeed.

src/Runner/ErrorHandler.php
sebastianbergmann: Do you propose falling back to `null` like so?

```patch
diff --git a/src/Event/Value/Test/Issue/IssueTrigger.php b/src/Event/Value/Test/Issue/IssueTrigger.php
index 5114b1c60..0de34dda2 100644
--- a/src/Event/Value/Test/Issue/IssueTrigger.php
+++ b/src/Event/Value/Test/Issue/IssueTrigger.php
@@ -26,7 +26,7 @@ public static function unknown(): self
         return new self(null, null);
     }
 
-    public static function from(Code $callee, Code $caller): self
+    public static function from(?Code $callee, ?Code $caller): self
     {
         return new self($callee, $caller);
     }
diff --git a/src/Runner/ErrorHandler.php b/src/Runner/ErrorHandler.php
index 8d194f63f..a8c72f996 100644
--- a/src/Runner/ErrorHandler.php
+++ b/src/Runner/ErrorHandler.php
@@ -301,8 +301,8 @@ private function trigger(TestMethod $test, bool $isUserland, ?string $errorFile
      */
     private function triggerForUserlandDeprecation(TestMethod $test, array $trace): IssueTrigger
     {
-        $callee = Code::ThirdParty;
-        $caller = Code::ThirdParty;
+        $callee = null;
+        $caller = null;
 
         if (isset($trace[0]['file'])) {
             $callee = $this->categorizeFile($trace[0]['file'], $test);
```

stof: I don't know in which cases those files would be unknown, but it might make sense to classify those as an unknown trigger.

Btw, `isUnknown` should be changed to `$this->callee === null || $this->caller === null` then (as the instantation won't guarantee anymore that they are both null or both non-null). Otherwise, we have weird classification for third-party callee with unknown caller, where none of the `is*` methods would return true.

sebastianbergmann: I hope I implemented this correctly in 6589f445de2a4abbc24812d0b81fe2d5423e95e6.

sebastianbergmann:
> I confirm that the implementation of `IssueTrigger::is*` methods is correct according to the semantic.

I added tests for all possible argument pairs for `IssueTrigger::from()` (although not all of them make sense or are used by `ErrorHandler`:

```
IssueTrigger
 ✔ (null, null) is classified as "unknown"
 ✔ (null, Code::Test) is classified as "unknown"
 ✔ (null, Code::FirstParty) is classified as "unknown"
 ✔ (null, Code::ThirdParty) is classified as "unknown"
 ✔ (null, Code::PHPUnit) is classified as "unknown"
 ✔ (null, Code::PHP) is classified as "unknown"
 ✔ (Code::Test, null) is classified as "self" and "unknown"
 ✔ (Code::Test, Code::Test) is classified as "self"
 ✔ (Code::Test, Code::FirstParty) is classified as "self"
 ✔ (Code::Test, Code::ThirdParty) is classified as "self"
 ✔ (Code::Test, Code::PHPUnit) is classified as "self"
 ✔ (Code::Test, Code::PHP) is classified as "self"
 ✔ (Code::FirstParty, null) is classified as "self" and "unknown"
 ✔ (Code::FirstParty, Code::Test) is classified as "self"
 ✔ (Code::FirstParty, Code::FirstParty) is classified as "self"
 ✔ (Code::FirstParty, Code::ThirdParty) is classified as "self"
 ✔ (Code::FirstParty, Code::PHPUnit) is classified as "self"
 ✔ (Code::FirstParty, Code::PHP) is classified as "self"
 ✔ (Code::ThirdParty, null) is classified as "unknown"
 ✔ (Code::ThirdParty, Code::Test) is classified as "direct"
 ✔ (Code::ThirdParty, Code::FirstParty) is classified as "direct"
 ✔ (Code::ThirdParty, Code::ThirdParty) is classified as "indirect"
 ✔ (Code::ThirdParty, Code::PHPUnit) is classified as "indirect"
 ✔ (Code::ThirdParty, Code::PHP) is classified as "indirect"
 ✔ (Code::PHPUnit, null) is classified as "unknown"
 ✔ (Code::PHPUnit, Code::Test) is classified as "direct"
 ✔ (Code::PHPUnit, Code::FirstParty) is classified as "direct"
 ✔ (Code::PHPUnit, Code::ThirdParty) is classified as "indirect"
 ✔ (Code::PHPUnit, Code::PHPUnit) is classified as "indirect"
 ✔ (Code::PHPUnit, Code::PHP) is classified as "indirect"
 ✔ (Code::PHP, null) is classified as "unknown"
 ✔ (Code::PHP, Code::Test) is classified as "direct"
 ✔ (Code::PHP, Code::FirstParty) is classified as "direct"
 ✔ (Code::PHP, Code::ThirdParty) is classified as "indirect"
 ✔ (Code::PHP, Code::PHPUnit) is classified as "indirect"
 ✔ (Code::PHP, Code::PHP) is classified as "indirect"
```

`(Code::Test, null)` and `(Code::FirstParty, null)` are the only two argument pairs for which more than one `is*()` method returns `true`.

@stof What do you think is the correct behaviour here?

stof:
I would probably classify them as `self`, as we know that this is a case triggered your own deprecation, which only you can fix, not an third-party maintainer not knowing your code (not counting the case of circular dependencies).

sebastianbergmann:
AFAICS, all discussions are resolved. Do you agree, @stof, that this can be merged now? Thank you!

stof (commented):
LGTM

sebastianbergmann:
From the bottom of my heart, thank you so much @stof for guiding me through this.

sebastianbergmann:
Merged into `11.5` and merged to `12.5`, `13.0`, and `main` from there. This will ship with the next bug fix releases of PHPUnit 11.5, PHPUnit 12.5, and PHPUnit 13.0.

PR #6490, opened 2026-02-04
Fix tests count when tests were skipped in `BeforeClass` hook

nikophil:
Hello,

when some tests are skipped in a "before class" hook, there is a problem in the output.

Let's say our test suite have:
- 100 tests in a class which is skipped using `#[RequiresPhp('^9.0')]`
- 100 other tests wich will run normally

here what its output looks like:
```
...............................................................  63 / 200 ( 31%)
.....................................SSSSSSSSSSSSSSSSSSSSSSSSSS 126 / 200 ( 63%)
SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS 189 / 200 ( 94%)
SSSSSSSSSSS                                                     200 / 200 (100%)
 
Time: 00:00.025, Memory: 10.00 MB
 
OK, but some tests were skipped!
Tests: 200, Assertions: 100, Skipped: 100.
```
(BTW, the output is the same if the tests are skipped by calling `self::markTestSkipped()` in a "before" hook)

and now, the very same suite, but `self::markTestSkipped()` is called in a "before class" hook:
```
...............................................................  63 / 200 ( 31%)
.....................................
 
Time: 00:00.016, Memory: 10.00 MB
 
OK, but some tests were skipped!
Tests: 100, Assertions: 100, Skipped: 1.
```

We can see several problems in the output, comparing to the previous one:
- the second line of the progress bar does not display any count
- skipped tests are not displayed in the progress bar
- there are only 100 dots int the progress whereas the total displayed at the end of the line is 200
- the summary seems to not reflect the total number of tests, nor the number of skipped tests

Here is a proposal to make this case coherent with any other mean to skip a full class of tests.

### DIFF
--- a/src/Runner/TestResult/Collector.php
+++ b/src/Runner/TestResult/Collector.php
@@ -233,6 +233,8 @@ final class Collector
         }
 
         $this->testSuiteSkippedEvents[] = $event;
+
+        $this->numberOfTestsRun += $event->testSuite()->count();
     }
 
     public function testSuiteStarted(TestSuiteStarted $event): void
--- a/tests/end-to-end/regression/5165.phpt
+++ b/tests/end-to-end/regression/5165.phpt
@@ -14,9 +14,14 @@ PHPUnit %s by Sebastian Bergmann and contributors.
 
 Runtime: %s
 
+SS                                                                  2 / 2 (100%)
+
+Time: %s, Memory: %s MB
+
 There was 1 skipped test suite:
 
 1) Issue5165Test
 message
 
-No tests executed!
+OK, but some tests were skipped!
+Tests: 2, Assertions: 0, Skipped: 2.
--- a/src/Runner/TestResult/TestResult.php
+++ b/src/Runner/TestResult/TestResult.php
@@ -9,6 +9,8 @@
  */
 namespace PHPUnit\TestRunner\TestResult;
 
+use function array_map;
+use function array_sum;
 use function count;
 use PHPUnit\Event\Test\AfterLastTestMethodErrored;
 use PHPUnit\Event\Test\BeforeFirstTestMethodErrored;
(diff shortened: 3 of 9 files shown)


src/Runner/TestResult/Collector.php
```diff
@@ -233,6 +233,12 @@ public function testSuiteSkipped(TestSuiteSkipped $event): void
         }
 
         $this->testSuiteSkippedEvents[] = $event;
+
+        if ($this->prepared) {
+            return;
+        }
```
nikophil: not sure if this condition is needed, but I've mimicked what's done in `testSkipped()`

tests/end-to-end/regression/5165.phpt
```diff
@@ -14,9 +14,14 @@ PHPUnit %s by Sebastian Bergmann and contributors.
 
 Runtime: %s
 
+SS                                                                  2 / 2 (100%)
+
+Time: 00:00, Memory: 10.00 MB
```
nikophil: is the impact on this case tolerable?

sebastianbergmann: I do not understand what impact you are refering to here. However: in the above, `00:00` and `10.00 MB` must be replaced with `%s`.

nikophil: yeah, I've already fixed this.

I was just underlying that in this case (only one test class which is skipped), the output has changed

src/Runner/TestResult/Collector.php
nikophil: codecov yells about it, and I don't know how to mave this condition true, so I'll remove the condition

PR #6518, opened 2026-02-16
chore: remove extensions that are enabled by default

drupol:
This PR removes the following `ext-*` requirements from `composer.json`, as they are enabled by default.
The `composer.lock` file was updated accordingly.

Motivation: downstream packaging / devshells (notably Nix) flag these as redundant constraints and it can create unnecessary friction when building minimal environments.

Origin of the issue: https://github.com/loophp/nix-shell/issues/295

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,12 +23,7 @@
     "prefer-stable": true,
     "require": {
         "php": ">=8.4.1",
-        "ext-dom": "*",
-        "ext-json": "*",
-        "ext-libxml": "*",
         "ext-mbstring": "*",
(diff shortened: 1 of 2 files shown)


sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it for the reason already given in https://github.com/sebastianbergmann/phpunit/pull/6428#issuecomment-3596393039.

drupol:
Fair enough, thanks ! Going to investigate on how to find a workaround.

PR #6531, opened 2026-02-28
Improve error message when code coverage filter has no matching files

SeeTheRianBow:
Changed 'Incorrect filter configuration' to 'Configured filter does not match any files' to be more accurate when filter is configured but no files exist.


sebastianbergmann:
Cherry-picked into `10.5` and merged to `11.5`, `12.5`, `13.0`, and `main` from there.

Issue #6493, opened 2026-02-06
Improve performance of issue trigger identification


sebastianbergmann:
### Context

When PHPUnit handles deprecation warnings during test execution, the `ErrorHandler` uses `SourceFilter::includes()` to determine whether the deprecation was triggered in first-party code or third-party code (self, direct, or indirect).

As reported in #6111 and #6469, for example, some users are experience a significant performance degradation since issue trigger identification was introduced.

### `SourceMapper::map()` is slow

`SourceFilter::includes()` uses a map that is generated by `SourceMapper::map()`. The latter scans the configured `<source>` directories to build a list of first-party files.

This file scanning operation can cause significant performance degradation, especially when:
- The `<source>` include/exclude patterns are overly broad
- The configured source directories contain many files
- Many deprecation warnings are triggered during a test run

### `SourceMapper::map()` is called multiple times when process isolation is used

`SourceMapper::map()` is called only once per PHPUnit process. However, when tests are run in separate processes (`--process-isolation` CLI option, `processIsolation="true"` XML configuration file setting, or `#[RunInSeparateProcess]` and `#[RunTestsInSeparateProcesses]` attributes) then `SourceMapper::map()` performs the potentially expensive file scanning operation multiple times, once in the parent process and once in each child process.

### Plan

- [X] #6488
- [x] #6495
- [ ] #6535

PR #6535, opened 2026-03-08
Use `sebastian/file-filter` in `SourceFilter::includes()` for issue trigger identification

sebastianbergmann:
`FileFilterMapper::map()` maps a `PHPUnit\TextUI\Configuration\Source` object to a `SebastianBergmann\FileFilter\Filter` object. `SebastianBergmann\FileFilter\Filter\SourceFilter` now uses this `SebastianBergmann\FileFilter\Filter` object.

### Backward Compatibility

The changes proposed here mostly preserve backward compatibility. The following sections describe backward compatibility breaks in detail.

- File and directory inclusion/exclusion semantics are preserved
- Prefix and suffix filtering works identically
- Explicit file includes are still overridden by directory excludes (same precedence as old implementation)
- The includes() method signature and return type are unchanged for all callers
- Code coverage filtering is unaffected (still uses SourceMapper::mapForCodeCoverage())
- All tests for `SourceFilter` pass, including the additional tests that were implemented on `main` while working on these changes

I think the backward compatibility breaks discussed below are acceptable, even in a minor version such as PHPUnit 13.x, considering the performance improvement.

The only way to [improve performance of issue trigger identification](https://github.com/sebastianbergmann/phpunit/issues/6493) I see is implemented by these changes. We *could* keep the old implementation, introduce the new implementation as an alternative that can be opted-in using the XML configuration file. In the future, we could then deprecate the old implementation and later remove it. However, I want to avoid having to maintain two implementations.

#### Symlinks

The old implementation resolved symlinks and relative paths via `realpath()`. The new implementation does string-based matching only.

#### Hidden directories

The old implementation excluded hidden directories as a side effect of `SebastianBergmann\FileIterator\Facade` skipping hidden directories during filesystem traversal. This only happened on non-Windows platforms.

The new implementation excludes hidden directories on all platforms, including Windows. This is a minor behavioral change that only affects Windows users who have source code in hidden directories. The new behavior is arguably more consistent across platforms.

### DIFF
--- a/src/TextUI/Configuration/FileFilterMapper.php
+++ b/src/TextUI/Configuration/FileFilterMapper.php
@@ -26,7 +26,6 @@ final class FileFilterMapper
             $this->files($source->includeFiles()),
             $this->directories($source->excludeDirectories()),
             $this->files($source->excludeFiles()),
-            false,
         );
     }
 
--- a/src/TextUI/Configuration/FileFilterMapper.php
+++ b/src/TextUI/Configuration/FileFilterMapper.php
@@ -9,6 +9,7 @@
  */
 namespace PHPUnit\TextUI\Configuration;
 
+use function realpath;
 use SebastianBergmann\FileFilter\Builder as FilterBuilder;
 use SebastianBergmann\FileFilter\Filter as FileFilter;
 
@@ -37,8 +38,10 @@ final class FileFilterMapper
         $result = [];
 
         foreach ($directories as $directory) {
+            $path = realpath($directory->path());
+
             $result[] = [
-                'path'   => $directory->path(),
+                'path'   => $path !== false ? $path : $directory->path(),
                 'prefix' => $directory->prefix(),
                 'suffix' => $directory->suffix(),
             ];
@@ -55,7 +58,9 @@ final class FileFilterMapper
         $result = [];
 
         foreach ($files as $file) {
-            $result[] = $file->path();
+            $path = realpath($file->path());
+
+            $result[] = $path !== false ? $path : $file->path();
         }
 
         return $result;
--- a/src/TextUI/Configuration/FileFilterMapper.php
+++ b/src/TextUI/Configuration/FileFilterMapper.php
@@ -26,6 +26,7 @@ final class FileFilterMapper
             $this->files($source->includeFiles()),
             $this->directories($source->excludeDirectories()),
             $this->files($source->excludeFiles()),
+            false,
         );
     }
 
--- a/src/TextUI/Configuration/FileFilterMapper.php
+++ b/src/TextUI/Configuration/FileFilterMapper.php
@@ -1,0 +1,63 @@
+<?php declare(strict_types=1);
+/*
+ * This file is part of PHPUnit.
+ *
+ * (c) Sebastian Bergmann <sebastian@phpunit.de>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+namespace PHPUnit\TextUI\Configuration;
+
+use SebastianBergmann\FileFilter\Builder as FilterBuilder;
+use SebastianBergmann\FileFilter\Filter as FileFilter;
+
+/**
+ * @no-named-arguments Parameter names are not covered by the backward compatibility promise for PHPUnit
+ *
+ * @internal This class is not covered by the backward compatibility promise for PHPUnit
+ */
+final class FileFilterMapper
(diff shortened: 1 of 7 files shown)


src/TextUI/Configuration/FileFilterMapper.php
```diff
@@ -0,0 +1,63 @@
...
+ * @no-named-arguments Parameter names are not covered by the backward compatibility promise for PHPUnit
+ *
+ * @internal This class is not covered by the backward compatibility promise for PHPUnit
+ */
+final class FileFilterMapper
+{
+    public function map(Source $source): FileFilter
+    {
+        return (new FilterBuilder)->build(
+            $this->directories($source->includeDirectories()),
```
dantleech, line 26: it seems the include directories are not "normalized" - when the config is in a sub-drectory (e.g. `config/phpunit.xml`):

```
    <source ignoreSelfDeprecations="true">
        <include>
            <directory suffix=".php">../src/</directory>
        </include>
    </source>
```

Results in the regex:

```
#^/home/daniel/www/dantleech/phpunit\-6111/config/\.\./src/(?:/|$)#
```

But the path needs to be normalized from `phpunit-6111/config/../src` to`phpunit-6111/src` in order for it to match the source paths.

(see [Symfony's Path component](https://github.com/symfony/symfony/blob/8.1/src/Symfony/Component/Filesystem/Path.php#L66) for an example)

sebastianbergmann: Good catch, should be handled by c9821a4ed4a1cecdd29c654f6cb93b6b433c5541.

sebastianbergmann:
The proposed changes replace the eager, filesystem-walking `SourceMapper` with a lazy, pattern-based filter backed by `sebastian/file-filter`. I want to explain why I am comfortable merging this in a minor version rather than holding it for the next major.

The old mechanism walked the filesystem at startup via `FileIteratorFacade`, materialised a set of canonical file paths, and answered `SourceFilter::includes()` with a simple `isset()` lookup. The new mechanism compiles the configured include and exclude directories into regular expression matchers, keeps the explicitly listed files in maps, and answers the same question by matching `dirname()` and `basename()` against those matchers.

The public API of `SourceFilter` is unchanged. The only user-visible signal is that `SourceFilter::includes()` now returns the same answer without first scanning the project tree.

I have gone through every shape of input the filter is expected to handle and convinced myself that, for every file that actually exists in the source tree, the two implementations produce the same decision:

- Both implementations reject files under a hidden subdirectory of an include root, and both accept a hidden file when it is listed explicitly via `includeFiles` or when the hidden segment is the include root itself.
- Both implementations canonicalise via `realpath()`. The new mechanism received this treatment in 8af052774; without that commit a regular expression built from a raw `/b/../a` would not have matched a canonical `/.../a/...` directory, and the behaviours would have diverged.
- Both implementations reject explicitly included files that live inside an excluded directory.
- Both implementations apply prefix and suffix filtering on included and excluded directories identically.

To keep myself honest I added the same parametrised scenarios to both `SourceFilterTest` (which exercises the new `FileFilterMapper`) and `SourceMapperTest` (which still exercises the old `SourceMapper`). Both suites pass. Any future drift between the two mechanisms will show up as a concrete test failure.

There are a handful of corner cases where the two mechanisms could in principle disagree:

- The old mechanism is a materialised set built at startup, whereas the new one is pattern-based. A query for a path that does not exist on disk could, in theory, be rejected by the old mechanism and accepted by the new one. PHPUnit only ever queries the filter with paths of source files that were actually loaded during test execution, so this case should be unreachable in practice.
- The new mechanism interprets `*`, `**`, and `?` in directory paths as glob wildcards. The old mechanism treated them as literal characters, and because `realpath()` on such a path returns `false`, effectively ignored them. This is an added capability, not a change in behaviour on valid existing inputs.
- Files created after the filter has been built would have been rejected by the old mechanism (they were not in the precomputed map) and would be accepted by the new one. This is not a scenario PHPUnit exercises.

None of these differences change the accept/reject outcome for any input PHPUnit actually feeds the filter.

The motivation for this change is performance: we no longer walk the entire configured source tree up front to answer a question that only needs a pattern match. The user-visible contract of `SourceFilter` is preserved, the behaviour on every realistic input is preserved, and the test suite now pins both implementations to the same set of expectations.

To the best of my knowledge, this does not break backward compatibility, and I am therefore comfortable merging it into the next minor release.

dantleech:
thanks @sebastianbergmann !

Issue #6542, opened 2026-03-16
Expose interfaces for extension bootstrap dependencies


laloona:
## Summary

When implementing and unit-testing a custom PHPUnit extension, the dependencies injected into

```php
public function bootstrap(
    Configuration $configuration,
    Facade $facade,
    ParameterCollection $parameters
): void
```

are concrete classes.

From a testing perspective, this makes extension code unnecessarily hard to unit-test in isolation. If these collaborators are final or otherwise not intended to be doubled, extension authors cannot easily mock/stub them and are pushed towards broader integration tests.

It would be helpful if PHPUnit exposed interfaces for these types, for example:

- PHPUnit\Runner\Extension\Facade
- PHPUnit\Runner\Extension\ParameterCollection
- PHPUnit\TextUI\Configuration\Configuration

or corresponding contracts such as:

- FacadeInterface
- ParameterCollectionInterface
- ConfigurationInterface

## Current behavior

Extensions receive concrete implementations in bootstrap(). This tightly couples extension code to PHPUnit internals and makes isolated unit tests more difficult.

## Expected behavior

Extension authors should be able to depend on interfaces instead of concrete classes, so their extension logic can be tested using test doubles.

## Why this would help

Providing interfaces would:

- improve testability of custom extensions
- reduce coupling to concrete implementations
- make extension code easier to maintain
- provide a clearer public contract for extension authors

## Example

Today an extension looks like this:
```php
use PHPUnit\Runner\Extension\Extension;
use PHPUnit\Runner\Extension\Facade;
use PHPUnit\Runner\Extension\ParameterCollection;
use PHPUnit\TextUI\Configuration\Configuration;

final class MyExtension implements Extension
{
    public function bootstrap(
        Configuration $configuration,
        Facade $facade,
        ParameterCollection $parameters
    ): void {
        // ...
    }
}
```

In practice, testing logic that depends on these collaborators is harder than necessary because there are no dedicated interfaces to mock against.

## Possible solution

Introduce public interfaces for the objects passed to Extension::bootstrap(), while keeping the current concrete implementations internal or implementation-specific.

sebastianbergmann:
Thank you for taking the time to open this issue and for the thoughtful write-up.

[...]

PR #6543, opened 2026-03-16
implemented facade as event and move facade class into extension faca…

laloona:
This could solve #6542

I implemented `Facade` as Interface so that current API stays and moved concrete implementation into `ExtensionFacade` which is used in the `TextUi\Application`.


sebastianbergmann:
Merged manually, thanks.

Issue #6546, opened 2026-03-17
Doubling properties with no or asymmetric hooks


derrabus:
In the light of #6545, I was playing with doubling hooks on (non-virtual) properties and I noticed a limitation I'd like to discuss.

Basically:

* If the property has no hooks, I cannot define behavior for it on the stub.
* If the property has a `set` hook only, I can define behavior for the `set` hook, but not the `get` hook (and vice versa).

For example:

```php
final class PropertyHookTest extends TestCase
{
    public function testGetHook(): void
    {
        $someClass = self::createStub(SomeClass::class);
        $someClass->method('$someProperty::get')->willReturn('Hello');

        $this->assertSame('Hello', $someClass->someProperty);
    }

    public function testSetHook(): void
    {
        $someClass = $this->createMock(SomeClass::class);
        $someClass
            ->expects(self::once())
            ->method('$someProperty::set')
            ->willReturnCallback(static function (mixed $value): void {
                self::assertSame('Hello', $value);
            });

        $someClass->someProperty = 'Hello';
    }
}

class SomeClass
{
    public string $someProperty;

    public function __construct()
    {
        $this->someProperty = 'some value';
    }
}
```

This limitation appears rather arbitrary to me, given that PHP allows you to add hooks to a property in a child class even though the parent class did not define any or only one hook on that property.

In the context of our example, the following declaration is perfectly valid:

```php
class SomeChildClass extends SomeClass
{
    public string $someProperty {
        get => $this->someProperty;
        set => $this->someProperty = $value;
    }
}
```

In fact, if I change my original test to mock `SomeChildClass` instead of `SomeClass`, the test passes.

My questions:
* Are there technical reasons against allowing mocks to add hooks that the parent class did not have?
* Would you be open to a PR that allows adding hooks to mocked properties?

sebastianbergmann:
> Are there technical reasons against allowing mocks to add hooks that the parent class did not have?

I do not know.

> Would you be open to a PR that allows adding hooks to mocked properties?

Thank you for taking the time to write this up so thoroughly, and for linking it to #6545. I can see you have thought carefully about this and I trust you that this is important.

I want to be upfront with you: I struggle with this one, and I want to explain why rather than just leave you without a response.

[...]

PR #6549, opened 2026-03-25
Allow to stub both hooks of non-virtual properties

### LINKED ISSUE #6545: Stubbing a class with set property hook leaves property uninitialized by default
| Q                   | A
| PHPUnit version     | 13.0.5
| PHP version         | 8.5.3

#### Summary

Hello Sebastian, I hope this bug report finds you well.

I'm not really sure if we should treat this as a bug because it uncovers a issue with badly written tests in the testsuite of the codebase I inherited. This bug (also) occurs because of a recent change in Symfony, but below you'll find a reproducer you can run on a fresh setup with Symfony being involved.

So, my testsuite mocks Symfony's `Request` class like this:

```php
$request = self::createStub(Request::class);
$request->headers = new HeaderBag([
    'some-header' => 'Some Value',
]);
```

This mock then is passed to the method under test that accesses the header bag.

```php
$myHeader = $request->headers->get('some-header');
```

While I generally disagree with mocking the `Request` class like this, this works fine. Currently.

The thing is: those public properties were not meant to be _set_ from outside. This is why Symfony decided to hook into the property (important: only a `set` hook, no `get` hook) and emit a deprecation notice if it's being set, so that in our next major release we can redeclare the property with async visibility (symfony/symfony#63541).

We assumed that adding a hook to a property is not a breaking change. However, it changes PHPUnit's behavior for that property. The set hook is now stubbed and the default implementation essentially discards any value passed to it, leaving the property uninitialized.

#### Current behavior

```
Error: Typed property TestStub_SomeClass_0461fa84::$someProperty must not be accessed before initialization
```

#### How to reproduce

```php
final class PropertyHookTest extends TestCase
{
    public function testHook(): void
    {
        $someClass = self::createStub(SomeClass::class);
        $someClass->someProperty = new \ArrayObject(['foo' => 'bar']);

        $this->assertSame('bar', $someClass->someProperty->offsetGet('foo'));
    }
}

class SomeClass {
    public \ArrayObject $someProperty {
        set => $this->someProperty = $value;
    }

    public function __construct()
    {
        $this->someProperty = new \ArrayObject();
    }
}
```

#### Expected behavior

Ideally, the test should pass with and without the `set` hook. But I assume that this has other implications that I haven't thought of.

derrabus:
Fixes #6545, addresses #6546 partly.

When creating a mock or a stub from a class with property hooks, PHPUnit currently creates invocation mockers for all configured hooks. This is sensible because the hooks are technically methods and should be treated as such. If a property only configures one of the hooks (get or set), PHPUnit will only create an invocation mocker for that hook and leave the other one unconfigured.

This behavior is perfect for virtual properties, for example:

```php
class MyClass
{
    public string $myProperty
    {
        set {
            $this->logger->debug('Write access to MyClass::$myProperty detected; the value has been discarded!', ['value' => $value]);
        }
    }

    public function __construct(
        private readonly LoggerInterface $logger;
    ) {}
}
```

This class has a virtual property that can only we written to. Any attempt to read from that property will result in a PHP error, `Property MyClass::$myProperty is write-only`. Same for the mock, all good.

For non-virtual properties however, this creates a problem. Take a look at the following class.

```php
class MyClass
{
    public string $myProperty
    {
        set {
            $this->logger->debug('Write access to MyClass::$myProperty detected!', ['value' => $value]);
            $this->myProperty = $value;
        }
    }

    public function __construct(
        private readonly LoggerInterface $logger;
    ) {}
}
```

This is an actual property with a set hook. There's no corresponding get hook, so any read access will read the raw value of the property directly. When creating a mock from this class, the property is in a practically unusable state:

* The constructor is not invoked, so the property is left unintialized.
* Read access will fail because we're reading an uninitialized property.
* Write access will go into our invocation stubber, so the underlying property won't be written to.
* If I wanted to work around this by configuring a return value for the property, PHPUnit won't allow me to do that because the get hook is not stubbed.

The only way I could still initialize this property would be through reflection.

With this PR, I'd like to propose to always create invocation stubbers for both hooks of a non-virtual property if the class has configured at least one of the hooks of a property.

### DIFF
--- a/src/Framework/MockObject/Generator/Generator.php
+++ b/src/Framework/MockObject/Generator/Generator.php
@@ -803,7 +803,7 @@ final class Generator
         }
 
         foreach ($propertiesWithHooks as $property) {
-            if ($property->hasGetHook()) {
+            if ($property->hasGetHook() || !$property->isVirtual() && $property->hasSetHook()) {
                 $configurable[] = new ConfigurableMethod(
                     sprintf(
                         '$%s::get',
@@ -815,7 +815,7 @@ final class Generator
                 );
             }
 
-            if ($property->hasSetHook()) {
+            if ($property->hasSetHook() || !$property->isVirtual() && $property->hasGetHook()) {
                 $configurable[] = new ConfigurableMethod(
                     sprintf(
                         '$%s::set',
@@ -888,6 +888,7 @@ final class Generator
                 $mapper->fromPropertyType($property),
                 $hasGetHook,
                 $hasSetHook,
+                $property->isVirtual(),
                 $setHookMethodParameterType,
             );
         }
--- a/src/Framework/MockObject/Generator/HookedProperty.php
+++ b/src/Framework/MockObject/Generator/HookedProperty.php
@@ -25,17 +25,19 @@ final readonly class HookedProperty
     private Type $type;
     private bool $getHook;
     private bool $setHook;
+    private bool $virtual;
     private ?Type $setterType;
 
     /**
      * @param non-empty-string $name
      */
-    public function __construct(string $name, Type $type, bool $getHook, bool $setHook, ?Type $setterType)
+    public function __construct(string $name, Type $type, bool $getHook, bool $setHook, bool $virtual, ?Type $setterType)
     {
         $this->name       = $name;
         $this->type       = $type;
         $this->getHook    = $getHook;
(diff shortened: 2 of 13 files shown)


sebastianbergmann:
Merged manually into `12.5` and to `13.0` and `main` from there.

derrabus:
Thank you!

PR #6555, opened 2026-03-30
Add an option to ignore undefined deprecation triggers

derrabus:
We would like to configure our deprecation triggers properly on Symfony's individual components.

There's just one problem though: The deprecation triggers may or may not be defined, depending on wheter or not one of our dependencies pulls one of the packages `symfony/deprecation-contracts` or `doctrine/deprecations` as a dependency. PHPUnit however triggers a warning if we configure a trigger that is undefined.

```
There were 3 PHPUnit test runner warnings:

1) Function trigger_deprecation cannot be configured as a deprecation trigger because it is not declared

2) Method Doctrine\Deprecations\Deprecation::trigger cannot be configured as a deprecation trigger because it is not declared

3) Method Doctrine\Deprecations\Deprecation::triggerIfCalledFromOutside cannot be configured as a deprecation trigger because it is not declared
```

However, if one of these triggers is undefined, it's okay for us that PHPUnit does not configure it. We'd like to silence those warnings because we're aware of the issue and we're happy with how PHPUnit handles it.

With this PR I'd like to propose an option that allows us to not emit a warning if one of the deprecation triggers is undefined:

```XML
<deprecationTrigger ignoreUndefinedTriggers="true">
    <function>trigger_deprecation</function>
    <method>Doctrine\Deprecations\Deprecation::trigger</method>
    <method>Doctrine\Deprecations\Deprecation::triggerIfCalledFromOutside</method>
</deprecationTrigger>
```

### DIFF
--- a/phpunit.xsd
+++ b/phpunit.xsd
@@ -372,6 +372,7 @@
                 <xs:element name="method" type="xs:string" minOccurs="0" maxOccurs="unbounded"/>
             </xs:choice>
         </xs:sequence>
+        <xs:attribute name="ignoreUndefinedTriggers" type="xs:boolean" default="false"/>
     </xs:complexType>
     <xs:complexType name="issueTriggerResolversType">
         <xs:sequence>
--- a/src/TextUI/Application.php
+++ b/src/TextUI/Application.php
@@ -793,14 +793,18 @@ final readonly class Application
             'methods'   => [],
         ];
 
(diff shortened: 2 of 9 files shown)


sebastianbergmann:
Don't you need a similar option for issue trigger resolvers?

derrabus:
Thank you for merging my changes. 🙂 

> Don't you need a similar option for issue trigger resolvers?

I'm not aware that we do. The issue trigger resolver that we're working on will be part of our PHPUnit bridge and as such always available to us.

Issue #4440, opened 2020-08-28
Misleading error message when no files present in coverage path


tlamy:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 9.3.8
| PHP version         | 7.4.8
| Installation Method | Composer

#### Summary
When running phpunit in a more or less empty project, "Incorrect filter configuration" is shown when there are no files to cover. This is misleadig.

#### Current behavior

A misleading error message `Warning:       Incorrect filter configuration, code coverage will not be processed` is shown when no source files present in covered path(s)

#### How to reproduce

- Clone https://github.com/tlamy/phpunit-misleading-error
- run `composer install`
- run `composer test-coverage`

#### Expected behavior

Show no error message (maybe a warning that there are no files to cover), but 0% coverage, and "No tests executed".

As soon as a (even an empty) .php file is added to the src directory, the expected behaviour is shown.

sebastianbergmann:
This is a known issue (at least to me). I agree that it is confusing, especially to new users. While I do not have time to work on this myself, I would gladly review/merge a pull request.

tlamy:
I'll have a look into this, maybe this weekend

alecwcp:
I just hit this myself on a non-empty project. Turns out my coverage section was wrong
I had
```
<coverage processUncoveredFiles="true">
  <include>
    <directory suffix=".php">src</directory>
  </include>
</coverage>
```
instead of
```
<coverage processUncoveredFiles="true">
  <include>
    <directory suffix=".php">../src</directory>
  </include>
</coverage>
```
(the directory is relative to the location of the config, so it needed to go up one and then into the src folder).

So I think a change to this reporting no files found as a separate message would be helpful to report the directories it actually searched (would have cottoned me onto my mistake sooner).

burzum:
It would be nice if we could also just skip this check. I have a project template repository that has no files inside but runs all the tools to see if my CI configs work.

msgpix:
Hello just landed here as a project I worked on suddenly stopped to work properly when asking to run tests with coverage. 

What was very strange to my mind was that no update of PHPUnit had been integrated lately in the project. Still, it did require the following new lines in phpunit.xml to work again:
```
    <coverage processUncoveredFiles="true">
        <include>
            <directory suffix=".php">app</directory>
        </include>
    </coverage>
```

The application being based on Laravel, the `app` directory is the conventional one, at least for this project where phpunit.xml is stored in the same directory as `app`.

[...]

PR #6556, opened 2026-03-31
Improve error message when configured code coverage filter is empty

Robert0303-C:
## Summary

Fixes the misleading error message shown when running PHPUnit with code coverage configured but no source files exist.

## Problem

When running `phpunit --coverage-text` with a configured filter that matches no files, the error shown was:

```
Incorrect filter configuration
```

This is misleading - the filter configuration is not incorrect, there are simply no files to cover.

## Solution

Modified `src/Runner/CodeCoverage.php` to check if the filter is empty before emitting the generic error message. When the filter is empty, a clearer message is shown:

```
Configured filter does not match any files, code coverage will not be processed
```

## Files Changed

- `src/Runner/CodeCoverage.php`: Added empty filter check in the catch block

## Testing

The fix follows the existing message style used elsewhere in the same file for the same scenario (line ~130).

### DIFF
--- a/src/Runner/CodeCoverage.php
+++ b/src/Runner/CodeCoverage.php
@@ -442,9 +442,15 @@ final class CodeCoverage
                 $filter,
             );
         } catch (CodeCoverageException $e) {
-            EventFacade::emitter()->testRunnerTriggeredPhpunitWarning(
-                $e->getMessage(),
-            );
+            if ($filter->isEmpty()) {
+                EventFacade::emitter()->testRunnerTriggeredPhpunitWarning(
+                    'Configured filter does not match any files, code coverage will not be processed',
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
Thank you for your contribution.

While the proposed changes look okay at first glance, I do have some concerns.

The pull request description, with its structured "Summary / Problem / Solution / Files Changed / Testing" format, appears to have been entirely written by an AI agent. Did you actually verify whether these changes fix #4440 as the commit message claims they do? I ask this because neither an existing test was updated nor a new test was added. The existing tests just pass with the proposed changes.

I am also concerned by the fact that your GitHub account was created only two days ago and does not have a real name associated with it.

Issue #6200, opened 2025-05-06
Regression in StringMatchesFormatDescriptionTest with 72ea2b54d


edorian:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 12.1.4
| PHP version         | any, tested on 8.3.20
| Installation Method | Composer

#### Summary

With https://github.com/sebastianbergmann/phpunit/commit/72ea2b54d we sadly introduced a regression that breaks %A matching in StringMatchesFormatDescriptionTest for multiline matches.

With %A greedily matching the whole document and thus changing the test result instead of improving the diff.

#### Current behavior

As it's annoying to paste diffs of diffs I made a reproduce:

The test can be seen here: https://github.com/sebastianbergmann/phpunit/compare/main...edorian:phpunit:regression-test-72ea2b54d#diff-ee7078c6148f6e7190b9f9eef5217f3986cbfc027b4608fc9bfab296f2605708R271-R332

#### How to reproduce

Copy the updated test(s) either from my branch or from the end of this ticket into tests/unit/Framework/Constraint/String/StringMatchesFormatDescriptionTest.php</summary>

```
git checkout 72ea2b5^ # revision before the issue
➜  phpunit git:(34e4489c8) ./phpunit tests/unit/Framework/Constraint/String/StringMatchesFormatDescriptionTest.php
PHPUnit 11.5.18 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.3.20
Configuration: /Users/edo/private/phpunit/phpunit.xml

.....................                                             21 / 21 (100%)

Time: 00:00.010, Memory: 10.00 MB

OK (21 tests, 88 assertions)
```

vs

```
➜  phpunit git:(72ea2b54d) ✗ ./phpunit tests/unit/Framework/Constraint/String/StringMatchesFormatDescriptionTest.php
PHPUnit 11.5.18 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.3.20
Configuration: /Users/edo/private/phpunit/phpunit.xml

..................F..                                             21 / 21 (100%)

Time: 00:00.010, Memory: 12.00 MB

There was 1 failure:

1) PHPUnit\Framework\Constraint\StringMatchesFormatDescriptionTest::testFailureMessageWithNewlinesAndAnythingMatcherMultilineMatches
Failed asserting that exception of type "PHPUnit\Framework\ExpectationFailedException" is thrown.
```

#### Expected behavior

[...]

PR #6557, opened 2026-03-31
Improve failure description

sebastianbergmann:
Improve the failure description for the `StringMatchesFormatDescription` constraint when `%a`/`%A` multiline placeholders match content spanning multiple lines.

Previously, `additionalFailureDescription()` compared expected and actual output line-by-line using a single shared index. When `%a` or `%A` consumed multiple actual lines, the arrays went out of sync and all subsequent lines appeared as false diffs — burying the real failure in noise.

This was attempted before in 72ea2b54d60c by modifying `regularExpressionForFormatDescription()` to remove the `$` end-of-string anchor for multiline patterns.

That approach caused regression [#6200](https://github.com/sebastianbergmann/phpunit/issues/6200) because `regularExpressionForFormatDescription()` is shared with `matches()`, which changed test pass/fail outcomes. It was reverted in 9bd312d59af5.

This implementation takes a different approach that avoids the regression:

- Rewrites only `additionalFailureDescription()`
- `matches()` and `regularExpressionForFormatDescription()` are unchanged
- Uses a two-pointer algorithm with anchor lookahead: when a `%a`/`%A` line is encountered, find the next non-multiline expected line (the "anchor") and search for it in the actual output to determine how many lines the placeholder consumed
- Builds a fresh synced array instead of mutating in-place with `array_splice()`

**Before** (multiline `%A` causes full desync):
```diff
 ## before first A
 some multiline
+text for
+A to match
 ## after first A
 *
 ## before second A
+more multiline text
+for A to match
+## after second A
 *
-## after second A
+Foo: s match
 *
-Foo: %s
+Additional Text that is not matched
```

**After** (only the real mismatch is shown):
```diff
 ## after second A
 *
 Foo: s match
+*
+Additional Text that is not matched
```

### DIFF
--- a/src/Framework/Constraint/String/StringMatchesFormatDescription.php
+++ b/src/Framework/Constraint/String/StringMatchesFormatDescription.php
@@ -121,13 +121,7 @@ final class StringMatchesFormatDescription extends Constraint
                             $synced[] = $actual[$i];
                         }
 
-                        if ($expected[$anchorExpectedIndex] === $actual[$anchorActualIndex]) {
-                            $synced[] = $actual[$anchorActualIndex];
-                        } elseif (@preg_match($this->regularExpressionForFormatDescription($expected[$anchorExpectedIndex]), $actual[$anchorActualIndex]) > 0) {
-                            $synced[] = $actual[$anchorActualIndex];
-                        } else {
-                            $synced[] = $expected[$anchorExpectedIndex];
-                        }
+                        $synced[] = $actual[$anchorActualIndex];
 
                         $expectedIndex = $anchorExpectedIndex + 1;
                         $actualIndex   = $anchorActualIndex + 1;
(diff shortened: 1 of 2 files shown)


sebastianbergmann:
@sebastianfeldmann @edorian Can you have a look, please? Thank you!

edorian:
Tested it against my .phpt test suites and everything works well.

The "%A" match failures that are common for us that I remembered and tested didn't change in output.

The tests look a lot nicer and the algo looks a lot more sensible than before.

All around 👍 from me @sebastianbergmann

sebastianbergmann:
Cherry-picked into `main`.

sebastianfeldmann:
Nice idea with the freshly synced array 👍 

All tests in my local branch passed as well. I will check if it makes sense to still commit the tests.

PR #6565, opened 2026-04-05
Optional `$skipWhenEmpty` parameter for `#[DataProvider]` and `#[DataProviderExternal]`

### LINKED ISSUE #6404: Add (opt-in?) support for data providers returning an empty iterable
When the data provider generates test cases based on some external sources, it might happen that there is no cases to be tested.
As PHPUnit [triggers an error](https://github.com/sebastianbergmann/phpunit/blob/02d9f1223c2748d1179e0c298997edb3e7eeae32/src/Metadata/Api/DataProvider.php#L245) when a data provider returns an empty iterable (technically when _all providers configured for a test do it, so that the merged result is empty, but most tests don't use multiple data providers), such projects end up doing hacks where the yield a single case in the data provider with special values that the test detects to call `markTestSkipped`.

Examples:
- async-aws having a test enforcing that modified services include updates to their changelog (a PR might ot modify any of the files requiring changelog entries): https://github.com/async-aws/aws/pull/1983
- the `IntegrationTestCase` of Twig has a dedicated test running the "legacy" test cases, where deprecations are ignored. A project running Twig integration tests might not have any legacy tests: https://github.com/twigphp/Twig/blob/d58c8be15ddbda8532df7ba5de283b51dd8939d2/src/Test/IntegrationTestCase.php#L191

It would be great to have a way to opt-out of that error, marking the tests as skipped due to no provided data instead.

sebastianbergmann:
This implements an optional `$skipWhenEmpty` parameter (defaulting to `false`) for the `#[DataProvider]` and `#[DataProviderExternal]` attributes. When set to `true`, a data provider that returns an empty iterable will cause the test to be skipped instead of triggering an error.

Usage:

```php
#[DataProvider('provider', skipWhenEmpty: true)]
public function testSomething(string $value): void
{
    // ...
}
```

Note: Parameter names are not covered by the backward compatibility promise for PHPUnit.

When all data providers for a test method return no data and at least one of them has `skipWhenEmpty: true`, the test is created as a single test case that skips itself with the message "The data provider for this test provided no data, which is explicitly permitted". The default behavior (error on empty data provider) is unchanged.

Closes #6404.

### DIFF
--- a/src/Framework/Attributes/DataProvider.php
+++ b/src/Framework/Attributes/DataProvider.php
@@ -24,14 +24,16 @@ final readonly class DataProvider
      */
     private string $methodName;
     private bool $validateArgumentCount;
+    private bool $skipWhenEmpty;
 
     /**
      * @param non-empty-string $methodName
      */
-    public function __construct(string $methodName, bool $validateArgumentCount = true)
+    public function __construct(string $methodName, bool $validateArgumentCount = true, bool $skipWhenEmpty = false)
     {
         $this->methodName            = $methodName;
         $this->validateArgumentCount = $validateArgumentCount;
+        $this->skipWhenEmpty         = $skipWhenEmpty;
     }
 
     /**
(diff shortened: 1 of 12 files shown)


sebastianbergmann:
@stof Is this what you had in mind?

stof:
@sebastianbergmann yes, exactly.

PR #6566, opened 2026-04-05
Allow `--stop-on-defect`, `--stop-on-error`, etc. to accept an optional threshold

### LINKED ISSUE #6459: Allow numeric value for stopOnDefect,...
https://docs.phpunit.de/en/12.5/configuration.html#the-stopondefect-attribute and others that have
>after the **first** error.

It would be great if these would also allow numeric (`int<1, max>`) values.
Especially for tests that are PHP version specific/dependent (e.g. PHP 8.4->8.5) or dependent on an external library will often result in 100s of issues after an update (that mostly have the same reason though, but that's irrelevant here) which result in lots of wasted CI time if stopOnDefect is not set.
However if it is set, it's wasting tons of build time when it just stops at the first error.

I'd like to be able to specify a reasonable number (e.g. 5 or 10), so it stops after the first 5 errors

sebastianbergmann:
Closes #6459.


sebastianbergmann:
@kkmuffme Is this what you had in mind? Thank you!

kkmuffme:
Yes, exactly! Thank you

PR #6571, opened 2026-04-06
Static analysis errors for `TestDoubleBuilder` method chaining

sax0n0xas:
After updating to the latest version of phpunit i got a lot of errors from psalm about getMock() invocation

`Method PHPUnit\Framework\MockObject\TestDoubleBuilder::getMock does not exist`

I think that there are issues with return types in TestDoubleBuilder. And return type must be `static`.

My test code looks like this
```
/** @var  MockObject&Obj $mock */
$mock = $this->getMockBuilder(Obj::class)
    ->onlyMethods(['method'])
    ->getMock();
```

### DIFF
--- a/src/Framework/MockObject/TestDoubleBuilder.php
+++ b/src/Framework/MockObject/TestDoubleBuilder.php
@@ -56,7 +56,7 @@ abstract class TestDoubleBuilder
      *
      * @return $this
      */
-    public function onlyMethods(array $methods): self
+    public function onlyMethods(array $methods): static
     {
         if ($methods === []) {
             $this->emptyMethodsArray = true;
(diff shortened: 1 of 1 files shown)


sebastianbergmann:
Cherry-picked into `12.5` and merged to `13.1` and `main` from there.

staabm:
I think its a bug in psalm, that this error is reported.
might be worth opening an issue on the psalm repo with a minimal reproducer.

From PHPDocs `$this->getMockBuilder(Obj::class)` is known to return `MockBuilder` and if a method is invoked on such type which is annotated as `return $this`, this is should be enough to know that the type after `->onlyMethods(['method'])` is still `MockBuilder`.

I think even `/** @var  MockObject&Obj $mock */` should not be necessary (actually PHPStan considers it bad-practice)

PR #6576, opened 2026-04-08
Improve invocation count message

HypeMC:
It's useful to have the actual number of calls displayed.


sebastianbergmann:
Thank you, I must have missed those while working on d82a0dc1234874c25e5016f5a8db86cd00d26fe9 last month.

I have cherry-picked your changes into `13.1 and merged them to `main` from there as `12.5` should only get bug fixes and I do not consider these changes to be bug fixes.

Issue #5860, opened 2024-06-11
Running tests with process isolation does not allow defining auto_prepend_file for child processes


azabolotnyi:
When running tests like this: `php -d auto_prepend_file=foo.php vendor/bin/phpunit` the `foo.php` file will not be loaded for the child `php` processes i.e. for each test case.

### What is expected
`foo.php` to be loaded for each test case in the suite.

### What actually happens
`foo.php` is not loaded because "-d auto_prepend_file" is not passed over to the child processes.

### Versions affected
Both latest 10.x and 11.x are affected. Did not check in 9.x.

### Sample code
https://github.com/azabolotnyi/phpunit-testcases

### Notes
Using `<ini>` in `phpunit.xml` has no effect either, because `auto_prepend_file` can't be affected by `ini_set()` - and that's how `<ini>` values are set by PHPUnit.

### Possible solution
Capture `-d` parameters from the parent `php` process and pass them into the `SeparateProcessTestRunner`.

### Composer info
```
myclabs/deep-copy                  1.11.1 Create deep copies (clones) of your objects
nikic/php-parser                   v5.0.2 A PHP parser written in PHP
phar-io/manifest                   2.0.4  Component for reading phar.io manifest information from a PHP Archive (PHAR)
phar-io/version                    3.2.1  Library for handling version information and constraints
phpunit/php-code-coverage          11.0.3 Library that provides collection, processing, and rendering functionality for PHP code coverage information.
phpunit/php-file-iterator          5.0.0  FilterIterator implementation that filters files based on a list of suffixes.
phpunit/php-invoker                5.0.0  Invoke callables with a timeout
phpunit/php-text-template          4.0.0  Simple template engine.
phpunit/php-timer                  7.0.0  Utility class for timing
phpunit/phpunit                    11.2.1 The PHP Unit Testing framework.
sebastian/cli-parser               3.0.1  Library for parsing CLI options
sebastian/code-unit                3.0.0  Collection of value objects that represent the PHP code units
sebastian/code-unit-reverse-lookup 4.0.0  Looks up which function or method a line of code belongs to
sebastian/comparator               6.0.0  Provides the functionality to compare PHP values for equality
sebastian/complexity               4.0.0  Library for calculating the complexity of PHP code units
sebastian/diff                     6.0.1  Diff implementation
sebastian/environment              7.1.0  Provides functionality to handle HHVM/PHP environments
[...]

PR #6592, opened 2026-04-17
INI metacharacters `;` and `"` are not preserved when forwarding settings to child processes

kayw-geek:
## The bug

`JobRunner` forwards every PHP setting to the child process as `-d name=value`. PHP CLI then parses the value as an INI string, where `;` starts a comment and `"` is a string delimiter. So if your value contains those characters, PHP silently eats parts of it.

I hit this with the `ddtrace` extension. Its default `datadog.appsec.obfuscation_parameter_value_regexp` is a long regex full of `"` (it matches JSON / URL-encoded keys). Every time PHPUnit forks a worker, that regex arrives mangled inside the child, and the extension stops working correctly.

## The fix

When the value contains `;` or `"`, wrap it in `"..."` and escape inner `"` as `\"`. Otherwise pass it through untouched, so values like `output_buffering=Off` or `error_reporting=E_ALL & ~E_NOTICE` keep their existing INI semantics (boolean keywords / bitwise expressions).

```
Before: -d highlight.string=(?:"foo";bar)   → child sees (?:foo
After:  -d highlight.string="(?:\"foo\";bar)" → child sees (?:"foo";bar)
```

## Security side-effect (defense in depth)

The pre-fix code also allowed INI directive injection: a value containing a literal newline would be parsed by PHP as multiple `-d` directives, letting the value override unrelated settings (`memory_limit`, `disable_functions`, `auto_prepend_file`, …).

In PHPUnit's threat model the values come from XML config or the host's own PHP runtime, so an attacker would already need filesystem write access to abuse it. Still, the new quoting closes that path because the value never escapes its quoted string.

## Test

Added one case to `JobRunnerTest`: feed the real ddtrace regex (634 bytes, lots of `"`) through a forked process, have the child `print ini_get(...)`, assert the output equals the input byte-for-byte. Fails on `main`, passes with this change. Full unit + end-to-end suites green on Ubuntu / Windows × PHP 8.4 / 8.5 / 8.6.

### DIFF
--- a/src/Util/PHP/JobRunner.php
+++ b/src/Util/PHP/JobRunner.php
@@ -26,6 +26,7 @@ use function is_file;
 use function is_resource;
 use function proc_close;
 use function proc_open;
+use function str_contains;
 use function str_replace;
 use function str_starts_with;
 use function stream_get_contents;
@@ -335,9 +336,14 @@ final readonly class JobRunner
     }
 
     /**
-     * Wraps the value portion of a "name=value" INI setting in double quotes
-     * so PHP's INI parser treats characters such as `;` (comment) and `"`
-     * (string delimiter) as literal data instead of metacharacters.
+     * Quotes the value portion of a "name=value" INI setting only when it
+     * contains characters PHP's INI parser would otherwise interpret as
+     * metacharacters (`;` starts a comment, `"` is a string delimiter).
+     *
+     * Quoting is avoided for plain values so that boolean keywords such as
+     * `On` / `Off` keep their special INI semantics; wrapping them in quotes
+     * turns them into the literal strings `"On"` / `"Off"` and breaks
+     * settings like `output_buffering`.
      */
     private function quoteSettingValue(string $setting): string
     {
@@ -347,8 +353,13 @@ final readonly class JobRunner
             return $setting;
         }
 
-        $name  = substr($setting, 0, $position);
         $value = substr($setting, $position + 1);
+
+        if (!str_contains($value, ';') && !str_contains($value, '"')) {
+            return $setting;
+        }
+
+        $name = substr($setting, 0, $position);
 
         return $name . '="' . str_replace('"', '\\"', $value) . '"';
     }
--- a/src/Util/PHP/JobRunner.php
+++ b/src/Util/PHP/JobRunner.php
@@ -26,8 +26,11 @@ use function is_file;
 use function is_resource;
 use function proc_close;
 use function proc_open;
+use function str_replace;
 use function str_starts_with;
 use function stream_get_contents;
+use function strpos;
+use function substr;
 use function sys_get_temp_dir;
 use function tempnam;
 use function trim;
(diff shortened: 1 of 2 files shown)


sebastianbergmann:
Is this a follow-up to #5860 which was recently addressed in f0ad610f8c9c8c28787d48dd10bc059f92ae135e? Even if not, I think this should target the `12.5` branch instead of `main`.

sebastianbergmann:
I will cherry-pick this to `12.5` and then merge to `13.1` and `main` from there.

sebastianbergmann:
Merged manually, thanks.

I also created https://github.com/sebastianbergmann/phpunit/security/advisories/GHSA-qrr6-mg7r-m243 for the security-related aspect of this.

ADmad:
Will a new release be done for PHPUnit 11 too with this security fix?

Currently composer refuses to install PHPUnit 11 on PHP 8.2 due to the security advisory published.

sebastianbergmann:
> Will a new release be done for PHPUnit 11 too with this security fix?

PHPUnit 11 is not affected. Only PHPUnit 12.5.21 and PHPUnit 13.1.5 are affected.

Please see https://github.com/github/advisory-database/pull/7430.

ADmad:
> PHPUnit 11 is not affected.

Hmm.. Then I wonder why composer is failing to install phpunit 11 on php 8.2 as seen below. It's as if it ignores the constraint for v11. This was not an issue until a few hours ago.

https://github.com/cakephp/cakephp/actions/runs/24597726343/job/71930898997?pr=19404#step:11:100

sebastianbergmann:
> > PHPUnit 11 is not affected.
> Then I wonder why composer is failing to install phpunit 11 on php 8.2 as seen below.

PHPUnit 11 is not affected by the issue this advisory is about.

The advisory I created yesterday at https://github.com/sebastianbergmann/phpunit/security/advisories/GHSA-qrr6-mg7r-m243 had the correct affected-version information.

When GitHub published it at https://github.com/advisories/GHSA-qrr6-mg7r-m243, they changed that information into something incorrect, which is what is now causing Composer to refuse installs it shouldn't be refusing.

I have opened https://github.com/github/advisory-database/pull/7430 to get this corrected on GitHub's side.

I also consulted @naderman and @Seldaek, the maintainers of Composer and Packagist. At their suggestion I submitted https://github.com/FriendsOfPHP/security-advisories/pull/762. That database takes precedence over the GHSA advisory for Composer, so the false positive on PHPUnit 11 should (have) go(ne) away. Note that https://packagist.org/packages/phpunit/phpunit/advisories shows the correct "affected versions" information.

There is nothing I can do from my end to fix this, and I'd appreciate you not directing the frustration at me. The original advisory I published was correct, and I did not introduce the error. I am also not available to look into specific CI failures like yours, especially not on a weekend.

ADmad:
Thank you for the clarification.

> ...I'd appreciate you not directing the frustration at me.

I am pretty sure I did nothing of the sort :) My comments were all inquires trying to figure out what was causing the failures.

sebastianbergmann:
> I am pretty sure I did nothing of the sort :)

I am sorry, you did not. That was supposed to go into a different comment.

PR #6598, opened 2026-04-24
`--disable-coverage-targeting` CLI option

sebastianbergmann:


sebastianbergmann:
@BackEndTea @fkupper @danon Any feedback from you on whether this addresses your need expressed in https://github.com/sebastianbergmann/php-code-coverage/issues/571 would be much appreciated. Thanks!

danon:
@sebastianbergmann is this exactly the same feature as in https://github.com/sebastianbergmann/php-code-coverage/issues/571 but the flag name is different? Because if so, that works for me!

sebastianbergmann:
> @sebastianbergmann is this exactly the same feature as in [sebastianbergmann/php-code-coverage#571](https://github.com/sebastianbergmann/php-code-coverage/issues/571) but the flag name is different? Because if so, that works for me!

Yes, to the best of my understanding this provides exactly what was asked for in https://github.com/sebastianbergmann/php-code-coverage/issues/571: a CLI-only option (`--disable-coverage-targeting`) that, for a single PHPUnit run, makes coverage collection behave as if the targeting metadata (`#[Covers*]` attributes, including `#[CoversNothing]`) were not present.

danon:
> @BackEndTea @fkupper @danon Any feedback from you on whether this addresses your need expressed in [sebastianbergmann/php-code-coverage#571](https://github.com/sebastianbergmann/php-code-coverage/issues/571) would be much appreciated. Thanks!

> Yes, to the best of my understanding this provides exactly what was asked for in [sebastianbergmann/php-code-coverage#571](https://github.com/sebastianbergmann/php-code-coverage/issues/571): a CLI-only option (`--disable-coverage-targeting`) that, for a single PHPUnit run, makes coverage collection behave as if the targeting metadata (`#[Covers*]` attributes, including `#[CoversNothing]`) were not present.

Then yes, this works perfectly for me.

PR #6604, opened 2026-05-03
Add support for partially ordered parameter sets

laloona:
Introduce new parameter matching rule for the MockObject framework:
`PartialOrderedParameterSets`: validates parameters across multiple invocations while allowing arbitrary order, except for explicitly fixed parameter sets.

Added `IndexedParameters` to associate parameter sets with a specific invocation index.
This allows distinguishing between fixed-position parameter sets and those that can be matched in arbitrary order.

Added `fixed()` to `Assert` as a convenience method to create indexed parameter sets without exposing the underlying implementation.

### DIFF
--- a/src/Framework/TestCase.php
+++ b/src/Framework/TestCase.php
@@ -97,6 +97,7 @@ use PHPUnit\TestRunner\TestResult\PassedTests;
 use PHPUnit\TextUI\Configuration\Registry as ConfigurationRegistry;
 use PHPUnit\Util\DifferBuilder;
 use PHPUnit\Util\Exporter;
+use PHPUnit\Util\Sanitizer;
 use PHPUnit\Util\Test as TestUtil;
 use ReflectionClass;
 use ReflectionMethod;
@@ -887,7 +888,10 @@ abstract class TestCase extends Assert implements Reorderable, SelfDescribing, T
                 return sprintf(' with data set #%s', $this->dataName);
             }
 
-            return sprintf(' with data set "%s"', $this->dataName);
+            return sprintf(
+                ' with data set "%s"',
+                Sanitizer::sanitizeBidirectionalControlCharacters($this->dataName),
+            );
         }
 
         return '';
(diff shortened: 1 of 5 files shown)


sebastianbergmann:
I would like to raise a few points before I merged this:

**1. `Assert::fixed()` exposes an `@internal` return type**

`IndexedParameters` is marked `@internal`, but `Assert::fixed()` is a public, documented entry point and its return type leaks the internal class into IDEs, stubs, and the API surface check. Could we either:

- introduce a small public marker interface (e.g. `FixedParameterSet`) that `IndexedParameters` implements, and return that from `fixed()`; or
- drop `@internal` from `IndexedParameters`, accepting it as part of the documented contract?

Ideally, we should be able to not need `Assert::fixed()` at all. This would also help with the next point.

**2. `Assert::fixed()` is not a good name**

`Assert::fixed()` is quite generic and does not hint that it is about pinning a parameter set to a specific invocation index. Again, ideally we solve this by not adding this method.

**3. Nullable `$index` vs. `at(): int`**

In `IndexedParameters`, `$index` defaults to `null`, but `at()` is typed `: int`. If `at()` is ever called before `index()` has been set, we will run into a type error, I think. Since both current callers (`PartialOrderedParameterSets`, `UnorderedParameterSets`) set the index immediately after construction, could the index simply be passed through the constructor? That removes the two-step init and the nullable field.

**4. Unrelated test renames**

The `Consecutive...` → `Ordered...` renames in `MockObjectTest.php` look like a sensible cleanup, but they are not related to the partial-order feature, at least not as far as I can see. Could those be split into a separate PR? It would make this one easier to review and revert if ever needed.

**5. Minor follow-ups**

- `OrderedParameterSets`' `@param` tag was loosened from `list<Parameters>` to `list<mixed>`. Was that intended? The class itself is not otherwise touched. Also note that I recently pushed changes affecting the types here (in an effort to reach PHPStan level 10).
- I am not sure about the message `"Too many parameter sets given, 2 out of 3 …"`. It seems to be emitted when *fewer* invocations than configured occurred. The wording appears to be inherited from `OrderedParameterSets`, but now it is duplicated.
- Test names alternate between `…WithPartialOrderedParameterSets…` and `…WithParameterSetsInPartialOrder…`, consistency would be much appreciated.

Thanks again for the contribution!

laloona:
Sorry it was closed when I fetched the new commits from main. I'll open a new one. Sorry for that.

PR #6606, opened 2026-05-08
Support for partially ordered parameter sets in mock object expectations

laloona:
Follow up from PR [6604](https://github.com/sebastianbergmann/phpunit/pull/6604).

I changed to API to that:
```php
$mock->expects($this->exactly(4))
    ->method('dispatch')
    ->withParameterSetsInPartialOrder(
        ['pinned' => $a],
        [$b],
        [$c],
        ['pinned' => $d],
    );
```
So `Assert::fixed` is removed.
Additional I removed the duplication of the exception. 
The index in `IndexedParameters` is moved to the constructor, so it can't be null now.

I hope the naming for the tests is now more consistent.

I noticed that the `@param list<mixed> $parameters` is correct. When we enter the construct $parameters are not clearly defined as they represent argument sets.


sebastianbergmann:
Thank you!

PR #6681, opened 2026-05-27
Keep XML comparison assertions comment-insensitive by default and add comment-aware variants

sebastianbergmann:
A recent change in `sebastianbergmann/comparator` made XML comments significant when comparing DOM documents (`C14N(false, true)`). Because PHPUnit's XML assertions delegate to that comparator, their behavior silently changed — two documents that differ only in comments are now treated as different. This surfaced in https://github.com/sebastianbergmann/comparator/issues/107#issuecomment-4553052370.

That change cuts both ways, and the second direction is the dangerous one. For two documents that differ only in comments:

| Assertion | Long-standing behavior | After the comparator change |
|---|---|---|
| `assertXmlStringEqualsXmlString()` | passes | fails (loud, you notice) |
| `assertXmlStringNotEqualsXmlString()` | fails | passes (silent, you do not notice |

The `Equals` regression is loud: a green test turns red and you fix it. The `NotEquals` regression is silent: a test written to prove that output actually changed now passes even when the only difference is a comment. This is a real assertion that quietly stops protecting you.

This pull request restores the long-standing, comment-insensitive behavior as the default for all six XML assertions, so existing test suites keep working as before:

- `assertXmlFileEqualsXmlFile()` / `assertXmlFileNotEqualsXmlFile()`
- `assertXmlStringEqualsXmlFile()` / `assertXmlStringNotEqualsXmlFile()`
- `assertXmlStringEqualsXmlString()` / `assertXmlStringNotEqualsXmlString()`

For the cases where comments should be part of the comparison, six new opt-in variants make that explicit:

- `assertXmlFileEqualsXmlFileConsideringComments()` / `assertXmlFileNotEqualsXmlFileConsideringComments()`
- `assertXmlStringEqualsXmlFileConsideringComments()` / `assertXmlStringNotEqualsXmlFileConsideringComments()`
- `assertXmlStringEqualsXmlStringConsideringComments()` / `assertXmlStringNotEqualsXmlStringConsideringComments()`

**Example**

```php
$expected = '<order><!-- generated at 2026-05-27 --><id>42</id></order>';
$actual   = '<order><id>42</id></order>';

// Default: comments are ignored — these are equal
$this->assertXmlStringEqualsXmlString($expected, $actual);

// Opt in to treating comments as significant — these differ
$this->assertXmlStringNotEqualsXmlStringConsideringComments($expected, $actual);
```

The default behavior matches what PHPUnit did before the comparator change, so no test-suite migration is required.

Only comments are ignored by the default assertions; element structure, attributes, and text content are always compared.

### DIFF
--- a/src/Framework/Assert.php
+++ b/src/Framework/Assert.php
@@ -2597,7 +2597,7 @@ abstract class Assert
     }
 
     /**
-     * Asserts that two XML files are equal.
+     * Asserts that two XML files are equal, ignoring comments.
      *
      * @throws Exception
      * @throws ExpectationFailedException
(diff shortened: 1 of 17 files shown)


sebastianbergmann:
I intend to backport the "Keep XML comparison assertions comment-insensitive by default" part to PHPUnit 12.5 and PHPUnit 13.1 but "add comment-aware variants" only for PHPUnit 13.2 (`main`).

derrabus:
Looks good, thank you!

Issue #5845, opened 2024-05-28
Error handlers registered before PHPUnit (e.g. via `auto_prepend_file`) cause false "risky test" warnings


kkmuffme:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 11.1.3
| PHP version         | 8.2.x
| Installation Method | both

#### Summary

When using php.ini `auto_prepend_file` where the file contains an error handler (I assume same issue with exception handlers) like

```
set_error_handler(static fn () => false);
```

test Issue5592Test fails:

```
-There was 1 risky test:
+There were 2 risky tests:
```

#### Current behavior

phpunit test suite fails

#### How to reproduce

`./phpunit` after adding error handler like in summary

#### Expected behavior

No error

---

This matters beyond this specific test, as the number of "risky" tests differs depending on whether or not an error/exception handler is defined in an `auto_prepend_file` or not.
Possibly this also happens when a PHP (PECL) extension defines an error/exception handler.

sebastianbergmann:
This problem has existed for a long time. Improvements made for PHPUnit 13.2 on the `main` branch, including additional test coverage for the error handler, made implementing a fix significantly easier. Therefore this fix will be implemented for PHPUnit 13.2 and not backported to PHPUnit 12.5 and PHPUnit 13.1.

PR #6708, opened 2026-06-05
Implement `deferToPreviousErrorHandler` XML configuration option

### LINKED ISSUE #6705: PHPUnit 13.2 breaks Drupal's approach for ignoring deprecations
| Q                   | A
| PHPUnit version     | 13.2.0
| PHP version         | 8.5

#### Summary

PHPUnit 13.2.0 breaks Drupal deprecation ignore process, that is based on the assumption that an error handler set in the bootstrap PHP file is the one active during the testing.

What we are doing in Drupal is we set in bootstrap the `\Drupal\TestTools\ErrorHandler\BootstrapErrorHandler` which itself falls back to PHPUnit's own one. So we can exclude some deprecations to be ignored (we ignore via a file containing a set of patterns to be ignored) and return to test execution without even letting PHPUnit know that a deprecation was triggered.

13.2 reverses that - it always sets its own error handler as the main one, and only falls back to `\Drupal\TestTools\ErrorHandler\BootstrapErrorHandler` **after** the error (including deprecation ones) have been already collected. So deprecations get through to PHPUnit and are reported.

sebastianbergmann:
Since PHPUnit 13.2, PHPUnit takes full ownership of the error handler while your tests run. It installs its own handler, processes every deprecation, notice, warning, and error itself (classifying it, applying the baseline, deduplicating it, and reporting it) and only then forwards the issue to any error handler that was registered before PHPUnit (for example in a bootstrap file).

This was an intentional consequence of #5845, and it is the correct behavior: deciding whether an issue is ignored or reported belongs to PHPUnit, in one place, so that the baseline, `--display-deprecations`, `failOnDeprecation`, and the issue-trigger classification all agree.

A side effect is that a custom error handler registered before PHPUnit can no longer suppress issues before PHPUnit sees them. Some setups relied on exactly that, for example Drupal's `BootstrapErrorHandler`, which filters deprecations against an ignore list and only forwards the rest (see #6705).

This adds a new opt-in boolean attribute on the root `<phpunit>` element of PHPUnit's XML configuration file:

```xml
<phpunit deferToPreviousErrorHandler="true">
```

When enabled, PHPUnit gives the previously registered error handler the first chance to handle each issue. If that handler reports the issue as handled, PHPUnit does nothing further with it: it is not processed, baselined, or reported. Only issues the previous handler declines are processed by PHPUnit as usual. This restores the pre-13.2 precedence and lets bootstrap-based handlers (such as Drupal's) work as they did before.

The option is off by default; existing behavior is unchanged unless you explicitly turn it on. It can only be set in the XML configuration file (there is no CLI flag).

When the option is enabled, any issue your handler suppresses becomes invisible to PHPUnit: to the baseline, deduplication, trigger classification, `--display-deprecations`, `failOnDeprecation`, and to loggers and extensions. PHPUnit can no longer guarantee that its reporting reflects what actually happened during the run.

Because of this, PHPUnit emits a notice on every run while the option is active, stating clearly that issue handling has been delegated outside of PHPUnit's control and that reporting integrity cannot be guaranteed. The notice is always shown. By default, PHPUnit does not exit with a shell exit code indicating failure just because the test runner triggered a notice.

Treat this as a compatibility escape hatch, not a recommended configuration. Prefer PHPUnit's built-in mechanisms for handling deprecations which keep PHPUnit as the single authority over what is reported. Use `deferToPreviousErrorHandler="true"` only when you specifically need a pre-existing error handler to take precedence.


sebastianbergmann:
Superseded by #6710.

PR #6710, opened 2026-06-06
Deprecation Filters

sebastianbergmann:
Since PHPUnit 13.2, PHPUnit takes ownership of the error-handler stack during a test run (see #5845). As a consequence, an error handler registered before PHPUnit runs, for example in a bootstrap file, can no longer silently drop deprecations before PHPUnit observes them. PHPUnit now observes, classifies, baselines, deduplicates, and reports every deprecation itself.

This is a deliberate and, in my opinion, correct change: deciding whether a deprecation is reported, ignored, or fails the suite is PHPUnit's responsibility, and that decision needs to be made consistently in one place so that the baseline, `--display-deprecations`, `failOnDeprecation`, deduplication, and the issue-trigger classification all agree.

However, this removed a capability that large, long-lived projects relied on. As discussed in #6705, Drupal needs to ignore deprecations using project-level rules, not one by one. Their real-world configuration uses roughly six regular expressions to suppress on the order of 3.5-4 million deprecation messages emitted across their test suite (cross-module return-type deprecations, a small set of test modules they will fix later, a few unavoidable internal/forward-compatibility warnings, and so on). Neither the hash-based baseline nor the per-test `#[IgnoreDeprecations]` attribute is a practical fit for numbers like that.

The existing escape hatch for this, letting a bootstrap error handler suppress issues before PHPUnit sees them, necessarily moves the decision outside of PHPUnit, which is exactly what 13.2 set out to fix, and means PHPUnit can no longer guarantee that its reporting reflects what actually happened.

This pull request introduces deprecation filters: user-supplied objects that decide, for each deprecation, whether PHPUnit should ignore it, with the decision made inside PHPUnit, after it has already observed and classified the deprecation.

You implement the `PHPUnit\Runner\DeprecationFilter\Filter` interface:

```php
use PHPUnit\Event\Code\IssueTrigger\IssueTrigger;
use PHPUnit\Runner\DeprecationFilter;

final class MyDeprecationFilter implements DeprecationFilter
{
    public function ignores(string $message, string $file, int $line, IssueTrigger $trigger): bool
    {
        // Return true to ignore this deprecation, false to let PHPUnit process it.
        return str_contains($message, 'might add "..." as a native return type');
    }
}
```

and register it in your XML configuration, alongside the existing `<source>` options:

```xml
<source>
    <deprecationFilters>
        <deprecationFilter className="App\Tests\MyDeprecationFilter"/>
    </deprecationFilters>
</source>
```

You can register more than one filter; a deprecation is ignored if any registered filter returns `true`. Each filter receives the deprecation message, the file and line, and the resolved issue-trigger classification (self/direct/indirect), so it can match on the message text, on a path, or on how the deprecation was triggered.

Because the filter runs inside PHPUnit, the result is recorded as a first-class outcome of the run rather than hidden from it. A filtered deprecation behaves exactly like one ignored via `#[IgnoreDeprecations]`:

- it does not appear in the test progress or the deprecation summary,
- it is not counted and does not trigger `failOnDeprecation`,
- but it is still observed and classified, so reporting stays internally consistent.

Crucially, this does not require disabling PHPUnit's own error-handler ownership, and it does not emit a "reporting integrity cannot be guaranteed" warning on every run (see #6708), because integrity is in fact preserved: PHPUnit still makes the decision. The filter simply expresses the project's policy about which deprecations are currently acceptable.

For a project like Drupal, this means the bespoke bootstrap deprecation handler can be replaced by a single registered filter that reads their existing ignore file, layering "project-level ignored deprecations" cleanly on top of the baseline and the per-test ignore mechanisms.

### DIFF
--- a/src/Event/Events/Test/Issue/DeprecationTriggered.php
+++ b/src/Event/Events/Test/Issue/DeprecationTriggered.php
@@ -44,6 +44,7 @@ final readonly class DeprecationTriggered implements Event
     private bool $suppressed;
     private bool $ignoredByBaseline;
     private bool $ignoredByTest;
+    private bool $ignoredByFilter;
     private IssueTrigger $trigger;
 
     /**
@@ -59,7 +60,7 @@ final readonly class DeprecationTriggered implements Event
      *
      * @internal This method is not covered by the backward compatibility promise for PHPUnit
      */
-    public function __construct(Telemetry\Info $telemetryInfo, Test $test, string $message, string $file, int $line, bool $suppressed, bool $ignoredByBaseline, bool $ignoredByTest, IssueTrigger $trigger, string $stackTrace)
+    public function __construct(Telemetry\Info $telemetryInfo, Test $test, string $message, string $file, int $line, bool $suppressed, bool $ignoredByBaseline, bool $ignoredByTest, bool $ignoredByFilter, IssueTrigger $trigger, string $stackTrace)
     {
         $this->telemetryInfo     = $telemetryInfo;
         $this->test              = $test;
@@ -69,6 +70,7 @@ final readonly class DeprecationTriggered implements Event
         $this->suppressed        = $suppressed;
         $this->ignoredByBaseline = $ignoredByBaseline;
         $this->ignoredByTest     = $ignoredByTest;
+        $this->ignoredByFilter   = $ignoredByFilter;
         $this->trigger           = $trigger;
         $this->stackTrace        = $stackTrace;
     }
@@ -122,6 +124,11 @@ final readonly class DeprecationTriggered implements Event
         return $this->ignoredByTest;
     }
 
+    public function ignoredByFilter(): bool
+    {
+        return $this->ignoredByFilter;
+    }
+
     public function trigger(): IssueTrigger
     {
         return $this->trigger;
@@ -154,6 +161,10 @@ final readonly class DeprecationTriggered implements Event
 
         if ($this->ignoredByTest) {
             $details[] = 'ignored by test';
+        }
+
+        if ($this->ignoredByFilter) {
+            $details[] = 'ignored by filter';
         }
 
         if ($this->ignoredByBaseline) {
--- a/src/Runner/DeprecationFilter.php
+++ b/src/Runner/DeprecationFilter.php
@@ -7,14 +7,14 @@
  * For the full copyright and license information, please view the LICENSE
  * file that was distributed with this source code.
  */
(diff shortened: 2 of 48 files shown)


src/Runner/DeprecationFilter.php
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * @no-named-arguments Parameter names are not covered by the backward compatibility promise for PHPUnit
+ */
+interface DeprecationFilter
+{
+    /**
+     * Return true to ignore the deprecation, false to let PHPUnit process it.
+     */
+    public function ignores(string $message, string $file, int $line, IssueTrigger $trigger): bool;
```
mondrake, line 22: Wondering if it could be possible/useful to also pass the test context where the deprecation happened, so that an implementing class could do things like ignoring just for some specific test `#[Group]`, test suite, etc

src/Event/Events/Test/Issue/DeprecationTriggered.php
```diff
@@ -44,6 +44,7 @@
     private bool $suppressed;
     private bool $ignoredByBaseline;
     private bool $ignoredByTest;
+    private bool $ignoredByFilter;
```
mondrake, line 47: Fully agree that deprecations ignored by the filter should be accounted for separately from baseline and individual test #[IgnoreDeprecation].

Ideally, in the future we could think of a way to report actually ignored deprecations vs expected ignored deprecations, to identify stale expectations, so allowing to cleanup baseline/test ignores (PHPStan does something like this, telling you that an error that you baselined or ignored is no longer present, and that's an error in itself). 

When and if we get to that, probably the deprecations ignored by filter should be kept out of that to avoid false positives, in case of test runs that execute only a portion of the code base.

sebastianbergmann: I am not convinced such functionality belongs in PHPUnit. It can easily be implemented either in a third-party extension or in a tool that analyses the OTR logfile.

src/Runner/DeprecationFilter.php
sebastianbergmann: While it should be possible, I would rather not do this just because it "might be useful". Unless you can provide a compelling, concrete use case, I would be against adding this.

src/Event/Events/Test/Issue/DeprecationTriggered.php
mondrake: @sebastianbergmann yes we are cooking something here for Drupal: https://www.drupal.org/project/drupal/issues/3523614.

PR #6725, opened 2026-06-08
Allow `#[CoversNothing]` on methods

### LINKED ISSUE #6722: Allow `#[CoversNothing]` on methods
This is a follow up of the comment https://github.com/sebastianbergmann/phpunit/issues/6235#issuecomment-3189085966 because having a discussion on a closed issue is not ideal.

The discussed change was introduced in https://github.com/sebastianbergmann/phpunit/commit/e3d2f1c38bee9796e08fae89a98e79332f6d349b, and AFAIU the rational was:

>For me, the problem here is that #[CoversNothing] can be used on a class and on a method whereas the other #[Covers*] attributes can only be used on a class.
>I am not convinced that there is a valid use case for using #[CoversNothing] on a method and therefore would rather sunset support for using #[CoversNothing] on a method.

Like @alexander-schranz, I've hit several cases and would like to offer a counter-point.

You may have a test `IndexXmlCoverageLocatorTest` for the class `IndexXmlCoverageLocator`, and have the `#[CoversClass]` attribute:

```php
#[CoversClass(IndexXmlCoverageLocator::class)]
class IndexXmlCoverageLocatorTest {
}
```

One advantage of being able to apply `#[CoversNothing]` on a method, is that it allows to easily mark a test as "not interesting" from a coverage perspective and effectively kicking it out of tools like infection, without requiring to extract the method to a second test:

```php
#[CoversClass(IndexXmlCoverageLocator::class)]
class IndexXmlCoverageLocatorTest {
  function setup() { // complicated setup }

  // Important sanity check, validating the setup, but otherwise not important.
  #[CoversNothing]
  function test_it_the_default_path_of_this_test_is_not_the_standard_location() {}

  ... other tests
}
```

The change is a bit old, so I'd understand there is no desire to change it back, but I wanted to offer a concrete use case.

sebastianbergmann:


theofidry:
👌 

Thank you very much

PR #6753, opened 2026-06-17
Persistent worker for native parallel test execution

sebastianbergmann:
This is the next step towards native parallel test execution in PHPUnit, building on the async-capable `JobRunner` that was implemented separately in d667e2d9e14449f29360d2b96da8296bb498e1bf.

Rather than introducing a new subsystem, the changes proposed here generalize the existing process isolation implementation, which already runs a test in a separate process and faithfully reconstitutes its outcome in the parent, from "one synchronous child per test" toward "a reusable worker that runs many tests."

This PR does not wire anything into the CLI. There is no scheduler, worker pool, or `--parallel` option here. Sequential execution remains the only behavior a user can trigger, and process isolation is unchanged.

This work depends on the `JobRunner` refactor that split process spawning (`proc_open()`) from reaping (`proc_close()`) via the `RunningJob` handle, and added `JobRunner::start()` to spawn a long-lived process whose stdin stays open as a control channel. That change is already on `main`; this PR leverages it.

### Design notes / trade-offs

- Tests run by one worker share a process and therefore do not get per-test global-state isolation; this is what allows the bootstrap cost to be amortized
- Code coverage is cleared per command; each envelope ships only that test's coverage, which the parent merges
- `PassedTests` is shipped cumulatively, which is harmless because its import is idempotent
- Concurrency uses `proc_open()` and `stream_select()` only, so there is no new runtime requirement beyond what process isolation already needs

### DIFF
--- a/src/Runner/Parallel/Exception/WorkerException.php
+++ b/src/Runner/Parallel/Exception/WorkerException.php
@@ -1,0 +1,22 @@
+<?php declare(strict_types=1);
+/*
+ * This file is part of PHPUnit.
+ *
+ * (c) Sebastian Bergmann <sebastian@phpunit.de>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+namespace PHPUnit\Runner\Parallel;
+
+use PHPUnit\Exception;
+use RuntimeException;
+
+/**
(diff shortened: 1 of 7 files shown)


staabm:
Looking at it my first idea was that RunInParallelProcess would be a good fit for a new attribute

PR #6777, opened 2026-06-24
Add negated string assertions for ignoring line endings

abolfazl-moeini:
## Summary

The `IgnoringLineEndings` assertion family was missing its negated counterparts. Every other string assertion family already provides both positive and negative variants:

- `assertStringContainsString` / `assertStringNotContainsString`
- `assertStringContainsStringIgnoringCase` / `assertStringNotContainsStringIgnoringCase`
- `assertStringEqualsStringIgnoringWhitespace` / `assertStringNotEqualsStringIgnoringWhitespace`

This adds the two missing methods:

- `assertStringNotContainsStringIgnoringLineEndings()`
- `assertStringNotEqualsStringIgnoringLineEndings()`

## Changes

- `src/Framework/Assert.php` — two new assertion methods
- `src/Framework/Assert/Functions.php` — corresponding global function wrappers
- Two new unit tests mirroring the existing whitespace negation test pattern with `DataProviderExternal`

## Testing

Verified with red/green cycle on the line-endings assertion tests:

- Without the implementation: 6 errors, 11 failures (34 tests)
- With the implementation: 34 tests, 51 assertions — all pass

Full `tests/unit/Framework/Assert/` suite also passes (1676 tests).

### DIFF
--- a/src/Framework/Assert.php
+++ b/src/Framework/Assert.php
@@ -2405,6 +2405,14 @@ abstract class Assert
     }
 
     /**
+     * @throws ExpectationFailedException
+     */
+    final public static function assertStringNotContainsStringIgnoringLineEndings(string $needle, string $haystack, string $message = ''): void
+    {
+        self::assertThat($haystack, new LogicalNot(new StringContains($needle, false, true)), $message);
+    }
+
+    /**
      * Asserts that two strings are equal except for line endings.
      *
      * @throws ExpectationFailedException
@@ -2412,6 +2420,22 @@ abstract class Assert
     final public static function assertStringEqualsStringIgnoringLineEndings(string $expected, string $actual, string $message = ''): void
     {
         self::assertThat($actual, new StringEqualsStringIgnoringLineEndings($expected), $message);
+    }
+
+    /**
+     * Asserts that two strings are not equal except for line endings.
+     *
+     * @throws ExpectationFailedException
+     */
+    final public static function assertStringNotEqualsStringIgnoringLineEndings(string $expected, string $actual, string $message = ''): void
+    {
+        self::assertThat(
+            $actual,
+            new LogicalNot(
+                self::stringEqualsStringIgnoringLineEndings($expected),
+            ),
+            $message,
+        );
     }
 
     /**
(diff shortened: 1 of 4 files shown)


sebastianbergmann:
Thank you for your contribution. I appreciate the time you invested in preparing this pull request. However, I have decided not to merge it because the omission was intentional rather than an oversight.

The premise that "every other string assertion family already provides both positive and negative variants" is not quite accurate. The families are deliberately asymmetric:

| Family | Contains | NotContains | Equals | NotEquals |
|---|:--:|:--:|:--:|:--:|
| IgnoringCase | yes | yes | yes | yes |
| IgnoringWhitespace | no | no | yes | yes |
| IgnoringLineEndings | yes | no | yes | no |

Only `IgnoringCase` is fully populated, because case-insensitivity is a genuinely symmetric relation: "ignoring case" means the same thing whether you assert presence or absence. `IgnoringWhitespace` has no `Contains` variant at all, in either polarity.

`IgnoringLineEndings` and `IgnoringWhitespace` are normalization-for-robustness helpers. Their purpose is to keep a positive assertion from failing over an irrelevant cosmetic difference (CRLF vs LF, indentation). In that direction the normalization does exactly what you want.

In a negated assertion the normalization inverts into something counter-intuitive:

- `assertStringNotEqualsStringIgnoringLineEndings("foo\r\n", "foo\n")` would fail, because after normalization the two strings are equal. Anyone reaching for a "not equals" between two strings that differ only in line endings almost certainly expects it to pass. The helper works against the intent of a negative comparison — it only passes when the strings differ in some way other than line endings, which is a narrow and hard-to-explain guarantee.

- `assertStringNotContainsStringIgnoringLineEndings()` is harder still: a double negative over a normalization ("the haystack does not contain the needle, but treat their line endings as equal"). There's no `Contains`-ignoring-whitespace precedent for it, and the cognitive load far outweighs any realistic use case.

For absence/inequality assertions you almost always want the strictest comparison, not a loosened one, so "not equal but ignore these differences" is close to a contradiction in intent. That's why these two methods were left out, and I'd prefer to keep it that way.

PR #6784, opened 2026-06-26
Parallel test execution

sebastianbergmann:
PHPUnit can now execute a test suite across several worker processes concurrently instead of one test after another. Parallel execution is opt-in via a new `--parallel=<n>` command-line option and changes nothing when it is not used: the sequential `TextUI\TestRunner` remains the default and `TextUI\ParallelTestRunner` is selected only when `<n> > 1`.

Native parallelism here is a generalization of process isolation rather than a new subsystem: a worker reconstructs and runs a unit of work and ships its outcome home in the very same serialized envelope that process isolation already uses, and the parent replays that envelope through its normal event pipeline. As a result the parent process remains the single source of truth for all output, logging, results, and code coverage, which are produced exactly as in sequential mode.

## How it works

- **Distribution unit = one test class.** All selected tests of a class are run together by a single worker, preserving `#[BeforeClass]`/`#[AfterClass]` and intra-class ordering. A `DataProviderTestSuite`, and the `IterativeTestSuite` that carries the repetitions of a repeated test or the attempts of a retried test, travel to the worker as atomic members of their class' unit, so their suite envelopes nest in logger output exactly as they do sequentially.
- **Chunks.** The top-level `<testsuite>` elements of an XML configuration are run one after another, just as in sequential mode: only tests that belong to the same top-level test suite ever run concurrently.
- **Worker pool.** `Runner\Parallel\WorkerPool` owns `n` persistent workers and pulls from a dynamic work-stealing queue, so the load self-balances against stragglers. A worker signals completion through the filesystem, which the parent polls; `stream_select()` is not used, because it does not work on the workers' output pipes on Windows.
- **Longest-first dispatch.** Within a chunk, `Runner\Parallel\Scheduler` dispatches units in the order of the durations recorded by the test run history, longest first, so that the longest-running work starts as early as possible instead of becoming the straggler the pool waits for. A unit with no recorded duration goes first. Only the dispatch order is affected; results are released in suite order either way.
- **Live progress.** While a unit is still running, its worker streams the events of every test that finishes, as length-prefixed frames appended to a stream file that the parent reads incrementally. Progress output therefore appears per finished test rather than stalling until a whole test class is done.
- **Ordered result aggregation.** Workers finish out of order, but `Runner\Parallel\ResultAggregator` buffers each unit and releases it only once every preceding unit (in suite order) has been released, so the event stream, and therefore every output format, including the default progress output, is byte-for-byte what sequential mode produces. Streamed events of the unit that is next in suite order are forwarded immediately; those of later units are buffered until their turn.
- **PHPT tests run concurrently in the main process.** They are not `PHPUnit\Framework\TestCase` instances and cannot be reconstructed in a worker (and nesting their child processes inside workers hung on Windows), so `Runner\Parallel\PhptRunner` runs them side by side in the main process, each as its own child process.
- **One budget of concurrent processes.** The worker pool and the PHPT runner are advanced side by side in a single polling loop and share one `ProcessBudget`, so `--parallel <n>` never executes more than `<n>` units at once, no matter how a chunk is composed.
- **Suite envelopes.** In a parallel run nothing produces the "test suite started"/"test suite finished" events of the root suite and the top-level test suites, so the runner emits them around the chunks. The loggers that reconstruct the suite hierarchy — JUnit XML, Open Test Reporting, TeamCity — depend on them.

## Running in the main process

Some tests cannot, or must not, run in a worker. These run in the main process at their correct suite position — the aggregator invokes them while releasing results, so global ordering is preserved — and their execution there is ordinary sequential execution:

- Tests attributed with **`#[DoNotRunInParallel]`** — a new attribute, valid on classes and methods, for tests that must not run alongside others (for instance because they share a machine-global resource). Such a unit runs *alone*: the worker pool and the PHPT runner first finish what they are executing and start nothing new until it is done.
- Tests that require **process isolation** (`#[RunInSeparateProcess]`, `#[RunTestsInSeparateProcesses]`, or global `--process-isolation`) — a shared worker cannot provide isolation, but the main process spawns the isolated child as usual.
- Tests with a **cross-class `#[Depends]`** — the result they depend on is produced by a different unit and is only visible in the main process, once that unit has been released.
- Tests whose **data cannot be serialized** for transport to a worker — closures and other non-serializable values (which make `serialize()` throw) as well as resources (which `serialize()` silently degrades to `0`).
- **Repeated or retried PHPT tests**, whose iterations are orchestrated by their suite and must run one after another.
- Any other test that is neither a `TestCase` nor a PHPT test.

Because a work unit is a whole test class, a single test method carrying `#[DoNotRunInParallel]`, requiring isolation, depending on another class, or providing non-serializable data takes its entire class out of the parallel phase.

PHPT tests declare their concurrency constraints with a **`--CONFLICTS--`** section instead, since a PHPT file cannot carry PHP attributes: while a test holding conflict key `K` runs, no other test declaring `K` is started, and the reserved key `all` means the test runs entirely on its own.

## Robustness

- **Crashed workers.** If a worker dies, its unit is retried once on a freshly booted worker process — the crash may have been caused by state the dead process accumulated, so the retry gets a pristine environment. The retry is vetoed when some of the unit's results were already streamed and forwarded; in that case, and when the retry crashes as well, every test of the unit is reported as errored (`The worker process running X ended unexpectedly`), the remaining units are redistributed across the surviving workers, and the run accounts for all of them.
- **Corrupted event streams.** A worker whose event stream fails verification is terminated and reaped, so its unit's retry boots a fresh process instead of aborting the run or leaking the compromised one.
- **`--stop-on-*`.** As soon as the collected results call for a stop, the aggregator releases nothing further and the units still executing are terminated (a grace period, then a forced kill). A PHPT test whose `FILE` section was terminated still runs its `CLEAN` section, and one terminated during `SKIPIF` does not start its `FILE` section.
- **Stray worker output.** A worker's output is redirected to a file, so a test that writes more than the operating system buffers in a pipe cannot deadlock the run.

## New public API

- `--parallel=<n>` CLI option (also listed in `--help`); a value that is not a positive integer is ignored with a test runner warning.
- `#[PHPUnit\Framework\Attributes\DoNotRunInParallel]` (`TARGET_CLASS | TARGET_METHOD`), with full metadata-layer support.
- The `--CONFLICTS--` PHPT section, as supported by the PHP project's `run-tests.php`: one conflict key per line, `#` starts a comment, blank lines are ignored, and `all` is reserved.
- Two worker-identity environment variables, exposed to the tests a worker runs so that fixtures can partition shared resources (a database, a port, a temporary directory, ...) per worker:
  - `PHPUNIT_WORKER_ID` — the small, stable ordinal (`0`, `1`, `2`, ...), ideal for indexing a fixed set of pre-provisioned resources. A worker restarted after a crash keeps its ordinal.
  - `PHPUNIT_WORKER_TOKEN` — a value of the form `<id>_<random>` that is unique across workers and across runs, for resources that must not collide with those left behind by a previous run. A restarted worker gets a fresh token.
- `Event\TestRunner\ChildProcessReason::ParallelWorker`, so that the events about child processes say why one was used.

## Shared with the sequential runner

Rather than duplicating the sequential runner, behavior the two must perform identically was extracted so it is expressed exactly once:

- `TextUI\TestRunnerLifecycle` — the test runner lifecycle both runners drive.
- `Event\Dispatcher\CollectionWindow` and the facade's dispatcher selection — the window during which events are collected instead of dispatched.
- `Framework\TestRunner\ChildProcessBootstrap` plus two templates — the boot code every kind of child process shares; the configuration and source map are written once per run for all of them, so isolated processes and parallel workers cannot drift apart.
- `Framework\TestRunner\ChildProcessResultEnvelope` — the result envelope's encoding and decoding, shared by its consumers.
- Test run history is now persisted once, on `TestRunner\ExecutionFinished`, instead of every time an outermost test suite finishes.

## Results for PHPUnit's own test suite

Measured on a 6-core machine:

| Suite | Sequential | `--parallel 10` | |
|---|---|---|---|
| `--testsuite unit` (5813 tests) | 11.0s | 6.7s | 1.6× |
| `--testsuite end-to-end` (1257 tests) | 249.5s | 48.5s | 5.1× |

Both suites report the same numbers of tests, failures, and skipped tests in both modes, stable across runs. The end-to-end suite benefits the most because it is dominated by child processes rather than by CPU work in the main process.

Which worker runs which test class is not deterministic; the work-stealing scheduler assigns classes by timing, but the aggregated output is identical regardless.

## Notes and limitations (possible follow-ups)

- Parallel execution is configurable only via the CLI option; there is no XML configuration setting yet.
- An in-process unit that is not exclusive may temporally overlap workers still running later units; this is invisible in the output (worker events stay buffered) and harmless for process-global state (separate processes), but a test contending on an external shared resource should use `#[DoNotRunInParallel]`.
- Chunk boundaries are a synchronization point: the units of the next top-level test suite are not dispatched before the current chunk has drained.


Slamdunk:
My 2 cents on the topic:

1. There is no logical possibility that Extensions support parallelization natively: I suggest to require extensions to explicitely provide support to it (in some way) and to stop the run if an extension without parallel support is detected
2. I've been requested to provide both simple worker ID like 1, 2, 3 as well as unique worker ID like `uniqid('1_')`,  `uniqid('2_')`,  `uniqid('3_')`
3. Coordinated bootstraping operations are crucial to the users, and many asked to be able to have `bootstrap` running both before the whole execution and within each worker, with a flag to distinguish between the two cases (see https://github.com/paratestphp/paratest#initial-setup-for-all-tests). For example users want to `CREATE DATABASE` in the parent process and `DELETE FROM table` in the workers to speed I/O up
4. I see that you chose `Distribution unit = one test class` which is sensible: be prepared to reply to a mass of users that wants/needs `Distribution unit = one test method`, what is ParaTest we provide with `--functional`

5.
> * `--testsuite unit --parallel 10` matches sequential exactly (same test and assertion counts), stable across runs.

Parallel execution can't and will never be deterministic, a tiny difference between how much a unit lasts results in the following test being run in worker X instead of Y.
Many test suites are not well isolated, and if the developer is unaware of that a test run may fail if the aforementioned test runs in worker X instead of Y.
To be able to debug such situations, the developers needs an option to replay the previous run with exactly the same test order in the exact same worker ID, with no concern on the test speed since you (correctly) chose to `pulling from a dynamic work-stealing queue so the load self-balances against stragglers`

sebastianbergmann:
Thank you, @Slamdunk, for taking the time to write all of this up. There's a lot of hard-won experience in here, and I appreciate it.

I want to set expectations honestly, though: for me, native parallel test execution in PHPUnit is still just an idea, and this pull request is a (remarkably well-working) proof of concept rather than a roadmap commitment.

If I ever decide to be serious about this and actually ship it, the feature will have limitations. And those limitations must (and will) be very well documented. Complaints about them will then be kindly refused. 🙂

Even in that case, I neither intend nor expect this to replace dedicated solutions such as ParaTest or Paraunit. The scope I have in mind right now is roughly:

PHPUnit supports parallel execution for well-architected test suites that properly deal with resource-usage conflicts like databases, etc.

So several of the needs you describe, coordinated bootstrapping, functional/method-level distribution, deterministic replay of a run, are exactly the kind of thing that lives outside that scope and is better served by the tools built specifically for it.

That said, your notes are genuinely useful for thinking about where the boundaries of that scope should sit, so thank you again.

sebastianbergmann:
I have briefly looked into whether the new I/O polling API that PHP 8.6 introduces ([RFC: poll_api](https://wiki.php.net/rfc/poll_api)) could replace the file-based completion polling used by the parallel test runner. The conclusion is that it cannot, because it has the exact limitation that forced the file-based design in the first place: **on Windows, it cannot wait on `proc_open()` pipes.**

`WorkerPool` and `PhptRunner` do not wait on the worker processes' output pipes with `stream_select()`, because `stream_select()` does not work on pipes on Windows (it only works on sockets there). Instead, a worker signals completion through the filesystem: after writing its result file, it creates a sibling `.done` file, whose appearance the parent detects by polling `is_file()`, sleeping 1000 microseconds between rounds.

`Io\Poll` looks like the natural replacement at first glance: a `Context` monitors multiple `StreamPollHandle`s and `wait()` blocks until one of them becomes readable, using epoll/kqueue/event ports/poll as the platform backend.

However:

* On Windows, the only backend that is registered is WSAPoll (`main/poll/poll_core.c`, the `#ifdef PHP_WIN32` branch). `WSAPoll()` is a Winsock function and accepts sockets only.
* `StreamPollHandle` obtains its descriptor via `php_stream_cast(PHP_STREAM_AS_FD_FOR_SELECT)` (`ext/standard/io_poll.c`). For a `proc_open()` pipe on Windows, this yields a CRT file descriptor, not a `SOCKET`, which WSAPoll immediately rejects with `POLLNVAL`. Readiness on a pipe can therefore never be waited on.
* The RFC's future scope lists `SocketPollHandle`, `CurlPollHandle`, `TimerHandle`, and `SignalHandle`, but no process or pipe handle.

In other words, `Io\Poll` is a unified, modern interface to the same platform primitives whose Windows limitation this branch already had to work around.

Therefore, the uniform file-based completion polling stays. This decision should be revisited only if a future PHP version ships a poll handle that can wait on pipes or process handles on Windows, which is what this feature actually needs.

Slamdunk:
I see that it's up to the user to select how many worker to use.

The `auto` flag has been a big request by the community, and currently https://github.com/theofidry/cpu-core-counter is serving ParaTest as well as PHP-CS-Fixer, PHPStan and Infection for that very purpose.

You might be interested in giving it a try, so PHPUnit can provide `--parallel=auto` out of the box.

sebastianbergmann:
> I see that it's up to the user to select how many worker to use.
> 
> The `auto` flag has been a big request by the community, and currently https://github.com/theofidry/cpu-core-counter is serving ParaTest as well as PHP-CS-Fixer, PHPStan and Infection for that very purpose.
> 
> You might be interested in giving it a try, so PHPUnit can provide `--parallel=auto` out of the box.

I am aware of `fidry/cpu-core-counter`, but thank you for your suggestion. Should I decide to implement `--parallel=auto` I will look into using that library.

sebastianbergmann:
> > I see that it's up to the user to select how many worker to use.
> > The `auto` flag has been a big request by the community, and currently https://github.com/theofidry/cpu-core-counter is serving ParaTest as well as PHP-CS-Fixer, PHPStan and Infection for that very purpose.
> > You might be interested in giving it a try, so PHPUnit can provide `--parallel=auto` out of the box.
> 
> I am aware of `fidry/cpu-core-counter`, but thank you for your suggestion. Should I decide to implement `--parallel=auto` I will look into using that library.

`--parallel auto` is now implemented.

sebastianbergmann:
# The plateau in worker count is expected behavior

Feedback from a user testing this branch on a real-world suite:

| Workers | 1 | 2 | 4 | 6 | 8 |
|---|---|---|---|---|---|
| Wall time | ~1:35 | ~1:22 | ~0:57 | ~0:58 | ~1:01 |

So: a good speedup up to 4 workers, a plateau at 6, and a slight regression at
8. The same shape shows up in our own unit test suite (`tests/unit`, 6117
tests in 582 test classes, on a machine with 8 logical cores, warm result
cache, without code coverage):

| Workers | 1 | 2 | 4 | 6 | 8 | 12 |
|---|---|---|---|---|---|---|
| Wall time | 9.26 s | 7.19 s | 5.79 s | 5.19 s | 5.20 s | 5.35 s |

The maximum speedup is 1.8x, and it is reached at 6 workers. This is a
suite in which practically nothing runs serially (6 of 582 classes carry
`#[DoNotRunInParallel]`, none require process isolation, one uses
`#[Depends]`).

This is not a defect to fix, it is rather what the design implies. Recording it here
so that it does not get investigated as a bug later, and as a reminder that
the documentation has to say something about it. Users might otherwise expect
`--parallel N` to be N times faster and read the plateau as a problem.

What limits the speedup, in the order in which it tends to matter:

1. **Per-unit overhead.** The distribution unit is one test class: it is
   serialized for transport to a worker, reconstructed there, and its events
   travel back. A suite of many small classes pays this per class.

2. **The main process is single-threaded and does all of the accounting.**
   Every event of every worker is decoded and replayed in the parent, passed
   tests are imported, and code coverage data is merged there.

3. **Units that cannot run in a worker run one at a time in the main
   process**, at their position in suite order: process isolation,
   `#[DoNotRunInParallel]` (which additionally drains the whole pool, because
   such a unit must run alone), cross-class `#[Depends]`, and test data that
   cannot be serialized.

4. **Each top-level `<testsuite>` is a barrier**: the suites of an XML
   configuration run one after another, so every one of them ends with its own
   straggler.

5. **Worker boot cost** is paid once per worker (bootstrap script included),
   and more workers than the machine has usable cores just adds context
   switching. This is where the regression at 8 and 12 workers comes from.

### Documentation

The manual's section on parallel test execution should state that

- `--parallel N` does not give an N-fold speedup, and that the useful value is
  usually well below the number of cores; finding it is a matter of measuring
  the concrete suite,
- code coverage reduces the achievable speedup, because the data is merged in
  the main process,
- which tests do not run in a worker (the four cases above) and that a suite
  dominated by them will barely profit,
- that `PHPUNIT_WORKER_ID` / `PHPUNIT_WORKER_TOKEN` exist for partitioning
  external resources such as test databases per worker, and that a shared
  resource that is not partitioned is the most likely reason for a suite that
  does not scale at all.

Slamdunk:
I've always had diminishing returns in ParaTest too:

image

sebastianbergmann:
`--CONFLICTS-- all` delays the reporting of everything behind it

Separate from the scaling question, and like it not a defect to fix.

A PHPT test that declares `--CONFLICTS-- all` must run entirely on its own, so the PHPT runner defers it until the worker pool and the other PHPT tests have drained. In practice: to the end of its chunk. The results, however, are released in suite order. A test like this that sits early in suite order therefore stalls the release sequence at its index for most of the run: everything after it has long finished, is buffered, and is reported in oneburst once the test finally runs at the end.

Our own end-to-end suite is a good example, and probably a worst case: 38 of its 1259 PHPT tests declare `--CONFLICTS-- all`, and the first one sits at suite position 266 (`cli/validate-configuration/`). Measured with `--parallel 4` and a warm test run history, as the share of the run's wall time by which half of the progress output had been written:

| | 25% of the output by | 50% by | 75% by |
|---|---|---|---|
| all 1259 tests | 67% of the run | 97% | 99% |
| the same minus the 38 `all` tests | 59% | 89% | 96% |

So these 38 tests alone account for roughly 8 percentage points of the back-loading.

**Decision: leave the scheduling as it is.** Draining the pool and running such a test at the moment the release sequence reaches its index would trade concurrency for output continuity, and for a declaration whose whole point is exclusivity that is the worse deal. Reserving more dispatch slots for the suite order does not help here either: such a test cannot run early no matter which slot would like to start it.

Two follow-ups instead:

1. The manual's section on `--CONFLICTS--` should say that a test which
   conflicts with `all` runs at the end of its test suite and that the results
   of the tests after it are reported only once it has run. It is a reason to
   prefer a specific conflict key over `all`.
2. For our own suite: check whether all 38 tests really need `all`. The
   `execution-order/` and `migration/` tests look like they conflict over a
   shared cache directory or configuration file, not over every other test; a
   specific key would let them run concurrently with everything else and would
   remove the reporting delay they cause.

sebastianbergmann:
Merging this before #6919 is preferable because it restructures how a test run is executed (`TestRunnerLifecycle`, `ParallelTestRunner`, `ChildProcessResultEnvelope::mergeCodeCoverage()`). The test impact analysis branch has to adapt to that structure anyway, and that adaptation is best done when merging `feature/test-impact-analysis`.

theofidry:
@sebastianbergmann is there plans to make it opt-out instead of opt-in? Asking to know if it's worth guarding against early as when orchestrating PHPUnit with infection, for mutation processes, we do not want parallelism (as the mutation processes are already parallelised)

sebastianbergmann:
> is there plans to make it opt-out instead of opt-in?

No; I believe this should remain opt-in.

bendavies:
Hi @sebastianbergmann 
i tried this branch on our real product testsuite.

It is a a Symfony 7.4 application running PostgreSQL, 1073 functional test classes with ~13k tests, using the DAMA doctrine-test-bundle (https://github.com/dmaicher/doctrine-test-bundle).
CI splits the tests across three jobs.
Today we use GNU parallel to run one PHPUnit process per test file, 5 at a time, each on its own database. 
For the comparison we kept our own three way split of the file list (see point 5) and ran each part with `phpunit --test-files-file <list> --parallel 5`. 
The integration was otherwise easy. `PHPUNIT_WORKER_ID` can be used instead of Paratests `TEST_TOKEN`

Feedback:

1. Extensions do not run in a worker 
Their subscribers only see the events replayed in the parent, so any extension that has to act inside the test process does nothing in a worker. 
This was a problem for DAMA. it begins a transaction on `Test\PreparationStarted` and rolls it back afterwards. this doesn't work inside a worker.
workaround: the base test class begins the transaction in a `#[Before]` hook and rolls it back in an `#[After]` hook when `PHPUNIT_WORKER_ID` is set, and the bootstrap file does the one time setup the extension would have done from its `TestRunner\Started` subscriber. 
Process isolation has the same limitation, but `--parallel` applies it to the whole suite at once, silently. 
Suggest either bootstraping the configured extensions in the worker as well, with a way for an extension to tell that it runs in a worker.

2. The fallback to the main process is silent. 
239 of our 1073 classes ran in the main process: 234 because a data provider yields a closure, 3 because of an `UploadedFile`, 1 because of an anonymous class, 1 because it requires process isolation.
Those 239 classes hold 6345 tests, half of the entire suite. 
Found by patching `ParallelTestRunner` to print its plan. 
A summary line, for example "239 of 1073 test classes ran in the main process: 238 not serializable, 1 process isolation", plus a per class runner warning with the reason and the offending type, would alert us to fix it.

3. Provider data with a custom `__serialize()` arrives incomplete. Our Symfony security user entity keeps only id and email when serialized, as the security component recommends for the session. 
classes that pass such a user through a data provide failed in the worker with "Typed property ... must not be accessed before initialization". 

4. A unit running in the main process stalls the pool. `ResultAggregator::release()` invokes the in process runner synchronously from `add()`, and `runChunk()` does not call `tick()` until it returns. 
Workers that finish meanwhile wait, and nothing new is dispatched. 
Checked this with a test suite: 2 workers, eight classes that sleep 1s, and one class at suite position 3 that runs in the main process (its provider yields a closure) and sleeps 10s. 
No worker test finished during those 10 seconds; both workers idled from t+1s to t+11s, and the run took 15.9s instead of about 11s. With half of our real tests in the main process this dominated the result, and made --parallel far slower than our current solution.

it looked like this:
```
t+ 0.0s  parallel1   worker 0
t+ 0.0s  parallel2   worker 1
t+ 0.0s  inprocess   START, main process
t+10.0s  inprocess   END
t+11.1s  parallel3   worker 0
t+11.1s  parallel4   worker 1
t+12.1s  parallel5   worker 0
t+12.1s  parallel6   worker 1
t+13.1s  parallel7   worker 0
t+13.1s  parallel8   worker 1
```

5. There is no sharding across machines. `--parallel` spreads one run over the cores of one machine; splitting a suite over several CI runners is left to the user. 
Today, we list the test files, split the list in three and pass each part with `--test-files-file`. 
A `--shard <n>/<m>` option that partitions the test classes deterministically, ideally by recorded duration, would remove our script.

6. Smaller notes.
- Memory: the parent peaked at 0.7 to 1.4 GB per chunk of about 4300 tests, and the largest process in the tree at 1.97 GB. Our per file processes peak at 72 MB (p90) and 236 MB at most.
- `JobRunner::startProcess()` exports every scalar `$_SERVER` entry to the worker, including values Symfony's Dotenv loaded from `.env` in the parent. Inside a worker, `getenv()` on a variable interpolated from the worker id returned the parent's value while `$_SERVER` held the worker's. This did not break anything for us, because our configuration reads `$_SERVER` first, but code that calls `getenv()` directly would see the parent's values and share a resource meant to be split per worker.

sebastianbergmann:
Thank you @bendavies for trying the branch on a real test suite and for the detailed report. Five of your six points are addressed on the branch now. The details are below, in your numbering.

I would be grateful if you could run the branch against your suite once more, ideally with DAMA adapted to `ParallelWorkerExtension`, and post the numbers again, in particular for point 4.

#### 1. Extensions do not run in a worker

An extension can now opt into being bootstrapped in the worker processes as well. It implements the new `PHPUnit\Runner\Extension\ParallelWorkerExtension` interface, which extends `Extension` with

```php
public function bootstrapWorker(Configuration $configuration, WorkerFacade $facade, ParameterCollection $parameters): void;
```

`bootstrapWorker()` is called once in every worker process, before the worker runs its first test. The subscribers registered through `WorkerFacade` receive the events of the tests the worker runs, live, inside the worker. The main process still bootstraps the extension through `bootstrap()` and still replays every event to the subscribers registered there, so an extension that acts in both places decides in `bootstrap()` what its main-process subscribers do when `$configuration->numberOfParallelWorkers()` is greater than one. No test runner lifecycle event is emitted inside a worker; what an extension would do on `TestRunner\Started`, it does in `bootstrapWorker()`. An extension that does not implement the interface behaves exactly as before.

For DAMA this means: implement `ParallelWorkerExtension`, register the subscribers that begin and roll back the transaction in `bootstrapWorker()`, do the one-time setup there, and skip registering them in `bootstrap()` when the run is parallel. Your `#[Before]`/`#[After]` workaround and the bootstrap-file setup should then no longer be needed.

Process isolation (`#[RunInSeparateProcess]`, `--process-isolation`) is deliberately not covered by this yet.

#### 2. The fallback to the main process is silent

Every test class that runs in the main process for a reason its author did not ask for is now reported with a PHPUnit test runner notice that names the class and the reason, for example:

```
The tests of class Foo\BarTest are run in the main process instead of a parallel worker because test testBaz depends on Foo\OtherTest::testQux, a test of another class
```

By default the notices are only counted in the result line; `--display-phpunit-notices` shows them, and `--fail-on-phpunit-notice` turns them into a failure. A configured `--process-isolation` and `#[DoNotRunInParallel]` are deliberate choices and are not reported. I did not add a separate summary line: the notice count is that summary, and the notices carry the information the summary could not.

#### 3. Provider data with a custom `__serialize()`

This one, together with the closures, `UploadedFile`, and anonymous classes from point 2, had one root cause: the data a data provider provides travelled from the main process to the worker by value, through `serialize()`. It no longer does. The worker invokes the data provider again and selects the data set by the name the main process selected. Closures, resources, anonymous classes, and objects with a lossy `__serialize()` now reach the worker intact, because they are created there, and "data cannot be serialized" is no longer a reason for a class to run in the main process. In your suite, this should move 238 of the 239 classes into the workers.

What this asks of a data provider is that it provides the same data set names in every process. A provider keyed by something that differs between invocations, such as a random value or the time, does not provide the selected name in the worker; the tests of that class are then reported as errored with a message that says so. A provider is invoked once per test method and unit in the worker, on top of the invocation the main process makes when it builds the suite.

#### 4. A unit running in the main process stalls the pool

Your analysis is correct: the in-process runner is invoked synchronously from the release sequence, and the pool is not ticked until it returns. I have not changed that. With point 3 in place, the only classes of your suite that still run in the main process are the one requiring process isolation and any with cross-class dependencies, so I expect the measurement to look very different on a second run. If it still dominates, the fix is to let workers hold a backlog of commands so they stay busy while the main process is occupied; I would rather do that on evidence from a real suite than pre-emptively.

#### 5. Sharding across machines

This is out of scope.

#### 6. Smaller notes

Memory: a worker can now be replaced by a fresh process after it has run a configured number of test classes: `numberOfTestClassesBeforeWorkerRecycling="…"` in the XML configuration, or `--recycle-workers-after <n>` on the command line. It is off by default, because a recycle repeats whatever per-worker setup the suite performs at boot, and PHPUnit cannot estimate that cost. Given the growth you observed, roughly 9 MB per class, a value between 25 and 100 should keep the workers well below what you saw. The main process holding 0.7 to 1.4 GB per chunk is mostly the buffered results of units that finished out of suite order; with far fewer classes stalling in the main process, that backlog should shrink as well.

Environment: the child environment is now built from `getenv()` instead of `$_SERVER`, so values that Dotenv wrote into `$_SERVER` in the main process no longer reach a worker through `getenv()`.

Slamdunk:
> #### 5. Sharding across machines
> This is out of scope.

Parallel execution across threads/cores is physically bounded to the CPU available in the market and afforable ones rarely exceed the dozen, while buying fifty quad-core computers is extremely cheap.

Sharding across machines is a key requirement for large setups and a long requested feature in ParaTest: I hope you can add it in the backlog

bendavies:
Hi Derek, Thanks for that. 

I tried it again.

Total wall time for our tests went from 1772s on 989c149 to 1284s on 62f92b1, 28% less.
Nice work.

**ParallelWorkerExtension** 

Dama, Sequentially it subscribes to three events: 
1. `TestRunner\Started` turns its static connections on, 
2. `Test\PreparationStarted` rolls back the previous test's transaction and begins one for the test about to run, and 
3. `TestRunner\Finished` rolls back the last transaction and turns static connections off. 

In a worker the `PreparationStarted` subscriber works unchanged, so every test still starts on a fresh transaction, and `bootstrapWorker()` takes over what `TestRunner\Started` did. 

Two issues remain:
1. There is no counterpart to `TestRunner\Finished`. So, the last transaction in each worker is never rolled back. it stays open until the worker exits and the database discards it. This is harmless for DAMA, but an extension that must flush or close something at the end of a worker has nowhere to do it.
2. Double events. A subscriber registered in `bootstrap()` also receives every event forwarded from the workers, with nothing to denote where it came from. This is ok for DAMA because an extra rollback on an idle connection is harmless. It's not ok for an extension that must act once per test. A flag on forwarded events would fix it.

<details>
<summary>Reproducer for 2: a counting extension, files and output verbatim</summary>

`phpunit.xml`

```xml
<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../vendor/phpunit/phpunit/phpunit.xsd" bootstrap="bootstrap.php" cacheDirectory=".phpunit.cache">
  <testsuites>
    <testsuite name="probe">
      <directory>tests</directory>
    </testsuite>
  </testsuites>
  <extensions>
    <bootstrap class="Probe\TestCounterExtension"/>
  </extensions>
</phpunit>
```

`bootstrap.php`

```php
<?php

declare(strict_types=1);

require __DIR__ . '/../vendor/autoload.php';
require __DIR__ . '/TestCounterExtension.php';
```

`TestCounterExtension.php`

```php
<?php

declare(strict_types=1);

namespace Probe;

use PHPUnit\Event\Test\Finished;
use PHPUnit\Event\Test\FinishedSubscriber;
use PHPUnit\Runner\Extension\Facade;
use PHPUnit\Runner\Extension\ParallelWorkerExtension;
use PHPUnit\Runner\Extension\ParameterCollection;
use PHPUnit\Runner\Extension\WorkerFacade;
use PHPUnit\TextUI\Configuration\Configuration;

final class TestCounterExtension implements ParallelWorkerExtension
{
    public function bootstrap(Configuration $configuration, Facade $facade, ParameterCollection $parameters): void
    {
        $facade->registerSubscriber(new CountingSubscriber());
    }

    public function bootstrapWorker(Configuration $configuration, WorkerFacade $facade, ParameterCollection $parameters): void
    {
        $facade->registerSubscriber(new CountingSubscriber());
    }
}

final class CountingSubscriber implements FinishedSubscriber
{
    public function notify(Finished $event): void
    {
        file_put_contents(sys_get_temp_dir() . '/count.txt', getmypid() . ' ' . $event->test()->name() . "\n", FILE_APPEND);
    }
}
```

`tests/WorkerOneTest.php`

```php
<?php

declare(strict_types=1);

namespace Probe\Tests;

final class WorkerOneTest extends TestCase
{
    {
        $this->assertTrue(true);
    }
}
```

`tests/WorkerTwoTest.php`

```php
<?php

declare(strict_types=1);

namespace Probe\Tests;

final class WorkerTwoTest extends TestCase
{
    public function testTwo(): void
    {
        $this->assertTrue(true);
    }
}
```

`tests/MainTest.php`

```php
<?php

declare(strict_types=1);

namespace Probe\Tests;

use PHPUnit\Framework\Attributes\DoNotRunInParallel;

#[DoNotRunInParallel]
final class MainTest extends TestCase
{
    public function testMain(): void
    {
        $this->assertTrue(true);
    }
}
```

`vendor/bin/phpunit -c probe/phpunit.xml` writes 3 lines to `count.txt`:

```
36966 testMain
36966 testOne
36966 testTwo
```

`vendor/bin/phpunit -c probe/phpunit.xml --parallel 2` writes 5 lines. 36976 is the main process, 36985 and 36986 are the workers. `testOne` and `testTwo` are logged in their worker and again in the main process:

```
36976 testMain
36985 testOne
36986 testTwo
36976 testOne
36976 testTwo
```

A subscriber that has to do something exactly once per test, would do it twice for every worker test, and the second time would fail or double count. Nothing on the event lets it skip the copy.
</details>

<details>
<summary>Our `DamaParallelWorkerExtension`, for reference</summary>

```php
final class DamaParallelWorkerExtension implements ParallelWorkerExtension
{
    public function bootstrap(Configuration $configuration, Facade $facade, ParameterCollection $parameters): void
    {
        // also receives every forwarded worker event, see 2.
        new PHPUnitExtension()->bootstrap($configuration, $facade, $parameters);
    }

    public function bootstrapWorker(Configuration $configuration, WorkerFacade $facade, ParameterCollection $parameters): void
    {
        StaticDriver::setKeepStaticConnections(true); // what TestRunner\Started did

        $facade->registerSubscriber(new class implements PreparationStartedSubscriber {
            public function notify(PreparationStarted $event): void
            {
                PHPUnitExtension::rollBack();

                if (PHPUnitExtension::hasSkipAttribute($event->test())) {
                    PHPUnitExtension::skipTransaction();

                    return;
                }

                PHPUnitExtension::unskipTransaction();
                PHPUnitExtension::beginTransaction();
            }
        });

        // nothing can do what TestRunner\Finished did, see 1.
    }
}
```
</details>

**Notices** 
A class whose tests carry `#[RunInSeparateProcess]` is reported, while `--process-isolation` and `#[DoNotRunInParallel]` are not. 

We want to keep `failOnPhpunitNotice="true"`, so a class we deliberately isolate with `#[RunInSeparateProcess]` should not produce a notice and fail the build in parallel mode. right now it does.

**sharding** 
native sharding would be great, and the quality of life features would follow from it, above all balancing the shards by the test durations recorded in historic runs. 
But it is easily worked around today, so it is not blocking me at least. This is all we do, on every runner:

```bash
# same list, same order on every runner: sort, then shuffle with a fixed seed
find tests -name '*Test.php' | LC_ALL=C sort \
  | shuf --random-source=<(openssl enc -aes-256-ctr -pass pass:seed -nosalt </dev/zero 2>/dev/null) \
  > all_tests

# runner N of M takes the Nth of M equal slices
split --number="l/${SHARD_NUMBER}/${SHARD_COUNT}" all_tests > my_tests

vendor/bin/phpunit --test-files-file my_tests --parallel 5
```

bendavies:
Hi Sebastian, Thanks for that. 

I tried it again.

Total wall time for our tests went from 1772s on 989c149 to 1284s on 62f92b1, 28% less.
Nice work.

**ParallelWorkerExtension** 

Dama, Sequentially it subscribes to three events: 
1. `TestRunner\Started` turns its static connections on, 
2. `Test\PreparationStarted` rolls back the previous test's transaction and begins one for the test about to run, and 
3. `TestRunner\Finished` rolls back the last transaction and turns static connections off. 

In a worker the `PreparationStarted` subscriber works unchanged, so every test still starts on a fresh transaction, and `bootstrapWorker()` takes over what `TestRunner\Started` did. 

Two issues remain:
1. There is no counterpart to `TestRunner\Finished`. So, the last transaction in each worker is never rolled back. it stays open until the worker exits and the database discards it. This is harmless for DAMA, but an extension that must flush or close something at the end of a worker has nowhere to do it.
2. Double events. A subscriber registered in `bootstrap()` also receives every event forwarded from the workers, with nothing to denote where it came from. This is ok for DAMA because an extra rollback on an idle connection is harmless. It's not ok for an extension that must act once per test. A flag on forwarded events would fix it.

<details>
<summary>Reproducer for 2: a counting extension, files and output verbatim</summary>

`phpunit.xml`

```xml
<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../vendor/phpunit/phpunit/phpunit.xsd" bootstrap="bootstrap.php" cacheDirectory=".phpunit.cache">
  <testsuites>
    <testsuite name="probe">
      <directory>tests</directory>
    </testsuite>
  </testsuites>
  <extensions>
    <bootstrap class="Probe\TestCounterExtension"/>
  </extensions>
</phpunit>
```

`bootstrap.php`

```php
<?php

declare(strict_types=1);

require __DIR__ . '/../vendor/autoload.php';
require __DIR__ . '/TestCounterExtension.php';
```

`TestCounterExtension.php`

```php
<?php

declare(strict_types=1);

namespace Probe;

use PHPUnit\Event\Test\Finished;
use PHPUnit\Event\Test\FinishedSubscriber;
use PHPUnit\Runner\Extension\Facade;
use PHPUnit\Runner\Extension\ParallelWorkerExtension;
use PHPUnit\Runner\Extension\ParameterCollection;
use PHPUnit\Runner\Extension\WorkerFacade;
use PHPUnit\TextUI\Configuration\Configuration;

final class TestCounterExtension implements ParallelWorkerExtension
{
    public function bootstrap(Configuration $configuration, Facade $facade, ParameterCollection $parameters): void
    {
        $facade->registerSubscriber(new CountingSubscriber());
    }

    public function bootstrapWorker(Configuration $configuration, WorkerFacade $facade, ParameterCollection $parameters): void
    {
        $facade->registerSubscriber(new CountingSubscriber());
    }
}

final class CountingSubscriber implements FinishedSubscriber
{
    public function notify(Finished $event): void
    {
        file_put_contents(sys_get_temp_dir() . '/count.txt', getmypid() . ' ' . $event->test()->name() . "\n", FILE_APPEND);
    }
}
```

`tests/WorkerOneTest.php`

```php
<?php

declare(strict_types=1);

namespace Probe\Tests;

final class WorkerOneTest extends TestCase
{
    {
        $this->assertTrue(true);
    }
}
```

`tests/WorkerTwoTest.php`

```php
<?php

declare(strict_types=1);

namespace Probe\Tests;

final class WorkerTwoTest extends TestCase
{
    public function testTwo(): void
    {
        $this->assertTrue(true);
    }
}
```

`tests/MainTest.php`

```php
<?php

declare(strict_types=1);

namespace Probe\Tests;

use PHPUnit\Framework\Attributes\DoNotRunInParallel;

#[DoNotRunInParallel]
final class MainTest extends TestCase
{
    public function testMain(): void
    {
        $this->assertTrue(true);
    }
}
```

`vendor/bin/phpunit -c probe/phpunit.xml` writes 3 lines to `count.txt`:

```
36966 testMain
36966 testOne
36966 testTwo
```

`vendor/bin/phpunit -c probe/phpunit.xml --parallel 2` writes 5 lines. 36976 is the main process, 36985 and 36986 are the workers. `testOne` and `testTwo` are logged in their worker and again in the main process:

```
36976 testMain
36985 testOne
36986 testTwo
36976 testOne
36976 testTwo
```

A subscriber that has to do something exactly once per test, would do it twice for every worker test, and the second time would fail or double count. Nothing on the event lets it skip the copy.
</details>

<details>
<summary>Our `DamaParallelWorkerExtension`, for reference</summary>

```php
final class DamaParallelWorkerExtension implements ParallelWorkerExtension
{
    public function bootstrap(Configuration $configuration, Facade $facade, ParameterCollection $parameters): void
    {
        // also receives every forwarded worker event, see 2.
        new PHPUnitExtension()->bootstrap($configuration, $facade, $parameters);
    }

    public function bootstrapWorker(Configuration $configuration, WorkerFacade $facade, ParameterCollection $parameters): void
    {
        StaticDriver::setKeepStaticConnections(true); // what TestRunner\Started did

        $facade->registerSubscriber(new class implements PreparationStartedSubscriber {
            public function notify(PreparationStarted $event): void
            {
                PHPUnitExtension::rollBack();

                if (PHPUnitExtension::hasSkipAttribute($event->test())) {
                    PHPUnitExtension::skipTransaction();

                    return;
                }

                PHPUnitExtension::unskipTransaction();
                PHPUnitExtension::beginTransaction();
            }
        });

        // nothing can do what TestRunner\Finished did, see 1.
    }
}
```
</details>

**Notices** 
A class whose tests carry `#[RunInSeparateProcess]` is reported, while `--process-isolation` and `#[DoNotRunInParallel]` are not. 

We want to keep `failOnPhpunitNotice="true"`, so a class we deliberately isolate with `#[RunInSeparateProcess]` should not produce a notice and fail the build in parallel mode. right now it does.

**sharding** 
native sharding would be great, and the quality of life features would follow from it, above all balancing the shards by the test durations recorded in historic runs. 
But it is easily worked around today, so it is not blocking me at least. This is all we do, on every runner:

```bash
# same list, same order on every runner: sort, then shuffle with a fixed seed
find tests -name '*Test.php' | LC_ALL=C sort \
  | shuf --random-source=<(openssl enc -aes-256-ctr -pass pass:seed -nosalt </dev/zero 2>/dev/null) \
  > all_tests

# runner N of M takes the Nth of M equal slices
split --number="l/${SHARD_NUMBER}/${SHARD_COUNT}" all_tests > my_tests

vendor/bin/phpunit --test-files-file my_tests --parallel 5
```

bendavies:
One more thing: suspiciously high memory in the parent. With `--parallel 5` on chunks of about 4,300 tests the parent peaked at 1.35, 1.81 and 2.48 GB; the same chunk sequentially ends at 0.8 GB.

Most of the bytes were our own fault: on a cold cache the first test in each worker triggers one deprecation about 28,000 times.
Not PHPUnit's problem.

What is PHPUnit's problem: results are reported in suite order, so when one slow class is still running, every class that finishes after it has its results in the parent's memory until the slow one is done.
In one chunk a single long running test was still running while 274 tests finished behind it, which resulted in the memory usage of 2.48 GB. 
If the parent only read a class's results at the moment it reports them, the backlog would sit on disk instead.

sebastianbergmann:
Thank you, @bendavies, for testing this again and for the detailed feedback. I've pushed the following changes to the branch.

**End-of-worker hook.** `ParallelWorkerExtension` has a new method, `shutdownWorker(): void`. It is called once in every worker process, on the same instance whose `bootstrapWorker()` succeeded, after the worker has run its last test class. That happens at the end of the run and also when a worker is replaced because of `--recycle-workers-after`. It is not called for a worker that crashes or is killed. If it throws, the exception is reported as a PHPUnit test runner warning. This is where the rollback that your `TestRunner\Finished` subscriber used to do can go now.

**Duplicate events.** Every event's telemetry information now records the ID of the process that emitted it, `$event->telemetryInfo()->processId()`. A subscriber in the main process can use this to skip the events it receives again when the main process replays what a worker already ran:

```php
public function notify(PreparationStarted $event): void
{
    if ($event->telemetryInfo()->processId() !== getmypid()) {
        return; // this test ran in another process
    }

    // ...
}
```

With this check, your counting reproducer logs each test exactly once, in the process that ran it. The same applies to tests that run in process isolation, whose events the main process has always received a second time.

**Notices.** A class that uses `#[RunInSeparateProcess]` or `#[RunTestsInSeparateProcesses]` no longer triggers the notice. Like `--process-isolation` and `#[DoNotRunInParallel]`, it is a deliberate choice. The only fallback that is still reported is a test that depends on a test of another class, so `failOnPhpunitNotice="true"` should work for you again.

**Memory in the main process.** You are right: results are held in memory until it is their turn in suite order, so one slow class keeps everything that finishes after it in memory.

**Sharding.** Thanks for sharing your `--test-files-file` workaround. If at all, I want to tackle sharding after parallel test execution has been released.

While working on this, I found that extensions are not bootstrapped in the child process of a test that runs in process isolation: #6995. This means that a `#[RunInSeparateProcess]` test currently runs without DAMA's transaction, whether or not the run is parallel.

bendavies:
@sebastianbergmann I ran the branch a third time on 3bb542a. 
All tests passed with no notices, so `failOnPhpunitNotice="true"` is on again. 
The peak memory of the parent process did not change. Expected, because the buffering is the same.

The branch is also 31% faster than our current runner, which starts one PHPUnit process for each test file and runs five at a time. 
I compared five runs of this branch with the last eight passing runs on our main branch, and the slowest run of the branch was faster than the fastest run on main.

The reason is the start cost of each process. 
A process for one test file uses about 2.4 CPU seconds more than its tests need. About one second of this is the process start, before the PHPUnit timer starts. 
The remainder is the setup of the container, the metadata caches and the database connections, etc...
Persistent workers pay this cost once, not once for each file.

This is awesome!

This is the DAMA adapter with the new API. 
It shows how an extension uses the new API, so you can decide whether any part of it needs to be simpler.

<details>
<summary>Our <code>DamaParallelWorkerExtension</code></summary>

```php
<?php

declare(strict_types=1);

use DAMA\DoctrineTestBundle\Doctrine\DBAL\StaticDriver;
use DAMA\DoctrineTestBundle\PHPUnit\PHPUnitExtension;
use PHPUnit\Event\Test\PreparationStarted;
use PHPUnit\Event\Test\PreparationStartedSubscriber;
use PHPUnit\Event\TestRunner\Finished;
use PHPUnit\Event\TestRunner\FinishedSubscriber;
use PHPUnit\Event\TestRunner\Started;
use PHPUnit\Event\TestRunner\StartedSubscriber;
use PHPUnit\Runner\Extension\Facade;
use PHPUnit\Runner\Extension\ParallelWorkerExtension;
use PHPUnit\Runner\Extension\ParameterCollection;
use PHPUnit\Runner\Extension\WorkerFacade;
use PHPUnit\TextUI\Configuration\Configuration;

final class DamaParallelWorkerExtension implements ParallelWorkerExtension
{
    public function bootstrap(Configuration $configuration, Facade $facade, ParameterCollection $parameters): void
    {
        $facade->registerSubscriber(new class implements StartedSubscriber {
            public function notify(Started $event): void
            {
                StaticDriver::setKeepStaticConnections(true);
            }
        });

        $facade->registerSubscriber(new TransactionPerTestSubscriber());

        $facade->registerSubscriber(new class implements FinishedSubscriber {
            public function notify(Finished $event): void
            {
                PHPUnitExtension::rollBack();
                StaticDriver::setKeepStaticConnections(false);
            }
        });
    }

    public function bootstrapWorker(Configuration $configuration, WorkerFacade $facade, ParameterCollection $parameters): void
    {
        StaticDriver::setKeepStaticConnections(true);

        $facade->registerSubscriber(new TransactionPerTestSubscriber());
    }

    public function shutdownWorker(): void
    {
        PHPUnitExtension::rollBack();
        StaticDriver::setKeepStaticConnections(false);
    }
}

final class TransactionPerTestSubscriber implements PreparationStartedSubscriber
{
    public function notify(PreparationStarted $event): void
    {
        if ($event->telemetryInfo()->processId() !== \getmypid()) {
            return;
        }

        PHPUnitExtension::rollBack();

        if (PHPUnitExtension::hasSkipAttribute($event->test())) {
            PHPUnitExtension::skipTransaction();

            return;
        }

        PHPUnitExtension::unskipTransaction();
        PHPUnitExtension::beginTransaction();
    }
}
```

</details>

The adapter repeats two things:

- It registers the subscriber for each test twice, once for each facade. The subscriber then compares process IDs to skip the events of other processes. A facade method that registers a subscriber only for the tests of the current process can remove that comparison.
- `bootstrapWorker()` and `shutdownWorker()` repeat the work that the `TestRunner\Started` and `TestRunner\Finished` subscribers do in the main process. If a worker sends these two events to its own subscribers, one `bootstrap()` can work in both processes. Then the interface only has to mark the extension as safe for workers.

Neither of these are blocking. We can use the API as it is.

sebastianbergmann:
@bendavies Once more: thank you for running the branch again, and for the numbers: 31% faster than a process per test file is a nice confirmation that persistent workers pay off. Thank you also for sharing the DAMA adapter; seeing a real extension use the API is exactly what I needed.

**Registering a subscriber for the events of the current process**

I have implemented this: the extension facade now has `registerSubscriberForEventsOfThisProcess()` (and `registerSubscribersForEventsOfThisProcess()`). A subscriber registered with it is only notified of events that were emitted in the current process. The events that the main process replays for tests that ran in a worker, or in a separate process, are not delivered to it.

You only need it in `bootstrap()`. A worker never replays events from another process: tests that require process isolation run in the main process, not in a worker. So `WorkerFacade::registerSubscriber()` already behaves this way.

Please note that the filter is based on where an event was emitted, not on where a test ran. When a worker crashes, the main process emits "test errored" and "test finished" events for the tests of that worker that never got to run. These events are emitted in the main process, so a subscriber registered with the new method receives them. `PreparationStarted` is not among them, so your subscriber is not affected.

**Emitting `TestRunner\Started` and `TestRunner\Finished` in the workers**

I would like to keep `bootstrapWorker()` and `shutdownWorker()`:: a single `bootstrap()` works nicely for DAMA because every subscriber it registers is about the process it runs in. That is likely not true for extensions in general:

- `bootstrap()` receives a `Facade` that can replace the output, register tracers, and require code coverage collection. None of this applies in a worker. We would have to ignore these calls silently or let them fail there.
- Every subscriber registered in `bootstrap()` would also be registered in the workers. An extension that also has a subscriber that writes a report from all events of the test run would then write partial reports in each worker. It would have to check whether it runs in a worker, so the branch between the two cases would move into the extension instead of being expressed by the interface.
- `TestRunner\Started` and `TestRunner\Finished` are emitted once per test run today. In the workers, they would be emitted once per worker, several times when workers are recycled, and `TestRunner\Finished` would not be emitted for a worker that crashes. A subscriber that does something once per test run, such as creating a database schema or truncating a log file, would do it several times.

The duplication in the adapter is two statements, which a private method on the extension can hold. Together with the new method, the adapter could look like this:

```php
final class DamaParallelWorkerExtension implements ParallelWorkerExtension
{
    public function bootstrap(Configuration $configuration, Facade $facade, ParameterCollection $parameters): void
    {
        $facade->registerSubscriber(new class implements StartedSubscriber {
            public function notify(Started $event): void
            {
                DamaParallelWorkerExtension::start();
            }
        });

        $facade->registerSubscriberForEventsOfThisProcess(new TransactionPerTestSubscriber());

        $facade->registerSubscriber(new class implements FinishedSubscriber {
            public function notify(Finished $event): void
            {
                DamaParallelWorkerExtension::finish();
            }
        });
    }

    public function bootstrapWorker(Configuration $configuration, WorkerFacade $facade, ParameterCollection $parameters): void
    {
        self::start();

        $facade->registerSubscriber(new TransactionPerTestSubscriber());
    }

    public function shutdownWorker(): void
    {
        self::finish();
    }

    public static function start(): void
    {
        StaticDriver::setKeepStaticConnections(true);
    }

    public static function finish(): void
    {
        PHPUnitExtension::rollBack();
        StaticDriver::setKeepStaticConnections(false);
    }
}

final class TransactionPerTestSubscriber implements PreparationStartedSubscriber
{
    public function notify(PreparationStarted $event): void
    {
        PHPUnitExtension::rollBack();

        if (PHPUnitExtension::hasSkipAttribute($event->test())) {
            PHPUnitExtension::skipTransaction();

            return;
        }

        PHPUnitExtension::unskipTransaction();
        PHPUnitExtension::beginTransaction();
    }
}
```

The process ID comparison in `TransactionPerTestSubscriber` is no longer needed.

theofidry:
>Notices. A class that uses #[RunInSeparateProcess] or #[RunTestsInSeparateProcesses] no longer triggers the notice. Like --process-isolation and #[DoNotRunInParallel], it is a deliberate choice. The only fallback that is still reported is a test that depends on a test of another class, so failOnPhpunitNotice="true" should work for you again.

So what happens with those cases actually? I.e. what is the effective behaviour when you use `#[RunInSeparateProcess]` or `#[RunTestsInSeparateProcesses]`?

sebastianbergmann:
> > Notices. A class that uses #[RunInSeparateProcess] or #[RunTestsInSeparateProcesses] no longer triggers the notice. Like --process-isolation and #[DoNotRunInParallel], it is a deliberate choice. The only fallback that is still reported is a test that depends on a test of another class, so failOnPhpunitNotice="true" should work for you again.
> 
> So what happens with those cases actually? I.e. what is the effective behaviour when you use `#[RunInSeparateProcess]` or `#[RunTestsInSeparateProcesses]`?

They run the way they do without `--parallel`: in the main process, one isolated child process per test, in their normal place in the suite order. Because the whole class is taken out of the worker pool, one `#[RunInSeparateProcess]` method takes its whole class out of parallel execution, and the pool cannot get new work while that class runs.

PR #6803, opened 2026-07-02
Remove dependency on ext-xml

TimWolla:
PHPUnit is not actually using the ext-xml extension. Its own test suite passes for a minimal PHP 8.5 build with the following configure line:

    ./configure --disable-all --enable-option-checking=fatal --enable-werror --enable-ctype --enable-tokenizer --enable-mbstring --enable-dom --with-libxml --enable-xmlwriter --enable-phar --enable-filter

with the following extensions loaded

    [PHP Modules]
    Core
    ctype
    date
    dom
    filter
    hash
    json
    lexbor
    libxml
    mbstring
    pcre
    Phar
    random
    Reflection
    SPL
    standard
    tokenizer
    uri
    xmlwriter
    Zend OPcache

    [Zend Modules]
    Zend OPcache

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -64,7 +64,7 @@ jobs:
         uses: shivammathur/setup-php@f3e473d116dcccaddc5834248c87452386958240 # 2.37.2
         with:
           php-version: ${{ env.PHP_VERSION }}
-          extensions: none, ctype, curl, dom, json, libxml, mbstring, openssl, phar, tokenizer, xml, xmlwriter, pcntl
+          extensions: none, ctype, curl, dom, json, libxml, mbstring, openssl, phar, tokenizer, xmlwriter, pcntl
           coverage: none
           tools: none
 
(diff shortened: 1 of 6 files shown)


TimWolla:
The ext-filter extension is only required to pass the `tests/end-to-end/extension-xml/phar-extension.phpt` test. The `vendor/phar-io/manifest/` dependency uses `filter_var()` without declaring an explicit dependency on `ext-filter`. It may make sense to also list `'filter'` in the `$requiredExtensions`.

sebastianbergmann:
Please also remove `xml` from `.github/workflows/ci.yaml`. Thank you!

TimWolla:
> Please also remove `xml` from `.github/workflows/ci.yaml`. Thank you!

Done; also from nightly.yaml, which I assume you meant to include.

sebastianbergmann:
Cherry-picked into `12.5` and merged to `13.2` and `main` from there.

sebastianbergmann:
Backported this now to `8.5`, `9.6`, `10.5`, and `11.5`, too.

PR #6812, opened 2026-07-04
adding temp files to gitignore, for git fork demonstration

h978412:
No comments


sebastianbergmann:
Please do not do something like this again.

PR #6825, opened 2026-07-08
Guard against recursive previous error handler forwarding

ostrolucky:
#### Summary

Fix an infinite recursion in `PHPUnit\Runner\ErrorHandler` when PHPUnit forwards an issue to a previously registered error handler and that handler re-enters PHPUnit's error handling.

#### Problem

PHPUnit since 75d896e4aa3a0a1eb7e60364103fef818ba7f108 changed its error handling so that a previously registered error handler is invoked before PHPUnit records an issue. This is important for compatibility with handlers that convert PHP warnings, notices, or deprecations into exceptions: in those cases, the error should become normal control flow observed by the test runner rather than being reported as a PHPUnit issue.

However, this forwarding can create a recursive call cycle when the previous handler triggers another PHP issue or otherwise calls back into PHPUnit's `ErrorHandler` while PHPUnit is already forwarding an issue to it. In that situation, PHPUnit can repeatedly forward between handlers instead of reporting the warning normally, which may lead to a hang, stack exhaustion, or a segmentation fault depending on the environment.

A real-world example of this showed up in a functional test using `mlocati/ip-lib` for CIDR parsing together with Symfony's error handling bridge. That combination interacts with PHP's error-handler stack while warnings are being processed, and with PHPUnit's previous-handler forwarding in place it could trigger the recursive error-handler path. The observed result was that the test process no longer completed reliably and could hang indefinitely or crash instead of producing a normal PHPUnit warning/error report.

### DIFF
--- a/src/Runner/ErrorHandler.php
+++ b/src/Runner/ErrorHandler.php
@@ -74,6 +74,7 @@ final class ErrorHandler
     private ExcludeList $excludeList;
     private bool $enabled                     = false;
     private ?int $originalErrorReportingLevel = null;
+    private bool $isForwarding                = false;
 
     /**
      * @var ?callable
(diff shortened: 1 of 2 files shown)


sebastianbergmann:
Thank you!

I refined your solution a bit in 120f4d2023058ef60fda47e244618261a808b623. Can you please check whether your problem is still solved after my changes? Thank you!

ostrolucky:
LGTM, nothing was hanging for me there at least

Issue #6484, opened 2026-01-27
Customize which deprecation type to fail the test run


D3strukt0r:
Hi there

We used to have the case with the Symfony PhpUnit bridge, where self and direct deprecations would cause the test run to fail, and indirect deprecation would simply be shown informatively.

I do not seem to be able to configure this now, though, since `failOnDeprecation` is for all types. Is there something that could be done to support this?

sebastianbergmann:
This is what "[limiting issues to your code](https://docs.phpunit.de/en/12.5/error-handling.html#limiting-issues-to-your-code)" is about.

D3strukt0r:
I see, that's also what we have configured in our projects. But there is no way do show me the indirect deprecations without also failing the test run.

If that's desired, then I guess, it is what it is; if not, however, that would be really nice. Otherwise, people at our company would never discover deprecations in our own Libraries

PS: That's the use case, discover deprecations in our libraries without failing our main project's tests

PPS: There are also often Symfony configuration deprecations, which fall under indirect deprecation but is actually very important, since it's the projects configuration which causes the deprecation. Stuff like this wouldn't be discovered until the next major upgrade

stof:
"Limiting issues to your own code" restricts both what we fail on and what we display. It would be great to allow distinguishing those (of course, it does not make sense to restrict the display if we don't restrict failures, so this should probably be a 3-state value, not 2 separate booleans)

D3strukt0r:
Also what I just remembered, when I was upgrading a legacy project at work and and came across the Symfony Bridge env variable for it (`SYMFONY_DEPRECATIONS_HELPER=max[total]=99999&max[self]=1&max[direct]=2127`), I can tell it how many errors of a specific type are allowed, as in "setting a baseline", so maybe instead of just a 3-state it should also allow a number to say "anything above this count will result in an error"

sebastianbergmann:
Closes via #6827.

PR #6827, opened 2026-07-10
Customize which deprecation trigger types fail the test run

sebastianbergmann:
## Problem

`failOnDeprecation` is currently all-or-nothing: when it is enabled, every deprecation that is collected fails the test run. The only way to exempt, for example, indirect deprecations is `<source ignoreIndirectDeprecations="true"/>`, which drops them entirely: they neither fail the run *nor* are they displayed.

Projects that follow the workflow of Symfony's PHPUnit Bridge want to fail on deprecations they can act on (`self`, `direct`) while still seeing indirect deprecations. With #6489 (issue trigger classification) and #6530 (custom issue trigger resolvers) in place, the trigger classification is now reliable enough to attach different policies to.

## Solution

This extends the existing `failOn*` / `doNotFailOn*` family with trigger-specific members:

| XML attribute (`<phpunit>`) | CLI options |
|---|---|
| `failOnSelfDeprecation` | `--fail-on-self-deprecation` / `--do-not-fail-on-self-deprecation` |
| `failOnDirectDeprecation` | `--fail-on-direct-deprecation` / `--do-not-fail-on-direct-deprecation` |
| `failOnIndirectDeprecation` | `--fail-on-indirect-deprecation` / `--do-not-fail-on-indirect-deprecation` |

The trigger-specific flags are overrides layered on top of the resolved `failOnDeprecation` value:

* Each trigger-specific flag is initialized from the resolved `failOnDeprecation` (which itself respects `failOnAllIssues` and `--do-not-fail-on-deprecation`).
* `failOn<Trigger>Deprecation` / `--fail-on-<trigger>-deprecation` forces failing for that trigger type, `--do-not-fail-on-<trigger>-deprecation` exempts it.
* Deprecations with an unknown trigger (no `<source>` configured, or `identifyIssueTrigger="false"`) always follow the base `failOnDeprecation` flag, so the trigger-specific options are only effective when issue trigger identification is active.

Together with the existing `ignore<Trigger>Deprecations` attributes on `<source>`, each trigger type now supports the three states requested in #6484: fail (report + fail), report (report without failing), and ignore.

Display behavior is unchanged: a deprecation that is exempted from failing is still counted, still shown with `--display-deprecations`, and still results in "OK, but there were issues!". The shell exit code is 0, though. As with `failOnDeprecation`, using one of the new `failOn*` options implicitly enables the display of deprecation details.

## Implementation notes

* The trigger classification was previously lost at aggregation: `Collector` folds deprecation events into `Issue` objects keyed by `file:line:message`, and one aggregated issue can be, for example, direct in one test and indirect in another (same callee, different callers). The `Collector` therefore tracks distinct issue IDs per trigger type at the event level; for test-triggered as well as test-runner-triggered deprecations (both `E_USER_DEPRECATED` and `E_DEPRECATED`), after the `IssueFilter` and baseline checks so that ignored deprecations do not count.
* `TestResult` exposes the new counts via `hasSelfDeprecations()`, `hasDirectDeprecations()`, `hasIndirectDeprecations()`, and `hasDeprecationsWithUnknownTrigger()` (plus `numberOf*()` counterparts). Because an aggregated issue can be counted under more than one trigger type, these buckets do not necessarily sum to `numberOfPhpOrUserDeprecations()`.
* `ShellExitCodeCalculator` replaces the single `$failOnDeprecation && $result->hasPhpOrUserDeprecations()` check with one check per trigger type plus the unknown-trigger check. With none of the new options used, all trigger-specific flags equal the resolved `failOnDeprecation` value and the union of the four checks is exactly the previous behavior (full backward compatibility).
* `IssueFilter`, `Collector` aggregation, `ProgressPrinter`, `SummaryPrinter`, `ResultPrinter`, baseline handling, events, and loggers are untouched.


sebastianbergmann:
@stof @D3strukt0r May I kindly request your review of these changes? I hope I understood your use case correctly. Thank you!

stof (commented):
LGTM

D3strukt0r (commented):
also looks good to me, i don't see any way to define a baseline though, like saying only fail above 10 self deprecations. you think that would fit in the same setting (instead of bool => bool|number?) or a separate setting in the future?

however, even as is, it would already help a lot. Super gemacht :)

Issue #6451, opened 2025-12-16
Incomplete version in `RequiresPhp` (e.g. `<=8.5`) is compared against full PHP version, causing unexpected skips


staabm:
| Q                   | A
| --------------------| ---------------
| PHPUnit version     | 10.5.60-17-g8978ea9aa
| PHP version         | 8.5.0
| Installation Method | Composer / PHAR

#### Summary

Test is skipped even though it should not.

#### Current behavior

test is skipped on .0 versions, when `RequiresPhp` is used with `<=`:

```
➜  phpunit git:(10.5) ✗ php phpunit TestFoo.php
Xdebug: [Step Debug] Could not connect to debugging client. Tried: 127.0.0.1:9003 (through xdebug.client_host/xdebug.client_port).
PHPUnit 10.5.60-17-g8978ea9aa by Sebastian Bergmann and contributors.

Runtime:       PHP 8.5.0
Configuration: /Users/staabm/workspace/phpunit/phpunit.xml

S                                                                   1 / 1 (100%)

Time: 00:00.005, Memory: 10.00 MB

OK, but some tests were skipped!
Tests: 1, Assertions: 0, Skipped: 1.
```

#### How to reproduce

TestFoo.php

```php
<?php

#[\PHPUnit\Framework\Attributes\RequiresPhp('<=8.5')]
final class TestFoo extends \PHPUnit\Framework\TestCase {
    public function testFoo() {
        $this->assertTrue(true);
    }
}
```

#### Expected behavior

test should not be skipped

---

I have [reported this issue on php-src](https://github.com/php/php-src/issues/20706) but atm its judged as working-as-expected (which I don't agree with)

staabm:
we could work arround, by filling the user supplied "8.5" to a "8.5.0"

sebastianbergmann:
> we could work arround, by filling the user supplied "8.5" to a "8.5.0"

We could do that, but not without emitting a warning.

sebastianbergmann:
I decided against "repairing data" by "by filling the user supplied `8.5` to a `8.5.0`. Instead we only emit a test runner warning.

PR #6828, opened 2026-07-11
Test caret/asterisk version operator completeness

staabm:
looking at https://github.com/sebastianbergmann/version-requirement/commit/fe6db86867cd73d4ec425c96743e963defd0bdcd#diff-48dfada243c3603061caeb6193619a86b2168a8e709da0903ceb85a50b86a0f0R61-R66 made me wonder whether the current stable phpunit release handles constraints like `8.5.*` or `^8.5` as complete or incomplete

### DIFF
--- a/tests/end-to-end/regression/6451.phpt
+++ b/tests/end-to-end/regression/6451.phpt
@@ -13,7 +13,7 @@ PHPUnit %s by Sebastian Bergmann and contributors.
 
 Runtime: %s
 
-....                                                                4 / 4 (100%)
+......                                                              6 / 6 (100%)
 
 Time: %s, Memory: %s
 
@@ -26,4 +26,4 @@ There were 3 PHPUnit test runner warnings:
 3) Version requirement ">= 1" used by PHPUnit\TestFixture\Issue6451\Issue6451Test::testIncompletePhpExtensionVersion() is incomplete, expected a version that consists of major, minor, and patch level ("8.5.0" instead of "8.5", for example)
 
 OK, but there were issues!
-Tests: 4, Assertions: 4, PHPUnit Warnings: 6.
+Tests: 6, Assertions: 6, PHPUnit Warnings: 6.
--- a/tests/end-to-end/regression/6451.phpt
+++ b/tests/end-to-end/regression/6451.phpt
@@ -13,7 +13,7 @@ PHPUnit %s by Sebastian Bergmann and contributors.
 
 Runtime: %s
 
-......                                                              6 / 6 (100%)
+..%c%c..                                                              6 / 6 (100%)
 
 Time: %s, Memory: %s
 
@@ -26,4 +26,4 @@ There were 3 PHPUnit test runner warnings:
 3) Version requirement ">= 1" used by PHPUnit\TestFixture\Issue6451\Issue6451Test::testIncompletePhpExtensionVersion() is incomplete, expected a version that consists of major, minor, and patch level ("8.5.0" instead of "8.5", for example)
 
 OK, but there were issues!
-Tests: 6, Assertions: 6, PHPUnit Warnings: 6.
+Tests: 6, Assertions: 6, PHPUnit Warnings: 6.%A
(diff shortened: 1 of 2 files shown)


tests/end-to-end/regression/6451.phpt
```diff
@@ -26,4 +26,4 @@ There were 3 PHPUnit test runner warnings:
 3) Version requirement ">= 1" used by PHPUnit\TestFixture\Issue6451\Issue6451Test::testIncompletePhpExtensionVersion() is incomplete, expected a version that consists of major, minor, and patch level ("8.5.0" instead of "8.5", for example)
 
 OK, but there were issues!
-Tests: 4, Assertions: 4, PHPUnit Warnings: 6.
+Tests: 6, Assertions: 6, PHPUnit Warnings: 6.
```
staabm: as I assumed, we now have 2 more warnings because the newly added versions are considered incomplete.

I wonder though, why this test only renders 3 warnings instead of 6?

sebastianbergmann:
Please send this against `13.2`. Also note that 85527ec5d7ffd997f1ea72bbc01180b8f18e72aa recently touched related areas of the code.

tests/end-to-end/regression/6451.phpt
```diff
@@ -13,7 +13,7 @@ PHPUnit %s by Sebastian Bergmann and contributors.
 
 Runtime: %s
 
-......                                                              6 / 6 (100%)
+..%c%c..                                                              6 / 6 (100%)
```
sebastianbergmann: PHPUnit 13 does not support PHP < 8.4; I do not understand why these placeholder are needed.

staabm: 8.4 reports `......`
while 8.5 reports `..SS..`

I played arround a bit, but atm have no idea how to make the tests properly pass in all versions

sebastianbergmann: Ah, sorry for the noise and thank you for explaining. In such a case, I usually split the test in two with separate skip conditions.

staabm: Should be good to go now

staabm:
to re-iterate the initial motivation for this PR.

I think the test expectations here are wrong:
https://github.com/sebastianbergmann/version-requirement/commit/fe6db86867cd73d4ec425c96743e963defd0bdcd#diff-48dfada243c3603061caeb6193619a86b2168a8e709da0903ceb85a50b86a0f0R61-R66

sebastianbergmann:
@staabm The test expectations in [sebastian/version-requirement](https://github.com/sebastianbergmann/version-requirement) are intentional, and they match what PHPUnit 13.2 does. Including what the test from this pull request now documents: no warning is emitted for `^8.3` or `8.*`.

I think two things came together here to cause confusion:

* "we now have 2 more warnings because the newly added versions are considered incomplete". That never actually happened. The expected output of this test does not contain a warning for the caret/asterisk version requirements, neither in your first commit nor in the merged version. And `PHPUnit Warnings: 6` was already in the expected output back when the test only had 4 test methods.

* "why does this test only render 3 warnings instead of 6?" The 6 are the same 3 warnings counted twice. `Requirements::requirementsNotSatisfiedFor()` emits the warning on each call, and it is called twice per test: once while the test suite is built (`TestBuilder::build()` checks whether the requirements are satisfied) and once when the test is run (`TestCase::checkRequirements()`). The result printer de-duplicates the list of warnings it displays, but the counter counts every emission. That is a quirk that might be worth fixing on its own, but it is unrelated to caret/asterisk version requirements.

As to why the expectations are the way they are: only version comparisons such as `>= 8.5` or `<= 8` are evaluated using `version_compare()`, which is where incomplete versions have surprising semantics (`version_compare('8.5.0', '8.5', '<=')` is `false`, that was #6451). Version constraints such as `^8.5` or `8.5.*` are evaluated by phar-io/version, which gives partial versions well-defined semantics: `^8.5` means `>= 8.5.0, < 9.0.0` and `8.5.*` means `>= 8.5.0, < 8.6.0`. There is no footgun to warn about, therefore `isComplete()` returns `true` for them.

[I have expanded the documentation of `Requirement::isComplete()` as well as the tests in sebastian/version-requirement to make this intent explicit.](https://github.com/sebastianbergmann/version-requirement/commit/5a1ef5c0d69a3f6f7762332e1b5ff70d679a4b58)

PR #6832, opened 2026-07-14
Allow doubling properties that do not declare property hooks

sebastianbergmann:
This pull request implements the remaining part of #6546.

PHPUnit automatically doubles the property hooks of a doubled class or interface: reading from and writing to a hooked property can be configured and verified just like calling a doubled method. Properties that do not declare hooks, however, could not participate in this: on a test double, they behave like regular properties, and there was no way to configure behavior for them or to verify how the tested code interacts with them.

PHP itself allows a child class to add hooks to a property that its parent class declared without hooks. This pull request makes that capability available for test doubles as an explicit opt-in:

```php
$message = $this->getMockBuilder(Message::class)
    ->doubleProperties(['recipient'])
    ->getMock();

$message
    ->expects($this->once())
    ->method(PropertyHook::set('recipient'))
    ->with('user@example.com');
```

The new `doubleProperties()` method is available on both `getMockBuilder()` and `getStubBuilder()`. A property that has been opted in supports everything hooked properties already support: `PropertyHook::get()` and `PropertyHook::set()` can be used to configure return values, callbacks, and expectations.

## Why opt-in?

Doubling a hook-less property changes how it behaves: writes no longer store a value and reads no longer return it, `$double->arrayProperty[] = $item` and references to the property stop working. Assigning values to public properties of test doubles is a widespread pattern (see #6545), so doubling every property unconditionally would break existing test suites. With the opt-in, nothing changes for existing code: properties that are not listed in `doubleProperties()` continue to behave exactly as before.

## Limitations

A property can only be doubled when it exists, is public, is not static, is not readonly, is not final, and declares a type. `doubleProperties()` fails with a clear error message when one of these requirements is not met.

### DIFF
--- a/src/Framework/MockObject/Exception/PropertyCannotBeDoubledException.php
+++ b/src/Framework/MockObject/Exception/PropertyCannotBeDoubledException.php
@@ -1,0 +1,32 @@
+<?php declare(strict_types=1);
+/*
+ * This file is part of PHPUnit.
+ *
+ * (c) Sebastian Bergmann <sebastian@phpunit.de>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+namespace PHPUnit\Framework\MockObject;
+
+use function sprintf;
+
+/**
+ * @no-named-arguments Parameter names are not covered by the backward compatibility promise for PHPUnit
+ *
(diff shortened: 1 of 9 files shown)


sebastianbergmann:
@derrabus Would be great to get your opinion on this. Thank you!

derrabus (commented):
This looks great, thank you!

PR #6863, opened 2026-07-28
Cache which tests a test file contains

sebastianbergmann:
Running a subset of a large test suite is slower than it should be. Before PHPUnit can decide that a test file has nothing to contribute to the run, it has to load that file, and loading a test file is most of the work. On a suite of a few hundred files, `--group something-narrow` spends nearly all of its time loading files whose tests are then filtered away again.

This pull request adds an opt-in cache that remembers which tests each test file contains and which groups they are in, so that on a later run PHPUnit can decide not to load a file at all.

## Using it

The cache is off by default. Turn it on in the XML configuration file:

```xml
<phpunit cacheDirectory=".phpunit.cache" cacheTestIndex="true">
```

or on the command line:

```
--cache-test-index
--do-not-cache-test-index
```

It needs a cache directory, the same one the test run history and the static analysis cache for code coverage use. When the option is enabled without one, PHPUnit says so instead of quietly doing nothing.

## What it does for you

Measured on PHPUnit's own `unit` test suite (524 test files), comparing a run with a warm cache against the same run without the cache:

| what you run | without | with | |
|---|---|---|---|
| `--group test-runner/test-index` (98 tests) | 981 ms | 242 ms | **4.1× faster** |
| `--group framework/assertions` (1481 tests) | 686 ms | 313 ms | **2.2× faster** |
| `--filter TestIndexTest` (42 tests) | 690 ms | 461 ms | **1.5× faster** |
| the whole suite, no filter | 5.46 s | 5.22 s | no measurable difference |

The narrower the selection, the more there is to save. The gain is in test discovery, so it is a fixed saving per run rather than a percentage. This makes it most noticeable exactly where it is most annoying, on the quick focused runs you do dozens of times an hour.

Building the cache on the very first run is not free, but that cost is repaid in full by the second run.

### The tests that run

A file is only skipped when the cache can prove that nothing in it could have been selected anyway. Selecting tests by group and selecting them by name are each sufficient on their own to establish that, and both are consulted. Every measurement above was checked to select exactly the same tests, in the same order, as the same command without the cache.

### What PHPUnit tells you

If loading a file makes PHPUnit warn about that file, the file is never skipped, so the same command keeps producing the same output. Otherwise a warning would appear on the first run and silently vanish on the second.

### Anything, when it is off

The cache is opt-in and defaults to off.

## Staying correct as you work

An entry is used only while every source file it was derived from still has the contents it had when the entry was written. That includes parent classes and traits, so adding a test to an abstract base class or changing a group on a trait is noticed in every test class that inherits from it.

Files that could not be loaded are never remembered, so a run that fails will fail the same way again rather than being skipped past.


staabm:
to make the user aware of the opt-in performance feature it might be worth sending a warning/notice/whatever when `--group` is used and the index-feature is not enabled.. otherwise people might miss the feature in the end

(maybe such thing is already implemented, but the PR is very huge, so I was not able to read thru all of it)

PR #6903, opened 2026-08-15
Apply test selection and global state backup settings in parallel workers

sbuerk:
Hi Sebastian,

you asked whether native parallel test execution would be usable for TYPO3 CMS, so we tried it out on the core test suites. Thank you for the invitation, it was an interesting exercise, and this pull request offers back the two fixes we needed to get there.

Short answer to your question: **yes**. With the two fixes below, the whole TYPO3 core unit suite (11694 tests) passes with `--parallel` with the same test and assertion counts as a sequential run, and the functional suites pass as well. In our CI, two parallel functional jobs replace four sequential ones and the slowest job goes from 342s to 211s.

The first two commits are the fixes, the last two only add tests and are kept separate so you can drop or rewrite them on their own.

Your distribution unit is what makes this work for us: one test class per worker is exactly the isolation unit our testing framework already uses, since it derives both the test instance directory and the test database name from the test class name. So workers do not collide on either, and we needed no change at all to that part.

## The two fixes

**1. Test selection was not applied to the tests executed in workers**

`--filter`, `--group` and `--exclude-group` were silently ignored for every test living inside a data provider, repeat or retry suite. Selecting a single test method of a class ran all 107 tests of that class; a `--filter` on our whole unit suite ran 7774 tests instead of 11.

The cause is that the descriptors asked the suite for its tests, and `TestSuite::tests()` returns the raw array, while selection is a filter iterator that only applies when the suite is iterated, as `TestSuite::takeTests()` does in a sequential run. Plain test methods were filtered correctly, which is what made this hard to spot.

The descriptors now iterate the suite. The collection walk additionally skips a suite that the selection has emptied, for which `TestSuite::run()` returns early in a sequential run.

This one mattered a lot to us: our whole database matrix builds on `--exclude-group not-<dbms>`, so a SQLite job was executing the MySQL-only tests.

**2. Per test global state backup was not applied to the tests executed in workers**

A test case recreated from its descriptor never received the settings `TestBuilder` derives from metadata and configuration, so `backupGlobals` and `backupStaticProperties` had no effect inside a worker and state leaked from one test of a class to the next. A single test class run with `--parallel 2` was enough to show it, so no interaction between classes is needed to trigger it. The `--globals-backup` and `--static-backup` options and the `#[BackupGlobals]` attribute were equally ineffective.

The descriptor now delegates to `TestBuilder` to derive those settings again in the worker, where the metadata and the configuration are available as well. `preserveGlobalState` and the exclude lists take the same path and are covered by the same change.

## Verification

Both fixes were checked against sequential runs and against your own suites:

- For each of `--filter`, `--group`, `--exclude-group`, `--repeat` and `--retry`, the parallel run now selects the same tests as the sequential one, and the full `--testdox` output is identical, not only the counts.
- On PHPUnit's own unit suite, `--testsuite unit --parallel 4 --filter toString` went from 2720 tests with 528 errors to the 9 tests a sequential run selects.
- For the state backup, a class whose first test writes `$GLOBALS` and whose second asserts it is clean now passes in parallel, as does the static property equivalent and the class level attribute variant.
- `tests/end-to-end`, `tests/end-to-end/parallel` and `tests/unit` show the same failures before and after both changes, and the failing sets are identical file by file. The five failures under `tests/end-to-end/parallel` (`exclusivity`, `stop-on-clean`, and three `stop-on-failure` variants) are pre-existing on the branch here and untouched by this.
- `php-cs-fixer --dry-run` and `phpstan` are clean on the changed files.

## The two test commits, offered separately

The last two commits only add tests, and they are deliberately kept apart from the two fixes so you can drop or rewrite either of them without touching the rest.

**`Add tests for test selection in parallel test runs`** adds `tests/end-to-end/parallel/selection/`: six pairs of PHPT files sharing one fixture. Each pair runs the same selection sequentially and with `--parallel=2`, and both files of a pair pin the *same* `--EXPECTF--` block, so a `diff` of a pair shows only the `--TEST--` sentence and the added `--parallel=2` line. The sequential file documents what a selection selects, the parallel file asserts a parallel run selects the same. The pairs cover `--filter` on a data provider method, `--filter` on a single data set, `--group`, `--exclude-group`, a `--filter` that matches nothing, and `--repeat` combined with `--filter`. They use `--testdox`, so the expectation names the selected tests rather than only counting them.

The fixture is data provider driven on purpose: a plain test method was selected correctly even before the fix, so a fixture of plain test methods would pass either way and prove nothing. Reverting the selection fix makes all six parallel files fail and leaves all six sequential ones passing; each of the six then runs the whole fixture of 7 tests instead of the 2, 1, 4, 3, 0 and 6 tests the selection asks for.

**`Add tests for the configuration of test cases recreated in a worker`** covers `TestBuilder::configure()` in `tests/unit/Framework/TestBuilderTest.php`, reusing the fixtures already there. Besides two tests pinning concrete settings, the main one asserts that `configure()` leaves a manually instantiated test case configured exactly as `build()` leaves the test case it builds for the same class and method, across every fixture, including the data provider one. Making the method a no-op fails all eight cases.

That second commit is what closes the Codecov report on this pull request: the six lines of `configure()` were the missing ones, and they are missed because that method only ever runs in a worker process. The one line reported missing in `ParallelTestRunner` is already covered by the selection tests above, `filter-matching-nothing-in-parallel.phpt` in particular, since a selection that matches nothing is exactly what empties a suite. Patch coverage of both files is 100% with the two commits in place. The `tests/unit` suite grows by the eight new cases and is otherwise unchanged, and the failing set of `tests/end-to-end/parallel` is the same before and after.

## Notes from using it, take or leave

- Extensions are bootstrapped only in the parent process, without a warning. It does not affect us, we use none, but it is the one thing a user cannot discover from the outside.
- Installing the branch with Composer needs `"no-api": true` on the repository entry, otherwise the GitHub API rate limit for anonymous requests turns into a `Could not authenticate against github.com` error in containers that have no token. Only relevant while this is a branch, but it cost us a moment.
- An XML configuration setting would be welcome eventually, so a suite can declare its own parallelism instead of every call site repeating it. Not a blocker, `--parallel` on the command line is fine.

Three kinds of change were needed on our side, and I mention them because two of them are arguably a compliment to the feature: test cases that flush a shared Redis or Memcached daemon now carry `#[DoNotRunInParallel]`, and parallel execution exposed several of our test cases that registered a shared scratch directory for recursive deletion and so removed the fixtures of other test cases. The latter was a real defect of ours that sequential execution had been hiding.

The evaluation on our side is at https://review.typo3.org/c/Packages/TYPO3.CMS/+/95337 for reference. It is a work in progress change that we do not intend to merge, it only exists to try this out and to gather the numbers above.

Happy to adjust anything here, in particular the comment wording, or to split the two commits into separate pull requests if you prefer.

### DIFF
--- a/src/Framework/TestBuilder.php
+++ b/src/Framework/TestBuilder.php
@@ -191,6 +191,29 @@ final readonly class TestBuilder
     }
 
     /**
+     * Applies to a test case the configuration that build() derives from the
+     * metadata of the test method and from the configuration of the test
+     * runner: process isolation, the preservation of global state, and the
+     * backup of global variables and static properties.
+     *
+     * This is for test cases that are not built here but recreated from their
+     * description, in a worker process of a parallel test run, so that such a
+     * test case is configured exactly as it would be in a sequential run.
+     *
+     * @param class-string<TestCase> $className
+     * @param non-empty-string       $methodName
+     */
+    public function configure(TestCase $test, string $className, string $methodName): void
+    {
+        $this->configureTestCase(
+            $test,
+            $this->shouldTestMethodBeRunInSeparateProcess($className, $methodName),
+            $this->shouldGlobalStateBePreserved($className, $methodName),
+            $this->backupSettings($className, $methodName),
+        );
+    }
+
+    /**
      * @param non-empty-string       $methodName
      * @param class-string<TestCase> $className
      * @param array<ProvidedData>    $data
--- a/src/Runner/Parallel/DataProviderSuiteDescriptor.php
+++ b/src/Runner/Parallel/DataProviderSuiteDescriptor.php
@@ -43,7 +43,11 @@ final readonly class DataProviderSuiteDescriptor extends TestDescriptor
     {
         $members = [];
 
-        foreach ($suite->tests() as $member) {
+        // The suite is iterated, not asked for its tests: iterating applies
+        // the filter that test selection (--filter, --group, …) injected into
+        // it, so that only the tests a sequential run would have run travel to
+        // the worker.
+        foreach ($suite as $member) {
             $members[] = TestDescriptor::from($member, $className);
         }
 
--- a/src/Runner/Parallel/RepeatSuiteDescriptor.php
+++ b/src/Runner/Parallel/RepeatSuiteDescriptor.php
@@ -49,7 +49,11 @@ final readonly class RepeatSuiteDescriptor extends TestDescriptor
     {
         $repetitions = [];
 
-        foreach ($suite->tests() as $repetition) {
+        // The suite is iterated, not asked for its tests: iterating applies
+        // the filter that test selection (--filter, --group, …) injected into
+        // it, so that only the repetitions a sequential run would have run
+        // travel to the worker.
+        foreach ($suite as $repetition) {
             assert($repetition instanceof TestCase);
 
             $repetitions[] = TestCaseDescriptor::fromTestCase($repetition, $className);
--- a/src/Runner/Parallel/RetrySuiteDescriptor.php
+++ b/src/Runner/Parallel/RetrySuiteDescriptor.php
@@ -11,6 +11,7 @@ namespace PHPUnit\Runner\Parallel;
 
 use function assert;
 use function count;
+use function iterator_to_array;
 use PHPUnit\Framework\RetryTestSuite;
 use PHPUnit\Framework\TestCase;
 
@@ -44,7 +45,11 @@ final readonly class RetrySuiteDescriptor extends TestDescriptor
      */
     public static function fromTestSuite(RetryTestSuite $suite, string $className): self
     {
-        $aggregated = $suite->tests();
+        // The suite is iterated, not asked for its tests: iterating applies
+        // the filter that test selection (--filter, --group, …) injected into
+        // it, so that a test the selection excluded does not travel to the
+        // worker.
+        $aggregated = iterator_to_array($suite, false);
 
         assert(count($aggregated) === 1 && $aggregated[0] instanceof TestCase);
 
--- a/src/Runner/Parallel/TestCaseDescriptor.php
+++ b/src/Runner/Parallel/TestCaseDescriptor.php
@@ -14,6 +14,7 @@ use function is_array;
 use function serialize;
 use function sprintf;
 use function unserialize;
+use PHPUnit\Framework\TestBuilder;
 use PHPUnit\Framework\TestCase;
 use Throwable;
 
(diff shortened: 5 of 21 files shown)


sebastianbergmann:
Thank you so much! I will look into this after FrOSCon.

sbuerk:
I have pushed two test commits on top, as a suggestion. They are separate from the two fixes on purpose, so you can drop either of them, or rewrite them in whatever shape you prefer, without touching the fixes themselves. The pull request description is updated accordingly.

**`Add tests for test selection in parallel test runs`** is the PHPT test I mentioned. `tests/end-to-end/parallel/selection/` holds six pairs over one shared fixture: each pair runs the same selection sequentially and with `--parallel=2` and pins the *same* `--EXPECTF--` block, so a `diff` of a pair shows only the `--TEST--` sentence and the added `--parallel=2` line. Covered are `--filter` on a data provider method, `--filter` on a single data set, `--group`, `--exclude-group`, a `--filter` that matches nothing, and `--repeat` with `--filter`. They use `--testdox` so the expectation names the selected tests instead of only counting them.

The fixture is data provider driven deliberately: a plain test method was selected correctly even before the fix, so a fixture of plain test methods would pass either way. With the selection fix reverted, all six parallel files fail while all six sequential ones still pass, and each parallel run executes the whole fixture of 7 tests instead of the 2, 1, 4, 3, 0 and 6 the selection asks for.

**`Add tests for the configuration of test cases recreated in a worker`** answers the Codecov report. The six lines it flags are `TestBuilder::configure()`, missed because that method only ever runs in a worker process, so no end-to-end test reaches it. The commit covers it in the existing `tests/unit/Framework/TestBuilderTest.php` with the fixtures already there: two tests pin concrete settings, and the main one asserts that `configure()` leaves a manually instantiated test case configured exactly as `build()` leaves the one it builds for the same class and method, across every fixture including the data provider one. Making the method a no-op fails all eight cases.

The single line Codecov flags in `ParallelTestRunner` needed nothing: the selection tests of the first commit already execute it, `filter-matching-nothing-in-parallel.phpt` in particular, since a selection that matches nothing is exactly what empties a suite. I measured that per file rather than assuming it. Patch coverage of both files is 100% with the two commits in place.

I did not write a unit test for that guard. `collect()` is private, `ParallelTestRunner` has no unit test, and reaching it would mean reflecting into a runner that owns worker processes — that seemed worse than the end-to-end coverage it already has, but say the word if you would rather have one.

One gap I left open knowingly: no `TestBuilder` fixture carries `#[PreserveGlobalState]`, so `configure()` propagating that attribute is compared but never exercised with a non-null value. Closing it needs a new fixture, which I did not want to add unasked.

Issue #6488, opened 2026-02-03
Allow disabling issue trigger identification for improved performance


sebastianbergmann:
### Problem

When PHPUnit handles deprecation warnings during test execution, the `ErrorHandler` uses `SourceMapper` to determine whether the deprecation was triggered in first-party code or third-party code (self, direct, or indirect). This identification currently requires scanning the configured `<source>` directories to build a list of first-party files.

As reported in #6111 and #6469, this file scanning operation can cause significant performance degradation, especially when:
- The `<source>` include/exclude patterns are overly broad
- The configured source directories contain many files
- Many deprecation warnings are triggered during a test run

### Proposed Solution

Add a new `identifyIssueTrigger` boolean attribute to the `<source>` element in `phpunit.xml`. This attribute defaults to `true` (preserving current behaviour), but can be set to `false` to disable the trigger identification feature entirely.

When disabled:
- PHPUnit skips the expensive file scanning operations in the error handler
- Deprecations are still reported, but with an "unknown" trigger origin instead of being classified as "self", "direct", or "indirect"
- Test execution performance is significantly improved for large codebases

### Configuration Example

```xml
<source identifyIssueTrigger="false">
    <include>
        <directory>src</directory>
    </include>
</source>
```

### Trade-offs

When `identifyIssueTrigger="false"` is set:
- The `ignoreSelfDeprecations`, `ignoreDirectDeprecations`, and `ignoreIndirectDeprecations` options will not function correctly since they depend on trigger identification
- PHPUnit will emit a warning if trigger identification is disabled while any of these ignore options are enabled
- Deprecation reports will show "unknown if issue was triggered in first-party code or third-party code" instead of the specific classification

### Use Cases

This option is intended for users who:
1. Have large codebases where source scanning is expensive
2. Don't need the self/direct/indirect classification of deprecations
3. Want to optimize CI/CD pipeline execution time
4. Are willing to accept less detailed deprecation reporting in exchange for faster test runs

PR #6905, opened 2026-08-16
Do not build the source map for a child process when identifyIssueTrigger is disabled

### LINKED ISSUE #6904: `SourceMap` is built in child process even though `identifyIssueTrigger` is disabled
| Q                   | A
| PHPUnit version     | 12.5.33 (reproduced, although first seen in 11.5.56)
| PHP version         | 8.3.32

#### Summary

* Ref https://github.com/sebastianbergmann/phpunit/issues/6488
* Ref https://github.com/sebastianbergmann/phpunit/commit/debf1b7b0baa8931f559120211cf7aeeecb75a38

The attribute `identifyIssueTrigger` was specifically introduced to prevent extensive filesystem operations (see linked issue/commit). This works fine as long as the test does not run separately, in this case the source map is still built in the child process.

#### Current behavior

While https://github.com/sebastianbergmann/phpunit/pull/6495 improved the situation in so far, that the parent shares the map with the child processes, it does not take into account when no source map was build in the parent.

#### How to reproduce

```php
<?php

use PHPUnit\Framework\Attributes\PreserveGlobalState;

final class IsolationTest extends TestCase {
	public function testPlain(): void {
		$this->assertTrue(true);
	}

	#[RunInSeparateProcess]
	#[PreserveGlobalState(enabled: false)]
	public function testIsolated(): void {
		$this->assertTrue(true);
	}
}
```

with 

```xml
<?xml version="1.0"?>
<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/12.5/phpunit.xsd">
    <testsuite name="repro">
        <directory>tests</directory>
    </testsuite>
    <source identifyIssueTrigger="false">
        <include>
            <directory>src</directory>
        </include>
    </source>
</phpunit>
```

#### Expected behavior

None of the tests should try to built a source map.

#### Test output

(depends on the number of source files of course, here I added random files to show the difference)

```
root@67c7f7d1d4ea:/var/www/html/repo# time vendor/bin/phpunit -c with-source.xml
PHPUnit 12.5.33 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.3.32
Configuration: /var/www/html/repo/with-source.xml

Time: 00:00.785, Memory: 12.00 MB

OK (2 tests, 2 assertions)

real	0m0.980s
user	0m0.162s
sys	0m0.122s

root@67c7f7d1d4ea:/var/www/html/repo# time vendor/bin/phpunit -c with-source.xml --filter testPlain
PHPUnit 12.5.33 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.3.32
Configuration: /var/www/html/repo/with-source.xml

Time: 00:00, Memory: 6.00 MB

OK (1 test, 1 assertion)

real	0m0.164s
user	0m0.071s
sys	0m0.028s
```

lazerg:
With `<source identifyIssueTrigger="false">`, PHPUnit still builds the source map before running any `#[RunInSeparateProcess]` test, even though neither the parent nor the child ever needs it in that mode. `identifyIssueTrigger` was introduced to avoid exactly this filesystem walk (#6488), but `SeparateProcessTestRunner::sourceMapFileForChildProcess()` only checked `Source::notEmpty()`, not `Source::identifyIssueTrigger()`, unlike `ErrorHandler::instance()`, which already gates on both.

This adds the same `identifyIssueTrigger()` check to `sourceMapFileForChildProcess()`, so the map is skipped whenever issue-trigger identification is disabled, isolated tests included.

Verified against the full unit and end-to-end suites, both green.

I used an AI coding assistant (Claude Code) to investigate this issue, write the fix, and run the tests. I reviewed the change and the results before opening this PR.

Fixes #6904

### DIFF
--- a/src/Framework/TestRunner/SeparateProcessTestRunner.php
+++ b/src/Framework/TestRunner/SeparateProcessTestRunner.php
@@ -172,7 +172,7 @@ final class SeparateProcessTestRunner
             return self::$sourceMapFile;
         }
 
-        if (!ConfigurationRegistry::get()->source()->notEmpty()) {
+        if (!ConfigurationRegistry::get()->source()->notEmpty() || !ConfigurationRegistry::get()->source()->identifyIssueTrigger()) {
             self::$sourceMapFile = '';
 
             return self::$sourceMapFile;


sebastianbergmann:
Thank you for looking into #6904. The problem described there is real, but I cannot merge this patch: it fixes one configuration by regressing another.

The source map that `SeparateProcessTestRunner` writes for a child process is not only used for the identification of issue triggers. It is also the pre-populated cache for the code coverage filter: `method.tpl` calls `CodeCoverage::instance()->init(..., true)`, which calls `CodeCoverageFilterRegistry::init()`, which calls `SourceMapper::mapForCodeCoverage()`, which calls `SourceMapper::map()`. `SourceMapper::loadFrom()` in the child exists so that this call hits the map the parent already built instead of walking the filesystem again. That is what #6495 was for.

Gating the map on `identifyIssueTrigger()` therefore means that with `identifyIssueTrigger="false"` and code coverage enabled, every child process rebuilds the map from scratch.

The mismatch is even clearer on the branches this would be merged up into. In PHPUnit 13, `SourceFilter` no longer uses `SourceMapper`; it uses `FileFilterMapper`, which does not touch the filesystem. There, the source map's only remaining consumer is the code coverage filter, so conditioning it on `identifyIssueTrigger` conditions a coverage-only artifact on a flag that has nothing to do with coverage.

The condition has to be about what the child actually needs:

```php
if (!ConfigurationRegistry::get()->source()->identifyIssueTrigger() &&
    !CodeCoverage::instance()->isActive()) {
    self::$sourceMapFile = '';

    return self::$sourceMapFile;
}
```

I have implemented this, with end-to-end tests for both branches of the condition.

One more thing for future pull requests: this one changes runtime behaviour and adds no test. The Codecov comment saying that all modified lines are covered does not mean anything here, because the only modified line was already covered before.

Closing in favour of the fix described above. Thank you again for the report and the investigation.

PR #6919, opened 2026-08-24
Test Impact Analysis

### LINKED ISSUE #6897: Test Impact Analysis
## Summary

PHPUnit should be able to work out which tests can possibly be affected by a change to the code under test, and offer to run only those tests.

The information needed for this already exists inside PHPUnit: when code coverage is collected, PHPUnit knows exactly which source files each individual test executed. Today that knowledge is used only to produce coverage reports and is thrown away at the end of the run. Kept around, it answers a question developers ask constantly: "I changed this file, what do I need to run?"

## Motivation

The larger a test suite becomes, the less it is actually used sometimes.

A suite that takes ten minutes is not run after every change; it is run before a commit, or before a push, or not at all until CI complains. The feedback loop that unit testing exists to provide gets stretched until it stops being a loop. Developers work around this by hand: running a single test class, guessing at a `--filter` pattern, remembering which directory "probably" covers the thing they touched. Those guesses are quick, and they are wrong often enough to matter and miss the test three layers away that also exercised the changed line.

PHPUnit is in a better position to answer this than the developer is, because it has observed what every test actually executed rather than what someone assumes it executed.

## Where this idea comes from

The idea is old and well studied, under the name regression test selection. Gregg Rothermel and Mary Jean Harrold formalised it in the [1990s](https://dl.acm.org/doi/10.1145/248233.248262) and gave the field the property that still matters most: a selection technique is called safe when it selects every test that could expose a fault introduced by the change. Their techniques achieved that by comparing control flow graphs before and after a change. Everything since has been a trade between how safe a technique is and how cheap it is to run.

The name "test impact analysis" comes from industry tooling rather than from the literature. Microsoft shipped a feature under that name in Visual Studio 2010, and Paul Hammant's 2017 article [The Rise of Test Impact Analysis](https://martinfowler.com/articles/rise-test-impact-analysis.html) carried it into wider use.

What made it practical was lowering the resolution. Ekstazi ([ISSTA 2015](https://dl.acm.org/doi/10.1145/2771783.2771784)) showed that tracking dependencies at file granularity, observed dynamically while tests run, is coarse enough to be cheap to collect and precise enough to be useful, which is the granularity this proposal assumes. A different branch of the field, beginning with Meta's [predictive test selection](https://engineering.fb.com/2018/11/21/developer-tools/predictive-test-selection/) in 2018, drops dependency tracking entirely and predicts from historical run data using machine learning. That is a different feature and is not what is proposed here.

Most ecosystems now have something in this space, whether built into the test runner, the build tool, or the CI vendor.

## Proposed Feature

Two capabilities, both opt-in:

### Recording

During a run in which code coverage is collected, PHPUnit records which source files each test depends on and stores that alongside the other information it already stores between runs. This costs nothing beyond the coverage collection that is already happening.

### Selection

On subsequent runs, PHPUnit can compare the current state of the source files against what was recorded, determine which tests can possibly be affected, and run only those.

From the command line this might look like:

```
$ phpunit --only-impacted
```

with the set of changed files determined by PHPUnit itself, or supplied explicitly so that it composes with whatever the developer already uses to answer "what changed":

```
$ phpunit --impacted-by src/Service/Invoice.php
$ git diff --name-only | phpunit --impacted-by-file -
```

And a read-only form that answers the question without running anything:

```
$ phpunit --list-impacted-tests --impacted-by src/Service/Invoice.php
```

### Keeping the feedback loop short while working

A developer editing one class re-runs the suite after every save. Instead of ten minutes, they wait for the handful of tests that touch what they just changed. The suite becomes something used continuously rather than something run at checkpoints.

### Pre-commit and pre-push checks

A hook that runs the full suite gets disabled within a week. A hook that runs the tests affected by the staged changes stays enabled, because it finishes in seconds and still catches the mistakes that matter.

### Fast signal on pull requests

CI runs the affected tests first and reports within a minute, then runs the full suite. Reviewers get an early answer without giving up the guarantee that the complete suite ran before merge.

### Understanding the impact before making a change

Before touching a class in unfamiliar code, a developer asks which tests cover it. Today the answer requires generating a coverage report and reading it backwards. This turns it into one command, and the answer is derived from what the tests really executed rather than from `#[CoversClass]` attributes that may be incomplete or absent.

### Spotting untested changes

A change to a source file that maps to zero tests is worth knowing about. It means either that the file is untested, or that whatever exercises it does so in a way coverage did not observe. Both are useful signals during review.

### Working productively in large or inherited suites

In a codebase where the full suite takes long enough that nobody runs it locally, this restores the ability to test at all during development, without requiring the suite to be reorganised or split first.

## What users must be able to rely on

This feature trades completeness for speed, and it is only worth having if that trade is made honestly and visibly.

- It is never enabled by default. Selecting a subset of tests is a decision the user makes deliberately, per run or per configuration.

- The report never claims a test passed when it was not run. Selecting fewer tests must change the number of tests reported, not fabricate results for the rest. A run of 42 selected tests reports 42 tests and says plainly how many were left out and why. Skipping the rest is not the same as passing them, and the output must not blur the two.

- When in doubt, the test runs. Anything PHPUnit does not have reliable information about is treated as affected: tests it has never seen before, tests whose own file changed, tests that did not pass on the previous run, and tests that other tests depend on.

- When the ground shifts, everything runs. A change to the PHPUnit configuration, to the project's dependencies, or to the PHP version invalidates what was recorded, and the full suite runs.

- The selection is explainable. Users can ask which tests would be selected, and why, without executing anything.

- CI is not the target. The intended use is shortening the local feedback loop and getting early signal. It does not replace running the full suite before a change is merged or released, and the documentation should say so directly.

## Known limitations

Impact analysis is based on observed execution sees PHP code. It does not see everything a test depends on: fixture files, templates, schema definitions, environment configuration, data files, code reached only through reflection, or a test that asserts something is absent. A change to any of those may affect tests that this feature will not select.

This is a real limitation, not a bug to be fixed later, and users need to understand it before they rely on the feature. It also argues for an extension point: PHPUnit itself cannot know that a change to a template or a migration should re-run a particular group of tests, but the frameworks and tools built on top of PHPUnit do, and they should be able to contribute that knowledge.

## Non-Goals

- Replaying previous results. Reporting cached outcomes for tests that were not executed would make the suite faster in exactly the way that matters least and misleading in the way that matters most.

- Version control integration. PHPUnit should determine what changed by looking at the files it already knows about, and should accept an explicit list of changed files from the user. It should not need to understand branches, rebases, or any particular VCS.

- Framework-specific knowledge. Rules about templates, routes, migrations or asset pipelines belong in extensions, not in PHPUnit.

## Relationship to work already underway

Two features currently in development change what this one would have to build, and one of them changes whether it is worth building at all.

### Caching what a test file contains (#6863, `feature/test-index`)

Selecting fewer tests does not by itself make a run fast. Before PHPUnit can decide that a test has nothing to contribute, it has to load the file that test lives in, and loading test files is most of the cost of a narrow run. That work measured a `--group` run that selects 98 of a suite's tests at 981 ms without the index and 242 ms with it — the saving is almost entirely in not loading files.

An impact-selected run has exactly that shape: a few dozen tests out of many hundreds. Without the index, the time saved by not running the unaffected tests would be spent discovering them anyway, and the feature would underdeliver on the one thing it promises. These two belong together: "affected by what changed" is a third way of establishing that a file has nothing to contribute, alongside the group and name criteria that work already handles.

It also settles several things this feature would otherwise have to invent, and settles them the same way:

- Deciding whether remembered knowledge is still valid: An index entry is used only while every source file it was derived from still hashes the same, parent classes and traits included. Impact data needs precisely that rule on the test side, and should use the same mechanism rather than a second one.
- Hashing contents rather than trusting timestamps, for reasons that apply here unchanged.
- Failing open: A file that cannot be read or loaded is never remembered, so a later run loads it again rather than skipping past it.
- Being explicit that not loading a file is not the same as skipping a test. That work states the distinction outright: a skipped test is part of the suite and is reported; a test in a file that was not loaded is neither. That is the same honesty requirement stated above, already articulated.

### Parallel test execution (#6784, `feature/parallel-test-execution`)

Workers ship their results home to the parent process, which merges each unit's coverage into its own and remains the single source of truth for output, logging, results and coverage. Per-test attribution survives that merge, so recording the data this feature needs works under `--parallel` without anything extra.

- Selection size should inform how many workers are started: Starting ten workers to run six selected tests loses more than it saves. The two features are both about making short runs short, and they have to agree on that.
- Work is distributed one test class at a time: When only some tests of a class are affected, the unit sent to a worker has to be the selected subset, not the class.
- Tests that must run in the main process (those marked as not parallelisable, those requiring process isolation, those whose data cannot be serialised, PHPT files declaring a conflict) must still have their dependencies recorded, or they will silently become tests that are never selected again.
- Scheduling: That work notes that recorded durations could seed a longest-first ordering and are not used for it yet. Durations and impact data are both per-test knowledge kept between runs; if this feature lands, they should feed the same scheduler rather than two.

Taken together the three describe one coherent end state: the index avoids loading what cannot matter, impact analysis decides what does matter, and parallel execution runs what is left across all available cores.

### Is the dependency data PHPUnit can record actually correct enough?

When a test declares `#[CoversClass]` or `#[UsesClass]`, PHPUnit deliberately narrows the coverage it records to the declared targets. That is the right behaviour for a coverage report and the wrong behaviour for impact analysis: a test that exercises a helper class but does not declare it would have no recorded dependency on that helper, and would therefore not be selected when the helper changes. Silently.

So recording would have to capture what the test executed, independently of what it declares it covers — meaning the impact data and the coverage report would disagree with each other by design. Is that acceptable? If not, this feature cannot be built on observed execution at all, and everything below is moot.

### Is requiring a coverage driver a price users will actually pay?

Recording needs pcov or Xdebug, and a recording run is meaningfully slower than a normal run. Users without a driver installed locally get nothing. How many are in that position, and does the feature still pay for itself once the cost of periodically re-recording is counted?

### What defines the set of files this feature watches?

`<source>` is the obvious boundary, but it describes first-party PHP code for the purpose of coverage, not "everything a test might depend on". Changes outside it would be invisible. Is `<source>` the right boundary, does this need its own configuration, or is the honest answer that anything outside `<source>` forces a full run?

### What is the contract when the selection is empty?

A change that maps to no tests currently has no defined outcome. Is that a successful run, a warning, or a failure? The same question applies when the recorded data is stale or missing: does PHPUnit run everything, or refuse and tell the user to re-record? Both are defensible; picking one late would be a BC problem.

### How large is the recorded data, and where does it live?

A per-test-to-per-file map for a suite of any size is substantially larger than what PHPUnit keeps between runs today, and larger than a test index entry. Does it sit alongside that data in the cache directory, or does its size argue for something else? Is it per-machine, or something a team would want to share? If shared, it becomes a format with compatibility obligations.

### What happens the first time it selects wrong?

A false negative here does not look like a bug — it looks like a green suite. The failure mode is "PHPUnit told me my tests passed and they did not", which is the single worst bug report this project can receive, and it will arrive regardless of how the feature is documented. Is there a formulation of the output that makes the incompleteness impossible to miss? 

### Which existing features must work with it on day one?

`#[Depends]`, data providers, process isolation, `--order-by`, groups and PHPT tests. Some of these are straightforward, some are not.

The harder version of this question is one of sequencing rather than compatibility: this feature needs the test index to deliver what it promises, and should agree with parallel execution about how short runs are made short. Does it wait for both, or is there a first version that stands on its own?

### When only some tests of a class are affected, is the class still the right unit of work to hand to a parallel worker?

sebastianbergmann:
The changes proposed here implement [test impact analysis](https://github.com/sebastianbergmann/phpunit/issues/6897): PHPUnit records which source files each test depends on, and can then run only the tests a change can affect.

Closes #6897.

### Recording

`--record-test-impact-data`, or `recordTestImpactData="true"`, records while the tests run which source files each test depends on. The data goes into its own file in the cache directory, separate from the test run history.

### Querying

`--list-tests-that-depend-on src/Money.php` queries what was recorded, without running a single test:

```
Recorded from what the tests executed.

Tests that depend on /…/src/Money.php as it is now:
 - Example\InvoiceTest::testTotalIsTheSumOfItsItems
 - Example\SumTest::testAdds#0

Tests that depend on an earlier version of /…/src/Money.php:
 - Example\FormatterTest::testFormatsAmount
```

The two lists are kept apart on purpose: the first describes the code as it is now, the second describes tests recorded against a version of that file which no longer exists.

### Selecting

`--only-impacted` runs only the tests a change can affect:

```
Impact:        26 of 177 tests can be affected by what changed; 151 tests are not run

OK (26 tests, 80 assertions)
```

The line says how many were left out, and the summary reports the tests that actually ran. Nothing pretends the other 151 passed.

PHPUnit works out what changed by comparing what is there now against what it recorded. `--impacted-by src/Money.php --impacted-by src/Service` tells it instead and it implies `--only-impacted`. What you name is the change set: the recorded hashes are not consulted at all. That matters on a fresh checkout, where comparing against the recording reports everything that changed since, while `git diff --name-only` reports what you actually did.

Because naming paths one at a time does not compose with the tool that knows what changed, `--impacted-by-file` reads a list, one path per line, `-` being standard input:

```
$ git diff --name-only | phpunit --impacted-by-file -
```

An empty list is an answer rather than a missing one: nothing changed, so nothing that depends on code runs. A list that cannot be read is not an empty list, and PHPUnit stops rather than guessing.

## Two ways of maintaining the data

Observation needs code coverage data collection and works no matter what a project declares. `--derive-test-impact-data-from-coverage-targets` is the alternative for projects that already take coverage targets seriously: it works out what each test depends on from the `#[CoversClass]` and `#[UsesClass]` attributes it declares, without code coverage data collection, and without the tests having been run even once. `beStrictAboutCoverageMetadata` is what makes it sound: under it, a test that executes code it does not declare is already marked risky, so declarations are a superset of execution. PHPUnit warns when the mode is used without the strict check, because then nothing has ever verified the declarations.

## Measurements

All from one real project of 177 tests that enables both `requireCoverageMetadata` and `beStrictAboutCoverageMetadata`.

| | wall clock |
| --- | --- |
| the suite, no coverage driver | 5.44s |
| recording derived from declarations | 5.47s |
| a coverage run, no recording | 33.5s |
| the same coverage run, recording | 87.8s |

Recording by observation costs 2.6× on a run that was already collecting coverage, and all of it lands in the tests declaring `#[CoversNothing]`, which have to be collected for and are usually the slow ones. Deriving from declarations is free.

For 165 of 165 tests that declare targets, the declarations covered everything the test was observed to execute — no exceptions beyond `#[CoversNothing]`. The price is a coarser selection:

| changed file | from what tests executed | from what tests declare |
| --- | --- | --- |
| the matrix class | 129 of 177 | 138 |
| the tuple class | 123 | 164 |
| the world class | 16 | 25 |
| an output mapper | 6 | 6 |

Declaration is never below observation on any file, and costs between nothing and half as many tests again. Five of the 177 tests declare no targets, are never recorded in that mode, and therefore always run: that is the floor.

Selecting after editing two source files gave 26 of 177 whether the change set was worked out from the hashes or piped in from `git diff --name-only`. With nothing edited, `--only-impacted` selected 0 of 177 in 0.14 seconds. The data is about 150 to 350 bytes per test — 27 KiB for this project.

## What you can rely on

Everything the analysis has no reliable information about causes the test to run:

- a test that was never recorded, including a test declaring `#[CoversNothing]` and a PHPT test
- a test whose own code changed, or whose data provider changed
- a test that did not pass when it was last run
- a test another selected test depends on through `#[Depends]`

And these cause every test to run:

- nothing has been recorded yet, or what was recorded is no longer usable
- a source file changed that no test is recorded as depending on
- a source file is there that was not there, or was not first-party code, when the recording was made
- a path that was named, with `--impacted-by` or in a list, is not among the files that were recorded
- the configuration file changed, the set of first-party code changed, or `composer.lock` changed

The last one matters more than it sounds: what was recorded describes one state of the world, and when that state is not what it was, the honest answer is not that some entries are stale but that none of them can be trusted.

`--only-impacted` refuses to run at all without a cache directory, or when the test run history is turned off, rather than quietly being less careful.

Nothing is kept forever, either. A test run forgets every entry it did not record itself, but only when that run ran every test there is. A run that filtered, that selected by impact, or that stopped at the first failure recorded nothing for the tests it did not run, and nothing tells those apart from tests that are gone, so such a run adds to what is known and takes nothing away.

## `#[UsesFixture]`

Neither observation nor declaration can see a data file: reading `invoices.csv` is not executing code, and a coverage target names code units. A new attribute closes that:

```php
#[UsesFixture('../fixtures/invoices.csv')]
public static function provideInvoices(): array
```

It goes on a test class, a test method, or a data provider method: the last is the one that earns its keep, because a provider shared by many tests declares the file once and every test using it inherits the dependency. Directories work too, watched as a whole, so adding a file to one counts as a change. A path that does not exist produces a warning and is ignored, the same treatment a coverage target that cannot be used gets. Unlike coverage targets, nothing can ever verify this attribute; it is a promise the project makes to itself.

## Decisions

- **Recording is never automatic**, not even when coverage is already being collected. A suite whose slow tests declare `#[CoversNothing]` would otherwise get 2.6× slower unasked.
- **Tests declaring `#[CoversNothing]` are collected for** in observation mode, with the targets bypassed so nothing reaches the coverage report. The alternative — treating them as affected by everything — would put a floor under every selected run.
- **Impact data is never filtered using coverage targets; the report always is.** The two disagree by design. Before this, what the coverage report kept was missing 75% and 94% of the dependencies tests really have on the two suites I measured, because the report is narrowed to what a test declares. That is why the data is collected before the narrowing, which needed php-code-coverage 14.4.0.
- **The data is per-machine and must not be committed.** It holds absolute paths and is discarded on a PHPUnit or PHP version change.
- **A test's own files are recorded in the impact data** rather than checked against the test index at selection time. The index is written at a different moment, so a run that refreshed the index without recording impact data would report a changed test as unchanged.
- **The assumption hash covers the `<source>` declarations, not the expanded file list.** Hashing the list would throw everything away each time a class is added to `src/`, and buys nothing: a new file can only be reached if an existing one changed.
- **`--only-impacted`, `--impacted-by` and `--impacted-by-file` are command line options only.** A switch that makes every run partial does not belong in a file a project commits.


sebastianbergmann:
### What is recorded for a test today

Three things:

1. the source files the test executed, as observed through the code coverage driver and restricted to what `<source>` says is first-party code (`CodeCoverage::recordTestImpactDataFor()`);
2. the paths the test declares with `#[UsesFixture]`;
3. the files the test class itself is made of: its own file, its parent classes, its traits, and the classes its data providers are methods of (`TestFiles::of()`).

### The gap

Code that a test executes, that is not first-party code under `<source>`, and that is not a parent class, trait or data-provider class of the test, is in none of those three. The obvious example is a test helper:

```
src/Calculator.php          <- <source>
tests/CalculatorTest.php
tests/Support/Helper.php    <- executed by the test, in neither <source> nor the test class
```

```php
public function testAdds(): void
{
    $this->assertSame(Helper::expectedSum(), (new Calculator)->add(1, 2));
}
```

What gets recorded for that test is `src/Calculator.php` and `tests/CalculatorTest.php`. `tests/Support/Helper.php` is not recorded, so changing it changes nothing that was recorded:

```
$ # edit Helper.php so that the test would now fail
$ phpunit --only-impacted
Impact:        0 of 1 tests can be affected by what changed; 1 test is not run

No tests executed!            # exit code 0

$ phpunit
FAILURES!
Failed asserting that 3 is identical to 4.
```

That is the bad shape: the suite reports success while a change that breaks it goes unrun. It is not a stale-data problem; the recording is up to date and correct about everything it covers.

### What this is not

- Not the bootstrap gap: That one was the same finding's other half and is fixed: `Assumptions` now hashes the bootstrap scripts, so changing one discards the whole recording. That mechanism does not generalise to helpers and discarding everything whenever any test file changes would defeat the point.
- Not a problem for `--impacted-by` / `--impacted-by-file`: Naming the helper falls back to running everything, because the path is not among the files that were recorded:

  ```
  Impact:        every test is run: .../tests/Support/Helper.php is not among the files that were recorded
  ```

  Only the automatic mode (`--only-impacted`, which works out for itself what changed) is silently wrong.

### Workaround

Adding the helper directory to `<source>` closes the gap for that project:

```
Impact:        1 of 1 tests can be affected by what changed; 0 tests are not run
```

The cost is that the helpers then count as first-party code for code coverage too, and show up in coverage reports.

### Directions considered, none chosen (yet)

- Record every file the test loaded: Complete and mechanical, but a test would then depend on the framework, vendor code and its whole autoload graph: very large recordings, and selections that rarely narrow.
- Record from a declared test-code root: A new configuration element, or reusing the test suite directories. Bounded and predictable, but it needs new configuration and a rule for what belongs in it.
- Extend `#[UsesFixture]`, or add something like it, to cover code: Cheapest to build and the most precise, but opt-in per test, so the default behaviour stays as it is today; which is exactly what this note is about.

The first two change what "what a test depends on" means, so they want a deliberate decision rather than being settled on the way past.

sebastianbergmann:
When `--only-impacted` runs more tests than you expected, the run tells you how many tests were selected, but not why any particular one is among them:

```
Impact:  44 of 2849 tests can be affected by what changed; 2805 tests are not run
```

Being affected by a change is only one of the reasons a test is run. A test is also run when nothing is known about it, and that number can dominate the selection without anything saying so. Read as it stands, the line above claims something about 44 tests that is only true of a few of them.

The new `--explain-impacted` CLI I just implemented reports which tests can be affected by what changed and why each of them can be, and it does not run anything:

```
$ phpunit --explain-impacted

Recorded from what the tests executed.

5 of 2849 tests can be affected by what changed.

2 tests depend on something that changed:
 - App\Tests\Api\ThemeControllerTest::testGetTheme
   src/Http/Controllers/ThemeController.php
 - App\Tests\Api\ThemeControllerTest::testGetThemeNotFound
   src/Http/Controllers/ThemeController.php

3 tests have never been recorded:
 - App\Tests\Models\RegionTest::testHasFallback
 - App\Tests\Models\RegionTest::testHasNoFallback
 - App\Tests\Services\ExportServiceTest::testExports
```

A test is reported under one of these:

- **depends on something that changed:** the file it executed, or that it declares it uses, is not what it was. This is the reason you are looking for, and the file that caused it is named.
- **has never been recorded:** nothing is known about the test. It is new, or it was skipped or marked incomplete when it last ran, and a test that executes nothing has nothing to record. Anything unknown is run rather than silently left out.
- **did not pass when it was last run:** what was recorded describes a run that did not get through the test.
- **is depended upon by another test that is run:** it cannot be left out without the test that depends on it being skipped.
- **is not a test method and can never be recorded:** PHPT tests, for example.

That second reason is usually the surprising one. Tests that are always skipped never record anything, so they stay in the selection no matter what you change. They cost almost nothing to run, but until now there was no way to see that this is what was happening.

The option reports exactly what a run with `--only-impacted` would select, so you can use it to understand a selection before spending the time on it. It also takes the paths you already know changed:

```
$ git diff --name-only | phpunit --explain-impacted --impacted-by-file -
```

sebastianbergmann:
This should be merged after #6784. It merges cleanly into `main` on its own, but once the parallel test execution has been merged there will be two functional problems that a textual conflict resolution does not solve:

### 1. The test selection must reach the parallel test runner

This branch passes the tests selected by `--only-impacted` from `Application` into `TestRunner::run()`, which hands them to `TestSuiteFilterProcessor::process()`.

The parallel branch replaces the body of `TestRunner::run()` with `TestRunnerLifecycle`, which calls `TestSuiteFilterProcessor::process()` without a selection, and adds `ParallelTestRunner`, which uses the same lifecycle and does not accept a selection at all. `Application` chooses between the two runners based on `numberOfParallelWorkers()`.

If the conflict is resolved by simply keeping both sides, `--only-impacted --parallel=N` runs every test: the selection is computed and printed, but never applied. The selection has to be threaded through `TestRunnerLifecycle::run()` into `TestSuiteFilterProcessor::process()`, and `ParallelTestRunner::run()` needs to accept and forward it just like `TestRunner::run()` does. With that in place, selection under `--parallel=2` matches the sequential result on the end-to-end fixtures of this branch.

### 2. The test impact data recorded inside a worker must reach the parent

Test impact data is recorded per test in `CodeCoverage::testImpactData()` of the process that runs the test. For tests that run in a separate process, this branch ships what was recorded as `testImpactData` in the child's result envelope (`method.tpl`) and records it in the parent in `ChildProcessResultProcessor`.

The parallel branch's workers do not do this. `src/Runner/Parallel/templates/worker.tpl` ships only `codeCoverage`, `events`, and `passedTests` in the envelope of a unit, and `ResultAggregator` only calls `ChildProcessResultEnvelope::mergeCodeCoverage()`. Everything a worker records is lost. On the `tests/end-to-end/cli/test-impact-data` fixture, a sequential run records three tests, while the same run with `--parallel=2` records only `IsolatedCalculatorTest::testAddsInAnotherProcess`, which survives because it goes through the separate-process path. Since a run that records overwrites the cache, a parallel run also destroys what a previous sequential run recorded, so subsequent `--only-impacted` runs select more tests than necessary.

To fix this:

* `worker.tpl` must include what was recorded (`CodeCoverage::instance()->testImpactData()->recorded()`) in the unit's result envelope, in the same way `method.tpl` does.
* The recording loop that this branch adds to `ChildProcessResultProcessor` should move into `ChildProcessResultEnvelope::mergeCodeCoverage()` (or a sibling method), so that both the separate-process path and `ResultAggregator` record it into the parent's `CodeCoverage::testImpactData()`.
* `TestImpactData` needs a way to forget what was recorded (the interface currently only has `record()` and `recorded()`). A persistent worker runs many units, and `worker.tpl` already clears the code coverage and calls `PassedTests::reset()` after shipping a unit's envelope. Without a reset, every envelope ships the data of all units the worker has run so far.
* Until this is done, `--record-test-impact-data` combined with `--parallel` should be refused, like the other invalid option combinations this branch already rejects, so that a parallel run cannot silently overwrite the cache with incomplete data.

Issue #6959, opened 2026-09-04
Implement a way to test order of method calls


josep11:
There is the need to assert that method on mocks are called in a specific order, which is not currently supported by PHPUnit. The golden example is, for instance, asserting that a transaction was started before any method calls.

This was already mentioned in this [issue](https://github.com/sebastianbergmann/phpunit/issues/5357).

For people wanting to do this, there is this [workaround](https://zaantar.eu/2020/07/verifying-the-order-of-calls-across-different-mock-objects-in-phpunit/?lang=en), which is a bit nasty, in my opinion, as compared to a natively supported PHPUnit method.

## Reproduction

```php
final readonly class StartTransactionMiddleware implements MiddlewareInterface
{
    public function __construct(private Connection $connection)
    {
    }

    public function handle(Envelope $envelope, StackInterface $stack): Envelope
    {
        $this->connection->beginTransaction();

        $envelope = $stack->next()->handle($envelope, $stack);

        $this->connection->commit();

        return $envelope;
    }
}
```

## Expected behaviour

Suggestion `InvokedCountOrder::assertOrder` (without being familiar with PHPUnit codebase).
Receives `InvocationStubber` array in the expected order. Cannot be used together with `InvokedCount` on the same method.

```php
public function testStartsTransactionBeforeHandlingTheNextMiddleware(): void
 {
   $invokedCountOrder = new InvokedCountOrder();
    $connection = $this->createMock(Connection::class)
        ->expects($invoker)
        ->method('beginTransaction');

    $stack = $this->createMock(StackInterface::class);
    
    // Receives invocation stubber array in the expected order.
    $invokedCountOrder->assertOrder([
        $connection->method('beginTransaction'),
        $stack->method('next'),
        $connection->method('commit'),
    ]);


    $startTransactionMiddleware = new StartTransactionMiddleware($connection );
    $startTransactionMiddleware ->handle(new Envelope(new stdClass()), $stack);
}
```

sebastianbergmann:
Superseded by #6960.

Issue #6537, opened 2026-03-13
Soft-deprecate `id()` and `after()` for mock object expectations


sebastianbergmann:
## What these methods do

`id(string $id)` assigns a string identifier to an expectation configured on a mock object.

`after(string $id)` makes an expectation contingent on another expectation (identified by its ID) having already been invoked at least once before this one will match.

Together they allow expressing invocation ordering constraints between methods on the same mock object.

```php
$mock = $this->createMock(Foo::class);

$mock
    ->expects($this->once())
    ->method('foo')
    ->id('foo-expectation');

$mock
    ->expects($this->once())
    ->method('bar')
    ->after('foo-expectation');

$mock->foo();
$mock->bar();
```

Here, the expectation on `bar()` will not match until `foo()` has been invoked at least once.
If `bar()` is called before `foo()`, it will not match.

## Why this should be removed

The `id()` / `after()` mechanism is obscure. A test that needs `id()` / `after()` to enforce call order is likely testing an implementation detail rather than observable behaviour.

If the ordering of `foo()` and `bar()` is significant then `withParameterSetsInOrder()` should be used instead.

## Plan

- **PHPUnit 13:** [Soft-deprecate](https://phpunit.de/backward-compatibility.html#soft-deprecation) `id()` and `after()`
- **PHPUnit 14:** [Hard-deprecate](https://phpunit.de/backward-compatibility.html#hard-deprecation) `id()` and `after()`, calling them now triggers a deprecation
- **PHPUnit 15:** Remove `id()` and `after()` from the `InvocationMocker` interface, calling them now results in an error

Issue #5357, opened 2023-05-09
Implement a way to specify expected method call order across multiple mocks


elnur:
I have no idea why PHPUnit still doesn't have support for this feature given how old and mature the project is. Every other test library I've used has this feature:

- [jMock has it](http://jmock.org/sequences.html),
- [Mockito has it](https://javadoc.io/doc/org.mockito/mockito-core/latest/org/mockito/Mockito.html#6),
- [Mockery has it](https://docs.mockery.io/en/latest/reference/expectations.html#expectation-declaration-utilities),
- etc...

Even before [`at()` got deprecated and removed](https://github.com/sebastianbergmann/phpunit/issues/4297), it's biggest problem was the fact that it only worked for a single mock.

Putting aside purist theoretical mind-numbing discussions that we should not be testing the order of method calls, it is obvious that there's a need for such a feature. Otherwise other test libraries wouldn't have implemented it.

I personally have been running into this problem for years. I hoped PHPUnit would eventually support it, but it never happened. So now I'm taking initiative and trying to push this forward.

So let's discuss the design of this feature and then implement it.

I'd say the jMock's way is kinda ugly. The Mockito's way is nicer, but thanks to Java it relies on the fact that mocks in Mockito are typed in such a way that you statically call methods being tested, which wouldn't work in PHPUnit. So maybe go with the Mockery's way? Or come up with something else/better?

Thoughts? Ideas?

sebastianbergmann:
I do not think that relying on call order makes sense.

elnur:
Care to elaborate, @sebastianbergmann?

reecefowell:
I depend on the features you are removing (both `withConsecutive(...)` and `at(...)`, but I do not see any recommended alternative solutions to solving the problem that these now deprecated functions solved. \\

Just removing functionality we depend on without further elaboration and no provided alternative solutions (that I could find) is extremely unhelpful.

How do you anticipate us to work around this? 
What are the solutions you are promoting in favour of these deprecated functions?

Xachman:
@reecefowell I agree with this. This seems rather dogmatic @sebastianbergmann. It also creates confusion and time lost. I can appreciate the reasoning but even a bad test is better than no test.

siamskoi:
You can use something like this:

```php
        $matcher = $this->exactly(3);

        $this->dumper->expects($matcher)->method('dump')
            ->willReturnCallback(function ($value) use ($matcher) {
[...]

PR #6960, opened 2026-09-05
Record the order in which methods of mock objects are invoked

sebastianbergmann:
This pull request adds `PHPUnit\Framework\MockObject\InvocationJournal`, `TestCase::createInvocationJournal()`, and a new `recordIn()` method for configuring mock object expectations:

```php
$journal = $this->createInvocationJournal();

$connection->expects($this->once())->method('beginTransaction')->recordIn($journal);
$stack->expects($this->once())->method('next')->recordIn($journal)->willReturn($next);
$connection->expects($this->once())->method('commit')->recordIn($journal);

// ...

$this->assertSame(
    [
        'Connection::beginTransaction()',
        'Stack::next()',
        'Connection::commit()',
    ],
    $journal->invocations(),
);
```

(The default label is built from the fully-qualified name of the doubled type; the examples in this description use unqualified names for readability.)

An `InvocationJournal` records, in the order in which they happen, the invocations of methods that were configured with `recordIn()`. It does not assert anything by itself: the assertion on the recorded order is an ordinary assertion, written by the developer, in the test.

This deliberately makes PHPUnit a "recording device" and not a "protocol assertion engine". The reasoning for that distinction is spelled out below.

## Motivation

[#6959](https://github.com/sebastianbergmann/phpunit/issues/6959), and [#5357](https://github.com/sebastianbergmann/phpunit/issues/5357) before it, ask for a way to verify the order in which methods are invoked, across multiple test doubles.

I remain of the opinion that most tests that want this are testing an implementation detail. But two things have changed my mind about leaving the gap open.

### The migration path documented for `id()` and `after()` does not work

[#6537](https://github.com/sebastianbergmann/phpunit/issues/6537) soft-deprecates `id()` and `after()`. It says:

> If the ordering of `foo()` and `bar()` is significant then `withParameterSetsInOrder()` should be used instead.

That advice does not hold: `withParameterSetsInOrder()` orders the parameter sets of one expectation, it says nothing about which method was invoked. Even when the method name rule is a constraint that spans both methods, the wrong order passes silently:

```php
$connection->expects($this->exactly(2))
    ->method($this->logicalOr($this->equalTo('beginTransaction'), $this->equalTo('commit')))
    ->withParameterSetsInOrder([], []);

$connection->commit();
$connection->beginTransaction();

// OK (1 test, 1 assertion)
```

So `id()` and `after()` are on their way out with a documented replacement that does not replace them. Either that has to be stated plainly, or something has to fill the gap. This pull request fills the gap. The wording in #6537 and in `DEPRECATIONS.md` will be corrected separately.

### There is a legitimate, narrow use case

There are classes whose observable contract is an order. For those, the order is not an implementation detail, but the specification. See below.

## Sensible use cases

The test for whether an ordering test is sensible is a single question:

> Is the order part of a published contract, of the system under test, or of the collaborator's interface, or is it an accident of how the method body happens to be written today?

If the order changes and no documented promise is broken, the test should not fail.

### Use case 1: the system under test exists in order to order things

An event emitter that guarantees listeners are invoked in order of their priority. The ordering is the behaviour under test.

#### PHPUnit 13.3

```php
public function testListenersAreInvokedInOrderOfTheirPriority(): void
{
    $calls = [];

    $low = $this->createMock(Listener::class);

    $low->expects($this->once())->method('handle')->willReturnCallback(
        static function () use (&$calls): void
        {
            $calls[] = 'low';
        },
    );

    $high = $this->createMock(Listener::class);

    $high->expects($this->once())->method('handle')->willReturnCallback(
        static function () use (&$calls): void
        {
            $calls[] = 'high';
        },
    );

    $emitter = new Emitter;
    $emitter->register($low, 1);
    $emitter->register($high, 10);

    $emitter->emit(new Event);

    $this->assertSame(['high', 'low'], $calls);
}
```

#### With this pull request

```php
public function testListenersAreInvokedInOrderOfTheirPriority(): void
{
    $journal = $this->createInvocationJournal();

    $low = $this->createMock(Listener::class);
    $low->expects($this->once())->method('handle')->recordIn($journal, 'low');

    $high = $this->createMock(Listener::class);
    $high->expects($this->once())->method('handle')->recordIn($journal, 'high');

    $emitter = new Emitter;
    $emitter->register($low, 1);
    $emitter->register($high, 10);

    $emitter->emit(new Event);

    $this->assertSame(['high', 'low'], $journal->invocations());
}
```

### Use case 2: the collaborator has a documented lifecycle protocol

A collaborator whose own interface documents a protocol and whose methods also have return values that the system under test uses. For example: `open()`, then `write()`, then `close()`.

This is where the PHPUnit 13.3 workaround hurts most: `willReturnCallback()` is the only way to observe an invocation, and it is also the only way to configure a return value. The two cannot be combined, so the return value has to be hand-rolled inside the closure and `willReturn()` becomes unavailable.

#### PHPUnit 13.3

```php
public function testWritesAllRecordsWithinOneSession(): void
{
    $calls = [];

    $writer = $this->createMock(Writer::class);

    $writer->expects($this->once())->method('open')->willReturnCallback(
        static function () use (&$calls): bool
        {
            $calls[] = 'open';

            return true;
        },
    );

    $writer->expects($this->exactly(2))->method('write')->willReturnCallback(
        static function () use (&$calls): int
        {
            $calls[] = 'write';

            return 1;
        },
    );

    $writer->expects($this->once())->method('close')->willReturnCallback(
        static function () use (&$calls): void
        {
            $calls[] = 'close';
        },
    );

    (new Exporter($writer))->export(['a', 'b']);

    $this->assertSame(['open', 'write', 'write', 'close'], $calls);
}
```

#### With this pull request

```php
public function testWritesAllRecordsWithinOneSession(): void
{
    $journal = $this->createInvocationJournal();

    $writer = $this->createMock(Writer::class);

    $writer->expects($this->once())->method('open')->recordIn($journal)->willReturn(true);
    $writer->expects($this->exactly(2))->method('write')->recordIn($journal)->willReturn(1);
    $writer->expects($this->once())->method('close')->recordIn($journal);

    (new Exporter($writer))->export(['a', 'b']);

    $this->assertSame(
        [
            'Writer::open()',
            'Writer::write()',
            'Writer::write()',
            'Writer::close()',
        ],
        $journal->invocations(),
    );
}
```

### Use case 3: relative order, not total order

Because the assertion is an ordinary assertion, the developer chooses how strict it is. PHPUnit does not have to grow an API for every shade of ordering:

```php
$invocations = $journal->invocations();

// strict, total order
$this->assertSame(['a', 'b', 'c'], $invocations);

// the same calls happened, order irrelevant
$this->assertEqualsCanonicalizing(['a', 'b', 'c'], $invocations);

// partial order: 'a' happened at some point before 'c'
$this->assertLessThan(
    array_search('c', $invocations, true),
    array_search('a', $invocations, true),
);

// how many invocations were recorded
$this->assertCount(4, $journal);
```

This is the point of the design. `withParameterSetsInOrder()`, `withParameterSetsInAnyOrder()` and `withParameterSetsInPartialOrder()` show how quickly this problem space proliferates once PHPUnit owns the assertion. Multiplying those three shapes by the cross-object dimension is not a road I want to go down.

### Asserting an order that is really a state invariant

This is the example from #6959:

```php
final readonly class StartTransactionMiddleware implements MiddlewareInterface
{
    public function handle(Envelope $envelope, StackInterface $stack): Envelope
    {
        $this->connection->beginTransaction();

        $envelope = $stack->next()->handle($envelope, $stack);

        $this->connection->commit();

        return $envelope;
    }
}
```

The stated goal is to assert that `beginTransaction()` happened before the next middleware ran. Asserting the ordinal position of the three calls is a poor way to express that: it passes for a middleware that opens a transaction, commits it, and opens a second one before delegating. The order is a symptom; it is not the invariant.

What the test actually wants to know is "was a transaction open, and not yet committed, while the next middleware ran" and "is nothing committed when the next middleware throws". Both are expressible today, and both say why the order matters:

```php
public function testTransactionIsActiveWhileTheNextMiddlewareRuns(): void
{
    $begun     = false;
    $committed = false;

    $connection = $this->createMock(Connection::class);

    $connection->expects($this->once())->method('beginTransaction')->willReturnCallback(
        function () use (&$begun): void
        {
            $begun = true;
        },
    );

    $connection->expects($this->once())->method('commit')->willReturnCallback(
        function () use (&$committed): void
        {
            $committed = true;
        },
    );

    $next = $this->createMock(MiddlewareInterface::class);

    $next->expects($this->once())->method('handle')->willReturnCallback(
        function (Envelope $envelope) use (&$begun, &$committed): Envelope
        {
            $this->assertTrue($begun, 'transaction was not started before the next middleware ran');
            $this->assertFalse($committed, 'transaction was committed before the next middleware ran');

            return $envelope;
        },
    );

    $stack = $this->createMock(StackInterface::class);
    $stack->expects($this->once())->method('next')->willReturn($next);

    (new StartTransactionMiddleware($connection))->handle(new Envelope(new stdClass), $stack);
}

public function testTransactionIsNotCommittedWhenTheNextMiddlewareThrows(): void
{
    $connection = $this->createMock(Connection::class);
    $connection->expects($this->once())->method('beginTransaction');
    $connection->expects($this->never())->method('commit');

    $next = $this->createMock(MiddlewareInterface::class);
    $next->expects($this->once())->method('handle')->willThrowException(new RuntimeException);

    $stack = $this->createMock(StackInterface::class);
    $stack->expects($this->once())->method('next')->willReturn($next);

    $this->expectException(RuntimeException::class);

    (new StartTransactionMiddleware($connection))->handle(new Envelope(new stdClass), $stack);
}
```

`recordIn()` is not the tool for this, and adding it to PHPUnit is not an endorsement of using it for this.

### Pinning down the order of unrelated collaborators

```php
$logger->expects($this->once())->method('info')->recordIn($journal, 'log');
$cache->expects($this->once())->method('set')->recordIn($journal, 'cache');
$metrics->expects($this->once())->method('increment')->recordIn($journal, 'metrics');

$this->assertSame(['log', 'cache', 'metrics'], $journal->invocations());
```

Nothing about this system breaks if the three calls are reordered. A test like this is a change detector: it fails on refactorings that change no behaviour, and it teaches readers that an accident of the current implementation is a promise. This is the "implicit protocol" smell, and no amount of API support makes it a good test.

### Substituting for `withParameterSetsInOrder()`

If you care about the arguments of consecutive invocations of one method, use `withParameterSetsInOrder()`, `withParameterSetsInAnyOrder()`, or `withParameterSetsInPartialOrder()`. By design, an `InvocationJournal` records that a method was invoked, not with which arguments.

## Why not the API proposed in #6959

The issue proposes:

```php
$invokedCountOrder = new InvokedCountOrder();

$invokedCountOrder->assertOrder([
    $connection->method('beginTransaction'),
    $stack->method('next'),
    $connection->method('commit'),
]);
```

I am against this for reasons that are independent of whether the feature is wanted:

1. It asks the user to instantiate a rule class directly: Invocation order rules are created through `expects()` and the factory methods on `TestCase` (`once()`, `exactly()`, …). `new InvokedCountOrder()` bypasses that entirely.

2. It calls `method()` a second time on an already-configured double: Each `->method()` call creates an additional matcher on the double. The builders passed to `assertOrder()` are therefore *not* the expectations configured above them; they are new, unrelated, unconfigured ones.

3. It passes fluent builders around as if they were values: `InvocationStubber` is a configuration builder. It is not a handle, an identity, or a value object, and making it one would freeze a large amount of internal structure into the public API.

4. It puts an assertion on a rule object: `assertOrder()` on an `InvokedCountOrder` mixes two responsibilities that are separate everywhere else in the test double API: describing what is expected, and reporting what happened.

5. It conflicts with `InvokedCount` and the issue says so itself: An ordering feature that cannot be combined with `once()` or `exactly()` is not orthogonal to the rest of the API, and orthogonality is the property that keeps the test double configuration surface comprehensible.

## Why a journal and not a `Sequence`

jMock and Mockery express this as a sequence object that expectations are attached to:

```php
$sequence = new Sequence;

$connection->expects($this->once())->method('beginTransaction')->inSequence($sequence);
$stack->expects($this->once())->method('next')->inSequence($sequence);
$connection->expects($this->once())->method('commit')->inSequence($sequence);
```

I chose the journal instead:

- It is orthogonal to everything else: A `Matcher` already carries an invocation count rule, a method name rule, a parameters rule and a stub. Recording is none of those. It composes with all of them, including `withParameterSetsInOrder()`, and it constrains none of them.

- It does not occupy the stub slot: Today, observing an invocation requires `willReturnCallback()`, which is also the only way to configure a return value. That is the actual friction this pull request removes.

- It adds no lifecycle: A `Sequence` has to be verified at the end of the test (were all of its steps reached?) so it would have to be registered with the `TestCase` next to the mock objects, and `verifyMockObjects()` would have to know about it. An `InvocationJournal` needs none of that: the developer's own assertion is the verification, and `createInvocationJournal()` is a plain factory with nothing to hook into.

- PHPUnit does not have to invent the semantics: No decisions to make about counts greater than one, partially consumed sequences, interleaved sequences, or partial orders across objects. The developer picks the assertion.

- The failure lands in the test, with a diff, on a line the developer wrote:

  ```
  --- Expected
  +++ Actual
   Array &0 [
       0 => 'begin',
  -    1 => 'commit',
  -    2 => 'handle',
  +    1 => 'handle',
  +    2 => 'commit',
   ]
  ```

  A sequence can only report "expected step 2, got X", from deep inside `InvocationHandler`.

- Declaration order stays non-semantic: With `inSequence()`, the order in which the `expects()` lines happen to be written silently becomes part of the assertion. That is a nasty property for a test to have.

## Details

- `recordIn()` is declared on `InvocationMocker`, next to `with()` and the `withParameterSetsIn*()` methods, and is therefore available on mock objects only. The only reason to observe the order of invocations is to assert it, and assertions about how a test double was used belong on a mock object, not on a test stub.

- `InvocationJournal` is an interface. It exposes `invocations()` and `Countable`, the reading side, and nothing else. Recording is done by the internal `InvocationJournalImplementation`, which is what `TestCase::createInvocationJournal()` returns. Nothing in the public API can write to a journal, and the way invocations are recorded internally stays free to change.

- The default label is `Class::method()`, using the fully-qualified name of the doubled type. Pass a label explicitly when two doubles of the same type record into the same journal, or when shorter labels make the expected array easier to read.

- `MockObject::method()` is declared to return `InvocationStubber`, so `recordIn()` is reachable only through `expects()`. Recording therefore always sits next to an explicit expectation, which is where it belongs.

- An invocation is recorded before the configured stub is invoked, so invocations made from within a `willReturnCallback()` are recorded after the invocation that caused them.

- An invocation is recorded even when it violates the expectation it matched, so that the journal is a faithful log of what happened.

- The same journal can be shared by any number of test doubles.

### DIFF
--- a/src/Framework/MockObject/Exception/EmptyInvocationJournalLabelException.php
+++ b/src/Framework/MockObject/Exception/EmptyInvocationJournalLabelException.php
@@ -1,0 +1,25 @@
+<?php declare(strict_types=1);
+/*
+ * This file is part of PHPUnit.
+ *
+ * (c) Sebastian Bergmann <sebastian@phpunit.de>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+namespace PHPUnit\Framework\MockObject;
+
+/**
+ * @no-named-arguments Parameter names are not covered by the backward compatibility promise for PHPUnit
+ *
+ * @internal This class is not covered by the backward compatibility promise for PHPUnit
+ */
+final class EmptyInvocationJournalLabelException extends \PHPUnit\Framework\Exception implements Exception
+{
+    public function __construct()
+    {
+        parent::__construct(
+            'The label for recording invocations in an invocation journal must not be empty',
+        );
+    }
+}
--- a/src/Framework/MockObject/Exception/InvocationJournalAlreadyRegisteredException.php
+++ b/src/Framework/MockObject/Exception/InvocationJournalAlreadyRegisteredException.php
@@ -1,0 +1,25 @@
+<?php declare(strict_types=1);
+/*
+ * This file is part of PHPUnit.
+ *
+ * (c) Sebastian Bergmann <sebastian@phpunit.de>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+namespace PHPUnit\Framework\MockObject;
+
+/**
+ * @no-named-arguments Parameter names are not covered by the backward compatibility promise for PHPUnit
+ *
+ * @internal This class is not covered by the backward compatibility promise for PHPUnit
+ */
+final class InvocationJournalAlreadyRegisteredException extends \PHPUnit\Framework\Exception implements Exception
+{
+    public function __construct()
+    {
+        parent::__construct(
+            'An invocation journal has already been registered for this test double',
+        );
+    }
+}
--- a/src/Framework/MockObject/Generator/Exception/MethodNamedMethodException.php
+++ b/src/Framework/MockObject/Generator/Exception/MethodNamedMethodException.php
@@ -10,6 +10,8 @@
 namespace PHPUnit\Framework\MockObject\Generator;
 
 /**
+ * @no-named-arguments Parameter names are not covered by the backward compatibility promise for PHPUnit
+ *
  * @internal This class is not covered by the backward compatibility promise for PHPUnit
  */
 final class MethodNamedMethodException extends \PHPUnit\Framework\Exception implements Exception
--- a/src/Framework/MockObject/Generator/Exception/MethodNamedRecordInvocationsInException.php
+++ b/src/Framework/MockObject/Generator/Exception/MethodNamedRecordInvocationsInException.php
@@ -1,0 +1,21 @@
+<?php declare(strict_types=1);
+/*
+ * This file is part of PHPUnit.
+ *
+ * (c) Sebastian Bergmann <sebastian@phpunit.de>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+namespace PHPUnit\Framework\MockObject\Generator;
+
+/**
+ * @internal This class is not covered by the backward compatibility promise for PHPUnit
+ */
+final class MethodNamedRecordInvocationsInException extends \PHPUnit\Framework\Exception implements Exception
+{
+    public function __construct()
+    {
+        parent::__construct('Doubling interfaces (or classes) that have a method named "recordInvocationsIn" is not supported.');
+    }
+}
--- a/src/Framework/MockObject/Generator/Exception/MethodNamedRecordInvocationsInException.php
+++ b/src/Framework/MockObject/Generator/Exception/MethodNamedRecordInvocationsInException.php
@@ -10,6 +10,8 @@
 namespace PHPUnit\Framework\MockObject\Generator;
 
 /**
+ * @no-named-arguments Parameter names are not covered by the backward compatibility promise for PHPUnit
+ *
  * @internal This class is not covered by the backward compatibility promise for PHPUnit
  */
 final class MethodNamedRecordInvocationsInException extends \PHPUnit\Framework\Exception implements Exception
--- a/src/Framework/MockObject/Generator/Generator.php
+++ b/src/Framework/MockObject/Generator/Generator.php
@@ -80,6 +80,7 @@ final class Generator
      * @param ?list<non-empty-string> $methods
      * @param array<mixed>            $arguments
      * @param list<non-empty-string>  $doubledProperties
+     * @param ?non-empty-string       $displayName
      *
      * @throws ClassIsAnonymousException
      * @throws ClassIsEnumerationException
@@ -92,7 +93,7 @@ final class Generator
      * @throws RuntimeException
      * @throws UnknownTypeException
      */
-    public function testDouble(string $type, bool $mockObject, ?array $methods = [], array $arguments = [], string $mockClassName = '', bool $callOriginalConstructor = true, bool $callOriginalClone = true, bool $returnValueGeneration = true, array $doubledProperties = []): MockObject|Stub
+    public function testDouble(string $type, bool $mockObject, ?array $methods = [], array $arguments = [], string $mockClassName = '', bool $callOriginalConstructor = true, bool $callOriginalClone = true, bool $returnValueGeneration = true, array $doubledProperties = [], ?string $displayName = null): MockObject|Stub
     {
         if ($type === Traversable::class) {
             $type = Iterator::class;
@@ -119,6 +120,7 @@ final class Generator
             $arguments,
             $returnValueGeneration,
             $mockObject,
+            $displayName,
         );
 
         assert($object instanceof $type);
@@ -196,6 +198,7 @@ final class Generator
             $intersectionName,
             $mockObject,
             returnValueGeneration: $returnValueGeneration,
+            displayName: implode('&', $interfaces),
         );
     }
 
@@ -294,12 +297,13 @@ final class Generator
     }
 
     /**
-     * @param array<mixed> $arguments
+     * @param array<mixed>      $arguments
+     * @param ?non-empty-string $displayName
      *
      * @throws ReflectionException
      * @throws RuntimeException
      */
-    private function instantiate(DoubledClass $mockClass, string $type, bool $callOriginalConstructor, array $arguments, bool $returnValueGeneration, bool $isMockObject): object
+    private function instantiate(DoubledClass $mockClass, string $type, bool $callOriginalConstructor, array $arguments, bool $returnValueGeneration, bool $isMockObject, ?string $displayName = null): object
     {
         $className = $mockClass->generate();
 
@@ -324,7 +328,7 @@ final class Generator
          */
         $reflector->getProperty('__phpunit_state')->setValue(
             $object,
-            new TestDoubleState($mockClass->configurableMethods(), $type, $returnValueGeneration, $isMockObject),
+            new TestDoubleState($mockClass->configurableMethods(), $type, $returnValueGeneration, $isMockObject, $displayName),
         );
 
         if ($callOriginalConstructor && $reflector->getConstructor() !== null) {
@@ -353,6 +357,7 @@ final class Generator
      * @throws ClassIsEnumerationException
      * @throws ClassIsFinalException
      * @throws MethodNamedMethodException
+     * @throws MethodNamedRecordInvocationsInException
      * @throws ReflectionException
      * @throws RuntimeException
      */
@@ -505,6 +510,11 @@ final class Generator
 
         if ($mockMethods->hasMethod('method') || $class->hasMethod('method')) {
             throw new MethodNamedMethodException;
+        }
+
+        if ($mockObject &&
+            ($mockMethods->hasMethod('recordInvocationsIn') || $class->hasMethod('recordInvocationsIn'))) {
+            throw new MethodNamedRecordInvocationsInException;
         }
 
         $traits[] = Method::class;
--- a/src/Framework/MockObject/Runtime/Api/MockObjectApi.php
+++ b/src/Framework/MockObject/Runtime/Api/MockObjectApi.php
@@ -9,6 +9,7 @@
  */
 namespace PHPUnit\Framework\MockObject;
 
+use function assert;
 use PHPUnit\Framework\MockObject\Rule\InvocationOrder;
 
 /**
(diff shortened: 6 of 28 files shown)


sebastianbergmann:
I have revised this pull request: the feature is the same: a journal that records the order in which methods of mock objects are invoked and an ordinary assertion on that order. But the recording hook has moved, and that changes the API.

### What was wrong

`recordIn()` configured recording on a single expectation, and the recording happened in `Matcher::invoked()`. That is the expectation-matching path. But "the order in which methods were invoked" is a property of dispatch, and the two are not the same thing. Four defects followed directly from the mismatch:

One call could produce several entries: `InvocationHandler::invoke()` calls `invoked()` on every matcher whose `matches()` returns true. Two expectations for the same method, both recording, turned one call into two entries:

```php
$double->expects($this->any())->method('doSomething')->recordIn($journal, 'A');
$double->expects($this->any())->method('doSomething')->recordIn($journal, 'B');

$double->doSomething();

// ['A', 'B']
```

Real invocations could be missing: A matcher configured with `after()` returns `false` from `matches()` until its predecessor has run, so `invoked()` was never reached and the invocation was not recorded. The journal could report the exact reverse of what happened:

```php
$double->expects($this->any())->method('one')->id('first');
$double->expects($this->any())->method('two')->after('first')->recordIn($journal, 'two');

$double->two();
$double->one();
$double->two();

// ['two'] — two() was invoked twice, and the journal claims it happened after one()
```

This one matters more than the others, because `id()`/`after()` is the API this feature is meant to replace, so people migrating hit it first.

The default label was not what it claimed to be: It was built from `$invocation->className()`, which is the declaring class, not the doubled type. Mocking a subinterface recorded the parent's name, contradicting the failure message the same `Matcher` produces:

```php
interface Base { public function doIt(): void; }
interface Derived extends Base {}

$this->createMock(Derived::class);

// recorded:  'Base::doIt()'
// verify():  'Derived::doIt()'
```

For property hooks it was worse: the label contained the randomly generated double class name, so it differed on every run and no assertion could be written against it at all.

Calls that failed `with()` were recorded as though they matched: Recording ran before `parametersRule->apply()`, so `->with($expected)->recordIn($journal)` could not distinguish "called with the right arguments" from "called at all".

### What changed

Recording now happens once per dispatch, in `InvocationHandler::invoke()`, and the journal is registered on the mock object rather than on an expectation:

```php
$journal = $this->createInvocationJournal();

$database   = $this->createMock(Database::class);
$dispatcher = $this->createMock(Dispatcher::class);

$database->recordInvocationsIn($journal, 'database');
$dispatcher->recordInvocationsIn($journal, 'dispatcher');

$database->expects($this->once())->method('execute')->willReturn(true);
$dispatcher->expects($this->once())->method('dispatch');

(new Publisher($database, $dispatcher))->publish('value');

$this->assertSame(
    ['database::execute()', 'dispatcher::dispatch()'],
    $journal->asArray(),
);
```

`InvocationHandler::invoke()` is the single point every doubled method and every property hook funnels through. Recording there makes all four defects structurally impossible rather than things to document: there is exactly one entry per call, `after()` and `with()` cannot suppress or distort it, and the label is decided where the doubled type is actually known.

### What this costs

Recording is no longer per-method: Registering a journal records every invocation of that mock object, not only the ones you singled out. This is deliberate and keeps the journal a faithful log. `only()` narrows what an assertion looks at without changing what was recorded:

```php
$this->assertSame(['write', 'dispatch'], $journal->only('write', 'dispatch'));
```

That asserts `write()` happened before `dispatch()` and says nothing about anything else that happened around them, so it is also how a relative order is expressed now, without `array_search()` arithmetic.

Labels name the double by default, not the invocation: Entries are `label::method()` rather than a free-form description of the step. A third argument restores the original ergonomics for the methods where it matters:

```php
$database->recordInvocationsIn($journal, 'database', ['execute' => 'write']);
$dispatcher->recordInvocationsIn($journal, 'dispatcher', ['dispatch' => 'dispatch']);

$this->assertSame(['write', 'dispatch'], $journal->asArray());
```

A label configured this way replaces the whole entry, exactly as `recordIn()`'s label did. Methods not named in the map keep `label::method()`. The map is validated at registration: naming a method the double does not have throws `MethodCannotBeConfiguredException`, so a typo cannot silently fail to apply. Property hooks are named as they are everywhere else, `'$property::get'`.

I consider the remaining loss acceptable. Selectivity is not worth a journal that silently drops invocations or records ones that did not match.

It also contradicts one thing I wrote in the original description: "Recording therefore always sits next to an explicit expectation, which is where it belongs." That is no longer true. Recording is now independent of expectations, which is what lets it be a faithful log.

### API changes since the first version

- `InvocationMocker::recordIn()` is gone. `MockObject::recordInvocationsIn(InvocationJournal $journal, ?string $label = null, array $methodLabels = [])` replaces it.
- `InvocationJournal::invocations()` is now `asArray()`, there is a new `only(...$labels)` for narrowing an assertion to selected labels, and the interface extends `IteratorAggregate` as well as `Countable`. This matches every other collection in `src/`, all of which are `Countable`, `IteratorAggregate` and expose `asArray()`. Previously `count($journal)` worked but `foreach` silently iterated zero times and `assertContains()` raised a `TypeError`.
- Recording is still mock-object-only, for the reason given in the original description.
- A mock object that only records no longer triggers "No expectations were configured for the mock object". That notice advises using a test stub, which is not possible for a recording double, so a journal now counts as configuration.
- The default label is now the doubled type in every case: inherited methods report the doubled type rather than the declaring class, property hooks are deterministic, and a double for an intersection of interfaces reports `I1&I2` instead of the generated `Intersection_I1_I2_<hash>` name.
- An empty label, an empty per-method label, an unknown method name in the label map, and a second `recordInvocationsIn()` call on the same double all throw instead of being silently accepted. The second `recordIn()` used to discard the first journal without a word.
- Doubling a class that declares its own `recordInvocationsIn()` method throws, mirroring the existing handling of a method named `method()`. This applies to mock objects only; such a class can still be doubled as a test stub.

### Performance

Recording used to cost an unconditional call plus a per-invocation `sprintf()` in `Matcher::invoked()`, which is the hot path of every mock method call in every suite. It cost about +21% on that path for a feature that is unconfigured in almost every matcher. The cost is now one null check per dispatch, and nothing at all in `Matcher`.

